PR #662, opened 2012-04-11
Error while restarting using SIGHUP

ivirabyan:
There was an error, saying "Pidfile is already exists" while restarting
using SIGHUP. The problem arises only when "--pidfile" option is used.

Related to: https://github.com/ask/celery/issues/26


TylerSheffels:
I think

PR #817, opened 2012-06-21
Feature: Dynamic tasks

steeve:
So I submit to you dynamic tasks, and dynamic chords. This is a really powerful concept, and I think it changes the game.

Subtasks returned by dynamic tasks are executed right after the first task executes. As if they were in a chain.
You can also return chains and chords, and they will be properly inserted.

This allows you to design a pipeline that can be completely dynamic, while benefiting from Celery's powerful idioms (subtasks, chains, chords...).

Our whole backend at @veezio is powered by these. They allow use to have extensively dynamic pipelines.
We have pipes inside chords, chords inside chords, tasks put before pipes etc....

To be honest, I can't think of something you can't do with this.

How to use:

```
import celery.contrib.dynamic

@celery.dynamic_task
def test(number):
    if number > 5:
        return more.s()
    return less.()


@celery.dynamic_task
def more():
    print "it's more !"

@celery.dynamic_task
def less():
    print "it's less !"
```

But you can do cool shit too!

```
@celery.dynamic_task
def one(n):
    if n:
        return three.si(n + 4) | four.s() # if you don't use an immutable task, the result of one (a subtask) will be sent to three.
    return n
pipe = one.s(1) | two.s()

pipe.apply_async() # will run: one(1) | three(1+4) | four | two
```

You can also use them in chords! And have chords in chords!:

```
from celery.contrib.dynamic import dynamic_chord as chord

@celery.dynamic_task
def resolve_url(url):
    # do some stuff, and paralellize all that
    return chord(map(do_stuff.si, data), on_finished.s())

resolve_urls = chord(map(resolve_url.si, urls), url_resolved.s())
```

In that case url_resolved will be called with the results from on_finished(), which is:

```
[ [..., ], [..., ] ]
```

### DIFF
--- a/celery/contrib/dynamic.py
+++ b/celery/contrib/dynamic.py
@@ -7,6 +7,7 @@ Dynamic Tasks
 Subtasks returned by dynamic tasks are executed right after the first
 task executes. As if they were in a chain.
 You can also return chains and chords, and they will be properly inserted.
+You can have chords in chords etc...
 
 This allows you to design a pipeline that can be completely dynamic, while
 benefiting from Celery's powerful idioms (subtasks, chains, chords...).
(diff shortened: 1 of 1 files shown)


celery/contrib/dynamic.py
```diff
@@ -0,0 +1,82 @@
...
+                    current_task.request.group = None
+                current_task.request.callbacks = [retval]
+            return retval
+        return self.task(*args, **kwargs)(_fn_wrapper)
+    if len(args) == 1 and callable(args[0]):
+        fn = args[0]
+        args = args[1:]
+        return do_wrap(fn)
+    return do_wrap
+setattr(Celery, "dynamic_task", dynamic_task)
```
ThiefMaster: Why not simply `Celery.dynamic_task = dynamic_task`?

PR #1434, opened 2013-06-27
Feature: let subclasses easily generate their own task_id

ludwigkraatz:
this way subclasses of BaseTask can better influence the task_id generation

### DIFF
--- a/celery/app/amqp.py
+++ b/celery/app/amqp.py
@@ -213,7 +213,7 @@ class TaskProducer(Producer):
         retry = self.retry if retry is None else retry
         _rp = (dict(self.retry_policy, **retry_policy) if retry_policy
                else self.retry_policy)
-        task_id = task_id or uuid()
+        task_id = task_id or self.app.tasks[task_name].generate_task_id() # is this valid?
         task_args = task_args or []
         task_kwargs = task_kwargs or {}
         if not isinstance(task_args, (list, tuple)):
--- a/celery/app/builtins.py
+++ b/celery/app/builtins.py
@@ -183,7 +183,7 @@ def add_group_task(app):
         def prepare(self, options, tasks, args, **kwargs):
             AsyncResult = self.AsyncResult
             options['group_id'] = group_id = (
-                options.setdefault('task_id', uuid()))
+                options.setdefault('task_id', self.generate_task_id())) #TODO: check if not generate_group_id
 
             def prepare_member(task):
                 task = maybe_subtask(task)
@@ -192,7 +192,7 @@ def add_group_task(app):
                 try:
                     tid = opts['task_id']
                 except KeyError:
-                    tid = opts['task_id'] = uuid()
+                    tid = opts['task_id'] = self.generate_task_id()
                 return task, AsyncResult(tid)
 
             try:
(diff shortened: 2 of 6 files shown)


malinoff:
I want to add my 2 cents here.

While digging Celery code, I was often confused by the naming. One of the examples is ``task_id``.
It seems that this name is related to a task, but it's not. Actually, it is related to a *task request*.
And Celery provides a way to specify a custom ``task_id`` via arguments to ``apply_async``, ``send_task`` and so on.

Also, I find it extremely verbose to explicitly specify ``@app.task(base=CustomTask)`` for each task you want to have a custom ``task_id`` generation.

That's why I don't think that exposing the internal ``task_id`` generation is OK.
I can imagine the single useful purpose of having such method - to centralize the use of a generator. But this is already accomplished by Celery, because ``uuid`` is a function that wraps *something that generates a unique identifier* (more on that in https://github.com/celery/kombu/blob/master/kombu/utils/__init__.py#L148-L154)

celery/app/amqp.py
```diff
@@ -213,7 +213,7 @@ def publish_task(self, task_name, task_args=None, task_kwargs=None,
         retry = self.retry if retry is None else retry
         _rp = (dict(self.retry_policy, **retry_policy) if retry_policy
                else self.retry_policy)
-        task_id = task_id or uuid()
+        task_id = task_id or self.app.tasks[task_name].generate_task_id() # is this valid?
```
malinoff: No, you can't do that - because at this point you don't have any task *object*, only the string representation. And you may send a string representation of a task to a remote worker that has that specific task implementation, while the caller side may has no such task implementation at all.

thedrow:
@ask @malinoff So in your opinion this PR should be rejected?

malinoff:
I think yes, it should be rejected. Although I'd like to figure out the best way to do that, probably by renaming ``task_id`` to ``request_id``.

ask:
I'm ambivalent about this patch, I guess I'm not able to see when it would be useful for a task to override how IDs are generated?

Renaming `task_id` to `request_id` would be nice, but to change it we would have to accept both of the arguments for backward compatibility, and that's some inconvenience just to change a name.

ludwigkraatz:
when i submitted this request, i was using the following custom uuid generation mehtod:

https://github.com/ludwigkraatz/django-model-registry/blob/master/model_registry/uuid.py

this way i was able to store some custom information within the uuid itself.
e.g. what data type the object is or what customer / user it belongs to.

it helped for resolving the db table for this uuid represented object (domain.org/uuid/********)
or use some security logic without accessing the db by parsong this custom information (eg. is the requesting user the owner).

thats why a centralized and customizable id generation method would have helped me - because i could have used this for celery tasks as well.

 i no longer use this approach of custom (smart) uuids.
but in general i think its good to allow custom id generation - as the structure (which might also be just uuid5 and not custom at all)  of an objects  id might be important for some projects which not only use but deeply integrate celery into the projects core.

Sent from my mobile device

Am 12.06.2015 um 20:34 schrieb "Ask Solem Hoel" <notifications@github.com>:

malinoff:
So as I said, you can pass any custom ``task_id`` when executing tasks if you want to have some smart mapping between task ids and something else.

Another approach is to populate such mapping *somewhere else*, outside of celery. You can retrieve the generated ``task_id`` right after executing a task.

I'm going to close this PR, but I'll have this in mind - maybe I'll find a good approach to do that.

PR #1608, opened 2013-10-30
Feature: Autoretry for task

amezhenin:
Consider the following task:

    @task(max_retries=5, default_retry_delay=1)
    def div(a, b):
        try:
            return a / b
        except ZeroDivisionError, exc:
            raise div.retry(exc=exc)

If we want to retry task on particular exception, we have to put ugly `try-except`s in all our tasks. The code would be much cleaner, if you can tell Celery to retry task automatically. like so:

    @task_autoretry(max_retries=5, default_retry_delay=1, autoretry_on=ZeroDivisionError)
    def div(a, b):
        return a / b

I wrote decorator for this purpose in my own project, and I decided to share it with you.

### DIFF
--- a/celery/task/base.py
+++ b/celery/task/base.py
@@ -11,6 +11,7 @@
 """
 from __future__ import absolute_import
 
+import functools
 from kombu import Exchange
 
 from celery import current_app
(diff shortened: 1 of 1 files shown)


ask:
Closed by #2112 Thanks!

PR #1651, opened 2013-11-15
Corrected `CELERYD_PID_FILE` and `CELERYD_LOG_FILE` format strings: %N should really be %n

ksze:
The `%N` placeholder in the `CELERYD_PID_FILE` and `CELERYD_LOG_FILE` format strings should really be `%n`.


ask:
Seems this was already merged or something, as the current document uses `%n`.

PR #1723, opened 2013-12-08
Search `/usr/local/etc` for config file in [Free]BSD

tahajahangir:
In FreeBSD, by default config files are/should be in `/usr/local/etc` rather than `/etc/default` or even `/etc`, so check for existence of `/etc/default`


ask:
Merged into master!

PR #1812, opened 2014-01-22
Add an signal, sent when task state is updated

kaizoku:
Add a signal `task_state`, that gets thrown whenever a task's state is updated.
I wanted to add a handler to get info from my custom progress state.

From a task:
```
@celery.task(bind=True)
def update(self):
    for stuff in mystuff:
        ... do stuff with stuff
        self.update_state(
                state="PROGRESS",
                meta={'current': i, 'total': numtotal}
                )

```
```
@task_state.connect
def task_state_handler(sender=None, state="", meta={}, *args, **kwargs):
    print "%s sent state: %s with args %s" % (sender, state, meta)
```


ask:
This seems to be rarely requested, and you can easily enable this manually right now by overriding either Task.update_state, or Backend.store_result

PR #1822, opened 2014-01-25
Fix Celerybeat from hanging when using --detach mode after dispatching 1st set of tasks.

rogerhu:
The change made in https://github.com/celery/celery/commit/022ee4ade9f7d03f995413212b135fe18451be95
caused file descriptor 0 (usually sys.stdin) not to be included in the list of file descriptors to keep.
Doing a filter(none) apparently removes this number from this list, which somehow causes sys.stdin to be closed.
Changed the logic to be filter(lambda: x is not None) to allow 0 to be included.

FYI if we printed out the results from the close_all_fds() (i.e. disabling the os.closerange() calls and
writing the for loops), here's what we would see:

Celery v3.1.4 (working):

low=-1, high=0
low=0, high=1
low=1, high=2
low=2, high=1024
os.closerange(): low + 1 = 3, high=1024

Celery v3.1.5 (breaking):

low=-1, high=1
os.closerange(): low + 1 = 0, high=1
low=1, high=2
low=2, high=1024
os.closerange(): low + 1 = 3, high=1024

Somehow this behavior caused problems with Celerybeat and the UUID generation for task ID's on Ubuntu v12.04.
Celerybeat upon dispatching a few tasks, sleeping, and waking up again, blocks  while trying to generate a UUID for
If the uuidd daemon is invoked before startup, a socket connection is established and task dispatching occurs normally.  After another wakeup period, no subsequent calls are needed to the UUID socket daemon and read calls to /dev/urandom proceed correctly.  (Given that the Python uuid library only tries to use uuid_get_random() calls, there should only be read attempts to /dev/urandom and not to any socket-based connections.)

The latest current libuuid library on Ubuntu 12.04 also has issues with leaking file descriptors for each
UUID call (https://github.com/karelzak/util-linux/commit/c544aa2c25095e6a4d4fca761eb15c46435086fc).  Also,
the /dev/urandom file descriptors are not closed after reuse, so Celery workers that restart themselves
by os.exec() (https://github.com/karelzak/util-linux/commit/a61cac5b6474494fe5c409b82d09236c45856b07)
could end up maxing out on file descriptors.  The long term fix is that Ubuntu v12.04 should update
to the latest version of util-linux that includes the patches for libuuid.

### DIFF
--- a/celery/platforms.py
+++ b/celery/platforms.py
@@ -266,7 +266,8 @@ def _create_pidlock(pidfile):
 if hasattr(os, 'closerange'):
 
     def close_open_fds(keep=None):
-        keep = list(uniq(sorted(filter(None, (
+        # using filter(None) causes 0 to be filtered out too
+        keep = list(uniq(sorted(filter(lambda x: x is not None, (
             maybe_fileno(f) for f in keep or []
         )))))
         maxfd = get_fdmax(default=2048)


ngo:
I would like to say that in celery 3.1.18 and Ubuntu 14.04 I can still reproduce this issue. The fix for me was https://github.com/celery/celery/pull/1822#issuecomment-33430437

ask:
@ngo, what's the difference?  The current code is slightly different but should do the same:
https://github.com/celery/celery/blob/3.1/celery/platforms.py#L272-L274

rogerhu:
It's a problem with the libuuid library.  Still not fixed in Ubuntu 14.10.
I believe Ubuntu 15.04 has this issue resolved.


http://engineering.hearsaysocial.com/2014/01/30/file-descriptor-issues-in-celery/

dhui:
Just got bit by this issue as well (Celery 3.1.18 on Ubuntu 14.04)...
Thanks for tracking down the issue over a year ago @rogerhu !

@ask What do you think about doing one of the following to prevent future people from having these issues:

1. changing the `/etc/init.d/celerybeat` script to not specify `--detach` and use `start-stop-daemon` instead: https://github.com/celery/celery/blob/master/extra/generic-init.d/celerybeat
1. blessing a version of an upstart script while we wait for systemd in an Ubuntu LTS release

I can open a PR to do either.

Issue #1780, opened 2014-01-11
Filesystem result backend


moshevds:
Hi,

As an experiment, I tried to implement a filesystem based result backend, you can see it at https://github.com/moshevds/celery/commit/f50045aeb30bc8f074216305ef9d8fe1cca13b7d
I would like to hear any comments you have on this.

Perhaps this could be useful as more than an experiment. There might be uses for celery where a distributed filesystem is readily available, and a good choice for a backend. It should work with distributed filesystems like glusterfs (in my case) and hdfs (for hadoop users), but also shared filesystems like nfs, cifs and even sshfs, or the local /tmp directory for development environments.

As of yet, I don't really have an answer about performance in different usecases (compared to the other backends), but I don't see any reason that it wouldn't be as good (or better) than other fsync()ing backends.

What would be needed to add something like this upstream? I only tried this with python 3.3, and am not really sure what setups I should try and/or run the unittests with.
Are there any other things I should know before I assign some time for myself to work on creating a Filesystem result backend?

PR #1844, opened 2014-02-03
Initial support for a filesystem based result backend.

moshevds:
This is an implementation of a filesystem based result backend, as discussed earlier in #1780.
I ran the unittests on python 2.6, 2.7, 3.3 and pypy, as well as tested the backend locally and with multiple nodes on GlusterFS.

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -156,6 +156,7 @@ Antoine Legrand, 2014/01/09
 Pepijn de Vos, 2014/01/15
 Dan McGee, 2014/01/27
 Paul Kilgo, 2014/01/28
+Môshe van der Sterre, 2014/01/31
 Martin Davidsson, 2014/02/08
 Chris Clark, 2014/02/20
 Matthew Duggan, 2014/04/10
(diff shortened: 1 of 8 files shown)


thedrow:
The build is failing for Python 3. https://travis-ci.org/celery/celery/jobs/45052922#L657
Let's fix that and it will be merged.

moshevds:
Hi @thedrow,

The backend was created more than a year ago, and I have rebased a few times to get it merged. It seems I have rebased over a change in the way values are serialized, without checking if everything still works.

It is a small fix to convert the parameter for set() from a string to a buffer, but will that parameter always be a string (in the current codebase)? I think that makes more sense than the old behavior actually, so perhaps you can confirm this?

If not, then I think I need to add more tests, any suggestions on what tests are needed?

Greetings,
Môshe van der Sterre

thedrow:
I think that's probably valid because you are either passing pickled objects or json or msgpack.
@ask @malinoff Any thoughts?

moshevds:
Actually, pickle and msgpack are bytes, only json is str. I just checked, and reverting 8d4354a "Default serializer is now json" makes the tests pass.

In conclusion, for now, I'll add the following:
```
 if not isinstance(value, bytes):
    value = value.encode('utf-8')
```
So it works with both str and bytes (and is also correct for python 2.6+).

But this also means my unit tests are incomplete when only checking the default serializer. I'm not sure how to best fix that.

moshevds:
I looked at this some more, and I found that I can use kombu.utils.encoding.ensure_bytes.

Another thing that I was worried about is the decoding, because get() will always return bytes, not str. I found that this is actually handled in kombu.utils.json.loads, so that is not a problem.

auvipy:
This branch has conflicts that must be resolved

thedrow:
Thanks!

PR #1878, opened 2014-02-19
Init script improvements: resolve symbolic link to config file

ifischer:
Current sanity check will fail if config file is a symbolic link: symbolic links are always writeable for all.
Therefore, read link target before checks.


ask:
I don't think the init scripts can be expected to run everywhere, and that people may have to modify i to change config location, su behavior, etc.

It's more of a suggestion for package system maintainers than a drag and drop solution.

PR #1897, opened 2014-02-28
ClassTask: a metaclass that turns any generic class into a celery task

kalefranz:
I would like to submit this pull request for _feedback only_ right now, with the possibility of merging into `celery:master` after further work. It is not currently ready to integrate into the celery code base (although as you'll see, all tests pass and I don't think it would create any issues).

My objective here was to create some type of class modifier (mixin, decorator, metaclass, etc) that turned the _entire_ class into a task.  A small example might best illustrate what I want.

```
class WelcomeEmail(ClassTaskMixin, EmailBase):

    def __init__(self, recipient_address, template_name, template_context):
        self.recipient_address = recipient_address
        self.template_name = template_name
        self.template_context = template_context

   def send(self):
       self.render_templates()
       self.construct_mime()
       self.archive_to_db()
       return self.send_smtp_email()

   def run(self):
       return self.send()
```

The `WelcomeEmail` class would then be used as

```
we = WelcomeEmail(recipient_address, template_name, template_context)
we.enqueue()
```

where the `enqueue` method is provided by `ClassTaskMixin` and is analogous to the celery `delay` or `apply_async`.  Here, since the `run` method redirects to `send`, `enqueue` would simply send the email in the asynchronous process.

It's important to note that one of my goals was/is **delayed initialization of the instance object**.  That is, the python standard call to `__init__` should be "lazy" and **not** called when the instance is created.  This is important.  With a _lazy_ `__init__`, the user need not worry about "heavy" and duplicitous work being done in `__init__` (duplicitous because without lazy `__init__` the work would be done in both the synchronous and worker processes).  In the `WelcomeEmail` example, the coder may have for whatever reason put the calls to `render_templates` and `construct_mime` within `__init__`; and here, calling `render_templates` and `construct_mime` in the synchronous process is wasteful.

Now, I am aware of `celery.contrib.methods`, and after the above explanation, you can see that what I'm after is a bit more extensive.

I believe the pull request I am submitting here accomplishes my goals.  Rather than a class mixin, it was necessary to create a metaclass to accomplish the lazy initialization.  The two files in this pull request contain the `ClassTask` metaclass and a retrofitting of the first 12 of 32 tests for `celery.Task` (in `celery/tests/tasks/test_tasks.py`)

I am _very_ much interested in community input.  This code will likely be used in production for my employer.  And if the community thinks this might be more broadly useful, I am happy to contribute it to the celery library.

Some questions to start with...
1) Are there better/more direct ways to test the `ClassTask` code than be retrofitting/duplicating what's in `celery/tests/tasks/test_tasks.py`?
2) For the methods in `ClassTaskBase` that are almost nearly just a call to `super` (_i.e._ `apply_async`, `apply`, `retry`, etc.), is there a better way to avoid the wrote code?  My guess is there will be even more of these methods once I work through all of the applicable unit tests.
3) Any general comments/critiques/criticisms/questions/insights to help me make this code better?

### DIFF
--- a/celery/contrib/classtask.py
+++ b/celery/contrib/classtask.py
@@ -1,0 +1,255 @@
+# -*- coding: utf-8 -*-
+"""
+celery.contrib.classtask
+========================
+
+Metaclass that supports creating tasks out of classes.
+
+Use of this metaclass turns a class and its subclasses into a celery task. With
+minimal differences, the task behaves similarly to a task created by the
+`app.task` function decorator.
+
+
+Key Concepts
+------------
+
+* This metaclass makes instance initialization "lazy".  That is, the
+  conventional python object instantiation process is intercepted, and
+  :meth:`__init__` is explicitly _not_ called.  This delayed _initialization_ of
+  the instance object prevents unnecessary work being done in the synchronous
+  process.
+* References to the `args` and `kwargs` given to __init__ are held by the
+  instance object. They are passed as parameters in the `task` message to fully
+  initialize the class in the external process.
+* After class instantiation, :meth:`enqueue` may be called on the instance
+  object analogously to :meth:`delay` or :meth:`apply_async`. :meth:`enqueue`
+  takes no `args` or `kwargs`. It does however accept the same `options` as
+  :meth:`apply_async`.
+
+
+Use
+---
+
+1) Add `__metaclass__ = ClassTask` to the class definition. Among other
+   things, your class is now a subclass of :class:`ClassTaskBase`.
+2) Override the :meth:`run` method.  This code, along with whatever code
+   is in :meth:`__init__`, will be executed by the external process.
+3) To use your class, instantiate it as normal--i.e.
+   `my_instance = MyClass(*args, **kwargs)`. Then call :meth:`enqueue` on the
+   instantiated object--i.e. `my_instance.enqueue()`. The code in
+   :meth:`__init__` is *not* ran in-process, but rather delayed for the external
+   process to run.
+4) If for some reason you need MyClass initialized in-process, call :meth:`init`
+   on the instance object--i.e. `my_instance.init()`.  In this case, the
+   :meth:`__init__` code will run both in-process *and* in the external process.
+5) One method decorator is provided as auxiliary to this metaclass.
+   If a method is decorated with :decorator:`auto_init`, that method can
+   be accessed without first calling :meth:`init`, and :meth:`init` will
+   _automatically_ be called if :meth:`__init__` has not been ran on the
+   instance object already.
+6) Access to any :decorator:`staticmethod` or :decorator:`classmethod` is
+   allowed even if :meth:`__init__` hasn't been called on the instance.
+
+
+Metaclass Strategy
+------------------
+
+* The metaclass fundamentally delays execution of :meth"`__init__`, and it
+  stores the `args` and `kwargs` passed to the class callable as references
+  :attr:`_args` and :attr:`_kwargs` on the new instance object.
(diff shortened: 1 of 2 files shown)


ask:
This is an interesting subject.

In the beginning of Celery I allowed any callable to be a task, so as long as it had `__call__` it could be registered and used as a task.

The code ended up being very vague though, as it had to check for every little attribute, feature or method, and that also made the task execution rather slow.

Since then I have rewritten the task execution code to "precompile" the steps required, so that there's a closure generated for every task.  Most of the attributes are already resolved etc, caching things like "task.handle_failure" so that the attribute lookups don't need to happen every time the task is executed.

It would be nice to consider exactly what should be required of tasks, as a an abstract interface
for 1) as a client, i.e.. the calling interface 2) the server part of the task (e.g. __call__, handle_failure, etc).

There's even some interest in redesigning this completely so that the calling side is separated form the server side, so I guess this would be part of that larger discussion.

ask:
And sorry for responding so late, I've just been swamped with work.  It's definitely not personal or an indication of not finding your patch interesting.  I've been neglecting my friends to the same degree as this patch :(

manihamidi:
hey @kalefranz been looking for exactly such a pattern and am shocked that not only is it not common, but your PR hasn't been merged for 4 years. Is there a good reason for this? e.g., the qeustion you ask in [your SOF post](https://stackoverflow.com/questions/21150760/subclassing-celery-task-for-a-classtask-mixin): 

>Is there any reason that what I'm trying to do is very, very wrong within the Celery framework?

kai3341:
Oh, it seems the idea of task classes is in the air. I've suggested a very similar idea recently: https://github.com/celery/celery/pull/5755

auvipy:
@thedrow @georgepsarakis what do you think about the approach of this PR? can something like this could be considered for celery 4.5 or in celery core at all?

georgepsarakis:
Perhaps I do not fully understand the use case but couldn't the same be accomplised using [Custom Task Classes](http://docs.celeryproject.org/en/latest/userguide/tasks.html#custom-task-classes)? 

Or, to rephrase, what would be the difference between a implementing custom task class and the proposal here?

The following point also resembles the [Instantiation section](http://docs.celeryproject.org/en/latest/userguide/tasks.html#instantiation):

> It's important to note that one of my goals was/is delayed initialization of the instance object. That is, the python standard call to __init__ should be "lazy" and not called when the instance is created. This is important. With a lazy __init__, the user need not worry about "heavy" and duplicitous work being done in __init__ (duplicitous because without lazy __init__ the work would be done in both the synchronous and worker processes).

auvipy:
> Oh, it seems the idea of task classes is in the air. I've suggested a very similar idea recently: #5755

can you explain Goerge's concerns?

kai3341:
@auvipy, okey, but it will be my viewpoint, explaining my way

@georgepsarakis, there are some problems with with Custom Task Classes:
* Custom Task Class instances creates once at application start
* Custom Task Class does not solve boilerplate problem: when i describe job in this class, i also need in starter function. So to solve 1 task I need to create 2 things

Next about #5755
* Custom Task Class contains celery-specific methods, so it's really bad idea to mix controller-level logic (the main job logic) and model-level logic
* I want to register jobs as soon easy as using `@app.task` decorator
* I want to isolate multiple celery *different* instances (with different worker pool) from each of them, so creating global objects like auto-registering metaclasses is really bad idea for me

auvipy:
I have a different plan for you! ping me if I forget!

thedrow:
This PR has become stale. I'm sorry we couldn't get the work done to merge it.
Thank you for your contribution.

PR #1913, opened 2014-03-07
Configuration to disable auto-declare queues in worker

cce:
We found that in our use case — binding to hundreds or thousands of queues, all of which have already been declared by the publisher or by other setup code — skipping the auto-declaration of queues can speed up worker startup time considerably by avoiding unnecessary round trips to call the AMQP `queue.declare` method for each queue.  This helps us get new capacity up and running faster when demand increases.

Does this look good to you?  Any advice or suggestions are welcome, this has only been lightly tested so far.  Usage is to set `CELERYD_DISABLE_AUTO_DECLARE` to True in your celery config.

### DIFF
--- a/celery/app/defaults.py
+++ b/celery/app/defaults.py
@@ -170,6 +170,7 @@ NAMESPACES = {
         'AUTOSCALER': Option('celery.worker.autoscale:Autoscaler'),
         'AUTORELOADER': Option('celery.worker.autoreload:Autoreloader'),
         'CONCURRENCY': Option(0, type='int'),
+        'DISABLE_AUTO_DECLARE': Option(False, type='bool'),
         'TIMER': Option(type='string'),
         'TIMER_PRECISION': Option(1.0, type='float'),
         'FORCE_EXECV': Option(False, type='bool'),
--- a/celery/worker/consumer.py
+++ b/celery/worker/consumer.py
@@ -589,8 +589,10 @@ class Tasks(bootsteps.StartStopStep):
 
     def start(self, c):
         c.update_strategies()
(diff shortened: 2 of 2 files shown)


thedrow:
@cce Ping.
This looks like really cool feature.
I'll understand if you don't have time to work on this any more but if you do please let me know.
@ask Can you please tell me exactly what is missing and what needs to be done?

cce:
HI @thedrow sure go ahead.  We are using this patch in produciton but sounds like there would need to be a matching change to kombu to add an `auto_declare` flag for the Queue type.

mverrilli:
Would it be possible to just assume the queue declaration is done, and only declare on the exception?  I have thousands of queues as well, and I very often get stuck waiting on the declare.

cce:
Well @mverrilli I actually implemented that for kombu (only declaring queue after receiving an exception) in: https://github.com/celery/kombu/compare/3.0...cce:v3.0.23-handle-undeclared-queue

This PR is a two-line change so it is not too hard to give it a try and see if it matches your needs.  The kombu patch I made is a companion to this one and if anyone wants to pick them up and polish them into shape to meet @ask's guidelines that would be fine with me.

ask:
Only declaring on exception was tricky before, because we always had to close the connection on any channel error (slow), now with pyamp we are able to distinguish between recoverable/irrecoverable connection errors and open a new channel instead.

cce:
Hi @auvipy, sorry, I cannot. This was a 6-year-old PR to add 2 lines of code (one to check a bool config value, and another to pass it to a constructor's `auto_declare` argument) but I do not remember much more than that.

PR #1990, opened 2014-04-17
mongodb native serialization and group results meta collection

ocean1:
a little more flexible approach on storing data effectively on mongodb allowing to map-reduce/aggregate later:
-   added a groupmeta_collection option to save group results in a different collection than results
-   added a 'bson' fake kombu encoder to allow pymongo to serialize natively data in mongodb
-   modified the way groups are saved/rebuilt to avoid serializing the whole object but just the id and recreating the AsyncResult object(s) when restoring the group.

### DIFF
--- a/celery/backends/mongodb.py
+++ b/celery/backends/mongodb.py
@@ -20,7 +20,7 @@ if pymongo:
         from bson.binary import Binary
     except ImportError:                     # pragma: no cover
         from pymongo.binary import Binary   # noqa
-    from pymongo.errors import InvalidDocument # noqa
+    from pymongo.errors import InvalidDocument  # noqa
 else:                                       # pragma: no cover
     Binary = None                           # noqa
     InvalidDocument = None                  # noqa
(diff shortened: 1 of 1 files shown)


AlJohri:
Is there any reason this hasn't been backported to the 3.1 release? The part of the code that does

```
if self.serializer in BINARY_CODECS:
    payload = Binary(payload)
```

fixes this issue: https://github.com/celery/celery/issues/2081

AlJohri:
actually, this still doesn't quite fix it

`'result': self.encode(result)` needs to be changed to `'result': result` as shown in https://github.com/celery/celery/issues/2081#issuecomment-70168357

AlJohri:
@ask what are your thoughts on adding a configuration option allowing to store to content as a raw dict instead of doing `self.encode(result)`?

thedrow:
@AlJohri Open a new issue about the configuration option.
We'll discuss it there.

matt-snider:
@AlJohri @thedrow Any update on this, or a new issue where we can discuss it? I definitely think having the raw json in mongodb makes sense,  otherwise there isn't a whole lot of benefit to using mongo in the first place.

PR #2016, opened 2014-04-29
honor hijack setting for root loglevel

aristeinberg:
if CELERYD_HIJACK_ROOT_LOGGER is False, then, in addition to avoiding changing the handler, celery should also not change the root logger's loglevel setting.


ask:
Yeah, we should have a Logging guide in the documentation!

Just need to get someone to work on this.

thedrow:
Thank you for your contribution. I'm sorry we couldn't proceed with merging it.

PR #2054, opened 2014-05-20
Fix Django sample.

SoftwareMaven:
Importing Django's settings module can't happen until after we've 
told it what settings module to import.


ask:
I think it should to import it as long as you don't access any attributes on settings, if they changed
that we would all be in trouble :)

PR #2058, opened 2014-05-25
AsyncResult based on promises

malinoff:


ask:
The backend is async now! I took parts of this code and added some more stuff.

You can do:

  `add.s(2, 2).delay().then(on_success)`

and it works for groups/chains also.

When you use eventlet/gevent there's now a single greenthread responsible for consuming
the results.

We probably need some more deep testing, so the promises based API is on the experimental side for now.

PR #2072, opened 2014-06-04
Properly run all due tasks with `celery.beat`.

### LINKED ISSUE #2071: heapq usage in celery.beat prevents multiple PeriodicTasks being called in the same interval/tick.
Currently `celery.beat.Scheduler.tick()` looks at only the first event on the heap, when what it *should* do is iterate through the heap executing events until it finds an event that is not yet meant to be executed.

The simplest solution I can think of is to add a list, `checked_H`, and pop events off of `H`, staging them in `checked_H` either a future event is found or the heap has been exhausted, at which point we push events from `checked_H` back onto `H`. This hopefully maintains most of the added efficiency of the heap whilst also fixing this issue.

I'm wary of attaching a pull request as I don't understand the necessity `verify is event` check, and thus can't be certain that I'm crafting a proper tick function that doesn't end up just swapping one bug for another. I'd be more than happy to submit a pull request if someone could walk me through the reasoning behind that check.

karanlyons:
Fixes #2071.


thedrow:
I can't merge this as this is failing the tests.
There seems to be an accuracy issue in the scheduler.
If that is fixed. I'll merge.

karanlyons:
This was meant to *solve* the accuracy issues (which, though my memory is very foggy, were that the scheduler failed to dispatch all tasks for a timeslice). But the internals have likely shifted around since the initial request.

I’ll likely have to get this working again in the near term since I’m moving all my dependencies forward on this project, so I’ll see what things are looking like then and hopefully be able to submit a revised patch.

thedrow:
In the meanwhile I'm closing this one since it fails to resolve the issue.
Thanks!
I would really love to see a new patch for this going into 3.2.0.

PR #2110, opened 2014-06-26
Dependency tracking

malinoff:
Initial version of module dependency tracking. I'm not sure about whether I chose a good place to put `dependency_tracker` instance.
Also, I wasn't able to test that code in develop branches (http://pastie.org/9327399).

The report is:
```
software -> celery:3.2.0a2 (DEV) kombu:3.1.0a1 py:3.3.4
            billiard:3.4.0.0a1 py-amqp:2.0.0a1
```


ask:
I guess we cannot merge this now that we have removed the --autoreload flag :(

If we have the resources to bring back the autoreload flag we should include this patch.

Issue #2122, opened 2014-07-05
Implement pattern and matcher arguments for remote control


malinoff:
Hi Ask,

As you said in IRC, this feature may be useful.
Example:

``` python
app.control.ping(pattern=r'.*(frontend|db).*', matcher='pcre')
```

thedrow:
@ask @malinoff Can we look at this again?

malinoff:
I'd like to think about this a little bit more. I don't know why, but I have a though about revising whole remote control...

jheld:
This looks really cool.

If I'm reading correctly, it means if we had N number of worker nodes with a similar name (say a matching prefix), we could specify just that set, without knowing ahead of time their names? So critical-node-0, critical-node-1, etc, then do a glob or regex for that pattern?

If so, is there any reason not to include this feature? It seems there is talk of refactoring remote control/pidbox...but until then, why not add this?

thedrow:
We didn't have the time to implement this.
PRs are welcome.

jheld:
I think this PR implements what is being asked of: https://github.com/celery/celery/pull/5319

auvipy:
> I think this PR implements what is being asked of: #5319

yup seems so. closing for now. thanks

PR #2123, opened 2014-07-05
Added pattern/matcher arguments to control

malinoff:
Refs #2122


ask:
Assigning this to 5.0, if you have the time then it would be great if we could rebase this against master.  AFAIR this also included changes to kombu.

Issue #2154, opened 2014-07-16
In the generic init.d script, su switch without hyphen fails silently.


fpruitt:
In a recent setup, I found that the command output by 
`sudo /etc/init.d/celeryd dryrun`
started my workers properly, while 
`sudo /etc/init.d/celeryd start`
failed silently (workers appeard to start, but no process was actually running).

This is using the generic /etc/init.d script.

thedrow:
@fpruitt ping?

PR #2155, opened 2014-07-16
su should be followed by a hypen when used by root, fixes issue #2154

fpruitt:


thedrow:
@fpruitt Can you please explain why this is more correct or a better approach?

ask:
Security wise: using a login shell when starting a service increases the attack vector considerably.

I  think people should be editing the script manually to do this, but be aware of the consequences.

It's definitely tricky to maintain an init script that is supposed to work on all platforms :(

PR #2217, opened 2014-08-27
pass options to route_for_task

alibozorgkhan:
Passed options can be used in lookup_route too. e.g. I'm trying to implement priority queue support like this:

Passing `priority=True` to `send_task`:
```
app.send_task('support.test', priority=True)
```

Then route requests using a custom Router:

```
class MyRouter(object):
    def route_for_task(self, task, args=None, kwargs=None, options=None):
        if options and options['priority']:
            return {'queue': task + ".priority"}
        return {'queue': task}
```

### DIFF
--- a/celery/app/routes.py
+++ b/celery/app/routes.py
@@ -45,7 +45,7 @@ class Router(object):
     def route(self, options, task, args=(), kwargs={}):
         options = self.expand_destination(options)  # expands 'queue'
         if self.routes:
-            route = self.lookup_route(task, args, kwargs)
+            route = self.lookup_route(task, args, kwargs, options)
             if route:  # expands 'queue' in route.
                 return lpmerge(self.expand_destination(route), options)
         if 'queue' not in options:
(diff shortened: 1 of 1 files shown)


asmodehn:
+1.
I just realized I was trying to do the same and failed...

My goal is to be able to extend the "semantic" of a task run, without modifying/duplicating the task itself.
Actually the arguments I want to add would only affect which queue the request is going in, and as a result which worker would pick it up.
Example : 
- simulated true/false ( would route to worker with simplified implementation, just for simulation purposes before doing the real heavy work. )
- test true/false ( would route to worker with full validation test suite and report ( might be different than simulated and actual workers )

So it s exactly the same task (same code), but one run might be targeted to a different type of worker (different configuration, hardware, etc.)... The routing doesn't depend on the task name, but rather on the intent of the caller of apply-async() ( which might not know/want to care about the queue design/layout - some REST apis also might call some tasks ).

Ideally, I wouldn't want to duplicate my tasks, or add arguments to it.
Also, since any worker can call any tasks, I would like all my workers to import all the tasks.
Only some worker would listen to some queues, so that only the worker with the configuration matching the intent of the task run would actually run it.

Being able to extend the options, and use them for routing, seems like the way to go, but I guess for now I ll implement some kwargs workaround like @cp2587 did...

ask:
Thanks! Merged into master

Issue #2225, opened 2014-08-30
Unable to create chord: "TypeError: group object got multiple values for keyword argument 'task_id'"


aneilbaboo:
On celery master (2fddb876e1), when attempting to run a chord :

```python
from celery import Celery, group

app = Celery()

@app.task
def add(x,y): 
    return x+y

@app.task
def xsum(seq):
    return sum(seq)

c = (group([add.si(1,1), add.si(2,2)]) | xsum.s())
c.delay()
```

The method _apply_chord_incr generates an exception ```TypeError: group object got multiple values for keyword argument 'task_id'``` on the call to header:

```
def _apply_chord_incr(self, header, partial_args, group_id, body,
                      result=None, options={}, **kwargs):
    self.save_group(group_id, self.app.GroupResult(group_id, result))
    return header(*partial_args, task_id=group_id, **options or {})
```

The options dict already has a value for task_id (a different value from group_id).

I tried removing the task_id key from options:

```
def _apply_chord_incr(self, header, partial_args, group_id, body,
                      result=None, options={}, **kwargs):
    self.save_group(group_id, self.app.GroupResult(group_id, result))
    if options.get('task_id'):
        del options['task_id']
    return header(*partial_args, task_id=group_id, **options or {})
```

Which fixes the immediate error, but when I attempt to .get() the result, the program hangs.

I also tried removing the task_id=group_id keyword argument, but still get a hang.

aarcro:
I think I'm seeing this too. My chord with two tasks in the header raises this twice:

    [2015-07-10 20:03:25,658: ERROR/ForkPoolWorker-1] Chord callback '3d0dd6ef-b966-40a5-9aff-213a9e38d8bc' raised: ValueError('3d0dd6ef-b966-40a5-9aff-213a9e38d8bc',)

Which happens because around line 550 in celery/backends/base.py, there are no deps returned.
```
547  	        gid = task.request.group
548  ->	        if not gid:
549  	            return
550  	        key = self.get_key_for_chord(gid)
551  	        try:
552  	            deps = GroupResult.restore(gid, backend=task.backend)
553  	        except Exception as exc:
```

I'm throwing darts, so while I was in the debugger I tried to get GroupResults for the corollation ID, but that failed as well. So I think either that `task.request.group` value needs to be different, or something needs to have gotten in to the backed (redis for me) before we get to this code.

PR #2226, opened 2014-08-30
Fix issue #2225

aneilbaboo:
Creating a chord no longer results in "TypeError: group object got
multiple values for keyword argument 'task_id'".  Chords now
complete without hanging.

https://github.com/celery/celery/issues/2225


thedrow:
Can you please rebase?

thedrow:
@aneilbaboo Ping?

thedrow:
Can you please rebase this so we can merge?

aneilbaboo:
@thedrow Done now, I think.  NB: I haven't worked on this stuff in a long, long time.  I stopped using celery almost a year ago when I didn't get responses from the maintainer for my pull requests.

thedrow:
@aneilbaboo I'm really sorry. The maintainer has been really busy so we teamed up to help maintain this project.

thedrow:
@ask @malinoff What do you guys think?

aneilbaboo:
There was one merge conflict.  Please check that my resolution makes
sense.  I don't really understand this code now -- I merely accepted my
lines over the conflict in the master.

thedrow:
Where exactly is the conflict?

aarcro:
I think the failing test is related to [1e3fcaa969de6ad32b52a3ed8e74281e5e5360e6]
I'd love to see this PR get merged.

aneilbaboo:
The changes are in base.py, 360-366:

<<<<<<< HEAD
        options['task_id'] = group_id
        result = header(*partial_args, **options or {})
=======
        fixed_options = dict((k,v) for k,v in options.items() if
k!='task_id')
        result = header(*partial_args, task_id=group_id, **fixed_options or
{})
>>>>>>> Fix additional issue #2225

My branch is below, the master branch is above.

Looks like there's another commit to master.  I will rebase again.

thedrow:
FYI you're merging instead of rebasing. I'm not gonna be pedant about it but rebases are better when it comes to pull requests because the history is cleaner.
Read https://www.atlassian.com/git/tutorials/merging-vs-rebasing/ if you want to learn more.

aneilbaboo:
I'd be very cautious with these commits.  They are a ducktape and wire fix.


Honestly, after spending days tracing through this code, I came to the
conclusion that the workflow system might benefit from a complete rewrite.

First, I'd start with a set of tests and then work on a module that cleanly
serializes/deserializes workflow.  Currently, it's impossible to test
because the code for serializing the workflow is intermingled in many
places with the code which runs the workflow.  It would also be REALLY
useful if someone would formalize & document the data structures produced
by this system.  There's some ambiguity & duplication of data which I think
contributes to confusion.

Glad someone is taking a crack at this.  The workflow system is a great
idea.  Good luck with it!



On Mon, Jul 13, 2015 at 7:35 AM, Aneil Mallavarapu <aneil.mallavar@gmail.com

aneilbaboo:
I thought I was rebasing.  Sorry, I'm a bit rusty --- I haven't been coding
for a year.

thedrow:
Thanks a lot. The build was fixed so I'm merging this.

thedrow:
@aneilbaboo The build is failing on Python 3. Can you please take a look? https://travis-ci.org/celery/celery/jobs/70858903#L283

aneilbaboo:
Hi Omer, Apologies, I don't have time, and it's been ages since I looked at
this stuff.

aarcro:
I was going to look at fixing this, but tox doesn't fail the same way in master for me locally

thedrow:
@aarcro That's concerning. How does it fail for you? Which OS are you on?

ginoledesma:
I can replicate this on the following OS environments:

Mac OS X with python 3.4.3 and the python:3.4.3-wheezy docker image.

Specifically:

```test_group_to_chord (celery.tests.app.test_builtins.test_chain) ... ERROR
```

From what I can tell, it fails here:

```
def maybe_unroll_group(g):
    """Unroll group with only one member."""
    # Issue #1656
    try:
        size = len(g.tasks)
    except TypeError as err:
        print('TypeError', err, type(g.tasks))
        try:
            size = g.tasks.__length_hint__()
        except (AttributeError, TypeError) as err:
            pass
        else:
            return list(g.tasks)[0] if size == 1 else g
    else:
        return g.tasks[0] if size == 1 else g
```

The map class object doesn't support both `len()` and `__length_hint__()`

ginoledesma:
Ok, after digging it a little bit deeper, the problem is actually introduced in https://github.com/celery/celery/blob/master/celery/canvas.py#L606 by https://github.com/celery/celery/commit/05f84b34b1c7dc6ec1024b12cf32e266736375bc:

```python
def _maybe_group(tasks):
    if isinstance(tasks, group):
        tasks = list(tasks.tasks)
    elif isinstance(tasks, Signature):
        tasks = [tasks]
    else:
        tasks = map(signature, regen(tasks))
    return tasks
```

This can be fixed by replacing it with:
```python
tasks = list(map(signature, regen(tasks))
```

ask:
That will copy the list twice though, for Python 3 compatibility I prefer just using:

    tasks = [signature(task) for task in regen(tasks)]

There's actually another bug here: the app is not passed to the signature.
I tried to find all of the cases where the app is not passed, but probably I did a `ack 'signature\('`, the list comprehension would've avoided that.

ask:
Oh, I didn't realize the `map(signature`, is part of this patch.

The point of using regen in the first place here is to avoid consuming the iterator/generator: this should
only happen as the last step of sending the task messages.

`regen` is a simple wrapper to allow for caching the evaluation so that calling the tasks does not fail
with StopIteration if you `repr` the signature before calling.

thedrow:
@ask I'm not sure I'm following.
What's the fix here?

PR #2238, opened 2014-09-05
 auto retry methods, fixes #1260 

### LINKED ISSUE #1260: Feature: MongoDB AutoReconnect
If MongoDB replica set changes master, next request would fail with AutoReconnect error. Actually it may take several attempts because master elections take several seconds.
As a result, after master reelection celery failes to write result to result backend and task forever remains in running state. Proposed patch wraps mongo conenction in such a way, that requests are repeated if ConnectionFailure occurs. As far as this particular requests are idempotent it should be safe to repeat them.
```
--- celery/backends/mongodb.py  2013-02-11 19:36:47.000000000 +0400
+++ celery/backends/mongodb.py  2013-03-05 15:42:08.219558802 +0400
@@ -9,6 +9,8 @@
 from __future__ import absolute_import
 
 from datetime import datetime
+from functools import wraps
+import time
 
 try:
     import pymongo
@@ -38,6 +40,26 @@
         self.__dict__.update(kw)
 
 
+def auto_retry(f):
+    ''' Forces decorated function to retry if mongo throws error '''
+    @wraps(f)
+    def wrapper(*args, **kwds):
+        retry_interval = kwds.pop("retry_interval", 1)
+        while True:
+            try:
+                return f(*args, **kwds)
+            except pymongo.errors.ConnectionFailure, e:
+                pass
+            time.sleep(retry_interval)
+    return wrapper
+
+def auto_retry_methods(*methods):
+    def decorate(klass):
+        for method in methods:
+            setattr(klass, method, auto_retry(getattr(klass, method)))
+        return klass
+    return decorate
+
 class MongoBackend(BaseDictBackend):
     mongodb_host = 'localhost'
     mongodb_port = 27017
@@ -207,9 +229,16 @@
         return self._get_database()
 
     @cached_property
+    @auto_retry
     def collection(self):
         """Get the metadata task collection."""
-        collection = self.database[self.mongodb_taskmeta_collection]
+        collection = auto_retry_methods(
+            "remove",
+            "find_one",
+            "save",
+            "count",
+            "ensure_index")(
+                self.database[self.mongodb_taskmeta_collection])
 
         # Ensure an index on date_done is there, if not process the index
         # in the background. Once completed cleanup will be much faster
```

buckensl:
fixes #1260


dewaegw:
+1. Been running this patch for a while and it's working fine.

thedrow:
@buckensl Why is the build failing?
Have you changed something in the code that might cause this?

celery/backends/mongodb.py
```diff
@@ -43,6 +46,33 @@ def __init__(self, **kw):
         self.__dict__.update(kw)
 
 
+def auto_retry(f):
+    ''' Forces decorated function to retry if mongo throws error '''
```
thedrow: This should probably be more generic.
I don't think that merging something that is specific to mongo exceptions is useful.

buckensl: Just for the mongo backend indeed. I merged in latest changes from master to resolve the conflicts

ask:
We currently don't have the resources to maintain the mongodb result backend :(

There's no one to review this diff that knows MongoDB well enough, but I guess if you can figure
out why the build is passing we can merge it for 5.0.

georgepsarakis:
From pymongo 3.6 and onwards the `retryWrites` parameter has been added, for details see [here](https://github.com/mongodb/specifications/blob/master/source/retryable-writes/retryable-writes.rst).

From the [documentation](https://api.mongodb.com/python/current/api/pymongo/mongo_client.html?highlight=retry):
> ... retryWrites: (boolean) Whether supported write operations executed within this MongoClient will be retried once after a network error on MongoDB 3.6+. Defaults to False.
> ... Changed in version 3.6: Added support for mongodb+srv:// URIs. Added the retryWrites keyword argument and URI option.

So this option can be passed through the URI.

As far as reads are concerned, I think the caller should be responsible for retrying, handling the exception during a call to `.get`.

How does that sound?

thedrow:
Sounds fine by me.

Issue #2272, opened 2014-09-23
enable_events actually disables some events


kouk:
Hi,
I am writing a patch for [flower](http://flower.readthedocs.org/) and I noticed something unexpected. When flower (or any client for that matter) sends the `enable_events` command this will [add `task` to the list of event groups](https://github.com/celery/celery/blob/7921e12682e896a9a5091d447098653e4ae59a51/celery/worker/control.py#L113). However this list is by default empty, which, according to the documentation and the source code, means that [all events should be sent](https://github.com/celery/celery/blob/7921e12682e896a9a5091d447098653e4ae59a51/celery/events/__init__.py#L222). So for example, the `heartbeat` events will no longer be sent after sending `enable_events`, which is I think totally unexpected. In `flower` this results in workers appearing offline after some time. Probably the `enable_events` handler should check if the group is empty before adding 'task' to the list. Or maybe I'm missing something.

PR #2273, opened 2014-09-23
don't clobber other event types in enable_events. Fixes: #2272

kouk:


jacky--wang:
Just a reminder, the change only cover the method **enable_events**, does the same correction need to applied to the method **disable_events** followed?

Issue #2013, opened 2014-04-28
"Unknown syntax" error in celery amqp basic.publish command


bluebomber:
In following the tutorial [here](http://celery.readthedocs.org/en/latest/userguide/routing.html#hands-on-with-the-api), I get the following:

```
16> exchange.declare testexchange direct
ok.
17> queue.declare testqueue
ok. queue:testqueue messages:0 consumers:0.
18> queue.bind testqueue testexchange testkey
ok.
19> basic.publish 'This is a message!' testexchange testkey
unknown syntax: "basic.publish 'This is a message!' testexchange testkey". how about some 'help'?
```

I'm using v3.1.11 (Cipater)

starplanet:
In following the tutorial here, I get the following:
11> basic.publish "This is a test message!' testexchange testkey
`string indices must be integers, not unicode`

I use redis as broker and backend. My environment is:
ubuntu 14.04, celery 3.1.19

PR #2334, opened 2014-10-24
Interpret basic.publish text as a Message object

jlynn:
py-amqp expects a Message object for Channel.basic_publish.
https://github.com/celery/celery/issues/2013


thedrow:
@jlynn librabbitmq

ask:
Yeah, we need this to choose str/Message based on the transport used.  Please resubmit it with that as this is currently low priority ATM with the resources we have available.

PR #2392, opened 2014-11-28
Fix variable name in Task Cookbook tutorial

Ch00k:


malinoff:
Thank you! Could you please also add your name to https://github.com/celery/celery/blob/master/CONTRIBUTORS.txt?

PR #2399, opened 2014-12-03
celery.contrib.conflator: add a Conflator task subclass

aidan:
Sometimes life gives you a periodic task which can get backed up
and you only really want one of them to execute and the others are
effectively noops (for instance because they'd poll a remote system
for updates which had just been retrieved). This task type will
not execute any tasks which were scheduled before the most recent
execution occured.

With a memory cache backend this will only apply to a single worker
however if memcached is configured then any pool of workers with
access to the same memcache will be able to share this information.

### DIFF
--- a/celery/contrib/conflator.py
+++ b/celery/contrib/conflator.py
@@ -1,0 +1,68 @@
+# -*- coding: utf-8 -*-
+"""celery.contrib.conflator
+======================
+
+.. versionadded :: 3.2.0
+
+A task which will check that it's enqueued time is after the last time
+a task with the same conflation ID was executed to clear backed up
+work that's already been done.
+
+This is useful for e.g. periodic tasks which occasionally take longer
+than their period to execute such as checking in with a remote system
+which may be down, experiencing performance issues or so on.
+
+It's important that things should work out satisfactorily at task
+execution time if all previously scheduled tasks execute or if they
+have all been conflated into 1.
+
+**Simple Example**
+
+.. code-block:: python
+
+    # sync from a remote system
+    @app.periodic_task(base=Conflator, run_every=timedelta(minutes=1))
+    def update_from_remote():
+        from remote_system import update
+        update()
+
+"""
+
+from datetime import datetime
+
+from celery import states, Task
+from celery.utils.log import get_logger
+
+logger = get_logger(__name__)
+
+class Conflator(Task):
+
+    abstract = True
+
+    conflation_key = None
+
+    def apply_async(self, *args, **kwargs):
+        conflation_key = self.conflation_key
+        if conflation_key is None:
+            conflation_key = kwargs.get('conflation_key', 
+                                        self.__class__.__name__)
+        pending = self.backend.get_task_meta(conflation_key)
+
+        # If there's a pending task just skip it
+        if not pending['result']:
+            self.backend.store_result(conflation_key, True, 
+                                      "conflating")
+            return super(Conflator, self).apply_async(args, kwargs)
+        else:
+            logger.info("Conflating task with key %s", conflation_key)
+
+    def __call__(self, *args, **kwargs):
+
+        conflation_key = self.conflation_key
+        if conflation_key is None:
+            conflation_key = self.__class__.__name__
+
+        # always clear the conflation key so tasks that error out
+        # don't block the queue
+        self.backend.forget(conflation_key)
+        return super(Conflator, self).__call__(*args, **kwargs)
--- a/celery/tests/contrib/test_conflator.py
+++ b/celery/tests/contrib/test_conflator.py
@@ -1,0 +1,111 @@
+import celery.contrib.conflator
+from celery.contrib.conflator import Conflator
+from celery.tests.case import AppCase, depends_on_current_app
+from celery.tests.tasks.test_tasks import TasksCase
+from celery.utils.log import get_logger
+
+logger = get_logger(__name__)
+
+
+class test_ConflatingTask(TasksCase):
+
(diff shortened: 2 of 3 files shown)


thedrow:
Overall this looks good.
Please address the review remarks and I'll get this merged.

aidan:
@thedrow thanks for those, using the results backend is definitely a cleaner approach. I've addressed those now.

thedrow:
Ok LGTM.
@ask @malinoff @PMickael What do you guys think?
If I don't get a response until the end of the week I will merge this.

thedrow:
@aidan Oh do you mind adding some documentation to https://github.com/celery/celery/blob/master/docs/userguide/periodic-tasks.rst?
Also the docstrings should mention that this is added to 3.2.0 using the ```.. versionadded :: 3.2.0``` Sphinx notation.

aidan:
@thedrow I added it as a section in tasks.rst as it's not strictly related to periodic tasks (that's just the use case which I've needed this most frequently in).

ask:
The big problem with this is that wall clock time is not reliable in a distributed system, so this can only be 100% reliable if everything is executing on the same machine.   This is also the reason why the enqueue time is not part of the message. 

Not that a unreliable version cannot be useful, but it could be worth investigating a solution using a mutex on send and receive.

aidan:
Reliability hasn't been a concern really: it's been intended for speeding up system recovery from temporary conditions more than anything else.

I'll have a look at rejigging this with a mutex

thedrow:
@ask What if we explicitly say that this won't work unless ntp has been set up?
How would you suggest to do it otherwise? I mean what kind of mutex do you have in mind?

ask:
You can round to the nearest 10 seconds maybe, and have a warning that NTP must be used for clients and workers.  Rounding could be configurable so the daring could round to the second.

aidan:
I think we can just get rid of the timestamp entirely and replace it with a scheduled flag on the conflated key in the result backend which is more reliable and prevents tasks which will be conflated   entering the queue at all.

thedrow:
Sounds good.

aidan:
Stuffing a key into the backend works well, I couldn't figure out how to terminate the task from __init__ so it'll still clog up the queue and cause a bit of worker churn. Not a huge issue unless there's a lot of them (in which case you probably have other problems) but it'd be nice to get rid of it earlier.

thedrow:
What happens if you call self.reject from __init__?

ask:
You cannot call reject from `__init__`.  Instantiation of the task happens when the task is registered, since the task class is used both for sending messages and executing a request.

thedrow:
@aidan Given what @ask said I don't understand why are you using ```__init__``` to conflate the task.

aidan:
@thedrow yeah, i think that's probably wrong - I always forget the task instance gets created at least twice. I think moving it to delay is the right thing but haven't had time to try that yet.

mverrilli:
Not sure if this helps, but what I do for this is create a queue to process these tasks with a TTL just short of the period.  Works with RabbitMQ at least.

thedrow:
@mverrilli Queues TTL are an extension in RabbitMQ and we need to use the management API to ensure they are enabled.
That might happen at some point but not right now.

thedrow:
@aidan LGTM.
@ask @malinoff Any objections?

celery/contrib/conflator.py
```diff
@@ -0,0 +1,68 @@
...
+
+class Conflator(Task):
+
+    abstract = True
+
+    conflation_key = None
+
+    def apply_async(self, *args, **kwargs):
+        conflation_key = self.conflation_key
+        if not conflation_key:
```
malinoff: It is better to explicitly check `is None` when you want to check if it is None.

celery/contrib/conflator.py
```diff
@@ -0,0 +1,68 @@
...
+            self.backend.store_result(conflation_key, True, 
+                                      "conflating")
+            return super(Conflator, self).apply_async(args, kwargs)
+        else:
+            logger.info("Conflating task with key %s", conflation_key)
+
+    def __call__(self, *args, **kwargs):
+
+        conflation_key = self.conflation_key
+        if not conflation_key:
```
malinoff: Same as previous comment.

malinoff:
I also don't like `task_count` global variable in tests. Can you explain its purpose, @aidan?

thedrow:
@aidan ping?

aidan:
@thedrow sorry, been tied up with work - will address code review comments tomorrow as I'm on holiday & have a long train journey ;)

aidan:
@the_drow I've addressed the comments from @malinoff (explict is None and replaced the global task_count with a reference to the test class in the task constructor)

thedrow:
I don't think that the failure is related to you but just in case can you please rebase?

aidan:
@thedrow I rebased it with master before pushing

tko:
I don't want to prolong this PR more than it already has but I wonder how to best handle very similar but not identical case as implemented here. I'd say the main difference is that my tasks are not regularly scheduled / repeated but rather triggered on external (non-repeating) events.

As I gather with `Conflator` new tasks that get queued while a task has already been queued _or is already running_ simply gets ignored.

What I'd like to do instead is that if a task with identical arguments has already been queued it gets ignored, but if one is already running then new task is queued so that it gets run (at most) one more time.

thedrow:
This contribution has gone somewhat stale. Sorry about that.
@aidan If you feel like adapting the code for Celery 4 please open a new pull request.

PR #2400, opened 2014-12-03
Fixed settings load order

pennersr:
Fixed loading settings before setting environment variable pointing to settings


malinoff:
Thanks! Could you please add your name in https://github.com/celery/celery/blob/master/CONTRIBUTORS.txt?

PR #2410, opened 2014-12-10
filter out latest messages for all tasks, not just the one being queried

tweedmorris:
Hello!
I was playing around with the RPC backend and noticed that groups of tasks were generally failing in interactive mode. Easy test case using classical add() task with RPC backend and AMQP storage:

`>>> t1 = tasks.add.delay(1,2)`
`>>> t2 = tasks.add.delay(3,4)`
`# ... wait a few milliseconds for both tasks to complete ...`
`>>> print(t1.ready(), t2.ready())`
`True False`
`# ohno!`

t2 will never report ready() despite having executed, and worse, a request to t2.get() will now hang indefinitely. In general, only the first task from any batch that is queried for its status will have a sane status.

The culprit is get_task_meta() in the amqp backend, inherited by the rpc backend. Since the amqp backend uses one channel per task, it safely assumes that it can discard all queued messages except the last one targeted to task_id.
The rpc backend cannot afford such profligacy: any time a task metadata is queried, all pending messages for other tasks are discarded, and subsequent metadata requests by those tasks will wait forever for a message that will never come, until perhaps its 128-bit GUID is duplicated some trillions of years hence.

This patch provides a suitable get_task_meta() implementation for amqp. I would be thrilled if you found it suitable.

cheers
Morris

### DIFF
--- a/celery/backends/rpc.py
+++ b/celery/backends/rpc.py
@@ -62,3 +62,41 @@ class RPCBackend(amqp.AMQPBackend):
     @cached_property
     def oid(self):
         return self.app.oid
+
+    def get_task_meta(self, task_id, backlog_limit=1000):
+        # Polling and using basic_get
+        with self.app.pool.acquire_channel(block=True) as (_, channel):
+            binding = self._create_binding(task_id)(channel)
+            binding.declare()
+
+            # Discard all but the latest message per task id
+            latest_by_id = dict()
+            prev = acc = None
+            for i in range(backlog_limit):  # spool ffwd
+                acc = binding.get(
+                    accept=self.accept, no_ack=False,
+                )
+                if not acc:  # no more messages
+                    break
+                _id = acc.payload.get('task_id')
+                if _id:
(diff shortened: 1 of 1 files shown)


ask:
Did you test this on the master branch or the stable version?

I'm not able to reproduce there, and I fixed some serious problems with the rpc result backend
that are still present in 3.1.x.

In master it's now keeping messages received out of band in a buffer so it does not throw them away.

>  and subsequent metadata requests by those tasks will wait forever for a message that will never 
> come, until perhaps its 128-bit GUID is duplicated some trillions of years hence.

lol :)

ask:
My version keeps all of the messages, not just the most recent one.  Maybe you could patch the new version of AMQPBackend.get_task_meta to do the same?

ask:
I just realized this pull request is extremely old.  I'm sorry for that, it's not personal I just have been unable to keep up and we don't have a lot of resources.

I'm going to pull this in as your commit but make some changes.

ask:
I fixed it up and merged it, and you commit is now: e90444189b8bd177612066eacb632afb2d70445a

PR #1888, opened 2014-02-24
Beat failover

magmax:
Related to the issue celery#1495

This is another approach for the celery beat failover: use a queue as mutex to allow/disallow the beat execution.

I created the FailoverService class in another file in order to reuse it if necessary in other services.

There are some issues to be solved:

- Make the check timeout customizable.
- Make the queue name customizable.
- Build automatic tests (I have no idea how to check this).
- There can be two near beats whenever the failover happens, because the failover beat will not wait the beat time before getting the master token and launch its own beat.

I checked it this way:
1. Launch a beat.
2. Clone the directory and launch another beat.
3. Shutdown the first one. The second one should start to work.
4. Launch the first one again.
5. Restart Rabbitmq. Any of them should get the mastery (and just one of them).

Any comment is welcome.

Issue #1495, opened 2013-08-01
Failover for celerybeat


kerwin:
I want to make schedule process run in multiple nodes to improve availability, for reducing single point issue.

thedrow:
Related to #251.

ask:
Closing this, as we don't have the resources to complete this task.

magmax:
Sorry, but I do not understand this.
The proposed solution relies on RabbitMQ exclusive queues, what works as a distributed Mutex used to have a semaphore during the bootstrap. So, testing this is just like testing RabbitMQ.
In addition, this request would allow HA for Celery beat.

This has just a drawback: if the current beat server sends a beat and immediately dies, the semaphore-awaiting new beat server will send the beat again, no worrying about the beat time period. This should be another problem, that can be solved by avoiding to store anything in localhost, but this should be another issue, in my opinion.

We have had this patch working in production for a couple of years without any problem.

Anyways, I think closing an issue about HA just because there are no resources is not a good deal.

ask:
If there's a pull request for the issue then that will stay open.

I think I have looked into the pull request you refer to, using exclusive queues, but I have not
yet had the time to verify that it works.   I was not aware that they are considered a sound implementation of a distributed mutex?  We would need to run a stress test on this implementation, to make sure (at the very least serious) edge cases are covered.

That you have been using this in production for years is certainly a very sound data point.  Have you tested it under stress? Killing nodes in different scenarios seeing how it recovers?
E.g. one thing I'm concerned about is that beat, by nature of being a scheduler, is dependent on a
single clock source to be consistent.  If there's failover it means there are multiple clocks, which means that periodic tasks can be sent in duplicate.  We need to figure out, what in a best case scenario with NTP synced clocks, and a maximum clock skew the maximum precision a periodic task can have before duplicates are sent.

The problem is that we have so many issues, but nobody to work on them.  Many of them we don't even know if is still an issue, or maybe if it's already fixed in master for the 4.0 release.  If the only progress is dead silence, then we can just as well close it.  In this case I knew there's at least one open pull request attempting to solve the issue.

I really would like to have failover for Beat, and it's definitely possible,  but we don't have any resources to put into it.  The only time I have is the weekend, but I also have to pay the rent and have a son to raise, so I have closed most of the feature request threads. 

I'm also open to reopening any of the issues that I have closed, you all have the ability to influence this!

magmax:
Thank you for the explanation. It sounds really coherent.

[...]

PR #2414, opened 2014-12-11
High available celery beat

javrasya:
This is related with the issue #1495 and PR #1888 . In PR #1888 requested by @magmax is still opened. 

I guess It is not accepted because it does not support other than `AMQP`. Actually the solution can not be implemented for other types of broker easily. 

In my solution, there is a method used like in broker backends. There will be many implementation as they are developed. There is implementation only for AMQP at the moment. I added Redis strategy to be implemented as soon.  The solution in @magmax 's PR is helped me a lot to implement it for AMQP. 

I saw, tasks are applying in schedulers. So I added a scheduler which is `HAScheduler`(High Available Scheduler). This one is checking whether current node is master or not according to chosen strategy. If current node is master defined by the chosen failover strategy, normal way is used. There is no change for that case. But, if the node is slave defined by the chosen failover strategy, beat will think, task is applied, but it won't actually be applied. This scheduler should be used to have high available celery beat instead of default `Scheduler` class.

To decide what strategy is used, broker backend is used by parsing `BROKER_URL` as default. This can also be configured with a celery config which is `CELERYBEAT_FAILOVER_BROKER_URL`. So you can use different broker for your celery cluster and beat failover strategy. If you don't give, it'll be using BROKER_URL.

Now, you can run multiple celery beat as you want. One of them will be master. If master goes down, one among still up an running slave beats, will be new master by taking the throne with the way defined in the chosen failover strategy. 

I thought, beat code is not structured well. Because of failover strategy and its factory class, I just moved **celery/beat.py** into **celery/beat/**init**.py** and I implemented all these codes under **celery/beat** module. It can still be accessed by `from celery import beat`or `from celery.beat import *`. I hope that doesn't  cause another problem. If it does, I can change beat module name which I created.


bubenkoff:
@javrasya curios what is the reaction from project collaborators?

thedrow:
This needs to be rebased.
@ask @malinoff @PMickael Care to review?

javrasya:
@bubenkoff , There is none unfortunately. I had to use this improvement as external third-party in my projects.

bubenkoff:
@javrasya, can you please clarify how exactly?

thedrow:
@javrasya Please rebase. I'll try to ensure someone looks at it.

magmax:
Rebased

thedrow:
How come it still has merge conflicts then?

magmax:
Oh, sorry. I didn't realize there were more commits.

magmax:
Now I've merged master on 811a122

thedrow:
@magmax There are still conflicts...

celery/beat/__init__.py
```diff
@@ -26,14 +26,14 @@
...
 
-from . import __version__
-from . import platforms
-from . import signals
-from .five import items, reraise, values, monotonic
-from .schedules import maybe_schedule, crontab
-from .utils.imports import instantiate
-from .utils.timeutils import humanize_seconds
-from .utils.log import get_logger, iter_open_logger_fds
+from .. import __version__
```
ask: The celery source code actually never uses more than one level of relative imports (so no more than one dot).
It used to, but it ended up being a mess

celery/beat/__init__.py
```diff
@@ -26,14 +26,14 @@
 from kombu.utils import cached_property, reprcall
```
ask: Converting a module to a package (or the other way around) can cause problems if the user installs the new version on top of the old one (since the old module will not be deleted, and .pyc files may also exist).

So something should be added to https://github.com/celery/celery/blob/master/setup.py#L17-L46, but upgrading instead of downgrade_packages.

bernieke:
I've just forked celery (from 3.1.15) to get this to work. (We need it kind of urgent, so couldn't wait for this to get released.) My work for this is here: https://github.com/Awingu/celery/compare/v3.1.15...v3.1.15-awingu

I used the messy version (https://github.com/magmax/celery/commit/d0ab5492b54b80eaf0fd062416d7823b3f60823e) since it was a lot less work to implement.

But I ran into a problem where when a slave takes over from the master, it would relaunch all jobs which had been scheduled since the slave had started.
I realized this is because the scheduler (PersistentScheduler) was still separate between the beat instances, so the slave wouldn't know about the last run times from the master.
So then I switched to using the DatabaseScheduler from djcelery, but had the same problem there. Looking at the code this appears because the database is only read for the last run times at initialization of the scheduler. (Which is fine as long as you only have one beat instance off course.)
So I added a call to self.scheduler.setup_schedule() after the mutex stuff. (See the diff on our fork I linked at the beginning.)

About half way through this process I also tried the complete patch from this pull request (manually applied against 3.1.15), but couldn't get this to work at all.
And I simply don't get how the scheduler would work either, or how the problem I was facing would be handled (the necessity of sharing the last run times), especially since the HAScheduler takes the place of any DatabaseScheduler you might otherwise have used? I even tried to change the inheritance of the HAScheduler to DatabaseScheduler instead of the base Scheduler, but that didn't help me much either.

So unless I'm really overlooking something here, I've got a feeling that an important piece is missing here? (A way to keep the last run times in sync between the master and slaves.)

mrochford:
This would great to have celerybeat somehow mingle with each other.  Any movement on this open issue would be fantastic.

ask:
@bernieke This is the reason why the patch isn't merged yet.  We haven't been able to verify that the failover works in the edge cases, and I currently don't have time to do so :(

bernieke:
Well, in case anyone is still interested in this, we've recently rebased our changes to 3.1.19: https://github.com/Awingu/celery/compare/v3.1.19...v3.1.19-awingu

We've been using our fork in production for the past year, and haven't had any problems with it. Our product is an unmanaged HA solution, so a lot of edge case testing goes into QAing it. (Hard powering off nodes and such.) And so far we haven't had any problems with our fork of celery.

I did notice we're importing our lock solution from our common library (we use this in other parts of our product as well), which isn't publicly accessible, so here it is:

`
from datetime import datetime, timedelta

from django.db import models, transaction, IntegrityError


class Lock(models.Model):
    name = models.CharField(max_length=127, unique=True)
    created = models.DateTimeField(auto_now=True)

    class Meta:
        verbose_name_plural = 'Locks'

    def __unicode__(self):
        return self.name


class LockError(Exception):
    """Exception thrown when the requested lock already exists."""

    pass


class locked(object):
    """A context manager to add a distributed mutex."""

    def __init__(self, name, timeout):
        self.name = name
        self.lock = None
        self.timeout = timeout

    def __enter__(self):
        # first delete any expired locks
        expired = datetime.utcnow() - timedelta(seconds=self.timeout)
        Lock.objects.filter(name=self.name, created__lte=expired).delete()
        # then try to get the lock
        try:
            Lock(name=self.name).save()
        except IntegrityError:
            transaction.rollback()
            raise LockError('Could not acquire lock: {0}'.format(self.name))

    def __exit__(self, *args):
        Lock.objects.filter(name=self.name).delete()
`

ask:
Thank you @bernieke,  I'm sure someone here will find it very useful!   Sadly, we would want to avoid depending on an SQL database for core Celery, but maybe it could be used for the Django database scheduler.

I wish I had time to verify any of the solutions we have open in pull requests right now, as  I regularly lose sleep over neglecting the people who contributed them.   But it will probably not be until after the 4.0 release at least.

jonashaag:
FWIW, we've solved this by starting multiple https://github.com/ybrs/single-beat wrapped beats. single-beat makes sure only one beat process is running at the same time.

You could also use uWSGI's [legion mode](http://uwsgi-docs.readthedocs.io/en/latest/AttachingDaemons.html#legion-support).

thedrow:
This PR has become stale. I'm sorry we couldn't get the work done to merge it.
We will consider implementing this in the near future.
Thank you for your contribution.

PR #2435, opened 2014-12-28
This makes action=append with --arg 1 --arg 2 work

sontek:


frgtn:
This looks like a useful thing to have.

sontek:
@ask Ping, any comments on this?

ask:
There may be more places where this is not supported, as we have a few argument parsers (eek!)
and in all my years of UNIX traveling I could swear I've never seen this format used before.

PR #2446, opened 2015-01-07
Update next-steps.rst

bjdixon:
Added note about running the worker in the directory above proj. If the worker is run within the proj directory resolving the error isn't intuitive from the resulting message


ask:

Issue #2453, opened 2015-01-14
Celery not closing database connections during worker init for django 1.6, 1.7


lukeburden:
Hi there,

Quick details:
- django 1.7 (checked with django 1.6, is the same)
- latest celery release - 3.1.17

A django web application with simple functions defined as tasks using @app.task, only using primary keys and primitives as arguments.

Running celery with concurrency of 4 in prefork mode.

When the celery workers are initialised, the django fixes are applied. This runs through and attempts to close existing database connections in each worker. The only problem is, the fixup step attempts to find connection close functions by iterating over django's ConnectionHandler class at django.db.connections:

(in celery/fixups/django.py:DjangoWorkerFixup:_close_database)
funs = [conn.close for conn in self._db.connections]

In django 1.6 and 1.7 (I didn't check earlier) this will always fail as it's just trying to find attribute 'close' on the conn variable, which at this point is a string alias to a database, ie, 'default'.

So for django 1.6 and 1.7 execution will always jump to the attribute error catch and django.db.close_old_connections will be called for every worker. This does not close the connections as desired because the connection is neither old nor obsolete. As a result, every worker shares the same database connection. Hilarity ensues. :)

I've thrown together a pull request with a simple fix that will work with django 1.6 and 1.7. Will attach it shortly. Let me know if you need more info or if my findings are off!

Best regards,

Luke

ionelmc:
Merged in master and cherrypicked it for 3.1 branch (I botched the author, sorry).

bimusiek:
Just for someone in future that search for bug we had, it was connected with this issue, however we dont know why.

We were getting error from Celery about `"Socket operation on non-socket"` when connecting with django orm to postgresql. 

After updating to 3.1.18 everything started to work. :+1:

PR #2454, opened 2015-01-14
Fixes issue #2453

lukeburden:
Please see issue #2453 for a description of the problem.

### DIFF
--- a/celery/fixups/django.py
+++ b/celery/fixups/django.py
@@ -227,7 +227,7 @@ class DjangoWorkerFixup(object):
 
     def _close_database(self):
         try:
-            funs = [conn.close for conn in self._db.connections]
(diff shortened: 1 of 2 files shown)


ionelmc:
For posterity, this is an issue only when [CONN_MAX_AGE](https://docs.djangoproject.com/en/1.7/ref/settings/#conn-max-age) is used.

ionelmc:
Thanks for this and the bug report.

ionelmc:
Can you also make a PR adding yourself to CONTRIBUTORS.txt ?

ionelmc:
Actually I already made the change, make a PR if I misspelled your name.

lukeburden:
Looks good, thanks for that and you're very welcome.

PR #2459, opened 2015-01-18
Fix task.__wrapped__ to work properly with inspect.Signature

malinoff:
```python
import celery
app = celery.Celery()
@app.task
def add(x, y):
    return x + y
```
Currently if you try to create a signature of ``add`` task, you will receive this:
```python
In [1]: from tasks import add; import inspect

In [2]: print(inspect.signature(add))
(y)

```

With this patch, it works as expected:
```python
In [1]: from tasks import add; import inspect

In [2]: print(inspect.signature(add))
(x, y)

```

### DIFF
--- a/celery/app/base.py
+++ b/celery/app/base.py
@@ -279,15 +279,16 @@ class Celery(object):
         base = base or self.Task
 
         if name not in self._tasks:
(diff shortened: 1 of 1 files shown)


ask:
Great!

PR #2489, opened 2015-02-06
bug fix for couchbase backend

amg84:
Converting key from bytes to str, using the kombu bytes_to_str util, as the couchbase lib raises a ValueFormatError when a byte string is supplied.


wil-langford:
It looks like I failed a Travis CI check, but it also looks like every build since the 1st of this month has failed the same check.

thedrow:
@Smirl We'd love to see a PR on this and since @amg84 did not respond and it's not the correct fix I'm closing this PR and opening issue #2654 to ensure we won't miss this before releasing the next version.

Issue #2496, opened 2015-02-11
celerybeat init script incorrectly assumes existence of bash


exelian:
The generic-init.d celerybeat script starts out with `#!/bin/bash`, which incorrectly assume bash is available from that location.

PR #2497, opened 2015-02-11
Update celerybeat to use sh instead of bash

exelian:
Updated the init script to work without assuming bash is always available from /bin/bash/
Also see issue #2496


ask:

PR #2500, opened 2015-02-12
Remove ROOT_DISCOURAGED warning

oeuftete:
This warning creates noise in valid root-using scenarios like docker
containers.  Having users specify C_FORCE_ROOT is sufficient
discouragement from running celery as root.

### DIFF
--- a/celery/tests/bin/test_worker.py
+++ b/celery/tests/bin/test_worker.py
@@ -287,22 +287,6 @@ class test_Worker(WorkerAppCase):
             worker = self.Worker(app=self.app)
             worker.on_start()
             _exit.assert_called_with(1)
-            from celery import platforms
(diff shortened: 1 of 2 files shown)


celery/tests/bin/test_worker.py
```diff
@@ -275,36 +275,6 @@ def test_unknown_loglevel(self):
         self.assertEqual(worker1.loglevel, 0xFFFF)
 
     @disable_stdouts
-    @patch('os._exit')
```
oeuftete: Realizing I gutted too much here.

malinoff:
Although I agree that there are valid use-cases for running celery worker as root, I disagree that we should completely remove this warning and strongly recommend to use [-W](https://docs.python.org/3/using/cmdline.html#cmdoption-W) option for python interpreter.

PR #2511, opened 2015-02-20
Optimized Redis result backend (publish/subscribe instead of polling)

### LINKED ISSUE #799: Feature: Redis backend optimizations
- .get/.join should use SUBSCRIBE to avoid polling
- GroupResult should use incr to count number of failed/successful tasks.

sangwa:
A set of improvements for Redis result backend aimed to eliminate polling as much as possible and make it responsive for very fast tasks (with execution times under 0.5 seconds - the default polling interval used before).

* If the initial retrieval of the task result determines its state as not ready yet, the new backend subscribes to notifications sent upon setting the task result, instead of polling Redis on the fixed intervals. As soon as notification is received, the result is returned.
* To eliminate the possibility of missing the published notification in a short window between retrieving the initial task state and subscribing to notifications, locks for the task results are introduced. The lock is placed before the set operation and released after publishing the notification, also the lock is placed before retrieving the initial task state and released after subscribing to notifications.
* To avoid blocking on these locks if some operation crashes in the progress, lock expiration timers and tunable settings for them are introduced. Millisecond precision expiration is used if supported by Redis server.
* All batch operations (set/publish, get/subscribe, lock acquisition/release) use either lua scripts or pipelined commands depending on the support of lua scripts by Redis server.
* Feature detection for Redis server is implemented to support the functions listed above.

Should close #799 (GroupResult improvements are already implemented in redis_join branch as far as I can see and are supported in this version by re-implementing RedisBackend.get_many()).

### DIFF
--- a/celery/backends/redis.py
+++ b/celery/backends/redis.py
@@ -204,7 +204,7 @@ class RedisLock(object):
             return 0
         end
     """.format(ACQUIRED_SIGNAL)
-    ACQUIRE_SCRIPT_SHA = sha1(ACQUIRE_SCRIPT).hexdigest()
+    ACQUIRE_SCRIPT_SHA = sha1(ensure_bytes(ACQUIRE_SCRIPT)).hexdigest()
 
     RELEASE_SCRIPT = """
         local token = redis.call('get', KEYS[1])
@@ -221,7 +221,7 @@ class RedisLock(object):
             return 0
         end
     """.format(RELEASED_SIGNAL)
-    RELEASE_SCRIPT_SHA = sha1(RELEASE_SCRIPT).hexdigest()
+    RELEASE_SCRIPT_SHA = sha1(ensure_bytes(RELEASE_SCRIPT)).hexdigest()
 
     def __init__(self, client, name, expire=None, timeout=None,
                  thread_local=True, signal_key=None, **kwargs):
@@ -347,7 +347,7 @@ class RedisLock(object):
                                      args=(token, self.expire or ''))
         pipe = self.client.pipeline()
         pipe.watch(self.name)
-        current_token = pipe.get(self.name)
+        current_token = ensure_bytes(pipe.get(self.name))
         if current_token != token:
             # lock is already held by someone else
             pipe.reset()
@@ -455,7 +455,7 @@ class RedisBackend(KeyValueStoreBackend):
         end
         redis.call('publish', KEYS[1], ARGV[1])
     """
-    SET_SCRIPT_SHA = sha1(SET_SCRIPT).hexdigest()
+    SET_SCRIPT_SHA = sha1(ensure_bytes(SET_SCRIPT)).hexdigest()
 
     #: redis-py client module.
     redis = redis
--- a/celery/backends/redis.py
+++ b/celery/backends/redis.py
@@ -643,6 +643,18 @@ class RedisBackend(KeyValueStoreBackend):
         script.sha = self.SET_SCRIPT_SHA
         return script
 
+    def _store_result(self, task_id, result, status,
+                      traceback=None, request=None, **kwargs):
+        store_result = super(RedisBackend, self)._store_result(task_id,
+                                result, status, traceback=traceback,
+                                request=request, **kwargs)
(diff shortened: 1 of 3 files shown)


thedrow:
The PyPy failure is not related to you.

thedrow:
LGTM. @malinoff @ask Any objections?

malinoff:
@thedrow I can't help with reviewing non-amqp backends - so I think @ask has the final word here.

randylayman:
Its been a while since there was any activity on this PR?  Is this dead?

thedrow:
I've marked this for 4.0 if possible.
Let's see if @ask responds. If he doesn't, let's rebase and merge this.

ask:
Great work!  The only reservation I have here is the frightening combination of having "redis" and "lock" appear in the same sentence :)

I realize that you have spent some time on the lock implementation, but I still would prefer if we didn't use it due to the difficulty of proving that it will work in all cases.

Instead of using the lock, would it be possible to hook into Backend.on_task_call to subscribe before the task is sent?  The rpc:// backend uses this to make sure the queue is declared before the task message is published.

ask:
Since I recently refactored the backends to have an async interface, it means that we can now keep the pubsub instance running instead of creating/destroying it for every task.

I wrote an implementation of this in  1f995b3, it seems to work but I have only tested it naively.

I don't have much experience with redis pubsub, so would be great to have it reviewed.

thedrow:
@sangwa @randylayman Can you guys take a look?

yingzong:
For those just looking for applying this change (3 commits) without waiting for the eventual rebase/merge, I pulled it out of that foreign branch as a patch:
https://gist.github.com/yingzong/4d07ed37c0d9c0a35331

I've been running this patch back-applied to 3.1.18 and it runs fine. The waiting time of a simple Redis-backed Chord I'm running has gone from 1s to about 50ms (which includes actual work). I had to also apply `CELERY_DISABLE_RATE_LIMITS=True`; otherwise it was still in the hundred ms range, see:
https://github.com/celery/kombu/issues/94

Another nice thing this set of changes seems to make is that results are deleted immediately from Redis instead of hanging around for `CELERY_TASK_RESULT_EXPIRES` seconds.

randylayman:
After a read this looks good.  Unfortunately I don't have the time to test and dig in deeper.

Does the comment by @yingzong need to be documented?

ask:
@yingzong 

`CELERY_DISABLE_RATE_LIMITS` should have absolute no effect anymore in 3.1 when using redis/amqp as the broker.

The task results cannot actually be removed if this is a persistent result backend, at the very least not by default.

ask:
Clarification: disabling rate limits will have no effect if you don't have any rate limited tasks.  As there's no dedicated thread for rate limits any more, there's no longer any overhead.

yingzong:
@ask I just checked again, and you're right on both counts. I must have misremembered some behavior. Thanks for clarifying.

ask:
I merged the async backend version, not using locks into master as it's now passing the stress test suite.

@sangwa Even though I didn't merge your commit, I consider this a collaboration as I used your diff as a guide.  Thank you for this, it's a feature I have been wanting for a long time.
Could you please add your name to https://github.com/celery/celery/edit/master/CONTRIBUTORS.txt ?

ask:
The changes merged into master are: 886faf6754a7da806f403df6005914649a19b955 + 456fd752dd1184f76d32935679790eff854b80e1

sangwa:
@ask Thank you for the feedback! I haven't been working with Celery for quite some time, so I didn't follow the new async backend and didn't have enough spare time to review and/or test your changeset thoroughly, but at the first quick glance it seems fine. I trust other contributors to this issue ensured it's working properly.

jheld:
Hello @ask, if this code is still on the trajectory to get merged into the proper code-base, what is remaining before it is feature and functionally complete? And, would you or anyone like additional help? Documentation and/or dev work.

ask:
@jheld it should already be merged into master, you can help us test it, or work on an issue in the 4.0 milestone so we can release it soon :)

PR #2512, opened 2015-02-20
Optimized Redis result backend for 3.1 branch

sangwa:
#2511 for 3.1 branch

### DIFF
--- a/celery/backends/redis.py
+++ b/celery/backends/redis.py
@@ -9,14 +9,19 @@
 from __future__ import absolute_import
 
 from functools import partial
+from hashlib import sha1
+from threading import local
+from time import time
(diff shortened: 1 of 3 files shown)


ask:
Thanks!  Looking good, but this really needs to target master, not 3.1.  Are you able to change that?

sangwa:
Sure, here is the pull request targeting master: #2511. This one is targeting 3.1 because we used the stable branch in our development, then backported to the master.
I also made some minor improvements for the case when a chain raises an exception early, these were not pushed yet. I think I can push them and update to the current upstream this week.

thedrow:
Superseded by #2511.

PR #2515, opened 2015-02-24
Fix failure when task has multiple callbacks

NotSqrt:
Exception was:

```
File "celery/app/trace.py", line 276, in trace_task
    group.apply_async((retval, ))
TypeError: unbound method apply_async() must be called with group instance as first argument (got tuple instance instead)
```

Due to a simple typo (group -> group_).

I fixed branch 3.1 only, because it was heavily rewritten in 3.2; but the test is probably useful for 3.2 (at the moment, it fails due to another exception).

### DIFF
--- a/celery/app/trace.py
+++ b/celery/app/trace.py
@@ -273,7 +273,7 @@ def build_tracer(name, task, loader=None, hostname=None, store_errors=True,
                                     else:
                                         sigs.append(sig)
                                 for group_ in groups:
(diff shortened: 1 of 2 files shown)


NotSqrt:
The test failure is an unrelated pypy error.

PMickael:
Thanks, If you want add your name to CONTRIBUTORS.txt

PR #2522, opened 2015-02-28
Removes what looks like a copy/paste artifact in batches.py.

### LINKED ISSUE #2528: Syntax errors in contrib/batches.py
There are syntax errors in contrib/batches.py in the master branch in lines 200-213. Pull request #2522 removes what appear to be copy/paste artifacts causing the errors.

wil-langford:
It looks like there are some lines in contrib/batches.py starting at line 200 or so that have a bunch of extra characters far out to the right.  My best guess is that it was a split view in GNU screen and the code got copied and pasted with some sort of ps output.

I'm still getting unresolved reference warnings in my IDE after removing the extra characters, but I also see that there's a **kw argument to the function, so I'm presuming that the unresolved things are being brought in that way.

While the extra characters were clearly wrong (syntax errors), I'm new to celery and don't understand its internals well enough to verify that the code without the extra characters is now correct.

Resolves #2528.

### DIFF
--- a/celery/contrib/batches.py
+++ b/celery/contrib/batches.py
@@ -197,20 +197,20 @@ class Batches(Task):
         flush_buffer = self._do_flush
 
         def task_message_handler(message, body, ack, reject, callbacks, **kw):
-            if body is None:                                                                                                                        31513 ?        S    125:09 /usr/bin/python -m celery worker --without-heartbeat -c 50 --pool=eventlet -n celery6@ns326150.ip-37-187-158.eu --app=mai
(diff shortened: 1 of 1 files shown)


wil-langford:
This pull request is failing a Travis CI build check, but it looks like every pull request since the 1st of February has failed the same check.

PMickael:
Hi, my mistake, I resolved the problem and copy/paste from multi bash screen in Github editor to push it, and didn't see this ! @ask ping !

Issue #2523, opened 2015-02-28
Django apps NamespacePath support ?


douglarek:
As Django docs (https://docs.djangoproject.com/en/1.7/ref/applications/#namespace-packages-as-apps-python-3-3) said, Django supports a _NamespacePath app.

But when I used _NamespacePath with django-celery, I got some errors:
```
/Users/dj/.virtualenvs/3.4.3/lib/python3.4/site-packages/djcelery/loaders.py in find_related_module(app, related_name)
    197 
    198     try:
--> 199         imp.find_module(related_name, app_path)
    200     except ImportError:
    201         return

/Users/dj/.virtualenvs/3.4.3/lib/python3.4/imp.py in find_module(name, path)
    269         # Backwards-compatibility
    270         raise RuntimeError("'list' must be None or a list, "
--> 271                            "not {}".format(type(name)))
    272 
    273     if path is None:

RuntimeError: 'list' must be None or a list, not <class 'str'>
```

I know celery itself also not support it, will support ?

Thanks

douglarek:
sorry, not right place

kmollee:
This error will occur when use `autodiscover_task` in celery and one of these app's folder did't have `__init__.py` file

pretty hard to find out what is going on.. ? debug information is lacky

aztrock:
Traceback (most recent call last):
  File "/opt/kibox/bin/celery", line 11, in <module>
    sys.exit(main())
  File "/opt/kibox/local/lib/python3.4/site-packages/celery/__main__.py", line 30, in main
    main()
  File "/opt/kibox/local/lib/python3.4/site-packages/celery/bin/celery.py", line 81, in main
    cmd.execute_from_commandline(argv)
  File "/opt/kibox/local/lib/python3.4/site-packages/celery/bin/celery.py", line 770, in execute_from_commandline
    super(CeleryCommand, self).execute_from_commandline(argv)))
  File "/opt/kibox/local/lib/python3.4/site-packages/celery/bin/base.py", line 311, in execute_from_commandline
    return self.handle_argv(self.prog_name, argv[1:])
  File "/opt/kibox/local/lib/python3.4/site-packages/celery/bin/celery.py", line 762, in handle_argv
    return self.execute(command, argv)
  File "/opt/kibox/local/lib/python3.4/site-packages/celery/bin/celery.py", line 694, in execute
    ).run_from_argv(self.prog_name, argv[1:], command=argv[0])
[...]

PR #2525, opened 2015-03-02
Added _NamespacePath (Python 3.3 +) module support #2523

douglarek:
#2523

### DIFF
--- a/celery/loaders/base.py
+++ b/celery/loaders/base.py
@@ -13,20 +13,16 @@ import importlib
 import os
 import re
 import sys
-
 from datetime import datetime
-
-from kombu.utils import json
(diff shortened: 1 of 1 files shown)


douglarek:
Travis CI pypy build does not work for a long time !

malinoff:
1. Your PR somewhy restructures import statements and add an extra newline after the class definition. Why?
2. From the [docs](https://docs.python.org/3/tutorial/modules.html#packages-in-multiple-directories), `__path__` is intended to be `list`, not `str`.

If you're still facing this issue and want it to be actually fixed, please show all related files and configs in the original issue.

I'm closing this for now, from my point of view it doesn't really solve anything.

PR #2538, opened 2015-03-17
Redis new_join optim now takes `CELERY_CHORD_PROPAGATES` param into account

tristan0x:
I guess this parameter should be supported in 3.1.

### DIFF
--- a/celery/backends/redis.py
+++ b/celery/backends/redis.py
@@ -178,12 +178,13 @@ class RedisBackend(KeyValueStoreBackend):
         return self.client.expire(key, value)
 
     def _unpack_chord_result(self, tup, decode,
+                             propagate,
                              EXCEPTION_STATES=states.EXCEPTION_STATES,
                              PROPAGATE_STATES=states.PROPAGATE_STATES):
         _, tid, state, retval = decode(tup)
         if state in EXCEPTION_STATES:
             retval = self.exception_to_python(retval)
-        if state in PROPAGATE_STATES:
+        if propagate and state in PROPAGATE_STATES:
             raise ChordError('Dependency {0} raised {1!r}'.format(tid, retval))
         return retval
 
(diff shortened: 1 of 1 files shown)


ask:
The chord propagates setting is more like a future import, the old behavior was never defined and was just an accident.  The plan was to deprecate this setting, so the new_join does not have to be compatible with it.

tristan0x:
From what I understand, setting the chord callback state to ERROR when a task fails will be the sole behavior in 3.2.
But in 3.1, as `CELERY_CHORD_PROPAGATE` exists, even if it is a mistake, I guess it should be consistent in the entire application.

I use redis new join, and needs to be notified when something went wrong in a bunch of tasks wrapped in a chord. How can I achieve that without CELERY_CHORD_PROPAGATE set to False?

ask:
In 4.0 at least you can add an errback to the chord callback:

    group(add.s(i, i) for i in range(10)) | xsum.s().set(errback=handle_chord_error.s())

    @app.task(bind=True)
    def handle_chord_error(self, task_id):
        print('Chord raised: %r' % (self.app.AsyncResult(task_id).exception))


You should be using a persistent result backend for this (so not `amqp` or `rpc`), since you need to fetch the previous result.

We're not able to pass the actual exception/traceback objects since they cannot be serialized by json, but the result backend is already trying to somewhat reconstruct an exception (but you cannot e.g. use `isinstance(x, KeyError)` since the deserialization creates fake exception classes).  We could use the same mechanism in the errback now.

ask:
I just had the most amazing idea..

What if we don't call the errback as a task, but instead have the worker call it locally?
That way we could pass the raw exception/traceback objects just perfectly.

You wouldn't be able to route the errback to a different queue etc, but I have a feeling that is not very common.

ask:
Now it's in master, where you can do:

    @app.task
     def handle_error(request, exc, traceback):
         print('Task id {0!r} raised exception: {1!r}'.format(request.id, exc))

    (group(add.s(i, i) for i in range(10)) | xsum.s().on_error(handle_error.s())).delay()

PR #2542, opened 2015-03-19
Support for ultramemcached (umemcache)

pjdufour:
This PR adds support for ultramemcached (on PyPi called umemcache).  ultramemcached is a fast C++ Memcache client with Python bindings.  ultramemcached and gevent work well together for high throughput applications.

The GitHub repo for ultramemcached is https://github.com/esnme/ultramemcache.  It is available on PyPI at: https://pypi.python.org/pypi/umemcache.

This is my first PR for celery.  Please let me know if anything else is required.  Thanks for your consideration.


PMickael:
Did you try in production, because umemcache get function return tuple, and memcache get function return string.

umemcache
```python
mc.set('test', '1')
'STORED'
mc.get('test')
('1', 0)
```

memcache

```python
mc.set('test', '1')
True
mc.get('test')
'1'
```

pjdufour:
I have used it in production.  It looks like ultramemcache is a lower level api than python-memcached as in it doesn't searilize to text.  I believe the second value is "flags" as specified at https://github.com/memcached/memcached/blob/master/doc/protocol.txt#L155.  In umemcache C++ code the second value in the tuple is set to `oflags` (https://github.com/esnme/ultramemcache/blob/master/python/umemcache.cpp#L451), which can be passed in via the `flags` kwarg (https://github.com/esnme/ultramemcache/blob/c2b02fc18aefe863009df433e9f47a8ece49d27e/python/umemcache.cpp#L1029).

Mozilla in `django-memcached-pool`, which uses umemcache, has a wrapper around `get` at https://github.com/mozilla/django-memcached-pool/blob/master/memcachepool/cache.py#L156 that only passes up the value rather than the `(value, flags)` tuple.

The README for umemcache does specify you should use `c.get('key')[0]`.  

It looks like python-memcached keeps flags internal (https://github.com/linsomniac/python-memcached/blob/master/memcache.py).

**Possible Soution**
So will the umemcache implementation break all the celery tests?  If so, maybe I could do a PR on umemcache, so `get` doesn't return flags by default and only returns if flags kwarg is equal to true, such as if `c.get('key', flags=True)`.  That might break some backwards compatibility, though.

thedrow:
I don't think umemcached is that maintained. You can do ```c.get('key')[0]``` in the results backend though.

ask:
Can we try to add some tests for this?

For the `.get` problem it's probably easier to just wrap the client, or add the ability to specify the get function that the Cache backend should use, so that we can override just the get function.

thedrow:
@ask I think that we should close this one. umemcached hasn't seen one commit in the last 4 years. If we have problems with it we might sink time into debugging code that we cannot maintain.

Feel free to reopen if I'm wrong.

PR #2550, opened 2015-03-23
on_message callback added

xBeAsTx:
For some tasks it's realy convenient to catch all task state changes. For example, on clients I want catch comment progress cahnges as soon as they received (on backends which support it, like AMQP). So in task-code we will do this:
```
self.update_state(meta = { 'comment': 'some comment 1...'})
...
self.update_state(meta = { 'comment': 'some comment 2...'})
```

But currently in clients we can catch only final status without every seconds sleep like this terrible code:
```
while task.state != 'SUCCESS' and task.state != 'FAILURE' and task.state != 'REVOKED':
	# Check if comment was changed
	curr_comment = result.get('comment', last_comment) if result else last_comment
 
	# Process new lines
	process_state_lines(last_comment, curr_comment)
 
	# Update last comment
	last_comment = curr_comment

	# Get new task result
	time.sleep(1)
	result = task.result
```
It's not a good way. We must sleep and check that comment is changed.

So very easy solution is allow set 'on_message' call back for get_many method. And we can use it like so:
```
	def on_message(body):
		print body

	r = group(task).apply_async()
	r.get(on_message=on_message)
```

Probably, it's also can be usefull for 'chorded' tasks. We can monitor every status change in all tasks. But need more digging for accept this.

### DIFF
--- a/celery/backends/amqp.py
+++ b/celery/backends/amqp.py
@@ -231,7 +231,7 @@ class AMQPBackend(BaseBackend):
     def _many_bindings(self, ids):
         return [self._create_binding(task_id) for task_id in ids]
 
-    def get_many(self, task_ids, timeout=None, no_ack=True,
+    def get_many(self, task_ids, timeout=None, no_ack=True, on_message=None,
                  now=monotonic, getfields=itemgetter('status', 'task_id'),
                  READY_STATES=states.READY_STATES,
                  PROPAGATE_STATES=states.PROPAGATE_STATES, **kwargs):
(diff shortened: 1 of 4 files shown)


PMickael:
its sounds great, can you update test case for ampq :+1: 
https://github.com/celery/celery/blob/master/celery/tests/backends/test_amqp.py#L263
Thanks.

PMickael:
Please add your name to https://github.com/celery/celery/blob/master/CONTRIBUTORS.txt if not already in.
Thanks.

PR #2553, opened 2015-03-25
Improve use of uri in mongo backend

samjy:
The issue
----

When providing a mongo URI to ``CELERY_RESULT_BACKEND``, only the host and port are used for the backend's connexion. Username, password, database and options, which can also be present in the URI aren't considered, and are expected to be provided in ``CELERY_MONGODB_BACKEND_SETTINGS``.

Example::

    CELERY_RESULT_BACKEND = 'mongodb://myuser:mypass@localhost:27017/mycelerydb'

will indeed connect to localhost:27017, but will use the default ``None:None`` user/password, and will connect to the default ``celery`` database.

Proposal
----
Get all the possible information from ``CELERY_RESULT_BACKEND`` (user, password, database, options), so that an URI like this::

    CELERY_RESULT_BACKEND = 'mongodb://myuser:mypass@mongo1:27017,mongo2:27017,mongo3:27017/mycelerydb?replicaSet=rs0'

won't need any duplicate conf in ``CELERY_MONGODB_BACKEND_SETTINGS``

### DIFF
--- a/celery/backends/mongodb.py
+++ b/celery/backends/mongodb.py
@@ -45,6 +45,7 @@ class Bunch(object):
 
 class MongoBackend(BaseBackend):
 
+    mongo_host = None
     host = 'localhost'
     port = 27017
     user = None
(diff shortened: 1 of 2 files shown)


PMickael:
Thanks, If you want add your name to CONTRIBUTORS.txt

PR #2558, opened 2015-03-31
contributor added

xBeAsTx:
By request from #2550


PMickael:
Added manually, your version is outdated

PR #2575, opened 2015-04-15
Changed _apply_chord_incr to use int value

wiliamsouza:
Example
```
>>> mc.set('dfdsgfdgfd', '0')                                                   
True
>>> mc.incr('dfdsgfdgfd')                                                       
  File "<console>", line 1, in <module>
_pylibmc.ClientError: 1 keys failed
>>>
```

### DIFF
--- a/celery/backends/cache.py
+++ b/celery/backends/cache.py
@@ -129,7 +129,7 @@ class CacheBackend(KeyValueStoreBackend):
         return self.client.delete(key)
 
     def _apply_chord_incr(self, header, partial_args, group_id, body, **opts):
-        self.client.set(self.get_key_for_chord(group_id), '0', time=86400)
+        self.client.set(self.get_key_for_chord(group_id), 0, time=86400)
         return super(CacheBackend, self)._apply_chord_incr(
             header, partial_args, group_id, body, **opts
         )


jaysonsantos:
:+1:  for this because, if using memcached plain text it does not saves flags to say which type is that, so memcached will try to cas that to integer.
When using binary protocol, memcached libs store some flags like, integer, long, serialized, compressed, etc...

malinoff:
This PR seems to be backwards-incompatible. We may end up with some users having '0' and some users having 0 as value.

@ask can you take a look at this please?

jaysonsantos:
If user is already working with '0' maybe means that he is in memcached
plain text protocol, so he wouldn't be affected.


Em qua, 15 de abr de 2015 às 23:17, Dmitry Malinovsky <
notifications@github.com> escreveu:

PMickael:
Seem's good for me I merge, like say @jaysonsantos, if it's working with '0', then it will with 0

Thanks @wiliamsouza, add you're name to contributor if you want.

wiliamsouza:
@PMickael Great! Next time I add my name on the contributor file :)
Any chance to have a 3.1.18 version with this fix in the near future?

PMickael:
@wiliamsouza, I don't know about version 3.1.18, I'm only a contributor too, @ask is the boss, but yes there is some fix like this one, that would be great to be released.

ask:
Merged into 3.1!

PR #2580, opened 2015-04-16
mongodb default options removed.

### LINKED ISSUE #2579: pymongo refuses backend/mongodb.py default settings
`self.options.setdefault('max_pool_size', self.max_pool_size)`
`self.options.setdefault('auto_start_request', False)`

These lines are refusing from `pymongo` and gives `ConfigurationError` exception. By the way pymongo version is 3.0 and for use mongodb as backend i have to remove that lines from mongodb.py.

fatihsucu:
Fixes #2579

### DIFF
--- a/celery/backends/mongodb.py
+++ b/celery/backends/mongodb.py
@@ -80,7 +80,7 @@ class MongoBackend(BaseBackend):
 
 
         # default options according to pymongo version
-        if pymongo.version_tuple >= 3:
+        if pymongo.version_tuple >= (3,):
             self.options.setdefault('maxPoolSize', self.max_pool_size)
         else:
             self.options.setdefault('max_pool_size', self.max_pool_size)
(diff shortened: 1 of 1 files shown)


thedrow:
I think that this code should branch based on PyMongo's version.
If it's version 3 and above don't set the defaults.

PMickael:
I think, It would be better to treat the exception !

ionelmc:
Turns out `max_pool_size` was renamed to `maxPoolSize`. The other option is removed in 3.0 (it had something to do with read-after-write preference, see https://github.com/mongodb/mongo-python-driver/blob/master/doc/changelog.rst#classpymongomongo_clientmongoclient-changes )

@fatihsucu will you update the PR to have version check and only set `maxPoolSize` on 3.0 ?

fatihsucu:
Yes, i will @ionelmc

PMickael:
Great thanks.
Add you're name to contributor if you want.

fatihsucu:
Thanks for comments and advice. It would be great.

ask:
Wow, these mongo folks sure love pointless and backward incompatible changes ;)

ask:
Also applied to 3.1

PR #2597, opened 2015-04-29
check for AppConfig object and use name attribute to get package

### LINKED ISSUE #2596: auto discover should read `name` attribute from Django AppConfigs
The AppConfig module is not always in the same package as the named app (https://docs.djangoproject.com/en/1.8/ref/applications/#for-application-users), but the `name` attribute is always present in that config object (https://docs.djangoproject.com/en/1.8/ref/applications/#django.apps.AppConfig.name).

The `find_related_module` method should check that attribute.

References #2248

stringfellow:
fixes #2596

### DIFF
--- a/celery/loaders/base.py
+++ b/celery/loaders/base.py
@@ -283,7 +283,20 @@ def find_related_module(package, related_name):
     try:
         importlib.import_module(package)
     except ImportError:
-        package, _, _ = package.rpartition('.')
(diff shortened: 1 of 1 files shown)


stringfellow:
Closed because `autodiscover_tasks` can do this more elegantly:
`app.autodiscover_tasks(lambda: [n.name for n in apps.get_app_configs()]`

PR #2598, opened 2015-04-29
Fix Exception marshalling with JSON serializer

### LINKED ISSUE #2518: CELERY_RESULT_SERIALIZER = 'json' breaks Exception marshaling
Setting `CELERY_RESULT_SERIALIZER = json` and raising an exception in the worker leads to this:

```
/path/to/lib/python2.7/site-packages/celery/result.py in get(self, timeout, propagate, interval, no_ack, follow_parents, EXCEPTION_STATES, PROPAGATE_STATES)
    173             status = meta['status']
    174             if status in PROPAGATE_STATES and propagate:
--> 175                 raise meta['result']
    176             return meta['result']
    177     wait = get  # deprecated alias to :meth:`get`.

TypeError: exceptions must be old-style classes or derived from BaseException, not dict
```

where the contents of `meta['result']` are (in my case):

```
{u'exc_message': u'unknown keys: nam', u'exc_type': u'ValueError'}
```

so it *looks* like celery could convert the dict to a real exception before raising, but it does not currently.  Changing back to `pickle` works as expected.

bug can be reproduced with the following:

```python
# jsonresults.py
from celery.app.base import Celery

CELERY_RESULT_SERIALIZER = 'json'
CELERY_RESULT_BACKEND = 'amqp'

app = Celery(config_source=__name__)

@app.task
def hello():
    raise ValueError('go away')
```

worker:

```
# python -m celery --app=jsonresults:app worker
```

caller:

```python
import jsonresults
jsonresults.hello.delay().get()
```

allardhoeve:
When using the JSON serializer, exceptions get serialized as dicts. The code in `drain_events` in `amqp.py` naively then sets the result dict to such a dict without transforming the dict structure back into an actual Exception through `exception_to_python` [[1]](https://github.com/celery/celery/blob/6592ff64b6b024a4b68abcc53b151888fdf0dee3/celery/backends/amqp.py#L198).

When a task raises an exception, `AsyncResult.get` and `AsyncResult.maybe_reraise` try to raise the
exception [[2]](https://github.com/celery/celery/blob/6592ff64b6b024a4b68abcc53b151888fdf0dee3/celery/result.py#L890) [[3]](https://github.com/celery/celery/blob/6592ff64b6b024a4b68abcc53b151888fdf0dee3/celery/result.py#L273), which is actually still a dict and fail with:

```
TypeError: exceptions must be old-style classes or derived from
BaseException, not dict
```

This patch makes `drain_events` call `meta_from_decoded` which is
responsible for that, just like it is called in `get_many` [[4]](https://github.com/celery/celery/blob/33e72fdbc7b07fc26d13bcdc36fb6f42c8291b66/celery/backends/amqp.py#L255). Then,
raising the exception in `AsyncResult` works fine.

To reproduce, see the testcase in #2518. Then, apply the patch and see
stuff start to work again.

closes #2518

### DIFF
--- a/celery/backends/amqp.py
+++ b/celery/backends/amqp.py
@@ -195,7 +195,7 @@ class AMQPBackend(BaseBackend):
 
         def callback(meta, message):
             if meta['status'] in states.READY_STATES:
-                results[meta['task_id']] = meta
+                results[meta['task_id']] = self.meta_from_decoded(meta)
 
         consumer.callbacks[:] = [callback]
         time_start = now()
--- a/celery/tests/backends/test_amqp.py
+++ b/celery/tests/backends/test_amqp.py
@@ -13,6 +13,7 @@ from celery import states
 from celery.backends.amqp import AMQPBackend
 from celery.exceptions import TimeoutError
 from celery.five import Empty, Queue, range
+from celery.result import AsyncResult
 from celery.utils import uuid
 
 from celery.tests.case import (
@@ -246,10 +247,20 @@ class test_AMQPBackend(AppCase):
         with self.assertRaises(TimeoutError):
             b.wait_for(tid, timeout=0.01, cache=False)
 
+    def test_drain_events_decodes_exceptions_in_meta(self):
+        tid = uuid()
+        b = self.create_backend(serializer="json")
+        b.store_result(tid, RuntimeError("aap"), states.FAILURE)
+        result = AsyncResult(tid, backend=b)
+
+        with self.assertRaises(Exception) as cm:
+            result.get()
+
+        self.assertEqual(cm.exception.__class__.__name__, "RuntimeError")
+        self.assertEqual(str(cm.exception), "aap")
+
     def test_drain_events_remaining_timeouts(self):
-
         class Connection(object):
-
             def drain_events(self, timeout=None):
                 pass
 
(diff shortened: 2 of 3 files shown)


allardhoeve:
Travis seems to fail on the pypy build, which has been failing consistently for all PRs. Other builds are fine.

thedrow:
I'm not sure how the second commit fixed anything.
Is there any way to write a test for this bug fix to ensure it won't regress?

allardhoeve:
So I tried to explain that in the commit message:

Attempt to pass Travis test on pypy …
- supress progress bar by pip when running pypy as this
  seems to crash the test on Travis:
  https://travis-ci.org/celery/celery/jobs/60549687

- this prevents dev.txt from being properly installed and the
  result is tests failing with import errors

allardhoeve:
I don't see how we would write a test that checks this. It is a weird combination of outputting a progress bar with utf-8 characters into Travis's console. Locally, everything works, but not in Travis. This seemed like the smallest and quickest fix.

thedrow:
Sorry I meant a test for the bug fix itself. Not the fact that pip is failing with PyPy3.

allardhoeve:
like so?

thedrow:
Yup.

allardhoeve:
Any feedback on this PR?

thedrow:
LGTM. @ask Please merge.

allardhoeve:
bump @ask, @thedrow

thedrow:
@PMickael Maybe you can take a look?

PMickael:
Thanks, please add you're name to contributor if it's not already in.

naggie:
I think this is still an issue -- running the development version from the snapshot repositories. I can reproduce it using any combination of redis/rabbitMQ and json/msgpack.

Issue #2137, opened 2014-07-10
self not passed to link_error for task methods


alexhayes:
It appears that for task methods (ie.. `celery.contrib.methods.task`) if used as `link_error` the first argument passed is the `task_id` not `self`, as one would naturally expect.

For example:

```python
from celery.contrib.methods import task
class MyClass(object):

    @task()
    def raises(self):
        raise Exception('Hi')

    @task()
    def on_raise(self, task_id):
        ...
```

If one was to create the following:

```python
m = MyClass()
m.raises.s().apply_async(link_error=m.on_raise.s())
```

It fails, complaining that `TypeError: on_raise() takes at least 2 arguments (1 given)`.

It appears that `task_id` is passed as the first argument and `self` is not passed at all.

As a workaround the signature to `on_raise` can be changed as follows:

```python
    @task()
    def on_raise(task_id, self):
        ...
```

And then `self` passed into the `link_error` callback:

```python
m = MyClass()
m.raises.s().apply_async(link_error=m.on_raise.s(m))
```

However this just seems a little odd for a task method.

Is this an issue or intended behaviour?

nuklea:
Have same issue, when pass subtask by `link` argument. @alexhayes did you find solution?

nuklea:
When `CELERY_ALWAYS_EAGER = True` all works fine.

alexhayes:
I decided to write a decorator to handle this, it's a bit of a hack
solution which is pretty ugly but at least it works - see
https://bitbucket.org/alexhayes/django-toolkit/src/c8c6e3e96627bfa8d8526b2cf324672b9291d10d/django_toolkit/celery/decorators.py?at=master#cl-3

I'm working on a much better solution in #2600 which works for
`apply_async` but not `apply` at the moment and it would be easy to make
this work for normal classes (as opposed to django models) however I
haven't had a chance to do this and I'm also waiting on feedback about my
approach.

All this may be in vain however because @ask has stated that unfortunately
method based tasks are considered a failed experiment and will be removed
in 3.2.

ask:
Closing this, as we don't have the resources to complete this task.  Task methods have been removed for 4.0

PR #2600, opened 2015-04-30
Support for Django Model methods as tasks.

alexhayes:
I'm submitting this PR more for review and suggestions for possible future inclusion.

Essentially I wanted a way to easily turn Django Model methods into tasks.

Up until now I've used `celery.contrib.methods.task` and a [@refresh](https://bitbucket.org/alexhayes/django-toolkit/src/c9aaf1b4ded94b239b91ac354192ffbaa2a9566e/django_toolkit/models/decorators.py?at=master#cl-3) decorator to ensure I don't have a stale model instance. However this approach meant the entire model is serialized and sent across the wire, obviously this is not very efficient and completely unnecessary considering that you then refresh the model in the worker... Obviously the solution outlined in [the docs](http://docs.celeryproject.org/en/latest/userguide/tasks.html?highlight=model#state) gets around both these issues however when you have an application that relies heavily on methods in models and those methods need to be turned into tasks, creating tasks for each is not very DRY and a pain when refactoring.

Thus, I wanted a solution that would only publish the minimum information required to the broker and using this data get a refresh model instance in the worker. The `@model_task` decorator achieves this by using base class `DjangoModelBaseTask` that relies on some slight refactoring of `Task` internals.

My thoughts/questions/comments thus far;

1. I'm yet to confirm if the solution executes the retrieval of the model instance in the parent worker (which obviously should be avoided).
2. Is the refactoring work in `Task` an acceptable approach?
3. Naming of methods `augment_args_for_run` and `augment_args_for_send`
4. Use of `augment_args_for_run` in `celery/app/trace.py` - note that previously it wasn't checking for `self.__self__` but now it will be. Will this be a problem? Note, all tests are passing on python 2.7 & 3.4.
5. The solution doesn't yet work for chains - I'm working on a solution in `celery.canvas.Signature`.
6. The import of `from django.db.models.loading import get_model` occurs within `DjangoModelBaseTask.augment_args_for_run` so that an import error does not occur when building docs - not ideal...
7. How to go about creating tests - this would probably require Django, which I imagine is not really desired.

I'd be happy to break `DjangoModelBaseTask` out into a separate project/repository which would get around the requirement for Django however it obviously relies on the slight refactoring of `Task` - it's also something that I imagine a lot of other users could benefit from (frankly, I can't believe I'm the first person to think of this...).

### DIFF
--- a/celery/app/builtins.py
+++ b/celery/app/builtins.py
@@ -291,7 +291,7 @@ def add_chain_task(app):
             app = self.app
             last, fargs = None, args  # fargs passed to first task only
             for task in kwargs['tasks']:
-                res = signature(task, app=app).clone(fargs).apply(
+                res = signature(task, app=app).clone(fargs, async=False).apply(
                     last and (last.get(), ),
                 )
                 res.parent, last, fargs = last, res, None
--- a/celery/app/task.py
+++ b/celery/app/task.py
@@ -225,7 +225,7 @@ class Task(object):
 
     #: This is the instance bound to if the task is a method of a class.
     __self__ = None
-
+    
     #: The application instance associated with this task class.
     _app = None
 
@@ -560,7 +560,7 @@ class Task(object):
             args = (self.__self__, ) + args
         return args, kwargs
 
-    def augment_args_for_send(self, args, kwargs):
+    def augment_args_for_send(self, args, kwargs, async=True):
         """
         Augment the args/kwargs prior to sending the task to the broker.
         """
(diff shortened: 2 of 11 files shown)


alexhayes:
Commit e3650bf2a250a79e730e4d07f83ab2194c29aa21 has a working `apply_async` implementation however `apply` is not working as yet.

I'm working on a potential solution.

thedrow:
What's the status here?

malinoff:
I don't think this will be merged, since @ask is saying that task methods are failed experience (and I'm completely agree with him).

thedrow:
Do you have an alternative implementation?

alexhayes:
Current status is I was waiting on input as to if task methods would definitely be removed in 3.2. I've spoken to @ask and he says that they won't be removed if the bugs can be sorted out with them.

So I don't forget...

`===8<===`
@alexhayes: is there any chance that you can be persuaded to not remove task methods in 3.2?
@ask: yeah, if the problems with it are fixed :)
@alexhayes: are there specific bugs?
@ask: like self not being set twice for retry
@alexhayes ok, well I'll make that part of my mission for #2600
@ask: task.__self__ is just added as first argument in apply_async
@ask: no way to know if it's already "bound"
@alexhayes: Ok, I'll keep working on #2600 then with the view that it's definitely not going to get merged unless I can solve that problem
@ask: to solve it properly, it would probably need to be sent as a message header in the new task protocol
`===>8===`

alexhayes:
I'm closing this as I have created [django-cereal](https://github.com/alexhayes/django-cereal) which is a much better solution.

There are still a number of other issues, such as #2137, when using task methods but these aren't really related to this issue.

My focus now will be to fix issues related to task methods so they can be retained in 3.2.

PR #2607, opened 2015-05-06
Retries should behave the same in eager and not eager mode.

feanil:
Right now if you use the retry method.  If you are not running in eager mode, than the value of the `throw` paramater is honored(the function throws an exception with the results).  However if you are running in eager moder, this paramater is not honored.

### DIFF
--- a/celery/app/task.py
+++ b/celery/app/task.py
@@ -595,6 +595,8 @@ class Task(object):
             # if task was executed eagerly using apply(),
             # then the retry must also be executed eagerly.
             S.apply().get()
+            if throw:
+                raise ret
             return ret
 
         try:


feanil:
Looks like the travis failure is only with PyPy, I'm not sure what's going on there.  

If I could get some guidance on how to write a test, I'd be happy to do that.  If there is a similar task that I can use as a template that would be very helpful.

feanil:
@ask does this change make sense?

joostdevries:
@feanil could you rename your PR + ammend your commit message to clearly reflect the change? Also: Maybe add a bit of details on why you're proposing this change?

feanil:
@joostdevries Updated, let me know if that is enough information or if something is confusing.

PMickael:
Thanks,
Please add you're name to contributor if you want.

Issue #2609, opened 2015-05-08
Can't control default_app when using django


jtillman:
When using Django and multiple celery apps, the default app gets overwritten by the last made celery application because of the DjangoFixup. Seems like the fixup was made to always set an app as the default app to enable runserver: https://github.com/celery/celery/commit/ee3019386b277e34e68594a3e32df3d3291b4b39

Example:
``` python
from celery.app import default_app
first_app = Celery("first-app")
first_app.set_default()

second_app = Celery("second-app")

default_app == first_app << False
default_app == second_app << True
```

PR #2610, opened 2015-05-08
Only set default app if there isn't one already

jtillman:
This is a PR to address issue 2609: https://github.com/celery/celery/issues/2609


joostdevries:
looks reasonable :-)

PMickael:
Merge, thanks.

Add you're name to contributor if it's not already in.

PR #2617, opened 2015-05-13
Sync documentation with reality

squfrans:
according to `celery.beat.install_default_entries()`,
the job is scheduled for 4am (instead of midnight),
and the default expire time is 12 hours (instead of 1 day).


PMickael:
That's right for 4am https://github.com/celery/celery/blob/master/celery/beat.py#L206
But not for expire time : https://github.com/celery/celery/blob/master/celery/app/defaults.py#L157

PMickael:
Merge, thanks.
Please add you're name to contributor is not already in.

PR #2629, opened 2015-05-23
Fix: Django fixup throws error when the app uses custom loader class

slapec:
This small bug appeared when I instantiated Celery with a custom loader class in my Django project. The problem was that the `celery.fixups.django.fixup` function expected the loader class to be a symbol string, but it can be also the class directly.
I've added this small function `celery.utils.imports.dot` which returns the module and class name as a string too so the condition haven't had to change drastically in the fixup.

### DIFF
--- a/celery/utils/imports.py
+++ b/celery/utils/imports.py
@@ -14,6 +14,7 @@ import os
 import sys
 
 from contextlib import contextmanager
+from inspect import getmodule
 
 from kombu.utils import symbol_by_name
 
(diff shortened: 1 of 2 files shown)


thedrow:
Is there a chance we can add a test that verifies that this bug is fixed?

slapec:
Sorry, I've not spent that much time on finding out how the existing tests work but I'll try to implement something for this case soon.

celery/utils/imports.py
```diff
@@ -112,3 +113,7 @@ def module_file(module):
     """Return the correct original file name of a module."""
     name = module.__file__
     return name[:-1] if name.endswith('.pyc') else name
+
+def dot(cls):
+    """It is basically kombu.utils.symbol_by_name reversed"""
+    return getmodule(cls).__name__ + ':' + cls.__name__
```
auvipy: there should be a newline at the end of code

thedrow:
@auvipy Could you try to write tests for this?

thedrow:
@auvipy Would you like to fix this PR and get it to mergable state?

PR #2631, opened 2015-05-28
Update celeryd SysV Init Script for CentOS to take advantage of multi

TauPan:
Best of both worlds for me. I can keep settings in /etc/sysconfig/celeryd and still use e.g. 'stopwait' or 'restart' to smoothly adapt to the changed configuration in an rpm-based installation.

I think this can also easily be ported to the main init script, I'm just commiting what I changed in my local installation here.


TauPan:
Thanks for asking!

Not sure. I've never tried to use the generic script. I just used the one for centos, as it was there.

Closing this is ok with me.

PR #2645, opened 2015-06-05
Added a check for syncing the schedule even if nothing is in the schedule

colinmcintosh:
Currently, if there is nothing in the Beat schedule the schedule won't re-sync unless Beat is restarted. In addition, if there is something in the schedule Beat won't re-sync until a task is run. This could potentially be a very long time. 

This change adds a check to see if Beat should sync the schedule every time it wakes up.

### DIFF
--- a/celery/beat.py
+++ b/celery/beat.py
@@ -504,6 +504,8 @@ class Service(object):
                     debug('beat: Waking up %s.',
                           humanize_seconds(interval, prefix='in '))
                     time.sleep(interval)
+                    if self.scheduler.should_sync():
+                        self.scheduler._do_sync()
         except (KeyboardInterrupt, SystemExit):
             self._is_shutdown.set()
         finally:


thedrow:
Did you break the tests or is there an intermediate failure in the build?

thedrow:
I don't see that you changed the code and now that I'm thinking about it this needs at least a unit test.

thedrow:
I dug into the code a bit. Why do you need to sync at every beat interval?

thedrow:
@colinmcintosh ping?

PMickael:
@thedrow maybe because it's suppose to be run every x minutes (sync_every variable), but if there isn't any task or restart, sync is not triggered.

thedrow:
@PMickael @malinoff LGTM, I think. What do you guys think?

PMickael:
:+1:

colinmcintosh:
Sorry I didn't get a chance to reply to this earlier. @PMickael nailed my issue exactly. Thanks!

PR #2650, opened 2015-06-10
Retry BUG in windows with prefork concurrency option after v3.1.12

allenling:
Retrying a task in Windows would not work with prefork concurrency option cause lacking of the broker url while child process be instantiated. so, child process would establish a default broker connection with the default broker url, not the broker url we have specified, after v3.1.12.

The broker url we have specified would be restored in self._preconf when our Celery class be instantiated in father process. we must invoke Settings.__setitem__ method, like setattr(s._preconf, k, v), than the broker url would be restored in Settings.change. eventually, father process could pass the Settings.change to child process for instantiating Celery class.

In Linux, child process would inherit resources from it`s father process, so the broker url.

### DIFF
--- a/celery/app/base.py
+++ b/celery/app/base.py
@@ -33,7 +33,7 @@ from celery._state import (
     _announce_app_finalized,
 )
 from celery.exceptions import AlwaysEagerIgnored, ImproperlyConfigured
-from celery.five import values
+from celery.five import items, values
 from celery.loaders import get_loader_cls
 from celery.local import PromiseProxy, maybe_evaluate
 from celery.utils import gen_task_name
@@ -508,6 +508,10 @@ class Celery(object):
         while pending_beat:
             pargs, pkwargs = pending_beat.popleft()
             self._add_periodic_task(*pargs, **pkwargs)
+        # Settings.__setitem__ method, set Settings.change
+        if self._preconf:
+            for key, value in items(self._preconf):
+                setattr(s, key, value)
         self.on_after_configure.send(sender=self, source=s)
         return s
 


malinoff:
I'm not sure how this commit changes anything, ``self._preconf`` object is merged into the configuration as defaults:
https://github.com/celery/celery/blob/master/celery/app/base.py#L496

allenling:
Thanks for reply.

The most import is we must invoke Settings.__setitem__ method to change Settings.change. 
In ConfigurationView, we find that:
 
    def __setitem__(self, key, value):
            self.changes[key] = value

malinoff:
And why do you need to change ``Settings.changes``?
``__getitem__`` will inspect ``defaults`` when searching for the key.
https://github.com/celery/celery/blob/master/celery/datastructures.py#L458-L473

allenling:
This is  we want that passing the broker url which we specified to child process.

And we find that  all initial parameters which we specified in Celery class, would be included in Settings.change to be passed to child process when we tracing the code in billiard.

So, if we want to pass the broker url to child process, we must have the broker url restored in Settings.change, so we must invoke Settings.\__setitem\__ method, like setattr(s, k, v)

thedrow:
@malinoff Do we merge this? Do you have another proposed solution?

thedrow:
@malinoff ping?

malinoff:
While I think that the actual problem may live somewhere else (e.g. wrong `current_app` in non-main processes), I guess we can merge this as temporal solution.

malinoff:
@allenling thanks!

allenling:
Thanks guys!
And i got something confused while i keep tracking the code.
When initializing the Settings object,  Settings.changes and Settings.defaults will both be set up in Settings.\__init\__ method.
```
    def __init__(self, changes, defaults):
        self.__dict__.update(changes=changes, defaults=defaults,
                             _order=[changes] + defaults)
```

And in Celery._load_config method, it just simply initialize Settings object like that:
```
        s = self._conf = Settings(
            {}, [self.prepare_config(self.loader.conf), defaults],
        )
```

Apparently that changes is always {} when we initialize the Settings object.

And expect invoking Settings.\__setitem\__method, invoking Settings.update is another way to set Settings.changes.
```
    def update(self, *args, **kwargs):
        return self.changes.update(*args, **kwargs)
``` 
But no one invokes Settings.update or invokes Settings.\__setitem\__ method to set Settings.changes in celery/ap/base.py.
So it seems that the Settings.changes would not be set in the code ever.

Like @malinoff says:
> I'm not sure how this commit changes anything, self._preconf object is merged into the configuration as defaults:
https://github.com/celery/celery/blob/master/celery/app/base.py#L496

It is true that self._preconf is merged into the configuration as defaults.

But cause all parameters we specified in Celery class, not in celeryconfig.py, would be set in self._preconf, so finally self._preconf need be merged into configuration as changes but not defaults to pass the parameter to a child process.

*And*

>  And why do you need to change Settings.changes?
__getitem__ will inspect defaults when searching for the key.
https://github.com/celery/celery/blob/master/celery/datastructures.py#L458-L473

Yes, Settings.\__getitem\__ will inspect defaults when searching for the key. and actually, it will inspect Settings._order, which is a list includes changes and defaults. 

That is, if we set the BROKER_URL in our celeryconfig.py, there is no any problem, cause the child process would get the broker url from the same celeryconfig.py. if we do not set the BROKER_URL in our celeryconfig.py, we must set Settings.changes up to pass broker url to child process.

**So, the point is that the way initializing a Celery object in a child process with the parameters what we specify in father process Celery class in celery, is that setting all parameters what we specified to Settings.changes(celery code), and then changes would be passed to the Celery class in child process in billiard with fd(only in windows, and prefork in linux is different).**

Then Celery class in child process would use all values in changes as initial parameters to be initialized.

**So, we must change Settings.changes to set everything we specified Celery class in father process.**

**Or using another way to initialize a child process, not Settings.changes.**

Maybe the actual problem may live somewhere else, but still i  wanna share something that i try to figure out.

Cause that i am not a celery expert, it will be very nice if your guys would share your ideas. 

Thanks again!

thedrow:
@allenling Can you please open a new issue about this so we won't forget?

allenling:
@thedrow 
ok, done for new issue #2732

PR #2651, opened 2015-06-12
Add CELERY_EMAIL_CHARSET option

### LINKED ISSUE #2648: Celery can't send error email when task has non-ascii arguments
There is the issue with celery 3.1.18, python 3.4 and tasks with non-ascii string arguments.

When task raises exception and CELERY_SEND_TASK_ERROR_EMAILS set to True, celery couldn't send mail because email.mime.MimeText instance raises UnicodeDecodeError('ascii', ...). Exception is raised [here](https://github.com/celery/celery/blob/dffb61c4d99f1ce5817be267104e9810e88391ee/celery/utils/mail.py#L56). 
Looks like Message constructor has charset argument with 'us-ascii' as default value, and it can't be changed without editing celery source code. When charset changed to 'utf-8', bug disappears.

Example app:

```python
from celery import Celery

app = Celery('app')

app.conf.CELERY_SEND_TASK_ERROR_EMAILS = True
app.conf.ADMINS = (
    ('Admin', 'test@example.com'),
)

@app.task
def test(arg):
    raise Exception()

test.delay('Щ')
```

vladimir-g:
This PR adds CELERY_EMAIL_CHARSET option that allows to set charset for outgoing emails. Default value is set to "us-ascii" (as it was in older versions, for compatibility), but that charset causes UnicodeDecodeError in python 3 when email contains unicode characters. With this option user can set charset to more appropriate encoding ("utf-8" or other).

Fixes #2648

### DIFF
--- a/celery/app/base.py
+++ b/celery/app/base.py
@@ -465,6 +465,7 @@ class Celery(object):
                 timeout=conf.EMAIL_TIMEOUT,
                 use_ssl=conf.EMAIL_USE_SSL,
                 use_tls=conf.EMAIL_USE_TLS,
+                charset=conf.EMAIL_CHARSET,
             )
 
     def select_queues(self, queues=None):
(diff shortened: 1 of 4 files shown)


malinoff:
@thedrow please also add ``.. versionadded::`` directive.

thedrow:
@malinoff Done at https://github.com/celery/celery/commit/b815f57de34feb796c844e30b12a86ae7509c39c

PR #2658, opened 2015-06-15
consumer should not retry connection when BROKER_CONNECTION_RETRY is False

natw:
Even when `BROKER_CONNECTION_RETRY` is set to `False`, workers will continuously attempt to reconnect to an unaccessible broker. This changes the consumer behavior so that it will just raise the offending exception and fail out.

This new behavior is particularly desirable in cases where the worker process is run under a process manager such as daemontools or supervisord. If a worker is continuously attempting to reconnect, then we should let the manager handle restarts as to involve existing things like flapping daemon reports.

On a more meta note, this feels like a somewhat less than ideal solution, given that we already have a check for `BROKER_CONNECTION_RETRY` in `Consumer.connect`. But it looks like `Consumer` is actually implementing two different kinds of retry logic. First with `Consumer.connect` calling `conn.ensure_connection`, and then again in `Consumer.start` where it's catching connection errors. So I dunno.

### DIFF
--- a/celery/tests/worker/test_consumer.py
+++ b/celery/tests/worker/test_consumer.py
@@ -115,6 +115,13 @@ class test_Consumer(AppCase):
             c.start()
             sleep.assert_called_with(1)
 
+    def test_no_retry_raises_error(self):
+        self.app.conf.BROKER_CONNECTION_RETRY = False
+        c = self.get_consumer()
+        c.blueprint.start.side_effect = socket.error()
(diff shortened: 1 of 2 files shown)


thedrow:
LGTM. @malinoff @ask What do you think?

malinoff:
It can potentially break something, but I think that current behavior is misleading.
+1 for merging this, but we definitely should cleanup worker code to avoid duplicate logic.

Issue #2654, opened 2015-06-15
The Couchbase results backend should normalize bytes keys to string 


thedrow:
According to #2489 the Couchbase python library raises ValueFormatError when attempting to set a key that using a bytes string.
However #2489 provided the wrong fix for it.
@Smirl suggested to use the str_t method that belogs to the result backend which is the correct fix.

Smirl:
@thedrow You can see my changes in PR #2660 :)

Smirl:
I have noticed that the couchbase tests are skipped on travis because the couchbase python client is not installed. @thedrow is this something that you can add? 

Otherwise we should raise a new issue for this and close this one?

thedrow:
You are right and I have already filed #2517 about it.
Feel free to help.

PR #2660, opened 2015-06-16
#2654 - couchbase - changing the key_t to str_t

Smirl:
Couchbase python API needs to have str type not bytes for the keys. We use kombu.utils.encoding.str_t to make it compatible with Python 2 and 3


thedrow:
This looks like the right fix but I'd love to have a test case covering this bug fix to ensure it won't regress.

Smirl:
@thedrow Ready for you to take a look at the unit test.

thedrow:
Merged! Thank you.

PR #2666, opened 2015-06-19
Always make timezones aware in the schedule even if UTC is disabled

### LINKED ISSUE #943: Celerybeat runs periodic tasks every 5 seconds regardless of interval
I recently upgraded to celery 3 and have been experiencing some strange behavior with celerybeat, which was acting normally before the upgrade.

My environment:

python 2.7.3 
django 1.4.1
virtualenv 
django-celery 3.0.6
celery 3.0.6

Regardless of what interval I define for my periodic tasks - which I adjust through the django admin - celerybeat fires off the periodic tasks every 5 seconds. This occurs regardless of the interval. Interestingly, however, when the periodic task is set disabled, celerybeat stops sending tasks, so it must be responsive to the scheduler at some level.

thedrow:
Fixes #943.
@monax Please verify that this works for you.
In any event, setting CELERY_UTC_ENABLE to true is a good idea if you are using a Django version newer than 1.4.

### DIFF
--- a/celery/schedules.py
+++ b/celery/schedules.py
@@ -134,9 +134,7 @@ class schedule(object):
         return schedstate(is_due=False, next=remaining_s)
 
     def maybe_make_aware(self, dt):
-        if self.utc_enabled:
-            return maybe_make_aware(dt, self.tz)
-        return dt
+        return maybe_make_aware(dt, self.tz)
 
     def __repr__(self):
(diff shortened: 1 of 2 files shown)


thedrow:
@ask @malinoff Since there was a test covering this I'm not really 100% sure that this should be merged.
What am I breaking here?

PR #2670, opened 2015-06-20
django.contrib.admin is already uncommented above

mikeumus:
And I don't think the order or where `django.contrib.admin` is placed is of consequence.


malinoff:
This comment won't affect anybody and since this file is not intended to be used as-is (this is just an example), anybody who is confused by this comment may remove it manually.

mikeumus:
Alright then please just note my confusion in the duplicate. So much so that I re-read more of the entire django admin docs to see if the order or it's placement mattered.

Issue #2081, opened 2014-06-07
MongoDB result backend does not support json serializer.


jaggerwang:
The related configuration as follows:
CELERY_RESULT_BACKEND = 'mongodb://localhost:27017/'
CELERY_MONGODB_BACKEND_SETTINGS = {
    'database': 'celery',
    'taskmeta_collection': 'task_meta'
}
CELERY_ACCEPT_CONTENT = ['json', 'msgpack']
CELERY_TASK_SERIALIZER = "json"
CELERY_RESULT_SERIALIZER = "json"
The error message is ".../lib/python3.4/site-packages/celery/app/trace.py:364: RuntimeWarning: Exception raised outside body: TypeError('data must be an instance of bytes',):".

When changed result serializer to msgpack, it works.
CELERY_RESULT_SERIALIZER = "msgpack"

saritasa:
Yes this is very confusing that everything in mongo is in binary format:

> db.jobs.find().pretty()
{
        "_id" : "efb85a6d-b126-4da9-aad1-846d369f0da0",
        "status" : "SUCCESS",
        "date_done" : ISODate("2015-01-15T21:20:27.345Z"),
        "traceback" : BinData(0,"gAJOLg=="),
        "result" : BinData(0,"gAJ9cQEoVQdtZXNzYWdlcQJVEnlvdSBnb3QgYSBuZXcgc2hmdHEDVQpyZWNlcGllbnRzcQRdcQUoVQZEbWl0cnlxBlUFQm9yeWFxB1UESGllbnEIZVUHY3JlYXRlZHEJVRoyMDE1LTAxLTE1IDEzOjIwOjI3LjM0NTAwMHUu"),
        "children" : BinData(0,"gAJdcQEu")
}

however configuration has this:
CELERY_TASK_SERIALIZER   = 'json'
CELERY_RESULT_SERIALIZER = 'json'
CELERY_ACCEPT_CONTENT    = ['json']
CELERY_TIMEZONE          = 'America/Los_Angeles'
CELERY_ENABLE_UTC        = True
CELERYD_PREFETCH_MULTIPLIER = 128

as a result there is no mongo shell usage to query results of tasks/jobs, instead you need to write Python script to "SCAN" the collection and if you have thousands of jobs, from which you want to extract results only of specific jobs - this is performance hit on Mongodb

```python
import pprint, pickle, json
from jobs.tasks.app import app, t_push_notification
pp = pprint.PrettyPrinter(indent=4)

[...]

PR #2691, opened 2015-07-06
Fix de-serialization of exception (when used with json serializer)

sukrit007:
This PR addresses issue with de-serialization of exception when used with amqp backend with json serializer.

A simple example to reproduce this issue: 
```
@app.task
def fail_task():
    raise IOError('something')

@app.task
def join_task():
    pass

@app.task
def chord_task(self):
    return chord(
        group(
            fail_task.si(),
            fail_task.si(),
        join_task.si()).delay()

# Client code
chord_task.si().delay()
```

The issue seems to be having because get_task_meta of amqp backend caches the result in form of dictionary  and during chord unlock, when [error is thrown] (https://github.com/celery/celery/blob/3.1/celery/result.py#L691),  it results in 
```
TypeError: exceptions must be old-style classes or derived from BaseException, not dict
```

### DIFF
--- a/celery/backends/amqp.py
+++ b/celery/backends/amqp.py
@@ -180,7 +180,8 @@ class AMQPBackend(BaseBackend):
                 raise self.BacklogLimitExceeded(task_id)
 
             if latest:
-                payload = self._cache[task_id] = latest.payload
+                payload = self._cache[task_id] = \
+                    self.meta_from_decoded(latest.payload)
                 latest.requeue()
                 return payload
             else:
--- a/celery/tests/backends/test_amqp.py
+++ b/celery/tests/backends/test_amqp.py
@@ -1,5 +1,6 @@
 from __future__ import absolute_import
 
+import json
 import pickle
 import socket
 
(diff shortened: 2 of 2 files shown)


malinoff:
I have no idea why you rewrite original test from using pickle, can you write another test dedicated for this exact issue?
@ask ping

malinoff:
Btw, @sukrit007 your test isn't passing, inspect TravisCI build.

sukrit007:
@malinoff  I have added separate tests for json.  In addition travis should be passing now - python 3 compatibility)

thedrow:
@malinoff What do you think?

malinoff:
@sukrit007 what about other backends?

thedrow:
Will this PR resolve #2081?

sukrit007:
@thedrow  This PR does not resolve issue in #2081 which seems to be related to python 3 ( as I am not seeing the same behavior with python 2).  
However, it does resolve the conversion of dict to object incase exception was thrown from the task.

sukrit007:
@malinoff   I looked at the base backend and Mongo backend and it seems to be decoding the values correctly:

https://github.com/celery/celery/blob/master/celery/backends/mongodb.py#L210

I haven't got a chance to try other backends but  should we create a separate ticket to address those if they do have this issue ?  Possibly I might be switching to Mongo backend (for results)  but thought will have this PR in-case anyone runs into same while using AMQP.

thedrow:
LGTM. @malinoff @ask ?

malinoff:
@sukrit007 thanks!

PR #2698, opened 2015-07-08
Improve the autoscale behavior

PhilipGarnero:
Autoscale was sometime scaling down -n processes and I figured out I would take a look.
I also changed the _last_action in favor of _last_scale_up as keepalive should prevent stopping processes too soon and not prevent stopping them when needed.
I also fixed what I think was a small bug in update (rescaling was being done even when not necessary)


PhilipGarnero:
Is there something I can do for that build fail ?
It fails also on master

thedrow:
I restarted the build.
If it's still failing I'll check.

thedrow:
The master build is failing with another error.
I have no idea why though.

thedrow:
Thank you!

PR #2699, opened 2015-07-08
How to execute a batches worker with prefetch_multiplier override

mverrilli:
This information seems almost required to use Batches effectively along with other tasks in your application.

### DIFF
--- a/celery/contrib/batches.py
+++ b/celery/contrib/batches.py
@@ -13,6 +13,12 @@ Experimental task class that buffers messages and processes them as a list.
 
     In the future we hope to add the ability to direct batching tasks
     to a channel with different QoS requirements than the task channel.
+    
+    Until then, you can create a worker solely for your batches task 
+    and override the default `CELERYD_PREFETCH_MULTIPLIER` setting by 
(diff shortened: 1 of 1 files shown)


thedrow:
English is correct but I'm not familiar with the mechanism.
@ask @malinoff What do you think?

malinoff:
@thedrow I can't say anything useful about batches, @ask we need your opinion here :)

thedrow:
@malinoff When you need @ask just assign the issue to him.
That way he'll have a list of issues he needs to attend to.

PMickael:
Hi,

I added a custom args some month ago : prefetch-multiplier
So the good syntax to use is

```
celery worker -l info --prefetch-multiplier=1000
@app.task(base=Batches, flush_every=200, flush_interval=2, ignore_result=True, compression='zlib')
```
then it will prefetch 1000 execute every 200 or every 2 sec

mverrilli:
@PaulMcMillan  I didn't realize this was added, very good! I'll close this out and if I don't see any documentation for this I'll add a reference to your new parameter.  Thanks!!

PR #2707, opened 2015-07-14
Fix typeo

aarcro:


aarcro:
That's the failure we see elsewhere. I'm at a loss for why py3 can return none for task, but py2 does not. In:

        while steps:
            task = steps.popleft()

            if not isinstance(task, Signature):
                task = from_dict(task, app=app)
            if isinstance(task, group):
                task = maybe_unroll_group(task)

ask:
Thanks! I fixed the failing tests in master yesterday, and I also now removed the unused propagates stuff (_new_join doesn't need to support this, since that option is only there to enable backward compatible behavior, and will eventually be removed).

ask:
When is task None there? Do you have  a traceback?

aarcro:
@ask task is None when python 3 is running test_group_to_chord https://travis-ci.org/celery/celery/jobs/70960987#L285

aarcro:
So this is closed, but not merged. The result -> results change really is needed.

PR #2715, opened 2015-07-17
Try to fix crontab infinite loop (invalid date)

### LINKED ISSUE #2714: Infinite loop in scheduler crontab
When a *"bogus"* date is used as crontab (eg: *every April, 31st at noon*) , trying to retrieve the  next occurrence of the cron will trigger an infinite loop

Example:

```python
cron = crontab(day_of_month=31, month_of_year=4)
cron.remaining_estimate(datetime.now())
```

romuald:
When next occurrence can never be reached (example, April, 31th), trying to reach the next occurrence would trigger an infinite loop.

Try fixing that by raising a RuntimeError after 2,000 iterations

(Also added a test for crontab leap years in the process)

should fix #2714

**Warning:** in case of a regression, the test will trigger an infinite loop

Feel free to modify exception / error message / # of tries in case anything doesn't feel right

### DIFF
--- a/celery/schedules.py
+++ b/celery/schedules.py
@@ -453,7 +453,7 @@ class crontab(schedule):
             return False
 
         def roll_over():
-            while 1:
+            for _ in range(2000):
                 flag = (datedata.dom == len(days_of_month) or
                         day_out_of_range(datedata.year,
                                          months_of_year[datedata.moy],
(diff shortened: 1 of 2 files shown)


PMickael:
I don't think it has to be fixed, it's more a chair-to-keyboard interface problem :), because you can't be sure than 2000 is enough, I don't know.
Or you need to find an other way

romuald:
IMHO not a PBACK

An infinite loop should never be possible to trigger with user input. It should either be validated or avoided with some kind or workaround (this patch :p)

(The case came from one of our customers making a typo and then blocking the beat process launching tasks)

PMickael:
Yes, a fix would be nice, I was just thinking if there is a better way, at init check maybe.

romuald:
Agreed it should be better at init. However I don't have enough understanding of the crontab logic to do that without being sure of not breaking legit cases :)

Simply validating the day per month would be nice. But it's kinda easy to "break" this by adding something like *day_of_month=31, month=4,6*

ask:
Thanks! Merged into master

PR #2716, opened 2015-07-19
Fixe for pymongo >= 3 and gevent/eventlet

srault95:
Error: ConfigurationError: Unknown option use_greenlets

bug only if use eventlet or gevent (worker -P gevent)


thedrow:
Superseded by #2744.

PR #2728, opened 2015-07-27
Add additional information about the backend_cleanup task

djmitche:
I think this content is fairly straightforward.  However, there's another bit I'd like to add.

I don't use beat, so I need to do the cleanup through other means.  Is it OK to call this task directly (`app.tasks['celery.backend_cleanup'].apply_async()`), and if so, is it OK to add that to the documentation here?

[edit: add `celery.` to task name]


djmitche:
I don't think that travis failure is mine, since I didn't modify code..

thedrow:
Thanks for your contribution.

PR #2753, opened 2015-08-08
Hint for CELERYBEAT_SCHEDULE args for 1-item tuple

darylyu:
It took me a long time to figure out(around 2 days of trying out different things) why I was getting

`celerybeat raised exception : TypeError('argument 2 to map() must support iteration',)`

I was testing it out with a Task that only had one argument and mapped `args` to `(3)` instead of `(3,)`


thedrow:
Thanks!

PR #2756, opened 2015-08-10
Added headers arg to apply_async docs to fix #2750

### LINKED ISSUE #2750: Document headers argument
At the moment there's no much information regarding headers argument for apply_async (and I think also applies to retry).

I found out about this reading the changelog for version 3.1, but maybe it could be specified as an optional property for apply_async in the documentation.

I can provide the PR for this if you want. Let me know if you agree.

juanrossi:


juanrossi:
I'll create a new clean PR

PR #2767, opened 2015-08-18
Make EventReceiver's queue prefix configurable

tkanemoto:
In some message broker instances the queues are access-controlled based on their names, especially when the said broker serves multiple systems.

Allow the event receivers' queue prefix to be configured on the app level, using CELERY_EVENT_QUEUE_PREFIX, rather than always defaulting to 'celeryev'.

### DIFF
--- a/celery/app/defaults.py
+++ b/celery/app/defaults.py
@@ -120,6 +120,7 @@ NAMESPACES = {
         'ENABLE_REMOTE_CONTROL': Option(True, type='bool'),
         'EVENT_SERIALIZER': Option('json'),
         'EVENT_QUEUE_EXPIRES': Option(60.0, type='float'),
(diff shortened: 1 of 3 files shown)


tkanemoto:
I don't understand why my PR specifically fails in the PyPy environment, while others' pass. The error experienced by pip, as far as I can see, is nothing to do with my commit.

    RuntimeError: cryptography 1.0 is not compatible with PyPy < 2.6. Please upgrade PyPy to use this library.

thedrow:
This PR should include tests that verify that this works.

ask:
Thanks! Merged into master

Please add your name to: https://github.com/celery/celery/edit/master/CONTRIBUTORS.txt !

Issue #2669, opened 2015-06-20
Create a new Cassandra results backend based on python-driver


thedrow:
[pycassa](https://github.com/pycassa/pycassa) is pretty old and it doesn't support CQL or Cassandra's native protocol.
We need to implement a new results driver based on DataStax's [python-driver](https://github.com/datastax/python-driver).

piotrmaslanka:
I am currently working on it. I will submit a pull request on ready.

thedrow:
Great! Thanks.

piotrmaslanka:
So far my working version is [here](https://github.com/piotrmaslanka/celery/tree/new-cassandra-backend). No tests written yet, will eventually get back to it.

thedrow:
Looks good. The old Cassandra backend should raise a deprecation warning.

piotrmaslanka:
I believe this issue should be closed.

thedrow:
Yes. Thanks.

PR #2782, opened 2015-08-24
New cassandra backend

### LINKED ISSUE #1763: Use Cassandra as backend to store  task results
setting
```
CELERY_RESULT_BACKEND = 'cassandra',
CASSANDRA_SERVERS = ['localhost:9160'],
CASSANDRA_KEYSPACE = 'celery',
CASSANDRA_COLUMN_FAMILY = 'task_results',
CASSANDRA_READ_CONSISTENCY = 'ONE',
CASSANDRA_WRITE_CONSISTENCY = 'ONE',
CASSANDRA_OPTIONS = {
    'timeout': 300,
    'max_retries': 10
}
```

then
```
CREATE KEYSPACE celery WITH 
    replication = {'class': 'SimpleStrategy', 'replication_factor' : 3};
```

when send a task to celery 
```
Cassandra error: NotFoundException(_message=None,   
    why='Column family task_results not found.')
```

Why didn't automatically creates column family 'task_results' like database backend does, 
and how do I create the column family and set what column_name in it?

piotrmaslanka:
As per #2669 this is a Cassandra backend using DataStax python-driver.

For compatibility reasons it's called new_cassandra, and not cassandra. Configuration options also differ, mainly to force users to understand that it's a different thing.

Using old cassandra backend will raise a deprecation warning.

Schema is different from Thrift one. Assuming keyspace exists, column family will be created by default (fixes #1763).

### DIFF
--- a/celery/backends/new_cassandra.py
+++ b/celery/backends/new_cassandra.py
@@ -36,9 +36,6 @@ class NewCassandraBackend(BaseBackend):
         module :mod:`cassandra` is not available.
 
     """
-    servers = []
-    keyspace = None
-    table = None
     supports_autoexpire = True      # autoexpire supported via entry_ttl
 
     def __init__(self, servers=None, keyspace=None, table=None, entry_ttl=None,
@@ -58,16 +55,13 @@ class NewCassandraBackend(BaseBackend):
 
         conf = self.app.conf
         self.servers = (servers or
-                        conf.get('CASSANDRA_SERVERS') or
-                        self.servers)
+                        conf.get('CASSANDRA_SERVERS'))
         self.port = (port or
                      conf.get('CASSANDRA_PORT'))
         self.keyspace = (keyspace or
-                         conf.get('CASSANDRA_KEYSPACE') or
-                         self.keyspace)
+                         conf.get('CASSANDRA_KEYSPACE'))
         self.table = (table or
-                      conf.get('CASSANDRA_TABLE') or
-                      self.table)
+                      conf.get('CASSANDRA_TABLE'))
         expires = (entry_ttl or conf.get('CASSANDRA_ENTRY_TTL', None))
 
         if expires is not None:
--- a/celery/backends/new_cassandra.py
+++ b/celery/backends/new_cassandra.py
@@ -105,7 +105,7 @@ class NewCassandraBackend(BaseBackend):
             # We are forced to do concatenation below, as formatting would
             # blow up on superficial %s that will be processed by Cassandra
             self._write_stmt = cassandra.query.SimpleStatement(
-                'INSERT INTO %s (task_id, status, result,'''
+                'INSERT INTO '+self.table+''' (task_id, status, result,'''
                 ''' date_done, traceback, children) VALUES'''
                 ' (%s, %s, %s, %s, %s, %s) '+self.cqlexpires+';')
             self._write_stmt.consistency_level = self.write_consistency
--- a/celery/backends/new_cassandra.py
+++ b/celery/backends/new_cassandra.py
@@ -9,6 +9,7 @@
 from __future__ import absolute_import
 
 import sys
+import six
 try:  # pragma: no cover
     import cassandra
 except ImportError:  # pragma: no cover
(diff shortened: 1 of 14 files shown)


celery/backends/new_cassandra.py
```diff
@@ -0,0 +1,199 @@
...
+
+    .. attribute:: servers
+
+        List of Cassandra servers with format: ``hostname``
+
+    :raises celery.exceptions.ImproperlyConfigured: if
+        module :mod:`cassandra` is not available.
+
+    """
+    servers = []
```
malinoff: Please, avoid mutable class-level objects. IMO we don't need all these connection-related parameters on class level.

celery/backends/new_cassandra.py
```diff
@@ -0,0 +1,199 @@
...
+
+        if expires is not None:
+            self.cqlexpires = ' USING TTL %s' % (expires, )
+        else:
+            self.cqlexpires = ''
+
+        read_cons = conf.get('CASSANDRA_READ_CONSISTENCY') or 'LOCAL_QUORUM'
+        write_cons = conf.get('CASSANDRA_WRITE_CONSISTENCY') or 'LOCAL_QUORUM'
+        try:
+            self.read_consistency = getattr(cassandra.ConsistencyLevel,
```
malinoff: Specify default value for `getattr` instead of catching `AttributeError`, same for the lines below

celery/backends/new_cassandra.py
```diff
@@ -0,0 +1,199 @@
...
+                                            read_cons)
+        except AttributeError:
+            self.read_consistency = cassandra.ConsistencyLevel.LOCAL_QUORUM
+        try:
+            self.write_consistency = getattr(cassandra.ConsistencyLevel,
+                                             write_cons)
+        except AttributeError:
+            self.write_consistency = cassandra.ConsistencyLevel.LOCAL_QUORUM
+
+        if not self.servers or not self.keyspace or not self.table:
```
malinoff: This check should be moved right after all needed parameters are evaluated to reduce unnecessary function calls.

celery/backends/new_cassandra.py
```diff
@@ -0,0 +1,199 @@
...
+        Prepare the connection for action
+
+        :param write: bool - are we a writer?
+        """
+        if self._connection is None:
+            self._connection = cassandra.cluster.Cluster(self.servers,
+                                                         port=self.port)
+            self._session = self._connection.connect(self.keyspace)
+
+            self._write_stmt = cassandra.query.SimpleStatement(
```
malinoff: Consider using a template string and `format` method instead of concatenation.

celery/backends/new_cassandra.py
```diff
@@ -0,0 +1,199 @@
...
+                except cassandra.AlreadyExists:
+                    pass
+
+    def _store_result(self, task_id, result, status,
+                      traceback=None, request=None, **kwargs):
+        """Store return value and status of an executed task."""
+        self._get_connection(write=True)
+
+        import sys
+        if six.PY3:
```
malinoff: Move to the module level, also use `celery.five` instead of `six`

thedrow:
Can you please rebase?

piotrmaslanka:
Rebased all right.

egon1024:
Just curious if there's any sort of timeline on getting this into Celery?  We're looking to use it in a new project that's ramping up right now (Celery with Cassandra results storage, but on Python 3 so pycassa isn't an option).

thedrow:
I'll speak with @ask about it first but it looks good.

ask:
How about we keep it as `new_cassandra` in 3.1, but drop the old one in 3.2?

ask:
Good work! Tested it and it seems to work.  Merged into master

thedrow:
@ask That would kinda break semantic versioning that I think we should adopt.
Keep it as new_cassandra for now.

egon1024:
Fantastic!  Thanks for merging this folks!  Any idea when the next release will be that will contain this?

thedrow:
@egon1024 Sorry, we don't yet. Unfortunately we have some pretty critical bugs to resolve before we release a new major version.
I'm also working to rewrite our test suite to be more extensive and include end to end tests.

egon1024:
No worries - keep up the great work, it's very much appreciated.  :)

ask:
@thedrow 3.2 should probably be 4.0 at this point.
What do we do when the `new_new_cassandra` backend comes?

semantic versioning is good, but adding experimental features to point releases can be useful, but only
if it's completely orthogonal to existing features (like the qpid transport, or new result backends).

thedrow:
If it's 4.0 then we can rename the backend. Yes.

davidharrigan:
@ask @piotrmaslanka My team is looking to use Cassandra for our Celery backend, but I'm not sure if our timeline matches for when this feature becomes stable, so I am looking to patch up our local mirror of Celery with this feature. What are the steps that I should take about doing that?  I also don't see any code diff in this PR that would enable new Cassandra backend.

piotrmaslanka:
@davidharrigan I guess celery/backends/__init__.py and celery/backends/new_cassandra.py are the only files to add/modify in this case. celery/master is fine to fork from, just please patch against #2901 or wait until this is integrated. I will be running this backend in production from 5th Nov, so if there are any bugs I'll get to fix them ASAP.

davidharrigan:
Awesome, thanks for the quick response :D

piotrmaslanka:
@davidharrigan There was a small revolution in the master - old cassandra is no more. Please be aware of that, and repatch against #2901. I rebased it for simpler inclusion.

PR #2789, opened 2015-08-26
Fixed #2785 -- Changed multi stopwait to wait on pidfiles instead of process.

### LINKED ISSUE #2785: multi stopwait: possible race: kill can return EPERM if pid gets recycled
I run celery as a non-root user. Occasionally when running `multi stopwait` I'll get an unhandled exception: "OSError: [Errno 1] Operation not permitted". Full trace below.

I investigated this a bit.

In `celery/bin/multi.py`, `stopwait()` eventually ends up in a loop inside `shutdown_nodes()`. Inside, this loop, a signal is sent to the celery process until `ESRCH` is received (signal sent in `node_alive()`). However, inside this loop, while the multi process is sleeping, the celery process could go away and a new process could appear with the same pid; as the OS recycles pids. If this new process, with the same pid, was started by a different user, the kill system call will return `EPERM` (this gets translated to a Python exception).

It appears, Celery is not able to handle this case and so results in an unhandled exception like the one below.

See http://linux.die.net/man/3/kill section "Errors" for a full list of errors returned by the `kill` system call.

```
celery multi v3.1.18 (Cipater)
> Stopping nodes...
    > company@company.com: TERM -> 16451
> Waiting for 1 node -> 16451.....Traceback (most recent call last):
  File "/path/to/venv/a9ec6e728550daa597ef368dc7fa04887aa8577b/bin/celery", line 11, in <module>
    sys.exit(main())
  File "/path/to/venv/a9ec6e728550daa597ef368dc7fa04887aa8577b/lib/python2.6/site-packages/celery/__main__.py", line 30, in main
    main()
  File "/path/to/venv/a9ec6e728550daa597ef368dc7fa04887aa8577b/lib/python2.6/site-packages/celery/bin/celery.py", line 81, in main
    cmd.execute_from_commandline(argv)
  File "/path/to/venv/a9ec6e728550daa597ef368dc7fa04887aa8577b/lib/python2.6/site-packages/celery/bin/celery.py", line 769, in execute_from_commandline
    super(CeleryCommand, self).execute_from_commandline(argv)))
  File "/path/to/venv/a9ec6e728550daa597ef368dc7fa04887aa8577b/lib/python2.6/site-packages/celery/bin/base.py", line 311, in execute_from_commandline
    return self.handle_argv(self.prog_name, argv[1:])
  File "/path/to/venv/a9ec6e728550daa597ef368dc7fa04887aa8577b/lib/python2.6/site-packages/celery/bin/celery.py", line 761, in handle_argv
    return self.execute(command, argv)
  File "/path/to/venv/a9ec6e728550daa597ef368dc7fa04887aa8577b/lib/python2.6/site-packages/celery/bin/celery.py", line 693, in execute
    ).run_from_argv(self.prog_name, argv[1:], command=argv[0])
  File "/path/to/venv/a9ec6e728550daa597ef368dc7fa04887aa8577b/lib/python2.6/site-packages/celery/bin/celery.py", line 97, in run_from_argv
    [command] + argv, prog_name,
  File "/path/to/venv/a9ec6e728550daa597ef368dc7fa04887aa8577b/lib/python2.6/site-packages/celery/bin/multi.py", line 209, in execute_from_commandline
    self.commands[argv[0]](argv[1:], cmd)
  File "/path/to/venv/a9ec6e728550daa597ef368dc7fa04887aa8577b/lib/python2.6/site-packages/celery/bin/multi.py", line 394, in stopwait
    return self._stop_nodes(p, cmd, retry=2)
  File "/path/to/venv/a9ec6e728550daa597ef368dc7fa04887aa8577b/lib/python2.6/site-packages/celery/bin/multi.py", line 372, in _stop_nodes
    callback=callback)
  File "/path/to/venv/a9ec6e728550daa597ef368dc7fa04887aa8577b/lib/python2.6/site-packages/celery/bin/multi.py", line 320, in shutdown_nodes
    if not self.node_alive(pid):
  File "/path/to/venv/a9ec6e728550daa597ef368dc7fa04887aa8577b/lib/python2.6/site-packages/celery/bin/multi.py", line 277, in node_alive
    os.kill(pid, 0)
OSError: [Errno 1] Operation not permitted
```

jdufresne:
Sending signals to kill a process has a race condition where a pid may
be recycled by a new, unrelated process.

Instead, now use watchdog to watch the directory containing the
pidfile. Hold a lock until the pidfile is deleted. Once all pidfiles
are deleted, all celery processes have finished.

Introduces a dependecy on watchdog.

Fixes #2785

---

All feedback welcome.

### DIFF
--- a/celery/bin/multi.py
+++ b/celery/bin/multi.py
@@ -106,7 +106,10 @@ import sys
 from collections import OrderedDict, defaultdict, namedtuple
 from functools import partial
 from subprocess import Popen
-from time import sleep
+import threading
+
+import watchdog.events
+import watchdog.observers
 
 from kombu.utils import cached_property
 from kombu.utils.encoding import from_utf8
@@ -275,16 +278,38 @@ class MultiTool(object):
             return False
         return True
 
-    def node_alive(self, pid):
-        try:
-            os.kill(pid, 0)
-        except OSError as exc:
-            if exc.errno == errno.ESRCH:
-                return False
-            raise
-        return True
+    def init_pidfile_observers(self, nodes):
+        handlers = []
+        observer = watchdog.observers.Observer()
+        for node in nodes:
+            _, _, _, pidfile = node
+            pidfile = os.path.abspath(pidfile)
+            handler = PidfileDeleteHandler(node, pidfile)
+            observer.schedule(handler, os.path.dirname(pidfile))
+            handlers.append(handler)
+        observer.start()
+        return observer, handlers
 
-    def shutdown_nodes(self, nodes, sig=signal.SIGTERM, retry=None,
+    def wait_pidfiles(self, nodes, observer, handlers, on_down):
+        def note_waiting():
+            left = len(nodes)
+            if left:
+                pids = ', '.join(str(pid) for _, _, pid, _ in nodes)
+                self.note(self.colored.blue(
+                    '> Waiting for {0} {1} -> {2}...'.format(
+                        left, pluralize(left, 'node'), pids)), newline=False)
+
+        for handler in handlers:
+            note_waiting()
+            handler.lock.acquire()
+            nodename, _, _, _ = handler.node
+            self.note('\n\t> {0}: {1}'.format(nodename, self.OK))
+            on_down(handler.node)
+
+        observer.stop()
+        observer.join()
+
+    def shutdown_nodes(self, nodes, sig=signal.SIGTERM, wait=False,
                        callback=None):
         if not nodes:
             return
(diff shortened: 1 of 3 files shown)


thedrow:
It seems like you need to adjust one of the celery multi shutdown tests.
Check the build: https://travis-ci.org/celery/celery/jobs/77433748#L907
Could we make watchdog optional, so it will only be required if celery multi is used?

jdufresne:
> Check the build: https://travis-ci.org/celery/celery/jobs/77433748#L907

Thanks, I'll take a look.

>  Could we make watchdog optional, so it will only be required if celery multi is used?

Sure thing, makes sense. Should it fall back to the previous implementation if watchdog does not exist?

thedrow:
I don't think it should since it introduces an inconsistent state.

jdufresne:
> I don't think it should since it introduces an inconsistent state.

I'm not sure I'm 100% what you're asking for.

So if watchdog is unavailable, does that mean the stopwait command is also unavailable from multi?

Or do you mean, continue to always import watchdog from multi.py. So it is always required when running the multi command, but no other part of celery?

thedrow:
Make watchdog required when using celery multi but not required when it is not used.

ask:
I would prefer to not use watchdog at all, as if we add more dependencies there will be a large mob with pitchforks screaming at us how complicated celery is because installing it logs so much, or something similar.

More importantly though, watchdog seems to rely on threading which we cannot use as one of the goals with multi is the ability to use it as a library. E.g. cyme uses it to start and stop processes, and you cannot use threads in combination with `fork()`.  Every time I have used threads, it ends up being a major pain, so we are trying to be completely free of them.

It's also a C library, which is something we would like to avoid depending on by default.

We also already have code to watch files in celery/worker/autoreload, perhaps that could be generalized enough to be reused here?

ask:
We should probably not cache the value of the pid for long, and instead reread the pidfile just before we kill the process, possibly that can be combined with checking /proc contents for args or something on Linux for extra safety.

Programs have used these mechanisms without filesystem event support for decades, so it must be possible to handle in some way compatible with old IRIX systems or whatever.

If using watchdog to watch the pid we will still have the race condition where the process is killed remotely, not giving it a chance to remove the pidfile.

ask:
it should be enough to check for parent pid come to think of it

PR #2790, opened 2015-08-27
Adding the __bool__ and __nonzero__ methods to ResultSet

colinmcintosh:
This is to fix an issue where a ResultSet or GroupResult with an empty
result list are not properly tupled with the as_tuple() method when it is
a parent result. This is due to the as_tuple() method performing a logical
and operation on the ResultSet.

A snippet from celery/result.py:AsyncResult.as_tuple()

```python
return (self.id, parent and parent.as_tuple()), None
```

This is testing to make sure parent is not boolean False; if it is not boolean False then return the tuple. Per the Python docs an 'and' operation first checks \__bool__ then \__len__ if \__bool__ is not defined.[1][2] In the case of ResultSet and GroupResult \__bool__ is not defined but \__len__ is which returns the length of the ResultSet.results. If the list of results is empty the \__len__ method will return 0 which equates to False. This results in the 'and' operation return the first item, in this case the ResultSet or GroupResult instance.

I have added the \__bool__ method so that any ResultSet instances equate to boolean True as long as they exist.

The \__bool__ method is aliased to the  \__nonzero__ method for backwards compatibility with Python2.

Here is some PoC code that demonstrates the issue

```python
import uuid
from celery.result import AsyncResult, GroupResult


def main():
	gr_id = str(uuid.uuid4())
	ar_id = str(uuid.uuid4())
	group_result = GroupResult(id=gr_id, results=[])
	async_result = AsyncResult(id=ar_id, parent=group_result)
	print(async_result.as_tuple())	# Would expect all tuples but it contains a GroupResult instance


def proof():
	gr_id = str(uuid.uuid4())
	group_result = GroupResult(id=gr_id, results=[])
	print(group_result and group_result.as_tuple())  # Would expect a tuple but it's a GroupResult instance


if __name__ == "__main__":
	main()
	proof()
```

References:
[1] [Explanation of how the and operator works](https://docs.python.org/3.4/reference/expressions.html#boolean-operations)
[2] [\__len__ is called if \__bool__ is not found](https://docs.python.org/3.4/reference/datamodel.html#object.__bool__)

### DIFF
--- a/celery/result.py
+++ b/celery/result.py
@@ -714,10 +714,6 @@ class ResultSet(ResultBase):
     def __len__(self):
         return len(self.results)
 
-    def __bool__(self):
-        return True
-    __nonzero__ = __bool__  # Included for Py2 backwards compatibility
-
     def __eq__(self, other):
         if isinstance(other, ResultSet):
             return other.results == self.results
@@ -803,6 +799,10 @@ class GroupResult(ResultSet):
 
     def __reduce_args__(self):
         return self.id, self.results
+
+    def __bool__(self):
+        return bool(self.id or self.results)
+    __nonzero__ = __bool__  # Included for Py2 backwards compatibility
 
     def __eq__(self, other):
         if isinstance(other, GroupResult):
--- a/celery/result.py
+++ b/celery/result.py
@@ -714,6 +714,10 @@ class ResultSet(ResultBase):
     def __len__(self):
         return len(self.results)
 
+    def __bool__(self):
+        return True
+    __nonzero__ = __bool__  # Included for Py2 backwards compatibility
+
     def __eq__(self, other):
         if isinstance(other, ResultSet):
             return other.results == self.results
--- a/celery/tests/app/test_builtins.py
+++ b/celery/tests/app/test_builtins.py
@@ -100,8 +100,10 @@ class test_group(BuiltinsCase):
         x = group(app=self.app)
         x.apply()
         res = x.apply_async()
-        self.assertFalse(res)
         self.assertFalse(res.results)
+        self.assertTrue(res)
+        res.id = None
+        self.assertFalse(res)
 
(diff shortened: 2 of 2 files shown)


thedrow:
I'm not sure that always evaluating the group result as true is a good idea.
See the test failure https://travis-ci.org/celery/celery/jobs/77433941#L300

colinmcintosh:
I agree that it always evaluating to True is not a good idea but I don't think that GroupResult should be False only because it doesn't have a populated result list. GroupResult also can contain a valid result ID which I feel should make it not False. I think that because ResultSet only represents the list of results and no ID that it can stay the way it is and equate to False if the list is empty.

I would propose changing the PR to move the added methods to GroupResult and checking to see if an ID is present. Something like this:

```python
class GroupResult(ResultSet):
    ...
    def __bool__(self):
        return bool(self.id or self.results)
    __nonzero__ = __bool__  # Included for Py2 backwards compatibility
    ...
```

This would also mean that the unittest would have the change. Either testing if the whole GroupResult is False can be removed or it can be adjusted to check for both boolean states. I made the latter change as an example in my updated PR.

thedrow:
Don't you think each item in the group should have at least a result of None?

colinmcintosh:
I'm not sure what you mean by every item in the group having a result of None. Do mean that every result object in .results should equate to None?

thedrow:
If there's no result in a task in the group then I think it should.
E.g.
```python
@app.task
def a():
  print("hello")

print(group(a.s(), a.s()).delay().get())
```
Should print ```[None, None]``` no?

ask:
Maybe easier to just use `parent.as_tuple() if parent else None`? Or am I misunderstanding this?

thedrow:
@ask I don't know, you tell me :)

ask:

PR #2794, opened 2015-08-28
Fix  example code demonstrating Custom Camera

schmiddy:
The existing example code demonstrating a Custom Camera has a buggy print statement, and will give output like:

> Total: 974 events, %s 50


thedrow:
Thanks!

PR #2796, opened 2015-08-31
Redis: experimental new_join setting ignores the propagate preference for redis backend.

dawalama:
When using chord(group(task_set))(callback) 

Without this patch, chord won't call callback if one of the task in task_set throws an exception.

### DIFF
--- a/celery/backends/redis.py
+++ b/celery/backends/redis.py
@@ -180,13 +180,13 @@ class RedisBackend(KeyValueStoreBackend):
     def add_to_chord(self, group_id, result):
         self.client.incr(self.get_key_for_group(group_id, '.t'), 1)
 
-    def _unpack_chord_result(self, tup, decode,
+    def _unpack_chord_result(self, tup, decode, propagate,
                              EXCEPTION_STATES=states.EXCEPTION_STATES,
                              PROPAGATE_STATES=states.PROPAGATE_STATES):
         _, tid, state, retval = decode(tup)
         if state in EXCEPTION_STATES:
             retval = self.exception_to_python(retval)
-        if state in PROPAGATE_STATES:
+        if propagate and state in PROPAGATE_STATES:
             raise ChordError('Dependency {0} raised {1!r}'.format(tid, retval))
         return retval
 
@@ -228,7 +228,7 @@ class RedisBackend(KeyValueStoreBackend):
                     .delete(tkey)               \
                     .execute()
                 try:
-                    callback.delay([unpack(tup, decode) for tup in resl])
+                    callback.delay([unpack(tup, decode, propagate) for tup in resl])
                 except Exception as exc:
                     error('Chord callback for %r raised: %r',
                           request.group, exc, exc_info=1)


thedrow:
We need a test case that verifies that the propagate preference is taken into consideration. Otherwise this issue might regress.

dawalama:
Okay, I will add a test case and push again

ask:
This was actually deliberate, as propagates re-enables the undefined behavior we no longer want to support.

Passing the exception to the callback is not universally useful, since exceptions can only be deserialized by pickle.  It's also easy to crash tasks that don't except the type of the argument to change.

propagate was more of an accidental feature, since I had literally forgotten to deal with errors in the first
iteration of chords ;)

thedrow:
The proposed behavior makes sense. If I set propagate to true I expect an exception to be raised. We can raise an exception if propagate is true and the serializer is not set to pickle.

dawalama:
The behavior is supported with base amqp backend. I spent quite a bit of time scratching my head why existing code started not working when switching to redis backend.

My use-case was do a bunch of tasks, collect any exceptions and valid results, pass them to callback. The setup is this way because the next actions taken depends on the collective results of the tasks run in chord group. Failures cannot be handled one off because of business requirement to bundle them together for reporting and alerting as well. 

Yes, there is a way to break up the above logic in other ways, but that would take a much longer effort.

dawalama:
@thedrow @ask I am little confused. Should we add a test case or should we abandon this path like @ask indicated?

thedrow:
@dawalama We seem to be in a disagreement at the moment.
@malinoff @PMickael @ionelmc What do you guys think?

ask:
CELERY_CHORD_PROPAGATES is completely gone in 4.0.

The problem with propagate is that, having a task with signature:

```
@app.task
def xsum(results):
     return sum(results)
```

A workflow like: `group(add.s(4, 4), raising.s(), add.s(8, 8) | xsum.s()`

will cause `xsum` to be called with an exception instance instead of a list, which means
that any chord callback must take this semi predicate into account or be catched by surprise:

```
@app.task
def xsum(results):
    if isinstance(results, BaseException):
        raise results
    return sum(results)
```

The initial chord implementation had this behavior entirely by accident.

Now when the json serializer is used by default it will be even harder to conform to this,
as you would have to do something like:

```
@app.task(bind=True)
def xsum(self, result):
    if isinstance(result, (dict, BaseException)):
        raise self.backend.exception_to_python(result)
    return sum(result)
```
Not exactly friendly, and highly ambiguous.

So new_join is an incentive to transition away from the problematic behavior giving better
performance.  It's gone completely in 4.0 anyway, and I think in no way should it be reintroduced
unless we have a clean solution.

ask:
You can think of `?new_join=1` as being `from __future__ import new_join`.

With the patch above you can use `link_error` to be notified of errors:

```
@app.task(bind=True)
def on_error(self, id):
    exc = self.AsyncResult(id).get(propagate=False)
    print('Chord raised error: {0!r}'  % (exc,))
    # do something

group(add.s(2, 2), raises.s()) | xsum.s().set(link_error=on_error.s())
```

PR #2802, opened 2015-09-08
Fix ImportError

azalea:
With 'kombu.transport.django.KombuAppConfig' in INSTALLED_APPS,
running any manage.py command throws:

ImportError: No module named KombuAppConfig

It is fixed by changing 'kombu.transport.django.KombuAppConfig' to 'kombu.transport.django'


ask:
Thanks!

PR #2810, opened 2015-09-16
Removed an extra on_retry entry in the docs.

sww:
There's an extra `on_retry` entry here: http://docs.celeryproject.org/en/master/userguide/tasks.html?highlight=on_retry#on-retry.


thedrow:
Thanks!

PR #2813, opened 2015-09-18
Set celery priority in supervisord higher

raratiru:
The setting for priority refers to ```[program:celery]```

However, the comments either refer to a configuration setting of an unreal rabbitmq section which guarantees that rabbitmq will start first, or they are plain wrong since according to the [docs](http://supervisord.org/configuration.html#program-x-section-values):

> Higher priorities indicate programs that start last and shut down first.

I propose to:

1) Set the celery priority higher than the default (999)
2) Provide an explanation why


ask:
Thanks!

PR #2821, opened 2015-09-23
Clearly explain how to use memory cache backend

asfaltboy:
Admittedly I am slow, but it shouldn't have taken me 10 minutes to figure out how to set up cache/memory (very useful for `unittest`s).

(Note: Resubmitted PR to master branch instead of 3.1, and based on master's HEAD - 3rd time's a charm)


thedrow:
Thanks!

Issue #1847, opened 2014-02-04
Worker hangs at startup when running without mingle and gossip and many messages in queue 


sabw8217:
See test case here:
https://github.com/sabw8217/celery_test

If I enqueue a number of tasks, and then start a worker with --without-mingle and --without-gossip, the worker seems to hang and not actually run any of the tasks I enqueued. I'm running this on debian. On my dev box I have pretty consistently been able to cause the worker to "wake up" and start consuming tasks by issuing a 'celery inspect' command. 'celery inspect active' will return something like:
-> celery@aaron-dev.localdomain: OK
    - empty -
And then worker the will go through the tasks. 

Debug output from the worker looks like:
[2014-02-04 13:53:47,628: DEBUG/MainProcess] | Worker: Starting Hub
[2014-02-04 13:53:47,628: DEBUG/MainProcess] ^-- substep ok
[2014-02-04 13:53:47,628: DEBUG/MainProcess] | Worker: Starting Pool
[2014-02-04 13:53:47,633: DEBUG/MainProcess] ^-- substep ok
[2014-02-04 13:53:47,634: DEBUG/MainProcess] | Worker: Starting Consumer
[2014-02-04 13:53:47,635: DEBUG/MainProcess] | Consumer: Starting Connection
[2014-02-04 13:53:47,644: INFO/MainProcess] Connected to amqp://guest@localhost:5672//
[2014-02-04 13:53:47,645: DEBUG/MainProcess] ^-- substep ok
[2014-02-04 13:53:47,645: DEBUG/MainProcess] | Consumer: Starting Events
[2014-02-04 13:53:47,659: DEBUG/MainProcess] ^-- substep ok
[2014-02-04 13:53:47,661: DEBUG/MainProcess] | Consumer: Starting Heart
[2014-02-04 13:53:47,663: DEBUG/MainProcess] ^-- substep ok
[2014-02-04 13:53:47,663: DEBUG/MainProcess] | Consumer: Starting Tasks
[2014-02-04 13:53:47,666: DEBUG/MainProcess] basic.qos: prefetch_count->4
[2014-02-04 13:53:47,667: DEBUG/MainProcess] ^-- substep ok
[2014-02-04 13:53:47,667: DEBUG/MainProcess] | Consumer: Starting Control
[2014-02-04 13:53:47,669: DEBUG/MainProcess] ^-- substep ok
[2014-02-04 13:53:47,670: DEBUG/MainProcess] | Consumer: Starting event loop
[...]

PR #2823, opened 2015-09-24
Explicitly drain events when gossip/heartbeat is disabled.

bryanhelmig:
This is the fix for issue #1847 - kudos to @sabw8217 who's fix I copied verbatim.

AFAIK this bug exists because usually you can rely on the new gossip mechanism to implicitly drain events even if you've grabbed your prefetch limit on startup - in the absence of gossip draining the events we should do it explicitly instead.

The included test case doesn't illustrate this at all though. I've done lots of local tests.

### DIFF
--- a/celery/tests/worker/test_loops.py
+++ b/celery/tests/worker/test_loops.py
@@ -120,6 +120,10 @@ class test_asynloop(AppCase):
             return x + y
         self.add = add
 
+    def test_drain_after_consume(self):
+        x, _ = get_task_callback(self.app)
+        x.connection.drain_events.assert_called_with()
+
     def test_setup_heartbeat(self):
         x = X(self.app, heartbeat=10)
         x.hub.call_repeatedly = Mock(name='x.hub.call_repeatedly()')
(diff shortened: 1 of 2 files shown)


bryanhelmig:
We're planning to port this back to 3.1.18 and push to production - happy to report back.

ask:
Nice work @sabw8217, @bryanhelmig!

Merged into master and 3.1, so please test to make sure it's working

ask:
Btw, can you also add your name to http://github.com/celery/celery/edit/master/CONTRIBUTORS.txt ?

bryanhelmig:
I think perhaps @sabw8217 should be on the contributor list if they wish - I doubt I can appropriately claim such any such title. ;-)

bryanhelmig:
Also @ask you might close the related https://github.com/celery/celery/issues/1847.

mverrilli:
Wow thanks so much for this fix.  This exactly explains the behavior I've run into.  I'll be testing this out this week.

Issue #1628, opened 2013-11-09
Message being acknowledged on WorkerLostError when CELERY_ACKS_LATE=True


siddharth96:
When using celery v3.0.24, with `CELERY_ACKS_LATE = True` , if the OOM killer kills the celery worker, then the worker acknowledges the message.
As per [this](https://github.com/celery/celery/commit/e810420c) commit. The `exc_info.internal` comes in as `false`, which means it is not a internal error, due to which the message is acknowledged.
The desirable behaviour, in such a case would be to not acknowledge the message (and be able to know, whether its a OOM error), so that some other worker can pick it up. 
As a workaround, I've commented out the [code](https://github.com/siddharth96/celery/commit/427695d1b23034dadda85fd7a48f7367831be4fa), where celery acknowledges the message, because in such a case, message will be lost.

mpermana:
https://github.com/celery/celery/pull/2840

mpermana:
@ask Added configuration option for not acking terminated tasks

ask:
We have an option for this in master now

siddharth96:
I was just trying this set of changes, and when I raised an error in the consumer, the message didn't even went through this flow. ```on_failure``` defined in ```celery/worker/jobs.py``` got called, while the change here is at another place.
Should we re-open this?

thedrow:
@siddharth96 Can you provide a test case that reproduces the problem?

odedfos:
Acknowledging the task on WorkerLost is very misleading, when acks_late=True.
Processes get killed for numerous reasons, and even when OOM, the conditions on the machine could be that the memory issue isn't related directly to the killed process.

At the least, I think this behaviour should be configurable, and allow for the task not to be acknowledged.
I find myself implementing a watchdog searching for lost workers and requeueing those tasks instead of having the framework do as it suggests (according to documentation)

Adamantish:
:+1: for configurability

thedrow:
I think we can accept a configuration knob which implements this but I'm not sure it's worth it.
Perhaps there should a hook instead which you can configure to customize acknowledgment behavior.
@celery/core-developers What do you think?

galCohen88:
Not sure how this can be implemented, if there is a power shutdown there will be no guarantee for acking acks late message. 
The problem originates from non-idempotent tasks

shubham088:
Hello, i have this use case.
[...]

PR #2840, opened 2015-10-06
Fix issue #1628

mpermana:
https://github.com/celery/celery/issues/1628

### DIFF
--- a/celery/worker/request.py
+++ b/celery/worker/request.py
@@ -353,7 +353,7 @@ class Request(object):
         if self.task.acks_late:
             reject_and_requeue = (self.task.reject_on_worker_lost and
                 isinstance(exc, WorkerLostError) and
-                not self.delivery_info.get('redelivered', False))
+                self.delivery_info.get('redelivered', False) is False)
             if reject_and_requeue:
                 self.reject(requeue=True)
             else:
--- a/celery/worker/request.py
+++ b/celery/worker/request.py
@@ -326,7 +326,6 @@ class Request(object):
     def on_failure(self, exc_info, send_failed_event=True, return_ok=False):
         """Handler called if the task raised an exception."""
         task_ready(self)
-
         if isinstance(exc_info.exception, MemoryError):
             raise MemoryError('Process got: %s' % (exc_info.exception,))
         elif isinstance(exc_info.exception, Reject):
@@ -352,8 +351,11 @@ class Request(object):
                 )
         # (acks_late) acknowledge after result stored.
         if self.task.acks_late:
-            if isinstance(exc, WorkerLostError):
-                self.reject(True)
+            reject_and_requeue = (self.task.reject_on_worker_lost and
+                isinstance(exc, WorkerLostError) and
+                not self.delivery_info.get('redelivered', False))
+            if reject_and_requeue:
+                self.reject(requeue=True)
             else:
                 self.acknowledge()
 
--- a/celery/worker/request.py
+++ b/celery/worker/request.py
@@ -352,7 +352,7 @@ class Request(object):
                 )
         # (acks_late) acknowledge after result stored.
         if self.task.acks_late:
-            if self.task.reject_on_worker_lost and isinstance(exc, WorkerLostError):
+            if isinstance(exc, WorkerLostError):
                 self.reject(True)
             else:
                 self.acknowledge()
--- a/celery/worker/request.py
+++ b/celery/worker/request.py
@@ -352,7 +352,10 @@ class Request(object):
                 )
         # (acks_late) acknowledge after result stored.
         if self.task.acks_late:
-            self.acknowledge()
+            if self.task.reject_on_worker_lost and isinstance(exc, WorkerLostError):
+                self.reject(True)
+            else:
+                self.acknowledge()
 
         if send_failed_event:
             self.send_event(
--- a/celery/app/defaults.py
+++ b/celery/app/defaults.py
@@ -132,6 +132,7 @@ NAMESPACES = {
         'REDIS_DB': Option(type='int', **_REDIS_OLD),
         'REDIS_PASSWORD': Option(type='string', **_REDIS_OLD),
         'REDIS_MAX_CONNECTIONS': Option(type='int'),
+        'REJECT_ON_WORKER_LOST': Option(type='bool'),
         'RESULT_BACKEND': Option(type='string'),
         'RESULT_DB_SHORT_LIVED_SESSIONS': Option(False, type='bool'),
         'RESULT_DB_TABLENAMES': Option(type='dict'),
--- a/celery/app/defaults.py
+++ b/celery/app/defaults.py
@@ -132,7 +132,6 @@ NAMESPACES = {
         'REDIS_DB': Option(type='int', **_REDIS_OLD),
         'REDIS_PASSWORD': Option(type='string', **_REDIS_OLD),
         'REDIS_MAX_CONNECTIONS': Option(type='int'),
-        'REJECT_ON_WORKER_LOST': Option(type='bool'),
         'RESULT_BACKEND': Option(type='string'),
         'RESULT_DB_SHORT_LIVED_SESSIONS': Option(False, type='bool'),
         'RESULT_DB_TABLENAMES': Option(type='dict'),
--- a/celery/app/task.py
+++ b/celery/app/task.py
@@ -220,6 +220,12 @@ class Task(object):
     #: :setting:`CELERY_ACKS_LATE` setting.
     acks_late = None
 
+    #: When CELERY_ACKS_LATE is set to True, the default behavior to
+    #: handle worker crash is to acknowledge the message. Setting
+    #: this to true allows the message to be rejected and requeued so
+    #: it will be executed again by another worker.
+    reject_on_worker_lost = None
+
     #: Tuple of expected exceptions.
     #:
     #: These are errors that are expected in normal operation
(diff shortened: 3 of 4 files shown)


thedrow:
The build failure is not related to you.
I think that this should be on by default. @ask @malinoff @ionelmc What do you guys think?
Can you add a test case that proves that the task is rejected if the worker is lost. You just have to create a mocked task that raises that exception.

malinoff:
+1 on setting `True` by default. However, since this is not a backwards-compatible change, I'd ask @ask in which version exactly we should include this change.
Also, setting `False` by default makes this code compatible so we can include it and change the default value later.

ionelmc:
Should we even have this option at all? Is there an usecase where you'd want `REJECT_ON_WORKER_LOST=True`?

thedrow:
I think that it's always the case if the worker crashed and the task is idempotent. Re-queuing it ensures that it gets executed somewhere else.

mpermana:
Seeing majority of people using acks_late wants the retry, and we already coument that acks_late require task to be idempotent, I changed the code to always reject if acks_late and the exception is worker lost. Will add the test case for this soon.

malinoff:
@mpermana without making this configurable, you're breaking backwards compatibility, thus your PR cannot be merged into 3.x.

ionelmc:
Ok but is it not a bug? I can't imagine how anyone would want to specify something as the current behavior. @ask may have different opinion, dunno, but to me it seems that fixing a bug is not breaking backwards compatibility. Does anyone even rely on such broken behavior?

ask:
Even though this behavior seems like a reasonable default, it's can be cause unexpected behavior in practice.

For example the task restarts if sysadmin kills a misbehaving task, or if the os kills a task because it tried to do something illegal, or because it consumed all the memory (possibly causing the rest of the processes to crash).

When we tested this before, we often saw a crashing task resulting in a high frequency infinite loop taking everything down.

For these reasons I deliberately chose to ack signalled tasks, but I wouldn't mind having a configuration option.

Currently the default behavior will save tasks if there's a power failure or hard reboot (depending on broker configuration), which is often the most important thing.  If a task crashes by signal you will at least have the log, and retrying something that was signalled can have unintended side effects, even if idempotent.

mpermana:
How we make decisions then, do you guys do webconference and decide? I can always revert to my original proposed commit where we have configuration options and behave exactly the same by default.

mpermana:
I also consider the possibility of checking number of retry, but I am not familiar enough with the code. Can anyone help me if we count the number of retry and reject after it exceed the retry count.

ask:
@mpermana AMQP sets a `redelivered` property, but it's not a counter, just a boolean.

I don't think there's any way for us to add that information, unless we republish the message manually (which may be a bad idea, not sure)

celery/worker/request.py
```diff
@@ -352,7 +352,10 @@ def on_failure(self, exc_info, send_failed_event=True, return_ok=False):
                 )
         # (acks_late) acknowledge after result stored.
         if self.task.acks_late:
-            self.acknowledge()
+            if isinstance(exc, WorkerLostError):
+                self.reject(True)
```
ask: please use `self.reject(requeue=True)` to make it more obvious what's happening

mpermana:
Good news guys, I try to check the delivery_info.redelivered, indeed rabbit is setting it to True for redelivered message. So I can use that information to determine whether to reject and retry. Only original delivered message is retried, that way we can avoid the infinite crash that Ask worries about.

Basically:
- i reintroduced the settings to keep the existing behavior
- check the redelivered, only retry if acks late, using reject retry config, and its not a redelivered message

I am so excited, I can see it works in my local machine, I kill worker 2x, and 2nd kill is not retried.

thedrow:
But what happens if you use another broker?

mpermana:
Not sure what other broker do, but the code will not break, I see the delivery_info.redelivered is set here:
https://github.com/mpermana/celery/blob/master/celery/worker/request.py#L154

I would assume they will not set redelivered flag, and the behavior of the code will be exactly the same as current behavior, it will not retry on WorkerLost, it will ack. I think I can update the documentation, this only works for broker that sets delivery info, i.e: rabbitmq. Let me know what you think.

mpermana:
I just tried with redis broker and print out the deliver info, the redelivered defaults to None value:
{u'priority': 0, u'redelivered': None, u'routing_key': u'celery', u'exchange': u'celery'}

So for redis, with this setting, it'll keep retrying. Gonna change the code so it will not retry if redelivered is None (unknown).

mpermana:
@thedrow @ask 
- added unit test,
- tested with RabbitMQ and Redis broker

Let me know what you guys think.

thedrow:
I think that this is ready for merge.
@ask @malinoff @ionelmc any objections?

malinoff:
LGTM

ask:
Looks good, but not documented in `docs/configuration.rst` :)

<pre>
(def)asksol@grain:/opt/devel/celery(master)$> make configcheck
PYTHONPATH=. python extra/release/verify_config_reference.py "docs/configuration.rst"

Error: found undocumented settings:
  * CELERY_REJECT_ON_WORKER_LOST
make: *** [configcheck] Error 1
</pre>

ask:
I added it, but also reworded it because it was pretty weak on describing the details.  Like saying that the requeued task will execute on "another worker", when it can also execute on the same worker.

mpermana:
Thank you @ask

mpermana:
Btw, the message loop is prevented by delivery_info.redelivered, for broker that sets that information such as rabbit, this setting will take affect, it will requeue. For broker that don't like Redis, it does not have redelivered info, so it won't take affect.

ask:
Wouldn't it be better to use reject+requeue=False in that case, so that a dead letter queue can be used?

mpermana:
I think it makes sense to use reject+requeue=False when redelivered is not known.

thedrow:
@mpermana @ask Can any of you file a new issue about it please?

mpermana:
I addressed it in PR 2872

ask:
So retry once then just delete the message? basic.ack means the message is handled, basic.reject means the worker couldn't handle the message.  What do you do after it just fails fast twice and the message is lost forever?

mpermana:
If it fails 2x, it goes to self.acknowledge(), which is the pre-existing behavior.

thedrow:
Why shouldn't the task be redelivered to the dead letters exchange if it's configured?

mpermana:
Sorry I am not familiar with the dead letters exchange works, can you give clue what is the configuration, and how i can send it to dead letters exchange?

ask:
basic.reject will delegate the message to any configured dead letter exchanges, see https://www.rabbitmq.com/dlx.html

tumb1er:
Current PR version doesn't work correctly, see https://github.com/rutube/celery/pull/4
1. It really re-runs task in celery worker
2. clients still got WorkerLostError without waiting for re-run task.

thedrow:
@tumb1er Can you open a new PR that fixes the issue and the tests?
The relevant tests pass so I'm not sure what the problem is.

tumb1er:
@thedrow 
The problem is that following line of code marks task as failed in result backend
```python
self.task.backend.mark_as_failure(...)
```
So the task is retried and then successfully finishes, but client thinks it has failed.

For PR need some answers.
1. Is it OK to create PR from tumb1er/celery branch 3.0 to celery/celery 3.0, or master is also affected?
2. I don't completely understand what's wrong with 'redelivered' detection, so my PR may be incorrect.

fillest:
Ohh... So what's the state of this "reject" setting implementation? Can I just enable it or it doesn't work with auto-retrying (and is undocumented) yet? @tumb1er cau you please describe the issue more clearly one more time? Does it happen with some custom task handling?

I think the current default behavior is nothing but expected when late_acks are enabled (which should have been enabled by default too but oh well, it's another story and at least it's configurable and clearly explained in the docs). The whole point of tuning for reliability (late_acks, auto-retrying, broker persistence and so on) is that in the real production environments a lot of bad things do happen and what is exactly expected is that task would be ack'ed _only_ when they clearly were processed, otherwise they will be retried if it is needed. If some task causes OOM-or-whatever-failure loop - I will have to deal with it but I will know that after it is dealt with, the system will automatically recover and continue to work and I won't have to manually fix nastier consistency issues.
I think having this setting is very important - without it one has to build an extra layer of reliability system which is redundant because Celery already has a nice flexible supervisor component for task execution, and such a layer will be a supervisor for a supervisor, adding more complexity than if it were built-in.

tumb1er:
@fillest reject setting state: needs PR )
We forked celery to https://github.com/rutube/celery/tree/v3.1.18-rutube and added some fixes for this issue and some others. https://github.com/rutube/celery/pull/4 describes the situation completely clearly:

1. Run test task with `sleep` 
2. Kill worker process
3. Celery will mark task as failed even with these settings

```python
conf = {
    'BROKER_URL': 'amqp://guest:guest@127.0.0.1:5672/',
    'CELERY_RESULT_BACKEND': 'redis://127.0.0.1:6379',
    'CELERY_ACKS_LATE': True,
    'CELERY_REJECT_ON_WORKER_LOST': True
}
```

I didn't test it with latest Celery versions, so information may be outdated.

As for expected behavior, it's OK. You can regulate with flag, reject or retry task. But it just doesn't work.

Looks like plans are:
1. Test master and latest 3.x against this test case
2. Deal with redeliveries and message loops (short description here: https://github.com/celery/celery/pull/2840#issuecomment-148180647)
3. Test it with other brokers
4. Create correct PR, merge, release, forget )

fillest:
>Celery will mark task as failed even with these settings
Ah, so it's broken. Sigh.

Btw I've spent a week to build an extra layer (decorator with flock + saving run status and flock path into sqlite + periodic task checking if lock is still held), nearly finished it and only just now I realized that I can't reliably detect if task has been revoked on QUIT signal... Arrgh.

auvipy:
@siddharth96 can you please come with a draft PR?

PR #2868, opened 2015-10-20
Set priority from message properties if not in delivery_info

pymonger:
The value of delivery_info["priority"] remains null despite passing in priority=<int> to apply_async(). It is stored in message properties so get the value there if not in delivery_info.
Add contributor.

### DIFF
--- a/celery/worker/request.py
+++ b/celery/worker/request.py
@@ -152,7 +152,8 @@ class Request(object):
             'delivery_info': {
                 'exchange': delivery_info.get('exchange'),
                 'routing_key': delivery_info.get('routing_key'),
-                'priority': delivery_info.get('priority'),
+                'priority': delivery_info.get('priority',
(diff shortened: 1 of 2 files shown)


celery/worker/request.py
```diff
@@ -152,7 +152,8 @@ def __init__(self, message, on_ack=noop,
             'delivery_info': {
                 'exchange': delivery_info.get('exchange'),
                 'routing_key': delivery_info.get('routing_key'),
-                'priority': delivery_info.get('priority'),
+                'priority': delivery_info.get('priority',
+                                              properties.get('priority')),
```
ask: Seems I made a mistake in the virtual kombu transports here, as they set this in `delivery_info`.
Since this is for the 4.0 release we can just as well fix it properly, and read it from properties directly.

ask:

PR #2872, opened 2015-10-20
use reject+requeue=False when redelivered is not known, 

mpermana:
so that a dead letter queue can be used

This is continuation for issue #1628
As discussed in PR https://github.com/celery/celery/pull/2840

@thedrow @ask


thedrow:
Do the existing tests cover this?

mpermana:
I added additional test case to cover case when redelivered is not known

PR #2879, opened 2015-10-23

bujniewicz:
fails with OSError 9. This is fixed by /dev/urandom's fd to


bujniewicz:
Actually I believe this belongs in 3.1 for now - closing and resubmitting there.

PR #2880, opened 2015-10-23

bujniewicz:
fails with OSError 9. This is fixed by adding /dev/urandom's fd to


thedrow:
Please create this PR against master. We'll cherry pick it.

bujniewicz:
https://github.com/celery/celery/pull/2879

PR #2881, opened 2015-10-23
Database Backend: Use configured serializer, instead of always pickling.

m4ddav3:
Use the BLOB as an sa.BLOB
Serialise the result an add to the db as bytes


ask:
Is this backward compatible?

celery/backends/database/models.py
```diff
@@ -57,7 +57,7 @@ class TaskSet(ResultModelBase):
     id = sa.Column(sa.Integer, sa.Sequence('taskset_id_sequence'),
                    autoincrement=True, primary_key=True)
     taskset_id = sa.Column(sa.String(255), unique=True)
-    result = sa.Column(PickleType, nullable=True)
```
ask: Should we then also remove PickleType, if it's no longer used?

m4ddav3: Sure, I've removed the line

m4ddav3:
I need to check this, but I believe it is backward compatible while the result encoding type is set to 'pickle'. If the user has result encoding set to 'json' for instance, then it will change the behaviour (I suddenly found readable data that I can access easily)
I guess a script to convert the column to the required encoding could be provided.

ask:
I was trying to merge this into master, and it does not pass tests,  it seems to be missing some critical parts, like deserialization of the task/group payloads :(

m4ddav3:
Ugh. I need to dig up a Vagrant/Docker celery + rabbitmq + db environment

ask:
@m4ddav3: `cd funtests/stress/run; vagrant up; vagrant shell`

ask:
No database in that environment though, but shouldn't be too difficult to add

m4ddav3:
@ask ,
I've had a dig around, and added tests for json serialisation with an sqlite db backend.
The existing tests expect that the serializer be able to handle stored Python objects, which I think isn't necessarily the right thing to do.

> Celery is written in Python, but the protocol can be implemented in any language

In order to keep in-line with the current methodology, but still be able to implement a JSON serializer, I suggest adding [jsonpickle](https://github.com/jsonpickle/jsonpickle). This should enable enough compatibility to the current scheme of encoding/decoding exceptions and/or other serializable Python objects. What do you think?

ask:
jsonpickle would be nice, but sadly it has the same security concern as pickle, which is the ability to deserialize arbitrary objects by attribute path.

The tests must be rewritten to use json-safe data only, not sure how much work that is :(

martincapello:
I created #5907, which is a rebased and slightly updated version of this PR.

PR #2883, opened 2015-10-26

bujniewicz:
fails with OSError 9. This is fixed by /dev/urandom's fd to

PR against master per comment on https://github.com/celery/celery/pull/2880


bujniewicz:
Master is actually failing now, but it looks like those changes didn't make any additional tests fail and their own test passes just fine apart from PyPy3, where obviously tests aren't even run.

Thanks for quick merge!

ask:
Thanks! I broke the tests in master, but since have fixed it :)

PR #2898, opened 2015-10-30
Fix LRUCache.update for Python 3.5

### LINKED ISSUE #2897: LRUCache.update broken on Python 3.5
In Python 3.5, OrderedDict does not allow mutating while it is iterated over. This breaks `LRUCache.update` when it is called with a map larger than the maximum size:

``` 
File "/home/dennis/src/celery_/celery/utils/functional.py", line 75, in update
    for item in islice(iter(data), len(data) - limit):
RuntimeError: OrderedDict mutated during iteration
```

PR with fix follows

brakhane:
Python 3.5's OrderedDict does not allow mutation while it is being
iterated over. This breaks "update" if it is called with a dict
larger than the maximum size.

This commit changes the code to a version that does not iterate over
the dict, and should also be a little bit faster.

Closes #2897

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -197,3 +197,4 @@ Gerald Manipon, 2015/10/19
 Krzysztof Bujniewicz, 2015/10/21
 Sukrit Khera, 2015/10/26
 Dave Smith, 2015/10/27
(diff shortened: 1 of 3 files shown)


ask:
Nice one! Merged into master and 3.1

PR #2905, opened 2015-11-02
fixes broken little-worker example

attiasr:
using the -L test breaks the ability to use multiple instances of init script as described in the top of file


ask:
Thanks!

PR #2923, opened 2015-11-13
Replace use of 'if [[...]]' not supported in sh

cjh1:
The use of 'if [[...]]' breaks on Ubuntu that uses dash as its
implementation of sh. Replace wildcard matching with grep.

### DIFF
--- a/extra/generic-init.d/celeryd
+++ b/extra/generic-init.d/celeryd
@@ -37,9 +37,14 @@ if [ $(id -u) -ne 0 ]; then
     exit 1
 fi
 
+origin_is_runlevel_dir () {
+    set +e
+    dirname $0 | grep -q "/etc/rc.\.d"
+    echo $?
+}
 
 # Can be a runlevel symlink (e.g. S02celeryd)
-if [[ `dirname $0` == /etc/rc*.d ]]; then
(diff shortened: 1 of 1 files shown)


extra/generic-init.d/celeryd
```diff
@@ -39,11 +39,14 @@ fi
 
 
 # Can be a runlevel symlink (e.g. S02celeryd)
-if [[ `dirname $0` == /etc/rc*.d ]]; then
+set +e
```
ask: `+e` is the default, but you are setting it to -e later?

If the set -e is propagated from something else calling the script then it would be better to
wrap it in a function:

    origin_is_runlevel_dir () {
        set +e
        dirname $0 | grep -q "/etc/rc.\.d"
        echo $?
    }

    if [ $(origin_is_runlevel_dir) -eq 0 ]; then

cjh1: It set to -e at the top ... I will make the change you suggest. Thanks

ask:
Thanks!

ask:
Btw, please add your name to https://github.com/celery/celery/edit/master/CONTRIBUTORS.txt if you haven't already!

cjh1:
Thanks, I have created a PR that adds my name as a contributor.

PR #2936, opened 2015-11-26
Added a note about `new_join` to chord primitive description

hareevs:
If using Redis result backend the counter increment approach to chord unlocking (`_new_chord_apply` and `_new_chord_return`) is used only when `new_join` is set to `True` in result backend connection params. This wasn't mentioned anywhere in the documentation.


ask:
It's definitely incrementing without new_join, but new_join is enabling further optimizations.

I don't think new_join should be documented in the docs, as it doesn't have any real setting.  It will be the default behavior in 4.0 anyway :)

PR #2955, opened 2015-12-09
change celerybeat init script to report service as down when no pid file can be found

ezarowny:
The celerybeat example init script currently reports that the service is up when no pid file can be found.


ask:

PR #2971, opened 2015-12-22
Fix git link

russkel:
Hi,
While reading I saw that this bit needed a tiny fix up. I haven't built docs but it's simple enough.

### DIFF
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -393,6 +393,8 @@ Working on Features & Patches
     However following these steps may make maintainers life easier,
     and may mean that your changes will be accepted sooner.
 
+.. _forking-and-setting-up:
+
 Forking and setting up the repository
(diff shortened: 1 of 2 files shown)


ask:
Sadly `includes/installation.txt` is included in the `README` file which must
only use regular restructured text without Sphinx markup, so
there this will end up being:

    See ``forking-and-setting-up``


To have something that will make sense in both formats the `extra/release/sphinx-to-rst.py` converter would need to change:

    :ref:`Contributing <forking-and-setting-up>`

into `Contributing` or simply reword it into

    Please see :ref:`forking-and-setting-up` in the Contributing section of the documentation.

russkel:
Oh hmm, seems more effort than what it is worth. I'm not sure I want to dig into the converter. Feel free to close the pull request.

PR #2984, opened 2016-01-05
Update optimizing.rst

bryson:
librabbitmq does not support Python 3, and that should be noted in the guide.


dbrgn:
Looks good to me.

thedrow:
Thanks!

Issue #2992, opened 2016-01-08
celery recreate connection to Mongo backend each time when task result is stored


vklindukh:
When celery configured to store task results into Mongo backend (CELERY_RESULT_BACKEND and CELERY_MONGODB_BACKEND_SETTINGS properties defined) it opens and closes connection to Mongo each time when new record is written to Mongo.

It causes too many sockets in TCP_WAIT state during high task execution rate and finally celery unable to connect to Mongo for a while because of no free sockets available:
```
python2.7/site-packages/pymongo/mongo_client.py", line 378, in __init__ raise ConnectionFailure(str(e)) ConnectionFailure: [Errno 99] Cannot assign requested address
```
To fix it we may want to eliminate process_cleanup() from celery/backends/mongodb.py and let celery process to reuse self._connection instead of recreating it each time when new task result is going to be stored into mongo backend:
```
index 8935d0d..e217639 100644
--- a/celery/backends/mongodb.py
+++ b/celery/backends/mongodb.py
@@ -162,14 +162,6 @@ class MongoBackend(BaseBackend):
 
-    def process_cleanup(self):
-        if self._connection is not None:
-            # MongoDB connection will be closed automatically when object
-            # goes out of scope
-            del(self.collection)
-            del(self.database)
-            self._connection = None
```

ask:
Closed by #2993

PR #2993, opened 2016-01-08
do not destroy connection to mongo #2992

vklindukh:
This PR prevents destroying connection to Mongo backend and fixes issue described https://github.com/celery/celery/issues/2992

### DIFF
--- a/celery/backends/mongodb.py
+++ b/celery/backends/mongodb.py
@@ -162,14 +162,6 @@ class MongoBackend(BaseBackend):
 
         return self._connection
 
-    def process_cleanup(self):
-        if self._connection is not None:
-            # MongoDB connection will be closed automatically when object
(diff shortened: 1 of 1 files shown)


ask:
process_cleanup is called after every task, so the name could have been easily confused for something that cleans up before the process exits (understandably).  Are you sure the connection does not need to be closed?  Obviously it cannot need to since that is not always possible, but maybe beneficial?

ask:
Let's test it in 4.0.

Can you please add your name to https://github.com/celery/celery/edit/master/CONTRIBUTORS.txt if you haven't already?

vklindukh:
This is currently been testing production env with celery 3.1.17

AvnerCohen:
@ask Any chance this can go into a 3.1.x patch?

PR #2994, opened 2016-01-08
More detail about prefetch and long-running tasks

veatch:
When reading docs, I thought T3 would be stuck behind T1, but other tasks would be sent to available workers. I've seen that queued tasks aren't sent to available workers unless -Ofair is used, so clarifying this description.

### DIFF
--- a/docs/userguide/optimizing.rst
+++ b/docs/userguide/optimizing.rst
@@ -218,7 +218,7 @@ waiting for long running tasks to complete::
 
     -> send T3 to Process A
     # A still executing T1, T3 stuck in local buffer and will not start until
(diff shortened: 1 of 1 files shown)


ask:
workers here refer to worker instances, not processes so this update is not really correct, since -Ofair is only about the internal routing inside the worker to its child processes.

veatch:
Would it be correct to say, "other queued tasks will not be sent to idle processes"? Or am I misunderstanding what's going on here?

ask:
Yeah, that's much better.  Writing 'idle workers' is easily confused by worker instances.

The relationship between broker prefetch and worker process prefetch is apparently tricky for me to expain, lol.

ask:

PR #2998, opened 2016-01-12
docs: remove duplicated word “acknowledgements”

wjt:
Introduced in 2e8b4de.

The rest of the sentence is still a bit hard to parse, though…


ask:
Thanks!  I tried to rewrite it now in a9a85, but not sure if better or not.

PR #3000, opened 2016-01-13
Fixes doc typo in celery.contrib.batches

dentafrice:
Should say "10 seconds" instead of just "seconds".


ask:

PR #3002, opened 2016-01-15
Add example for broadcast queue and celerybeat

jbkkd:


ask:
Does this work?  Maybe I have implemented this by accident or something, but how can you set the schedule in `task_routes`?

jbkkd:
woops, my bad. I used the previous `CELERYBEAT_SCHEDULE` setting locally when testing this, and copied the wrong one from the mapping table when making the documentation. Fixed now.

ask:
Thanks!

PR #3012, opened 2016-01-20
Fix missing parameters passing through the _new_redis_client method

jeremyll:
The app configuration (host, port, etc..) are not passed through the `_new_redis_client` method.


ask:
Thanks! Merged into 3.1!

PR #3014, opened 2016-01-21
Let celery make tasks from functions with type hints

Adaptification:
Making a task from a function involves call head_from_fun in utils/functional.py, which uses getargspec. This will fail if said function has type hints (PEP 0484), so call getfullargspec instead -- but only when using python 3+


ask:
Great! Merged into master.

Issue #3021, opened 2016-01-23
group and chord: subtasks are not executed until entire generator is consumed


chadrik:
When passing a generator to `group` or `chord`, sub-tasks are not submitted to the broker for execution until the entire generator is consumed.  The expected result is that subtasks are submitted during iteration, as soon as they are visited.

Here is the test code I'm using:

``` python
from celery import Celery, group
from celery.utils.log import get_task_logger

app = Celery(__name__, broker='redis://localhost:6379/0',
             backend='redis://localhost:6379/0')

logger = get_task_logger(__name__)

def longyield():
    import time
    for i in xrange(5):
        time.sleep(2)
        print "yielding", i
        yield add.s(i, i)

@app.task
def add(x, y):
    logger.info("adding {} + {}".format(x, y))
    return x + y

def itertest():
    it = longyield()
    g = group(it)
    result = g.delay()
    print result.get()
```

I'm testing with 4.0rc1.

syradium:
Your example is really strange. Why do you need a group when you expect synchronous behaviour?

Well, I think you are wrong. Because first - you create a signature for a task which you call later. That task is a group which consists of `add` subtasks. This is created `yielding` from a `logyield`.

Later when the group is created and saved to `g` you call it a `add` subtasks are executed in parallel.

chadrik:
@syradium Hopefully my example is not pointless because it is taken almost directly from the celery docs:

```python
from celery import group
res = group(add.s(i, i) for i in xrange(10))()
res.get(timeout=1)
```

I used `group.delay()` but that doesn't make a difference.  By extending the time it takes to yield results (e.g. `longyield` vs `xrange`) I'm trying to make apparent that `group` is consuming the entire generator before submitting subtasks for execution.

> Why do you need a group when you expect synchronous behaviour?

I don't understand your question.  I want all sub-tasks executed in parallel.  I thought that was the point of the `group` primitive. 

I'm also not the only person who has had this problem. I found [this old reference](https://groups.google.com/forum/#!topic/celery-users/9asJ1vCwmmE) from the celery-users google group, but the suggestions from Ask seemed out of date for the 3.1.x codebase, and the ones that I tried did not help.   I was surprised to find this issue has not been address in 4.x.

> ```python
> result = [add.delay(i, i).get() for i in longyield()]
> print result
> ```

Yes, that's the workaround I'm using for now, but the point is that I *want* to use the canvas primitives so that I can combine them into complex topologies.

syradium:
[...]

PR #3043, opened 2016-02-06
group and chord: subtasks are not executed until entire generator is consumed

chadrik:
When passing a generator to `group` or `chord`, sub-tasks are not submitted to the backend for execution until the entire generator is consumed.  The expected result is that subtasks are submitted during iteration, as soon as they are yielded.  This can have a big impact on performance if generators yield subtasks over a long period of time.

I also opened issue #3021 on the subject.  [This explanation](https://github.com/celery/celery/issues/3021#issuecomment-176729202) from @eli-green I think is pretty on point.

So far I've only added support for redis.  I started looking at the other backends but I ran out of time.  It would be great to get some feedback on what I have so far and to get some thoughts on how difficult it will be to add for the other backends.

### DIFF
--- a/celery/utils/functional.py
+++ b/celery/utils/functional.py
@@ -17,7 +17,7 @@ try:
     from inspect import isfunction, getfullargspec as getargspec
 except ImportError:  # Py2
     from inspect import isfunction, getargspec  # noqa
-from itertools import chain, islice
+from itertools import chain, islice, tee, izip_longest
 
 from amqp import promise
 from kombu.utils.functional import (
@@ -28,7 +28,8 @@ from celery.five import UserDict, UserList, keys, range
 
 __all__ = ['LRUCache', 'is_list', 'maybe_list', 'memoize', 'mlazy', 'noop',
            'first', 'firstmethod', 'chunks', 'padlist', 'mattrgetter', 'uniq',
-           'regen', 'dictfilter', 'lazy', 'maybe_evaluate', 'head_from_fun']
+           'lookahead', 'regen', 'dictfilter', 'lazy', 'maybe_evaluate',
+           'head_from_fun']
 
 IS_PY3 = sys.version_info[0] == 3
 IS_PY2 = sys.version_info[0] == 2
@@ -310,6 +311,22 @@ def uniq(it):
     """Return all unique elements in ``it``, preserving order."""
     seen = set()
     return (seen.add(obj) or obj for obj in it if obj not in seen)
+
+
+def lookahead(it):
+    """Yield pairs of (current, next) items in `it`. 
+
+    `next` is None if `current` is the last item.
+
+    Example:
+
+        >>> list(lookahead(x for x in range(6)))
+        [(0, 1), (1, 2), (2, 3), (3, 4), (4, 5), (5, None)]
+
+    """
+    a, b = tee(it)
+    next(b, None)
+    return izip_longest(a, b, fillvalue=None)
 
 
 def regen(it):
--- a/celery/utils/functional.py
+++ b/celery/utils/functional.py
@@ -353,10 +353,11 @@ class _regen(UserList, list):
         return self.__it.__length_hint__()
 
     def __repr__(self):
+        # override list.__repr__ to avoid consuming the generator
         if self.__done:
             return repr(self.__consumed)
         else:
-            return '[{}]'.format(', '.join(
+            return '[{0}]'.format(', '.join(
                 [repr(x) for x in self.__consumed] + ['...']))
 
     def __iter__(self):
@@ -384,6 +385,8 @@ class _regen(UserList, list):
                 return self.__consumed[index]
 
     def __nonzero__(self):
+        # nonzero for list calls len() which would consume the generator: 
+        # override to consume maximum of one item.
         if len(self.__consumed):
             return True
         try:
@@ -402,6 +405,7 @@ class _regen(UserList, list):
         return self.__consumed
 
     def fully_consumed(self):
+        """Return whether the iterator has been fully consumed"""
         return self.__done
 
 
--- a/celery/utils/functional.py
+++ b/celery/utils/functional.py
@@ -382,6 +382,9 @@ class _regen(UserList, list):
         list(iter(self))
         return self.__consumed
 
+    def fully_consumed(self):
+        return self.__done
+
 
 def _argsfromspec(spec, replace_defaults=True):
     if spec.defaults:
--- a/celery/utils/functional.py
+++ b/celery/utils/functional.py
@@ -17,14 +17,14 @@ try:
     from inspect import isfunction, getfullargspec as getargspec
 except ImportError:  # Py2
     from inspect import isfunction, getargspec  # noqa
-from itertools import chain, islice, tee, izip_longest
+from itertools import chain, islice, tee
 
 from amqp import promise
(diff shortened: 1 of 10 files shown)


thedrow:
Do the tests already cover this usecase?

chadrik:
> Do the tests already cover this usecase?

@thedrow no they don't.  I haven't added any new tests, but I plan to once I get a nod of approval on the general idea from @ask or another maintainer (is there another maintainer?).  Any help I could get on implementing the feature for the other backends would be appreciated as well.

thedrow:
I'm also maintaining the project but I never used celery with redis so I'm not familiar with that part of the code.

chadrik:
@thedrow Can you weigh in on how this might be done for the other backends?  If they were completed, do you see any reason not to merge this PR?

thedrow:
I don't see any reason why this shouldn't be merged. It just requires tests.

chadrik:
I'm getting pretty close to having all the tests working, but for now I've decided to only support this feature for redis.  It seems like the other backends require that all tasks be submitted immediately to allow the `chord_unlock` task to work (though I still have more investigation to find out why that is). 

Before I can finish this, I have three questions:  

1)  Are there other backends like redis that do not use `chord_unlock`?

2) is it safe to assume that all of the 'chord' attributes on tasks processed by [`group._apply_tasks()`](https://github.com/celery/celery/blob/master/celery/canvas.py#L733) will have the same value?  To put it  another way, is it possible for the callbacks within a group to be heterogenous?  It seems to me that this should never happen, because in the redis backend each task caries the chord callback forward, and the last to complete triggers it, so the callback would be effectively random if the callbacks were not the same for all tasks in the group.  That said, I'm not sure what happens when a `chord` has another `chord` as a subtask.  I'll have to try that out.

3) why does [`group._freeze_unroll()`](https://github.com/celery/celery/blob/master/celery/canvas.py#L809) use `maybe_signature()` to ensure that the tasks are actually signatures (as opposed to dicts or lists), while [`group._traverse_tasks()`](https://github.com/celery/celery/blob/master/celery/canvas.py#L952) does not, even though both perform a similar recursive "unrolling" of a group's tasks?  At least in the case of the `chord.apply_async()` entry point, `_traverse_tasks` is called before `_freeze_unroll`, so if the latter is expecting non-signatures, the former will error before they get there. Is there any reason *not* to call `maybe_signature` in both?

Generally it seems there are a lot of different ways to do the same thing in the celery code and few comments to explain whether the subtle differences are intentional or not.

Last note, `maybe_signature` can return a list, but as far as I can tell, just about everywhere that it is called would result in an error if that actually happened.

chadrik:
sorry, one more question:  what is the point of the last 3 arguments to `group._prepared`?

```python
    def _prepared(self, tasks, partial_args, group_id, root_id, app, dict=dict,
                  CallableSignature=abstract.CallableSignature,
                  from_dict=Signature.from_dict):
```

There is nowhere within the celery code that uses these arguments, and no comment explaining this bizarre behavior.  And in fact, the `dict` argument isn't even used by the function itself.

thedrow:
I'm not familiar enough with the project so I'm assigning this to @ask to answer your questions.

ask:
>1) Are there other backends like redis that do not use chord_unlock?

only the cache backend

>2) is it safe to assume that all of the 'chord' attributes on tasks processed by group._apply_tasks() will >have the same value?

In the case of `chord`: yes
There's always the possibility that people are building custom chord like constructs,
which is something that I'd like to support, but not sure if there's any use case for
having tasks in the same group have different chord/group_id headers.
If it makes it easier, or it's a good optimization I'd say we can disallow it.

>3) why does group._freeze_unroll() use maybe_signature() to ensure that the tasks are actually >signatures (as opposed to dicts or lists), while group._traverse_tasks() does not, even though both >perform a similar recursive "unrolling" of a group's tasks?

I would guess this is because `_traverse_tasks()` is only used by `__length_hint__`,
and it doesn't matter if the object is a Signature or a dict, as long as the count is the same.
chord.apply_async calls __length_hint__, but it only cares about the count, using maybe_signature
would be expensive for no benefit as far as I can see.

>Generally it seems there are a lot of different ways to do the same thing in the celery code and few >comments to explain whether the subtle differences are intentional or not.

Do you have any examples of this?  In general comments related to the code in the function only,
not about when to use what (that would belong to the docstring).

>Last note, maybe_signature can return a list, but as far as I can tell, just about everywhere that it is >called would result in an error if that actually happened.

I'd guess this is something leftover from when all signatures had a builtin task.

>4) what is the point of the last 3 arguments to group._prepared?

It's localizing globals that are used within the loop, as for every iteration Python needs to
do a LOAD_GLOBAL for these.  `dict` is indeed not used, but it's not always easy to keep it up to date with the code.

Usually small savings, but it's used a lot in the worker inner loop (together with inlining), the end result
being able to process a thousand or so tasks more per/second.

Another reason for doing this is for threads, since when threads wake up during interpreter shutdown
some globals may have been garbage collected already and the LOAD_GLOBAL ends in a NameError

celery/utils/functional.py
```diff
@@ -334,8 +352,21 @@ def __reduce__(self):
     def __length_hint__(self):
         return self.__it.__length_hint__()
 
+    def __repr__(self):
+        if self.__done:
+            return repr(self.__consumed)
+        else:
+            return '[{}]'.format(', '.join(
```
ask: Please use `{0}` here, as position-less format is not supported by all 2.7 versions

ask:
Btw, I like your changes here.  I'm not sure I have understood all of them yet as I just looked at it briefly, maybe you could add some comments to describe what's happening in regen?

Good idea putting the length of the chord into redis, that totally makes sense since it's the only result backend that actually needs it.  I guess you can also remove _traverse_result and __length_hint__, since they are not used anymore.

I think one problem with using set_chord_size separately is that it complicates error handling, and this is maybe something that we should think about as currently everything depends on that operation actually succeeding, and there is no retry or anything.

ask:
I.e. If the set_chord_size fails we should probably retry, and on giving up it would need to set an error status for the chord?

ask:
Oh, `RedisBackend.set` already retries, but not sure what to do when giving up, maybe just raising a ConnectionError is good enough even though it already started n-1 tasks.

chadrik:
Hi @ask 
I addressed your notes.  And thanks for the answers.  Unfortunately, celery no longer figures into my development plans, so I'm not going to be able to invest much more time in this, but I think this is a pretty killer feature so I'll try to answer whatever questions you have.

ask:
I tried to merge this into master, but I have been unable to get the tests to pass so far.

The work is in the canvas-generator branch, commit: https://github.com/celery/celery/commit/23abf3576bc9142de9f3c9aaa98d6cc189a04d36

Need to postpone this to 4.1 for now as need to continue with getting 4.0 release out.

auvipy:
I am going to make it work on master. but you will definitely get the credit

chadrik:
Deal ;)
let me know if you need any help once you get it merged / rebased against
master.

auvipy:
re base seems problematic, so I am trying to implement it cleanly step by step with smaller chunks. could you review it https://github.com/celery/celery/pull/4459 here later? might need your suggestions about testing part

auvipy:
closing in favor of new patch

Issue #1895, opened 2014-02-28
CentOS init.d script does not work with 3.1.9


mattlong:
I'm having a difficult time getting the CentOS init.d script to work for me. Looking through it and the accompanying example sysconfig file, it appears that it's still half configured for the 3.0.x in that it does not use `celery multi` for daemonization:

`CELERYD_MULTI="$CELERYD_CHDIR/bin/celeryd-multi"`

Another example: I've also had to remove the escaped double quotes on this line from the init.d script to get it to work at all:

`DAEMON_OPTS="$DAEMON_OPTS --workdir=\"$CELERYD_CHDIR\""`

Is this script still officially support or am I just missing something? Note that I am using Scientific Linux 6 and not CentOS, but my understanding is that the two are very similar.

PR #3044, opened 2016-02-09
Deleted docs on removed CentOS init script

veatch:
For additional info, see https://github.com/celery/celery/issues/1895


ask:
Thanks!

PR #3053, opened 2016-02-14
Requeue tasks that failed due to Worker Lost Error

joemarct:
When setting CELERY_ACKS_LATE to True, one would expect that the tasks that failed due to WorkerLost error will be sent back to queue. This change implements what I think should be the expected behavior.


linar-jether:
+1 for this issue, any chance of merging this?

ask:
The worker should acknowledge the task if the process is killed, even if acks_late is enabled,
or you're likely to end up with a high frequency loop bringing down the worker.

See this new setting in 4.0: http://docs.celeryproject.org/en/master/configuration.html#task-reject-on-worker-lost

PR #3055, opened 2016-02-14
Fix the issue that the INIT script will ignore the environment variables

ljm625:
Fix the issue that the INIT script will ignore the environment variable settings, for example in /etc/profile, such as $PYTHONPATH, and causing error ( but the script continue says start successful ).


ask:
I think PYTHONPATH should be set in the configuration file /etc/default/celeryd

It's not common for daemons to load a full user login shell AFAIK

Issue #3068, opened 2016-02-20
Worker crash on startup when specifying multiple cache result backends


mbox:
Seen with Celery 3.1.20, Django 1.9

Steps to reproduce:

* Set CELERY_RESULT_BACKEND to multiple memcache servers e.g. `CELERY_RESULT_BACKEND = 'cache+memcached://127.0.0.1:11211;127.0.0.10:11211/'`
* Start a celery worker using ./manage.py celery worker

Result is a crash with stacktrace:

```
[2016-02-20 14:58:41,566: ERROR/MainProcess] Unrecoverable error: ValueError("invalid literal for int() with base 10: '11211;127.0.0.10'",)
Traceback (most recent call last):
  File "/Users/malcolmbox/dev/tellybug/webserver/lib/python2.7/site-packages/celery/worker/__init__.py", line 206, in start
    self.blueprint.start(self)
  File "/Users/malcolmbox/dev/tellybug/webserver/lib/python2.7/site-packages/celery/bootsteps.py", line 119, in start
    self.on_start()
  File "/Users/malcolmbox/dev/tellybug/webserver/lib/python2.7/site-packages/celery/apps/worker.py", line 170, in on_start
    string(self.colored.cyan(' \n', self.startup_info())),
  File "/Users/malcolmbox/dev/tellybug/webserver/lib/python2.7/site-packages/celery/apps/worker.py", line 232, in startup_info
    self.app.conf.CELERY_RESULT_BACKEND or 'disabled',
  File "/Users/malcolmbox/dev/tellybug/webserver/lib/python2.7/site-packages/kombu/utils/url.py", line 63, in maybe_sanitize_url
    return sanitize_url(url, mask)
  File "/Users/malcolmbox/dev/tellybug/webserver/lib/python2.7/site-packages/kombu/utils/url.py", line 58, in sanitize_url
    return as_url(*_parse_url(url), sanitize=True, mask=mask)
  File "/Users/malcolmbox/dev/tellybug/webserver/lib/python2.7/site-packages/kombu/utils/url.py", line 24, in _parse_url
    return (scheme, unquote(parts.hostname or '') or None, parts.port,
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/urlparse.py", line 113, in port
    port = int(port, 10)
ValueError: invalid literal for int() with base 10: '11211;127.0.0.10'
```

[...]

Issue #3045, opened 2016-02-09
Unrecoverable error: ValueError invalid literal for int() with base 10: '53473,ds053473-a1.{snip}.mongolab.com'


m-vdb:
This breaks mongo backend, because of https://github.com/celery/celery/commit/33f1dc829223da5499e05e271a6d49edde7ffc68. I believe `maybe_sanitize_url` doesn't play well with replica sets...

Full backtrace:
```
ValueError: invalid literal for int() with base 10: '53473,{snip}.mongolab.com'
  File "celery/worker/__init__.py", line 206, in start
    self.blueprint.start(self)
  File "celery/bootsteps.py", line 119, in start
    self.on_start()
  File "celery/apps/worker.py", line 170, in on_start
    string(self.colored.cyan(' \n', self.startup_info())),
  File "celery/apps/worker.py", line 232, in startup_info
    self.app.conf.CELERY_RESULT_BACKEND or 'disabled',
  File "kombu/utils/url.py", line 63, in maybe_sanitize_url
    return sanitize_url(url, mask)
  File "kombu/utils/url.py", line 58, in sanitize_url
    return as_url(*_parse_url(url), sanitize=True, mask=mask)
  File "kombu/utils/url.py", line 24, in _parse_url
    return (scheme, unquote(parts.hostname or '') or None, parts.port,
  File "python2.7/urlparse.py", line 113, in port
    port = int(port, 10)
```

thedrow:
Is that a problem with the results backend or the broker or both?
Also can you share a failing test case? Preferably in a form of a PR?
I'm tempted to blame Python's urlparse but I'd like to verify.

m-vdb:
In this case I guess only the result backend is impacted, but I guess whenever `maybe_sanitize_url()` is used with a replicaset connection URI it could fail. I'll try providing a PR today

m-vdb:
there you go @thedrow :smile:

asiebert:
Note that this issue is very similar to https://github.com/celery/kombu/issues/375
Indeed, `urlparse` should not be used for Mongo URIs

thedrow:
Any suggestion on how to fix this?

m-vdb:
[...]

PR #3069, opened 2016-02-20
Fix for issue #3068 - Worker crash with multiple cache results backends

mbox:
Correctly format the CELERY_RESULT_BACKEND setting if it's not a single URL

### DIFF
--- a/celery/apps/worker.py
+++ b/celery/apps/worker.py
@@ -200,15 +200,20 @@ class Worker(WorkController):
         if not self.send_events:
             events = 'OFF (enable -E to monitor this worker)'
 
+        result_backend = self.app.conf.result_backend or 'disabled'
+        if result_backend.startswith('cache+memcache://'):
+            fmt_results = ";".join([maybe_sanitize_url(part) for part in result_backend.split(';')])
+            fmt_results.replace('/;', ';')
+        else:
+            fmt_results = maybe_sanitize_url(result_backend)
+
         banner = BANNER.format(
             app=appr,
             hostname=safe_str(self.hostname),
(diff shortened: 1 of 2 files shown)


thedrow:
While this PR fixes your problem it feels kinda hackish.
We need a more generic solution to the problem.
People using MongoDB experienced the same issue in #3045.
If we added a per results backend parsing mechanism for DSNs that would resolve the issue for all results backends and brokers.
I'm going to close this pull request for now but feel free to provide compelling arguments for merging this.
In any case, this is a problem that must be resolved.
Would you mind issuing a PR with just the failing test. I'll be happy to merge that.

mbox:
I'll issue a pull request for the failing test.

How would you like this fixed in a more generic way? The fundamental problem is that `maybe_sanitize_url` is being passed things that aren't URLs. There's three ways to solve that:

- Change `maybe_santitize_url()` to cope with all the URL-like formats that celery actually uses, either by changing the function itself or replacing with a `maybe_sanitize_url_like_thing()`
- Change the callers that know they have something that isn't a URL to only pass URLs to `maybe_sanitize_url`

I took the second approach. Clearly if CELERY_RESULTS_BACKEND is passed to `maybe_sanitize_url` in multiple places, the logic to sanitise it could be wrapped into a `maybe_sanitize_results_backend()`, but a quick inspection didn't find anywhere else doing that. Would you prefer that approach?

Then the mongo issue you link to would be solved by adding logic to `maybe_sanitize_results_backend` to detect mongo "urls" and handle those correctly.

Would that be an acceptable patch?

thedrow:
Yes. That would be an acceptable patch.
It will also fix the same problem for other results backends.

PR #3072, opened 2016-02-23
Added code to source the $CELERYD_USER's environment on startup

mikeengland:
Added code to source the $CELERYD_USER's environment on startup. This is necessary for Python projects which require special environment variables to be set, for example in the case of Theano where the $THEANORC and $THEANO_FLAGS environment variables. Without sourcing the environment in this case, Celery crashes and core dumps.

### DIFF
--- a/extra/generic-init.d/celeryd
+++ b/extra/generic-init.d/celeryd
@@ -149,6 +149,10 @@ if [ -n "$CELERYD_CHDIR" ]; then
     DAEMON_OPTS="$DAEMON_OPTS --workdir=$CELERYD_CHDIR"
 fi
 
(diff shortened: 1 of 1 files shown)


ask:
While convenient, I don't think it's safe to source `.bashrc` in a init script, it also assumes
the user is using the bash shell which may not be the case.

A better way to solve this is to add the required environment variables to `/etc/default/celeryd`
(or source the .bashrc there if you think its content is safe).

ask:
Btw, to propagate environment variables to Python in the `/etc/default/celeryd` file you need to export them:

    export THEANORC="/etc/theanorc"

PR #3073, opened 2016-02-23
Add failing testcase for issue #3068

mbox:


ask:
Thanks, but I don't think we want to add failing test cases unless we also have a patch to make it pass.
We keep the CI build passing not because it means all bugs are fixed, but because we want to see if
new commits breaks the build.

mbox:
@ask I was explicitly asked to do a PR for a failing test case by @thedrow on PR #3068 which *did* have a patch to make the tests pass.

ask:
@mbox Oh, that makes sense, I would guess he meant adding it to the existing pull request.

The pull request in #3079 solves this problem in a generic way for both mongodb and memcache, and your test case has been included there!

thedrow:
@ask As a practice I make it a habit of merging failing tests so we don't forget to fix them before the release.

PR #3079, opened 2016-02-25
Mvdb/backend as uri method

### LINKED ISSUE #3045: Unrecoverable error: ValueError invalid literal for int() with base 10: '53473,ds053473-a1.{snip}.mongolab.com'

m-vdb:
fixes #3045 #3068

### DIFF
--- a/celery/backends/base.py
+++ b/celery/backends/base.py
@@ -24,6 +24,7 @@ from kombu.serialization import (
     registry as serializer_registry,
 )
 from kombu.utils.encoding import bytes_to_str, ensure_bytes, from_utf8
+from kombu.utils.url import maybe_sanitize_url
 
 from celery import states
 from celery import current_app, group, maybe_signature
@@ -93,7 +94,7 @@ class BaseBackend(object):
 
     def __init__(self, app,
                  serializer=None, max_cached_results=None, accept=None,
-                 expires=None, expires_type=None, **kwargs):
+                 expires=None, expires_type=None, url=None, **kwargs):
         self.app = app
         conf = self.app.conf
         self.serializer = serializer or conf.result_serializer
@@ -107,6 +108,7 @@ class BaseBackend(object):
         self.accept = prepare_accept_content(
             conf.accept_content if accept is None else accept,
         )
+        self.url = url
 
     def mark_as_started(self, task_id, **meta):
         """Mark a task as started"""
@@ -392,6 +394,13 @@ class BaseBackend(object):
 
     def __reduce__(self, args=(), kwargs={}):
         return (unpickle_backend, (self.__class__, args, kwargs))
+
+    def as_uri(self, include_password=False):
+        """Return the backend as an URI, sanitizing the password or not"""
+        # when using maybe_sanitize_url(), "/" is added
+        # we're stripping it for consistency
+        return self.url if include_password else maybe_sanitize_url(self.url).rstrip("/")
+
 BaseDictBackend = BaseBackend  # XXX compat
 
 
--- a/celery/backends/base.py
+++ b/celery/backends/base.py
@@ -688,7 +688,7 @@ class DisabledBackend(BaseBackend):
             'Please see the documentation for more information.')
 
     def as_uri(self, *args, **kwargs):
-        return 'disabled'
+        return 'disabled://'
 
     get_state = get_status = get_result = get_traceback = _is_disabled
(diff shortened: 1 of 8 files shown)


celery/backends/cache.py
```diff
@@ -149,3 +149,12 @@ def __reduce__(self, args=(), kwargs={}):
                  expires=self.expires,
                  options=self.options))
         return super(CacheBackend, self).__reduce__(args, kwargs)
+
+    def as_uri(self, *args, **kwargs):
+        """
+        Return the backend as an URI. It properly handles the
+        case of multiple servers. It doesn't try to sanitize
+        password because memcached URIs doesn't support them.
```
m-vdb: I wasn't sure about that, LMK if I'm completely off

celery/backends/base.py
```diff
@@ -678,5 +686,9 @@ def _is_disabled(self, *args, **kwargs):
         raise NotImplementedError(
             'No result backend configured.  '
             'Please see the documentation for more information.')
+
+    def as_uri(self, *args, **kwargs):
+        return 'disabled'
```
ask: Should be `disabled://` ?

celery/backends/cache.py
ask: Yeah, I think you can technically using SASL but the Celery cache backend doesn't let you do so via the URL

celery/backends/base.py
m-vdb: I kept the [previous behaviour](https://github.com/celery/celery/pull/3079/files#diff-5057f62807f26f9a701a0d9bf9a2a2b3L209) but makes more sense yes

celery/backends/cache.py
m-vdb: that's what I figured :white_check_mark:

m-vdb:
@ask @thedrow looks like this build is gonna pass. LMK what you think :smile:

m-vdb:
build is failing on appveyor because of missing `pymongo`, strange the issue wasn't there on travis

ask:
@m-vdb Oh yeah, MongoDB is only installed as a dependency on Travis, it's also not installed when running the unit tests using python setup.py test.

They are more like optional tests, so the two test cases should be moved into the mongodb and memcache tests respectively, and skipped if the dependency is not installed

m-vdb:
@ask I [mocked the tests](https://github.com/m-vdb/celery/blob/mvdb/backend-as-uri-method/celery/tests/backends/test_cache.py#L37-L41) for memcached so that it doesn't raise when memcached is not installed. Should I removed that too? Is this safe enough, do you have tests suites that run only mongodb / memcached / other 3rd parties plugin-like code? Because it would be nice to avoid a possible regression for those :)

m-vdb:
what do you think about my comment @ask ?

ask:
Merged into master and 3.1!

@m-vdb The Travis CI build runs mongodb backend tests, and I just changed that to include cache backend tests.

Could you also add your name to https://github.com/celery/celery/edit/master/CONTRIBUTORS.txt ?

m-vdb:
thanks @ask :) So cool. Are you planning to do a PyPi release?

m-vdb:
https://github.com/celery/celery/pull/3083

PR #3081, opened 2016-02-27
Mention that event capturing can be stopped by using should_stop

dessant:


ask:

PR #3088, opened 2016-03-03
[datastructures] Fix LimitedSet.discard()

### LINKED ISSUE #3087: LimitedSet.discard() fails to remove items from self._heap
The `LimitedSet.discard()` method fails to remove items from `self._heap`. It looks like a simple typo:

```python
    def discard(self, value):
        """Remove membership by finding value."""
        try:
            itime = self._data[value]
        except KeyError:
            return
        try:
            self._heap.remove((value, itime)) # Bug here. These args are swapped.
        except ValueError as e:
            pass
        self._data.pop(value, None)
    pop_value = discard  # XXX compat
```

The call to `self._heap.remove()` should be like this:

```python
self._heap.remove((itime, value))
```

You can see in the `add()` method that items are added to the heap this way:

```python
heappush(self._heap, (inserted, key))
```

As a result, the LimitedSet has unbounded memory growth for callers who call `discard()` to remove items.

djsmith42:
This was raising ValueError every time it was called, because the
argument order was backward, resulting in unbounded memory growth for
callers using discard() to remove items from LimitedSet.

Closes #3087

### DIFF
--- a/celery/datastructures.py
+++ b/celery/datastructures.py
@@ -633,7 +633,7 @@ class LimitedSet(object):
         except KeyError:
             return
         try:
-            self._heap.remove((value, itime))
+            self._heap.remove((itime, value))
         except ValueError:
             pass
         self._data.pop(value, None)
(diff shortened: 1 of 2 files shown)


ask:
Thanks! Already merged

PR #3089, opened 2016-03-03
fixed reversed min max values

raducc:
The max and min values are reversed


ask:

Issue #3095, opened 2016-03-07
LimitedSet is not really limiting as it should be


alekibango:
in worker/control.py:
```py
Panel.register                                                                 
def hello(state, from_node, revoked=None, **kwargs):                            
    if from_node != state.hostname:                                             
        logger.info('sync with %s', from_node)                                  
        if revoked:                                                             
            worker_state.revoked.update(revoked)                                
        return {'revoked': worker_state.revoked._data,                          
                'clock': state.app.clock.forward()}
```
worker_state.revoked  is  LimitedSet.... used  in worker/state.py:
```py
revoked = LimitedSet(maxlen=REVOKES_MAX, expires=REVOKE_EXPIRES)```
```
where LimitedSet is defined in datastructures.py.

Problem is this: using update on LimitedSet will not purge old entries out.
When we add one item into LimitedSet, it will stay the same size (which could be > maxlen).
The only  way to make the size limited is to call method `purge()`

Unit tests on LimitedSet (in tests/utils/test_datastructures.py)  are not using `expires`, they only use LimitedSet(maxlen=...). Also there are no tests on `s2.update(s1._data)` -- like it is used in file `worker/control.py`

Examples:
```py
In [155]: s1 = LimitedSet(maxlen=5)

In [156]: for i in range(200): s1.add(i)

In [157]: s1
Out[157]: LimitedSet(5)
```

now with expires:

```py
In [158]: s_exp=LimitedSet(maxlen=5, expires=5)

In [159]: for i in range(300): s_exp.add(i)

In [160]: s_exp
Out[160]: LimitedSet(300)

In [161]: s_exp.add(500)

In [162]: s_exp
Out[162]: LimitedSet(300)
In [163]: s_exp.purge()

In [164]: s_exp
Out[164]: LimitedSet(5)
```
As you can see, until we call purge, the LimitedSet class is not really living up to its name. Sadly this can eat lots of memory,

**Proposed solution:**

1.  Few more Unittests are needed.
2. when LimitedSet object has expires, 
  - run purge after every update which will make it longer than maxlen.
  - run purge(limit=None) after every add,  if we are longer then maxlen when method add starts.

alekibango:
the above code is with stable version, git version is different, but also wrong.
Git version will not remove items if they are not expired, even if we have too many of them.
It deletes only one on adding (if the first is expired), so the list can grow too.

[...]

Issue #3086, opened 2016-03-02
Revoke list causes unbounded memory growth


djsmith42:
Celery version tested: **3.1.16** and **3.1.20**
OS: **Ubuntu 12.04** (OS version should not matter)
Note: I am **not** using `--statedb`.

**Background:**

According to `celery/worker/state.py`, the revoke list should be limited to 50,000 entries after 3 hours:

```python
REVOKES_MAX = 50000
REVOKE_EXPIRES = 10800
revoked = LimitedSet(maxlen=REVOKES_MAX, expires=REVOKE_EXPIRES)
```

But this does not happen in practice.

**To reproduce:**

1. Revoke a lot of tasks (any ID will work):

  ```python
  app.control.revoke(['test-task-id-{}'.format(n) for n in xrange(100000)])
  # Revoking tasks one-at-a-time also reproduces the problem, but slower
  ```

2. Dump the revoke list:

  ```python
  for host, revokes in app.control.inspect(timeout=2).revoked().items():
        print host, len(revokes)
  ```

3. Come back 3+ hours later and revoke 1 task, and note that the list is still over 200,000.

**Workaround:**

If I modify the code like this, the tasks are limited to 50,000 in memory, no matter how many tasks I revoke:

```python
REVOKES_MAX = 50000
REVOKE_EXPIRES = None # Changed from 10800
revoked = LimitedSet(maxlen=REVOKES_MAX, expires=REVOKE_EXPIRES)
```

**Impact:**

Our production servers were using over 1GB of memory due to an accumulated 1 million revoked tasks after a few weeks of uptime. This became 15GB because we have 15 pre-fork worker processes which eventually re-spawned and inherited the 1GB of memory from the master process. In our cluster of 5 servers, restarting a server does not help because it mingles with the other servers on startup and reloads all the revokes into memory. We had to shut down every server in our cluster to get them all purged from memory.

**Quirks:**

When I play with `LimitedSet` in a Python shell, the expiration feature seems to behave as I expect. If I create a `LimitedSet(maxlen=3, expires=3)` and add 3 integers, wait 3 seconds, and add 1 more, there are only 3 integers in the set, as expected. I don't understand why this issue is manifesting like it is.

djsmith42:
Here is a failing unit test. Did I understand how `LimitedSet` is supposed to work? If this test were passing, I think the unbounded memory growth would not be happening:

```python
from celery.datastructures import LimitedSet
from time import time

def test_limited_set():
[...]

PR #3102, opened 2016-03-09
Improvements and fixes for LimitedSet

alekibango:
Getting rid of leaking memory + adding minlen size of the set.
Minlen is minimal residual size of set after operating for long time.
Minlen items are kept, even if they should be expired by time, until
we get newer items.

Problems with older and even more old code:

1. Heap would tend to grow in some scenarios
   (like adding an item multiple times).

2. Adding many items fast would not clean them soon enough (if ever).

3. When talking to other workers, revoked._data was sent, but
   it was processed on the other side as iterable.
   That means giving those keys new (current)
   timestamp. By doing this workers could recycle
   items forever. Combined with 1) and 2), this means that in
   large set of workers, you are getting out of memory soon.

All those problems should be fixed now,
also some new unittests are added.

This should fix issues #3095, #3086.

### DIFF
--- a/celery/datastructures.py
+++ b/celery/datastructures.py
@@ -686,7 +686,7 @@ class LimitedSet(object):
         now = now or time.time()
         if item in self._data:
             self.discard(item)
-        entry = (now, item)
+        entry = [now, item]  # this must be a list, we are using pointers.
         self._data[item] = entry
         heappush(self._heap, entry)
         if self.maxlen and len(self._data) >= self.maxlen:
--- a/celery/datastructures.py
+++ b/celery/datastructures.py
@@ -601,8 +601,7 @@ class LimitedSet(object):
     You can also configure minlen, which is the minimal residual size
     of the set.
 
-    All  arguments are optional, with exception of minlen, which must
-    be smaller than maxlen.  Unconfigured limits will not be enforced.
+    All arguments are optional, and no limits are enabled by default.
 
     :keyword maxlen: Optional max number of items.
 
@@ -611,10 +610,12 @@ class LimitedSet(object):
 
     :keyword expires: TTL for all items.
 
-        Items aging over expiration are purged as keys are inserted.
+        Expired items are purged as keys are inserted.
 
     :keyword minlen: Minimal residual size of this set.
         .. versionadded:: 4.0
+
+        Value must be less than ``maxlen`` if both are configured.
 
         Older expired items will be deleted, only after the set
         exceeds minlen number of items.
@@ -693,7 +694,9 @@ class LimitedSet(object):
             self.purge()
 
     def update(self, other):
-        """Update this LimitedSet from other LimitedSet, dict or iterable."""
+        """Update this set from other LimitedSet, dict or iterable."""
+        if not other:
+            return
         if isinstance(other, LimitedSet):
             self._data.update(other._data)
             self._refresh_heap()
@@ -701,7 +704,7 @@ class LimitedSet(object):
         elif isinstance(other, dict):
             # revokes are sent as a dict
             for key, inserted in items(other):
-                if isinstance(inserted, list):
+                if isinstance(inserted, (tuple, list)):
                     # in case someone uses ._data directly for sending update
                     inserted = inserted[0]
                 if not isinstance(inserted, float):
@@ -720,7 +723,6 @@ class LimitedSet(object):
         # mark an existing item as removed. If KeyError is not found, pass.
         entry = self._data.pop(item, sentinel)
         if entry is not sentinel:
-            entry[-1] = sentinel
             if self._heap_overload > self.max_heap_percent_overload:
                 self._refresh_heap()
     pop_value = discard
@@ -793,6 +795,10 @@ class LimitedSet(object):
     def __reduce__(self):
         return self.__class__, (
             self.maxlen, self.expires, self.as_dict(), self.minlen)
+
+    def __bool__(self):
+        return bool(self._data)
+    __nonzero__ = __bool__  # Py2
 
     @property
     def _heap_overload(self):
--- a/celery/datastructures.py
+++ b/celery/datastructures.py
@@ -673,7 +673,6 @@ class LimitedSet(object):
         """Time consuming recreating of heap. Do not run this too often."""
         self._heap[:] = [
             entry for entry in values(self._data)
-            if entry is not sentinel
         ]
         heapify(self._heap)
 
@@ -751,7 +750,8 @@ class LimitedSet(object):
         """Remove and return the oldest item, or :const:`None` when empty."""
         while self._heap:
             _, item = heappop(self._heap)
-            if self._data.pop(item, None) is not sentinel:
+            if item is not sentinel:
+                del self.data[item]
                 return item
         return default
 
--- a/celery/datastructures.py
+++ b/celery/datastructures.py
@@ -719,7 +719,10 @@ class LimitedSet(object):
                 self.add(obj)
 
     def discard(self, item):
-        """Remove item from LimitedSet. Return None if item was not found."""
+        """Remove item from :class:`LimitedSet`.
+
+        :keyword item: item to delete. If item is not found, nothing happens.
+        """
 
         entry = self._data.pop(item, sentinel)
         if entry is not sentinel:  # item was found in _data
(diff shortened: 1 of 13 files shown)


celery/datastructures.py
```diff
@@ -583,120 +583,217 @@ def values(self):
...
+        return 'LimitedSet(maxlen={0}, expires={1}, minlen={2})' \
+            ' Current size:{3}'.format(
+                self.maxlen, self.expires, self.minlen, len(self._data))
 
     def __iter__(self):
-        return (item[1] for item in self._heap)
+        # return (item[1] for item in
+        #         self._heap if item[-1] is not self.REMOVED)
+        # ^ not ordered, slow
+        return (i for _, i in sorted(self._data.values()))
```
ask: The heap is ordered, no? Surely sorting the data must be much slower than iterating over the heap?

celery/datastructures.py
```diff
@@ -583,120 +583,217 @@ def values(self):
...
     def __contains__(self, key):
         return key in self._data
 
     def __reduce__(self):
-        return self.__class__, (self.maxlen, self.expires, self._data)
+        """Pickle helper class.
+
+        This object can be pickled and upickled."""
+        return self.__class__, (
+            self.maxlen, self.expires, self.minlen, self.as_dict())
```
ask: This is actually backwards incompatible with data serialized with older versions.

All new keyword arguments must be added to the end, so `minlen` must be positioned after `data`

celery/datastructures.py
```diff
@@ -583,120 +583,217 @@ def values(self):
...
+                inserted_time, _ = self._heap[0]
+                if inserted_time + self.expires > now:
+                    break  # end this right now, oldest item is not expired yet
+                self.pop()
+
+    def pop(self):
+        'Remove and return the lowest time item. Return None if empty.'
+        while self._heap:
+            _, item = heappop(self._heap)
+            if item is not self.REMOVED:
```
ask: I cannot see any case here where item should be `REMOVED`?

Unless you count _refresh_heap, and I think we can just as well filter REMOVED values there

celery/datastructures.py
alekibango: well, heap is actually not ordered as you might think...
https://docs.python.org/2/library/heapq.html#priority-queue-implementation-notes says:
The interesting property of a heap is that its smallest element is always the root, heap[0].
Try it yourself: 
```py 
h=[x for x in range(10,1, -1)]
heapq.heapify(h)
print(h)
[2, 3, 4, 7, 6, 5, 8, 9, 10]
```
You might use heapq.nsmallest(2*len(h), h), which you think might be fast. But then you need to filter out removed items.   Still, sorted(self._data.values()) is as fast as heapq.nsmallest(2*len(s._heap), s._heap).  I tried to measure few things before i settled on that line :)

All the time i was coding this, while being angry at heapq, the only possible method i see to improve this code is to just use fifo queue of fixed limited size, and forget extra expiration ideas. It might make the code simpler and faster. But adding revoked items from other workers should not go to the front of the queue... it should merge in correct time -- which was not yet done correctly, by the way.

celery/datastructures.py
alekibango: this is prolly mistake on my side... i didnt notice any code using that.
I will reorder it, along with arguments of init()

celery/datastructures.py
alekibango: I am trying to call _refresh_heap()  only if size of _heap is bigger than acceptable (see variable _MAX_HEAP_PERCENTS_OVERLOAD , which is hardcoded to be  15% over the size of _data).
Reason: _refresh_heap is taking way too much time. It really makes the speed difference to make this refresh a bit lazy.  

Until we clean up the heap, REMOVED is needed to mark removed entries in the _heap.
Previous code relied on update to make (sometimes) refresh.  But by the way the code was used in the wild (*passing _data directly*), it failed to do update; instead it did many adds, without correct insertion timestamp. And the refresh (purge) method was never actually used there. 

Now we are rehashing, just when ammount of garbage in _heap requires it, on every add or update operation, along with purging all items which should be gone.. 
My first attempts to fix this code was to run purge everytime... but now it is really thousands times faster.

alekibango: to see REMOVED items in _heap, try this:
```py 
s = LimitedSet(maxlen=20)
for i in range(8): 
    s.add(i)
s.discard(2)
print(s._heap)
```
Note you must add at least 8 items, as 15 % of 8 is 1.2, which means we can have 1 discarded item marked as REMOVED. Discarding 2 of 8 items will initiate garbage collection (_refresh_heap()).

alekibango: On the other though, 
maybe it is possible to effectively remove items from heap and avoid using marks.
I will play with it a bit and report results to you here.

alekibango: Good reading about [deleting from heapq  at stackoverflow](http://stackoverflow.com/questions/5484929/removing-an-item-from-a-priority-queue)

There is a post showing how to remove item effectively if we know its position. But finding that position is also up to O(n) slow. So i guess my solution with lazy removing is best we can do with heapq.  Maybe we rather should take look at [treap](http://stromberg.dnsalias.org/~dstromberg/treap/) if we need something faster. Table comparing speed of different methods is very nice on that page.  

Speed of my code:
```py
def speedtest():
    s = LimitedSet(maxlen=50000, minlen=20000, expires=1000)
    start = time.time()
    for i in range(1000000, 0, -1):
        s.add(i)
    middle = time.time()
    for i in s:
        s.discard(i)
    end = time.time()
    print('adding 1M items, keeping only 50k last:', middle - start)
    print('deleting 50k items', end - middle)

adding 1M items, keeping only 50k last: 8.85692286491
deleting 50k items 0.271380186081
```

ask: Oh, yeah, I definitely think lazily purging the heap is a huge improvement.
The treap module looks interesting, hope I get the chance to use it some time.

AFAICS the only place where items in _heap can end up being `REMOVED` is in `refresh_heap`:

```
self._heap[:] = [entry for entry in self._data.values()]
```

But then this is doing the wrong check?:

celery/datastructures.py
ask: Oh yeah, this actually bit me before, but I guess not painful enough for me to remember ;)

Nothing is actually using `__iter__` as far as I can tell, so I guess it doesn't matter much anyway.

ask:
Merged into master!  Can you please go over the changes I made in 7c4c6eb to make sure I didn't miss anything?

You should also add your name to https://github.com/celery/celery/edit/master/CONTRIBUTORS.txt

ask:
and also this one: 32a088

celery/datastructures.py
alekibango: What the check is doing: it makes sure we remove from data oldest item, ordering provided by _heap. Exactly one existing item from data.    

Items are only deleted from _heap when we do refresh_heap() or pop() (which removes exactly one).

They are marked 'REMOVED' inside the _heap, when we discard(item) or  add(item) we already have in the set.

alekibango:
@ask: sorry for that chaos with git, i never did so many merges and diffs at one day, it somehow made me do some mistakes. Last version should be ok now.

ask:
Ah, I see.  But we don't actually need to mark it as removed in the heap do we? Pop can just continue if the key is not in `_data`:

```
    def pop(self, default=None):
        """Remove and return the oldest item, or :const:`None` when empty."""
        while self._heap:
            _, item = heappop(self._heap)
            try:
                self._data.pop(item)
            except KeyError:
                pass
            else:
                return item
        return default
```

This way `sentinel` does not have to be used at all, if we also handle KeyError in `discard`:

```
    def discard(self, item):
        # mark an existing item as removed. If KeyError is not found, pass.
        try:
            entry = self._data.pop(item)
        except KeyError:
            pass
        else:
            if self._heap_overload > self.max_heap_percent_overload:
                self._refresh_heap()
```

ask:
Btw, I did a diff against our branches, and master is very close to yours now, except for removing the sentinel.

alekibango:
Yes, i expected something like this.
Removing sentinel was interesting move. changes functionality a bit (purging readded item is possible), but that  is probably not a big problem...  i will double check the code during weekend.

ask:
The branch is already merged btw into master, but forgot to close.

Again, please add your name to https://github.com/celery/celery/edit/master/CONTRIBUTORS.txt

PR #3106, opened 2016-03-10
Describe the `virtual_host`  parameter

alexkiro:
Add details about changing the database number while using UNIX 
socket with a Redis broker.


ask:
Thanks! Merged into master and 3.1 :)

PR #3108, opened 2016-03-10
Make gevent.TaskPool to tell it's pool limit

oblalex:
This is a little improvement. It makes gevent.TaskPool to return information about pool size. Such info is currently totally missing due to default `BasePool` implementation. `eventlet`, `prefork` and `solo` implementations already have this.

This will allow to get pool size via `control.inspect().stats()`. This is essential to have such feature, for example, to know how much tasks a single pool can consume at one time.

### DIFF
--- a/celery/concurrency/gevent.py
+++ b/celery/concurrency/gevent.py
@@ -121,3 +121,8 @@ class TaskPool(BasePool):
     @property
     def num_processes(self):
         return len(self._pool)
+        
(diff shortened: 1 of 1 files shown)


ask:
Good idea!  Merged into master.

As all the pools have the `limit` attribute, I moved it to the base class.

oblalex:
Great! Thanks!

PR #3109, opened 2016-03-10
Prevent duplicates in LimitedSet

djsmith42:
I think this will help prevent dups that are currently reported for me in 3.1.22 by `app.control.inspect().revoked()`

### DIFF
--- a/celery/datastructures.py
+++ b/celery/datastructures.py
@@ -618,6 +618,7 @@ class LimitedSet(object):
         # and it will end up in the correct order.
         self.purge(1, offset=1)
         inserted = now()
+        self.discard(key) # remove this key if it exists to prevent dups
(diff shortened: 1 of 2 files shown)


djsmith42:
@ask I think this PR is pretty important. Without it, Celery has unbounded memory growth if callers revoke tasks, especially if mingle is enabled and there are multiple servers, and especially in prefork mode. Once a server restarts, it re-mingles with the others, and they all store a copy of all revoked task IDs.

In my production environment, this has led to 10GB of memory usage on a server with concurrency 10. Restarting the servers actually makes the problem worse, and there's no good workaround that I can find.

ask:
This was already fixed by #3102, but I merged in your test case!

PR #3114, opened 2016-03-11
Database backend url passing

ZoranPavlovic:
Fixed bug where database backend not passing through URL to BaseBackend __init__.


ask:
Thanks! `self.url` is set a few lines down, but this is definitely more correct.

Issue #1967, opened 2014-04-09
autoscaler: Scaling down -N processes


tschottdorf:
I've been using the autoscaler both on two gevent workers and, as of recently, on another worker using prefork. In the logs, I see a lot of "negative scale-downs", such as:
```
 Apr 01 19:40:45 app/worker.1: [2014-04-02 02:40:45,280: INFO/MainProcess] Scaling down -9 processes.
Apr 01 21:04:38 api app/worker.1: [2014-04-02 04:04:38,253: INFO/MainProcess] Scaling up 12 processes.
```
(There are also "expected" scale-downs with the other sign)

This appears to begin with the following pattern:
The worker scales up a bunch of times as a wad of tasks comes in. After the tasks are done, it scales down but it looks like it's scaling down too much. Here's an example of the pattern:
```

Apr 08 11:11:08 api app/cpuworker.1: [2014-04-08 18:11:07,973: INFO/MainProcess] Scaling up 1 processes.
Apr 08 11:11:32 api app/cpuworker.1: [2014-04-08 18:11:32,290: INFO/MainProcess] Scaling up 1 processes.
Apr 08 11:12:28 api app/cpuworker.1: [2014-04-08 18:12:27,979: INFO/MainProcess] Scaling down -3 processes.
Apr 08 11:12:58 api app/cpuworker.1: [2014-04-08 18:12:58,199: INFO/MainProcess] Scaling down -4 processes.
Apr 08 11:13:55 api app/cpuworker.1: [2014-04-08 18:13:55,506: INFO/MainProcess] Scaling up 1 processes.
Apr 08 11:14:28 api app/cpuworker.1: [2014-04-08 18:14:28,531: INFO/MainProcess] Scaling down -4 processes.
Apr 08 11:14:46 api app/cpuworker.1: [2014-04-08 18:14:46,059: INFO/MainProcess] Scaling up 1 processes.
Apr 08 11:15:28 api app/cpuworker.1: [2014-04-08 18:15:28,779: INFO/MainProcess] Scaling down -4 processes.
Apr 08 11:15:46 api app/cpuworker.1: [2014-04-08 18:15:46,864: INFO/MainProcess] Scaling up 1 processes.
Apr 08 11:16:29 api app/cpuworker.1: [2014-04-08 18:16:28,955: INFO/MainProcess] Scaling down -4 processes.
[...]

PR #3120, opened 2016-03-16
test to ensure that celery does not scale down by negative values

komuW:
This is in relation to this issue: https://github.com/celery/celery/issues/1967

So, it looks like that issue has been fixed in the current master(tip) but there has not been any testcase. This fixes that.
@insertjokehere had intimated that he had a fix but didn't have a good testcase; https://github.com/celery/celery/issues/1967#issuecomment-196491985 

So the gist of this pull request is to add the neccesary testcase

### DIFF
--- a/celery/tests/worker/test_autoscale.py
+++ b/celery/tests/worker/test_autoscale.py
@@ -209,7 +209,6 @@ class test_Autoscaler(AppCase):
             x.body()
             total_num_processes.append(self.pool.num_processes)
 
-        self. assertTrue(all(i <= x.min_concurrency for i in total_num_processes)
-                        )
-        self. assertTrue(all(i <= x.max_concurrency for i in total_num_processes)
-                        )
+        self. assertTrue(
+            all(x.min_concurrency <= i <= x.max_concurrency for i in total_num_processes)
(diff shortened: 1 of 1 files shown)


thedrow:
This needs to be rebased. I'll review it later.

komuW:
Hi @thedrow.
I don't mean to bug you, but have you had time to review my pull request? 
I would appreciate any feedback.

thedrow:
I didn't. I'll have time in a bit. Thanks for the ping.

celery/worker/autoscale.py
```diff
@@ -133,11 +135,13 @@ def scale_down(self, n):
 
     def _grow(self, n):
         info('Scaling up %s processes.', n)
+        self.growth_numbers.append(n)
```
ask: I think this part should be in the test, not in the original class

thedrow:
LGTM. Please rebase and I'll merge.

komuW:
@thedrow I have rebased. 
Also following a comment by @ask , I've refactored the test to be self-sufficient and thus this pull request does not make any changes to celery/worker/autoscale.py

PR #3122, opened 2016-03-18
updated docs about migrating celery in Django

gitaarik:
South is deprecated because Django has migrations native now


ask:
Thanks!

PR #3125, opened 2016-03-20
Update and fix links to external pages

tgwizard:
This PR updates links to external pages. It:

1. Changes http:// to https:// for many pages, like GitHub and Wikipedia, which always redirect to https.
2. Changes https://github.com/celery/pyamqp/issues, which returns a 404, to https://github.com/celery/py-amqp/issues
3. Updates links to other services, where to skip unnecessary redirects.

One thing I didn't fix is the link to http://leapfrogdevelopment.github.com/rcelery/. That project, and organization, doesn't seem to exist any more.

What do you think? I'd be happy to squash the commits, or skip some of them if you think this changes too much.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -245,7 +245,7 @@ issue tracker.
 
 * Celery: http://github.com/celery/celery/issues/
 * Kombu: http://github.com/celery/kombu/issues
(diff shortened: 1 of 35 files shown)


tgwizard:
I don't think my changes are the cause of the appveyor test failures. It seems as if some dependencies are not installed there.

malinoff:
Thanks!

PR #3136, opened 2016-03-29
Correct help for migrating in Django

MichaelAquilina:
The help provided does not work for Django 1.9 and should be telling the user to use `migrate` rather than the deprecated `syncdb` by default (https://docs.djangoproject.com/en/1.8/ref/django-admin/#syncdb)

### DIFF
--- a/docs/getting-started/brokers/django.rst
+++ b/docs/getting-started/brokers/django.rst
@@ -36,6 +36,10 @@ configuration values.
 
 .. code-block:: console
 
+    $ python manage.py migrate kombu_transport_django
+
+ Or if you are using a version of Django lower than 1.7
+
     $ python manage.py syncdb
 
 .. _broker-django-limitations:


MichaelAquilina:
@adamchainz

docs/getting-started/brokers/django.rst
```diff
@@ -37,6 +37,10 @@ configuration values.
 .. code-block:: console
 
     $ python manage.py syncdb
+    
+ Or if you are using Django 1.9 or later
+ 
+    $ python manage.py migrate kombu_transport_django
```
MichaelAquilina: As @adamchainz has already pointed out, this should be the preferred mechanism from Django 1.7 onwards. Django 1.9 just removed the `syncdb` command.

docs/getting-started/brokers/django.rst
```diff
@@ -37,6 +37,10 @@ configuration values.
 .. code-block:: console
 
     $ python manage.py syncdb
+    
+ Or if you are using Django 1.9 or later
```
adamchainz: I would put this first, and leave the `syncdb` for Django < 1.7 (is this even supported any more?). `migrate` was added in 1.7.

MichaelAquilina: I think this is probably the better solution

MichaelAquilina:
The test failure on appveyor looks unrelated to my change?

MichaelAquilina:
Any feedback on this or an ETA when it will be reviewed?

ask:

PR #3162, opened 2016-04-12
Avoid unicode in task.__repr__ in python 2

vissh:


thedrow:
@ask Should we port this to master as well?

ask:
I wouldn't feel safe enabling unicode_literals in 3.1, as I had to fix a huge amount of bugs for that in 4.0.  It's impossible to know in advance where it will be a problem, so better to leave that to 4.

We should probably have a `@python_2_repr_compatible` class, like we have with `@python_2_unicode_compatible`

ask:
Oh, wait! Our implementation of `@python_2_unicode_compatible` already fixes `__repr__` :D

thedrow:
So we don't need this PR?

ask:
The patch fixes support for unicode in task names when doing `repr(task)`.

I'm not sure if that works in master either, as __repr__ on 2.7 must always return bytes, not unicode and we don't use safe_str there, only `unicode.encode()` which will crash if there are unicode characters.

`kombu.utils.encoding.safe_str` and `.safe_repr` should mostly be used for strings that needs to be printed/written to file, as it uses `sys.getdefaultencoding` to try and guess what needs to be done as best as possible (words cannot even explain how difficult this is in Python 2).  It may have too many side effects for me to comfortably use it in `__repr__`.

We could use `unicode.encode('ascii', 'replace')` instead, to just replace the unicode characters with
garbage, but I would like to see that patch applied to master  then we can decide if we want to fix it in 3.1.

ask:
Let's just focus on 4.0 for now.  Moving forward :)

PR #3181, opened 2016-04-25
docs: broker_connection_timeout disabled for gevent

veatch:
Mentioned in 3.0.10 release notes: http://docs.celeryproject.org/en/3.0/changelog.html#version-3-0-10
Implemented here: https://github.com/celery/celery/blob/d23c8c/celery/worker/consumer/consumer.py#L205-L209


thedrow:
Thanks!

PR #3186, opened 2016-04-27
Update the demo project so it works with the latest Django and celery

batiste:
Hi Ask how are you doing?

Using a bit of Celery at work and I had to update my knowledge.
Some changes in the demo app for Django seems necessary!

Cheers to you and hope everything is well!

### DIFF
--- a/examples/django/README.rst
+++ b/examples/django/README.rst
@@ -53,7 +53,7 @@ Running a task
 
 .. code-block:: console
 
-    $ python ./manage.sh shell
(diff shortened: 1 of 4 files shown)


ask:
Thanks! Merged into master, with some slight changes.

I'm doing well, nice to see you around here :)

The autodiscover was changed not accept any argument to account for the new Django Appconfigs.
I'm guessing you used the master example with celery 3.1.

PR #3192, opened 2016-05-02
Grammar

jeffwidman:


jeffwidman:
I'm not totally confident this is the proper fix. It's certainly better grammar, but the explanation for `map` says: 

> [map/starmap] differ from group in that: 
1. only one task message is sent 
2. the operation is sequential.

Which is confusing, because the docs that I fixed appear to say that `group` executes tasks sequentially as well.

thedrow:
If you call the group using () it will be performed eagerly. If you call with .delay() the group will execute the subtasks in parallel.

PR #3203, opened 2016-05-12
passing callbacks to accumulate when replacing tasks

### LINKED ISSUE #2722: Replace only works inside a chord
When using task.replace() as shown below, the task must be called in a chord not a chain. While I can work around this, it's probably best if replace worked in chains as well.

```python
@app.task(bind=True)
def lazy_async_apply_map(self, items, d, runner):
    """Make new instances of runner for each item in items, and inject that
    group into the chord that executed this task.

    :param items: itterable of arguments for the runner
    :param d: data to operate on (probably returned by a previous task)
    :param runner: task signature to execute on each item. def runner(item, data, *a, **kw)
    """

    subtasks = []
    for item in items:
        r = runner.clone()
        r.args = (item, d) + r.args
        subtasks.append(r)

    raise self.replace(group(*subtasks))
```

Tests:
```
t = generator.s(items) | lazy_async_apply_map.s(d, runner=generator.s()) | dict_reducer.s()
t.delay().get()
```
Hangs forever because the `celery.accumulate` task doesn't get connected to `dict_reducer`.

BUT:

```
c = generator.s(items) | chord(lazy_async_apply_map.s(d, runner=generator.s()), dict_reducer.s())
c.delay().get()
```
Using chord instead of chain, this works completely as expected. I can post code for generator and dict_reducer if needed, but they are pretty trivial.

nicolasunravel:
Attempt to fix #2722 
debug mode showed that the accumulate didn't have the callbacks of the replaced task.

Now works for me, unit tests suite passes as well.

### DIFF
--- a/celery/tests/tasks/test_tasks.py
+++ b/celery/tests/tasks/test_tasks.py
@@ -408,7 +408,19 @@ class test_tasks(TasksCase):
         with self.assertRaises(Ignore):
             self.mytask.replace(sig1)
 
-    def test_replace__group(self):
+
+    def test_replace_callback(self):
+        c = group([self.mytask.s()], app=self.app)
+        c.freeze = Mock(name='freeze')
+        c.delay = Mock(name='delay')
+        self.mytask.request.id = 'id'
+        self.mytask.request.group = 'group'
+        self.mytask.request.root_id = 'root_id',
+        self.mytask.request.callbacks = 'callbacks',
+        with self.assertRaises(Ignore):
+            self.mytask.replace(c)
+
+    def test_replace_group(self):
         c = group([self.mytask.s()], app=self.app)
         c.freeze = Mock(name='freeze')
         c.delay = Mock(name='delay')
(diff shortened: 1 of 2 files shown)


thedrow:
Do you think that the existing tests cover this behaviour? If not, do you mind adding a new test?

nicolasunravel:
I don't mind adding a unit test in this regard when I got some time for it. Yet the unit tests around that are pretty much all mocked up, I'm not sure I'll test more than the attribute is being set.

thedrow:
If you don't mind, I'd like to ensure that this won't break somehow.

thedrow:
There's a small lint error https://travis-ci.org/celery/celery/jobs/132265004#L171
Other than that, looks good.

celery/tests/tasks/test_tasks.py
```diff
@@ -124,7 +131,6 @@ class MyCustomException(Exception):
 
 
 class test_task_retries(TasksCase):
-
```
ask: there *should* be space between methods in a class, that includes the first method :)

ask:
Merged into master! Please add your name to https://github.com/celery/celery/edit/master/CONTRIBUTORS.txt !

PR #3207, opened 2016-05-15
Fix syntax error in one example at the Tasks userguide doc.

tvieira:
While going through the examples in the Tasks user guide, I found this syntax error in the last example, for the blog/tasks.py file - the variable should be 'current_domain' and not 'domain'. I also added an additional line so we don't have a horizontal scroll in the code box.


thedrow:
Thanks!

Issue #3212, opened 2016-05-17
BROKER_URL and CELERY_RESULT_BACKEND not passed to worker processes properly when CELERYD_FORCE_EXECV 


jzafran:
Celery worker app config doesn't have `BROKER_URL` or `CELERY_RESULT_BACKEND`  set when `CELERYD_FORCE_EXECV` is used and RabbitMQ is not on the local host, so any calls to `self.retry` in a bound task case an `ConnectionError` to be raised.  This occurs on Celery 3.1.23.

> [2016-05-17 13:57:18,571: WARNING/MainProcess] Task add[f00f4535-c5bc-4f61-836e-5d559c2461c5] reject requeue=False: Error opening socket: a socket error occurred
Traceback (most recent call last):
  File "/Users/user/celery/celery/app/trace.py", line 240, in trace_task
    R = retval = fun(*args, **kwargs)
  File "/Users/user/celery/celery/app/trace.py", line 438, in __protected_call__
    return self.run(*args, **kwargs)
  File "/Users/user/celery/examples/testapp/testapp.py", line 25, in add
    self.retry()
  File "/Users/user/celery/celery/app/task.py", line 694, in retry
    raise Reject(exc, requeue=False)
Reject: (ConnectionError('Error opening socket: a socket error occurred',), False)

# Steps to reproduce

1. Use Celery 3.1.23. This originally occurred on Ubuntu 12.04.5 LTS, but was also reproduced on OS X 10.11.2 with Python 2.7.10.
2. Grab a copy of https://gist.github.com/jzafran/e6eb82afe79d64685a0620874a53b648
3. Set `BROKER_URL` to a RabbitMQ server that is *not* on the localhost. 
4. Run a Celery worker:
```bash
python testapp.py worker -l info
```
5. In a Python shell, import and enqueue the task
```python
from testapp import test_task
test_task.s(1, 2).delay()
```
In the worker logs, you should see above stacktrace.

jzafran:
This looks related to https://github.com/celery/celery/issues/3157

ask:
Fixed by #3214 :)

jzafran:
@ask  thanks for merging into master.  What are the changes this could be merged into the 3.1 branch to be included with v.3.1.34?

PR #3214, opened 2016-05-17
Update the Celery app config with preconfig as non-defaults so they will be pickled

jzafran:
Issue: https://github.com/celery/celery/issues/3212

This PR updates `Celery._get_config` to update the `conf` with `self._preconf` as non-defaults so they will be included in the pickled app.

Without this fix, any Celery tasks that (such as `task.retry`) fail because the BROKER_URL is not set properly in the unpickled app.  Additionally, results from these tasks are not saved in the result backend, because `CELERY_RESULT_BACKEND` is also not saved.  See https://github.com/celery/celery/issues/3212

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -197,3 +197,4 @@ Krzysztof Bujniewicz, 2015/10/21
 Sukrit Khera, 2015/10/26
 Dave Smith, 2015/10/27
 Dennis Brakhane, 2015/10/30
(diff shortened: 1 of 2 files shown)


ask:
Thanks! Merged!

jzafran:
@ask  thanks for merging into master.  What are the changes this could be merged into the 3.1 branch to be included with v.3.1.34?

ask:
Seems this is already in the 3.1 branch!

Issue #3232, opened 2016-06-03
[4.0.0rc2] Tasks not being executed 


honux:
The task 'Single.Task' (foo) will not be executed at all

```python
# The worker
from .celery import app

@app.task(name='Single.Task')
def single_task(from_task, outter='None'):
  print "Single.Task from %s - %s" % (from_task, outter)
  to_task = "Single.Task(%s)" % from_task
  return to_task

@app.task(bind=True, name='Double.Task')
def double_task(self, from_task):
  print "Double.Task from %s" % from_task
  to_task = "Double.Task(%s)" % from_task
  return self.replace(app.signature('Single.Task', args=(to_task,)) | app.signature('Single.Task'))
```

```python
# The client
t = app.signature('Double.Task', args=('1',)) | app.signature('Single.Task', args=('foo',))
t()
```

I believe it is something related to the self.replace, since 
```python
# The client
t = app.signature('Single.Task', args=('1',)) | app.signature('Single.Task', args=('foo',))
t()
```
runs OK.

BTW, is there any official way to cancel a chain?

honux:
Fixed the problem, at least for this specific problem and to the possible non-execution of a chord on the actual task.

BTW, use the latest version of the code, since the codding style on the first version was different (2 spaces instead of 4)

PR #3236, opened 2016-06-04
Updated Tasks.py

honux:
Fixed issue #3232, adding the signature to the chain (if there is any).
Fixed the chord suppress if the given signature contains one.


ask:
Thanks! Merged into master.

Please also add your name to https://github.com/celery/celery/edit/master/CONTRIBUTORS.txt !

Issue #2546, opened 2015-03-20
ALWAYS_EAGER: TypeError: run() got multiple values for keyword argument 'header'


jonblack:
I have a complex nested workflow that works fine until `CELERY_ALWAYS_EAGER` is set to `True` (which is required for testing). Then I get the following error:

    Traceback (most recent call last):
      File "/home/jon/.virtualenvs/hope-flask/local/lib/python2.7/site-packages/behave/model.py", line 1456, in run
        match.run(runner.context)
      File "/home/jon/.virtualenvs/hope-flask/local/lib/python2.7/site-packages/behave/model.py", line 1903, in run
        self.func(context, *args, **kwargs)
      File "tests/features/steps/report.py", line 47, in step_create_report
        r = context.app.post('api/create/', data=payload)
      File "/home/jon/.virtualenvs/hope-flask/local/lib/python2.7/site-packages/werkzeug/test.py", line 772, in post
        return self.open(*args, **kw)
      File "/home/jon/.virtualenvs/hope-flask/local/lib/python2.7/site-packages/flask/testing.py", line 108, in open
        follow_redirects=follow_redirects)
      File "/home/jon/.virtualenvs/hope-flask/local/lib/python2.7/site-packages/werkzeug/test.py", line 736, in open
        response = self.run_wsgi_app(environ, buffered=buffered)
      File "/home/jon/.virtualenvs/hope-flask/local/lib/python2.7/site-packages/werkzeug/test.py", line 659, in run_wsgi_app
        rv = run_wsgi_app(self.application, environ, buffered=buffered)
      File "/home/jon/.virtualenvs/hope-flask/local/lib/python2.7/site-packages/werkzeug/test.py", line 855, in run_wsgi_app
        app_iter = app(environ, start_response)
      File "/home/jon/.virtualenvs/hope-flask/local/lib/python2.7/site-packages/flask/app.py", line 1836, in __call__
        return self.wsgi_app(environ, start_response)
      File "/home/jon/projects/hope-flask/hope/middleware.py", line 36, in __call__
        return self.app(environ, start_response)
      File "/home/jon/.virtualenvs/hope-flask/local/lib/python2.7/site-packages/flask/app.py", line 1820, in wsgi_app
        response = self.make_response(self.handle_exception(e))
      File "/home/jon/.virtualenvs/hope-flask/local/lib/python2.7/site-packages/flask/app.py", line 1403, in handle_exception
        reraise(exc_type, exc_value, tb)
[...]

PR #3244, opened 2016-06-09
Fix chord arguments

tayfun:
This would fix https://github.com/celery/celery/issues/2546

### DIFF
--- a/celery/canvas.py
+++ b/celery/canvas.py
@@ -581,6 +581,15 @@ class chord(Signature):
         )
         self.subtask_type = 'chord'
 
+    def apply(self, partial_args=(), args=(), kwargs={}, **options):
+        """Apply the chord task with partial_args."""
+        # partial_args which will passed on to the header tasks are put in
+        # kwargs to be passed on.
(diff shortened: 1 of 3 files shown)


thedrow:
I wonder why the tests passed before? Is this covered?

tayfun:
@thedrow it wasn't covered before. I have added one test which has a chord providing an argument to its header tasks. That should cover this part.

thedrow:
LGTM.

ask:
I had to revert this as it broke #3504 :(

We cannot add new arguments to an existing API in a backward incompatible manner
such as this:

```
apply(args, kwargs, ...) -> apply(partial_args, args, kwargs)
```

New arguments must always be added as keyword arguments, and be added at the end of
the signature.

thedrow:
@ask I think it's safe to break API compatibility for Celery 4.x since it's a new major version.

tayfun:
@thedrow , the bug report is for 3.1.24. I thought 3.1.23 was the last 3.x version. Would https://github.com/celery/celery/pull/3269#r68267544 have fixed this @ask ?

PR #3254, opened 2016-06-15
Test PyPy with 5.3

thedrow:
Do we want to test other versions as well or just the most recent one?


ask:
I guess we could test the two latest versions?  I don't know how fast people upgrade PyPy, but I'm sure some people will be careful

PR #3259, opened 2016-06-17
task-cookbook: Note on cache backends

mgalgs:
The `cache.add` method is not actually atomic in all Django cache backends.
Some users get confused when the locking scheme proposed here
doesn't actually work since their backend doesn't actually support an atomic
`.add` operation.  Make a note to help avoid confusion.

References:
http://stackoverflow.com/questions/35034885/django-celery-cache-lock-did-not-work
http://stackoverflow.com/questions/37849647/using-django-cache-add-for-locking-in-a-celery-task-is-not-working


ask:

PR #3269, opened 2016-06-23
Complex graph fixes

tayfun:
I've added a few fixes for the complex canvases that might be in use. These are the cases when we clone chains (the first task in the chain has to be cloned if an arg is added, so as to not change the original task), and when partial args is used.

I have dug deep into code but I don't have too much experience with celery at the moment, have been back using it for the last few months now (I did have another stint a few years back), so I think it would be very helpful to check I'm not breaking some other stuff.

I have added a few tests as well that would not be passing in current 3.1

Let me know how this looks. I also know that core devs are focusing on the version 4, so let me know if this is too much of a work to merge in.

### DIFF
--- a/celery/canvas.py
+++ b/celery/canvas.py
@@ -365,7 +365,11 @@ class chain(Signature):
     def from_dict(self, d, app=None):
         tasks = [maybe_signature(t, app=app) for t in d['kwargs']['tasks']]
         if d['args'] and tasks:
-            # partial args passed on to first task in chain (Issue #1057).
+            # Partial args passed on to first task in chain (Issue #1057).
+            # We need to clone first task because we're changing it by adding
+            # args. The task in the chain should change, but not the original
+            # task.
+            tasks[0] = tasks[0].clone()
             tasks[0]['args'] = tasks[0]._merge(d['args'])[0]
         return _upgrade(d, chain(*tasks, app=app, **d['options']))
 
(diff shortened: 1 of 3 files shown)


celery/canvas.py
```diff
@@ -629,7 +633,8 @@ def delay(self, *partial_args, **partial_kwargs):
         return self.apply_async(partial_args)
 
     def apply_async(self, partial_args=(), args=(), kwargs={}, task_id=None,
-                    **options):
+                    producer=None, publisher=None, connection=None,
+                    router=None, result_cls=None, **options):
```
tayfun: This fixes one regression added in https://github.com/celery/celery/pull/3244/

I thought these parameters were not being used and removed them, when in fact they were being removed from `options` which is later handed over.

ask:
Could you please apply this to master?  Canvas has been refactored there and as you say we are focusing efforts on getting 4.0 out the door.  I'm not sure but maybe some issues that this is fixing will already have been fixed in master also.

tayfun:
Hey @ask I've checked the 4.0 version a bit and understand that there was a lot of refactoring done. I'm not too familiar with that version though, but I'll see if I can find bugs there and if I can fix them.

Thanks for input.

auvipy:
@tayfun closing as per conversation. looking forward to your contribution to the master branch.

PR #3271, opened 2016-06-23
Add -c parameter to make /bin/sh process the command

tomachalek:
The original file has been causing the following error:

```
... /bin/sh: 0: Can't open /usr/local/bin/celery ...
```

The *-c* argument (as documented in sh manual pages):
```
Read commands from the command string operand instead of from the standard input...
```


ask:
Thanks! :)

PR #3275, opened 2016-06-24
Add signal: `heartbeat`

kfr2:
* :data:`celery.signals.heartbeat`

     Dispatched when Celery sends a worker heartbeat.
     Sender is the :class:`celery.worker.heartbeat.Heart` instance.

---

I am interested in having a Celery worker regularly perform an action. In this specific case, I want the worker to update its corresponding health check in Consul. Thus, this change adds the `heartbeat` signal to allow a user to tap into Celery's heartbeat process to perform an action on a regular interval.

For instance, this signal might be used in the following manner:

```
import consul
from celery.signals import heartbeat


consul_client = consul.Consul(host='consul')


@heartbeat.connect
def notify_consul(sender=None, conf=None, **kwargs):
    consul_client.agent.check.ttl_pass('service:xyz')
```

I'm not aware of any existing mechanisms Celery might have to implement such a feature (nor do I know the overhead of adding a signal that may be called many times per minute) so please let me know if this feature does not make sense or could be implemented in a better way.

### DIFF
--- a/celery/signals.py
+++ b/celery/signals.py
@@ -22,9 +22,9 @@ __all__ = ['before_task_publish', 'after_task_publish',
            'celeryd_after_setup', 'worker_init', 'worker_process_init',
            'worker_ready', 'worker_shutdown', 'setup_logging',
            'after_setup_logger', 'after_setup_task_logger',
-           'beat_init', 'beat_embedded_init', 'eventlet_pool_started',
-           'eventlet_pool_preshutdown', 'eventlet_pool_postshutdown',
-           'eventlet_pool_apply']
+           'beat_init', 'beat_embedded_init', 'heartbeat',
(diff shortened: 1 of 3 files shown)


ask:
Thanks! Merged into master

I changed:

  Renamed signal to `heartbeat_sent` to be more consistent with the other signals
  Moved the documentation out of the celerybeat section into the worker section.
  Added a simle test

I also optimized it so that the signal will be a noop until someone actually connects to it.
I think doing this every 2 seconds will not do much for performance, but it would be important
to note that the signal handler cannot block the process for very long.

My changes can be found in: 67c1caed7f1e1f078e151d3eb060681e044489fc

Thanks for contributing this!
Please add your name to https://github.com/celery/celery/edit/master/CONTRIBUTORS.txt ASAP

PR #3289, opened 2016-07-04
contributor

astewart-twist:


malinoff:
Sorry, I can't merge this because there are two separate changes - please, create two separate PRs for CONTRIBUTORS change and for couchdb backend change.

astewart-twist:
oops sorry about that.  I'll try reverting the last commit and doing a PR on that, then follow up with the couchdb backend changes.  Let me know if that's the wrong way to do it.

PR #3293, opened 2016-07-05
Deserialize json string prior to inclusion in CouchDB doc

astewart-twist:


celery/backends/couchdb.py
```diff
@@ -91,13 +93,16 @@ def get(self, key):
             return None
 
     def set(self, key, value):
-        data = {'_id': key, 'value': value}
+        # data = {'_id': key, 'value': value}
+        data = json.loads(value)
```
ask: Why is this necessary? I feel if the couchdb needs to inject anything into the generated structure it should do so before the structure is serialized, e.g. by overriding `encode_result()` or `encode()`.

astewart-twist: I'm not familiar with the anatomy of a results backend class, so I'm not at all surprised if there's a better place to put that code.

The result of this code (wherever it goes) should be necessary, however, unless the intention is that the value of 'value' is supposed to be serialized within the serialized json string.  Like this:

```
{
  "_id": "celery-task-meta-2c040203-a07a-4a92-9fa1-3269839e2c54",
  "_rev": "1-6b9803d5b1a4ed51a5274ef7c567a332",
  "value": "{'status': 'SUCCESS', 'traceback': null, 'result': {}, 'task_id': '2c040203-a07a-4a92-9fa1-3269839e2c54', 'children': []}"
}
```

If that's not the intention, my proposed change would instead result in the CouchDB document looking like this:

```
{
   "_id": "celery-task-meta-2c040203-a07a-4a92-9fa1-3269839e2c54",
   "_rev": "1-6b9803d5b1a4ed51a5274ef7c567a332",
   "status": "SUCCESS",
   "result": { },
   "task_id": "2c040203-a07a-4a92-9fa1-3269839e2c54",
   "traceback": null,
   "children": [
   ]
}
```

thedrow:
I apologize we haven't got to this contribution yet.
So if I understand correctly when we retrieve the result from couchdb we don't get a dictionary but a string?
Does it break anything?

PR #3294, opened 2016-07-05
Wrong version number in readme.

KptMarchewa:
Pretty self-explanatory. Consistent with docs at http://docs.celeryproject.org/en/master/whatsnew-4.0.html

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -46,7 +46,7 @@ by `using webhooks`_.
 What do I need?
 ===============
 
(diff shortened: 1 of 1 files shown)


ask:
Thanks!

The README.rst file is actually generated from `docs/includes/introduction.txt` so I changed it in the master document also (you render it with `pip install -r requirements/pkgutils.txt; make readme`)

PR #3299, opened 2016-07-06
Fix chain links with always eager enabled

stevenh:
Ensure that link and link_error are called for chains when CELERY_ALWAYS_EAGER is True.


stevenh:
I'm guessing the appveyor error is not a problem with the PR but something to do with it running a test that's only configured for HEAD?
```
Build started
git clone -q https://github.com/celery/celery.git C:\projects\celery
git fetch -q origin +refs/pull/3299/merge:
git checkout -qf FETCH_HEAD
Specify a project or solution file. The directory does not contain a project or solution file.
```

ask:
Thanks!  Could you please apply this patch to the master branch instead?

auvipy:
@stevenh please rebase your work against master branch and send a new PR

stevenh:
Then I looked master was majorly different, I'll see what I can do.

PR #3333, opened 2016-07-20
Remove Amazon SQS "Experimental Status" disclaimer

amirrustam:
Since Amazon SQS broker is now identified with a "Stable" status on the broker docs page, then is the "Experimental Status" disclaimer still necessary? If not, then the disclaimer should be removed.


ask:
Thanks!

PR #3357, opened 2016-08-04
Fix to allow use of heartbeats with publishing connections

mazimi:


auvipy:
please re work on a new branch against master and send your PR again.

PR #3365, opened 2016-08-07
Fix reST markup in whatsnew-4.0.rst

berkerpeksag:


auvipy:
thanks, could you fix the travis failure?

berkerpeksag:
Unless I'm missing something, test failures are not related to this PR and they should be handled in separate pull requests. I've opened https://github.com/celery/celery/pull/3367 to fix the failed test in Appveyor.

berkerpeksag:
The new version of pyenchant library does not provide a pre-built binary for Linux: https://pypi.python.org/pypi/pyenchant However, it's a dependency of sphinxcontrib-spelling (which is a dependency of sphinx_celery.)

Issue #3356, opened 2016-08-04
Chaining on a chord gives the wrong result to the chain


tbodt:
A code block is worth a thousand words.

```python
@app.task
def one():
    return 1

@app.task
def xsum(stuff):
    print 'xsum got {}'.format(stuff)
    return sum(stuff)

@app.task
def callback(*stuff):
    print 'callback got {}'.format(stuff)
    return stuff
```

```python
>>> ~(group(one.s(), one.s()) | xsum.s())
xsum got ([1, 1],)
2
# expected result
>>> ~(group(one.s(), one.s()) | xsum.s() | callback.s())
callback got (1,)
callback got (1,)
xsum got [1, 1]
(1,)
# whaaat?
# should have been
>>> ~(group(one.s(), one.s()) | xsum.s() | callback.s())
xsum got [1, 1]
callback got (2,)
```

This is not an issue in Celery 3.x, but it is in 4.0.0rc3.

ask:
Seems this version works just fine:

```
>>> g = (group(one.s(), one.s()) | (xsum.s() | callback.s()))
>>> g().get()
[2]
```

So possibly something to do with how the ``|`` operator upgrades the group chain into a chord.

PR #3421, opened 2016-08-31
don't remove the task following a chord from a chain

tbodt:
The if statement isn't there in master, and it was causing bugs for me in complex workflows. I can't just start using Celery 4 (even though I'd like to) because of bug #3356, so I'd like this to be in 3.1.24.

### DIFF
--- a/celery/app/builtins.py
+++ b/celery/app/builtins.py
@@ -262,9 +262,8 @@ def add_chain_task(app):
                     if not res.parent:
                         res.parent = prev_res
 
-                if not isinstance(prev_task, chord):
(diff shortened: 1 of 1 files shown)


tbodt:
The build failures don't have anything to do with the patch, as far as I can tell...

tbodt:
Dammit. I didn't want my latest push to get added to the PR.

The new commit is still useful, though, for people like me who actually use headers...

thedrow:
@ask I'm not sure what we should do with this one? Are we still maintaining 3.x?

PR #3432, opened 2016-09-06
Fix doc for relative flag of periodic task

PierreF:
[doc](https://github.com/celery/celery/blob/2267553bb1011092adb86ed03a854a9bde667ef6/docs/userguide/periodic-tasks.rst#available-fields) say that "relative" field by default round the timedelta value, and that with relative set to True, rounding is disabled.

But in fact, when relative is True, [celery.schedule.schedule](https://github.com/celery/celery/blob/2267553bb1011092adb86ed03a854a9bde667ef6/celery/schedules.py#L71) will do the rounding, which is the opposite of what doc said.

And indeed, by default a schedule with timedelta(seconds=60) is not rounded to nearest minute, but run every minute from the start of celery beat. Setting relative to True do the rounding and task is run exactly when minute change.

I go with the update of documentation, instead of code to keep old behavior.

### DIFF
--- a/docs/userguide/periodic-tasks.rst
+++ b/docs/userguide/periodic-tasks.rst
@@ -179,12 +179,12 @@ Available Fields
 
 * `relative`
 
-    By default :class:`~datetime.timedelta` schedules are scheduled
+    If `relative` is true :class:`~datetime.timedelta` schedules are scheduled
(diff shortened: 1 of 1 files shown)


ask:
Thanks! This is much better

PR #3450, opened 2016-09-18
Document that -B should be used for development only

berkerpeksag:


auvipy:
looks ok but travis is failing

PR #3463, opened 2016-09-23
Update __init__.py

andytwoods:
Without that explicit location, unit tests fail in Django. Please see http://stackoverflow.com/questions/39588380/django-celery-unit-tests-with-pycharm-no-module-named-celery/39655850#39655850


auvipy:
build is failing

ask:
There's nothing called taskapp in the Django example.

I'm guessing this is for a project that uses a separate Django app for the celery app...

PR #3477, opened 2016-09-27
possibility to set on_message callback for single task

xBeAsTx:
Functionality was decribed and tests was added in https://github.com/celery/celery/pull/2550

Here is just added possibility to set on_message callback without creating group of tasks.
```python
def on_raw_message(body):
	print(body)

s = signature('test_task.hello', queue='test', kwargs={'a': 2, 'b': 8})
r = s.apply_async()
print(r.get(on_message=on_raw_message))
```

### DIFF
--- a/celery/backends/base.py
+++ b/celery/backends/base.py
@@ -428,9 +428,13 @@ class SyncBackendMixin(object):
         )
 
     def wait_for_pending(self, result, timeout=None, interval=0.5,
-                         no_ack=True, on_interval=None, callback=None,
-                         propagate=True):
+                         no_ack=True, on_message=None, on_interval=None,
+                         callback=None, propagate=True):
         self._ensure_not_eager()
+        if on_message is not None:
+            raise ImproperlyConfigured(
+                'Backend does not support on_message callback')
+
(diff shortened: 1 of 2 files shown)


celery/backends/base.py
```diff
@@ -428,9 +430,13 @@ def iter_native(self, result, timeout=None, interval=0.5, no_ack=True,
         )
 
     def wait_for_pending(self, result, timeout=None, interval=0.5,
-                         no_ack=True, on_interval=None, callback=None,
-                         propagate=True):
+                         no_ack=True, on_message=None, on_interval=None,
+                         callback=None, propagate=True):
         self._ensure_not_eager()
+        if on_message is not None:
```
ssudake21: Wonder if this condition can cause following issue? Mq backed is not supported for on_message ? @xBeAsTx @auvipy 
http://stackoverflow.com/questions/40784593/celery-raise-improperlyconfigured-exception

xBeAsTx:
@ssudake21, yes, something wrong with this feature. It must work with Redis and RabbitMQ backends.

PR #3464, opened 2016-09-23
Bytes/unicode treated differently by Shelve bysed on Python version

fridex:
relates: #3346

Fixes:
```
Traceback (most recent call last):
  File "/usr/lib/python3.5/site-packages/kombu/utils/objects.py", line 40, in __get__
    return obj.__dict__[self.__name__]
KeyError: 'scheduler'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/lib64/python3.5/shelve.py", line 111, in __getitem__
    value = self.cache[key]
KeyError: b'entries'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/lib/python3.5/site-packages/celery/apps/beat.py", line 107, in start_scheduler
    service.start()
  File "/usr/lib/python3.5/site-packages/celery/beat.py", line 522, in start
    humanize_seconds(self.scheduler.max_interval))
  File "/usr/lib/python3.5/site-packages/kombu/utils/objects.py", line 42, in __get__
    value = obj.__dict__[self.__name__] = self.__get(obj)
  File "/usr/lib/python3.5/site-packages/celery/beat.py", line 566, in scheduler
    return self.get_scheduler()
  File "/usr/lib/python3.5/site-packages/celery/beat.py", line 561, in get_scheduler
    lazy=lazy,
  File "/usr/lib/python3.5/site-packages/celery/beat.py", line 406, in __init__
    Scheduler.__init__(self, *args, **kwargs)
  File "/usr/lib/python3.5/site-packages/celery/beat.py", line 205, in __init__
    self.setup_schedule()
  File "/usr/lib/python3.5/site-packages/celery/beat.py", line 436, in setup_schedule
    self._store[b'entries']
  File "/usr/lib64/python3.5/shelve.py", line 113, in __getitem__
    f = BytesIO(self.dict[key.encode(self.keyencoding)])
AttributeError: 'bytes' object has no attribute 'encode'

```

Additional info:
As stated in Python3 docs [1]: "The keys are ordinary strings."
```
software -> celery:4.0.0rc4 (0today8) kombu:4.0.0rc4 py:3.5.1
            billiard:3.5.0.1 py-amqp:2.1.0
platform -> system:Linux arch:64bit imp:CPython
loader   -> celery.loaders.default.Loader
```

[1] https://docs.python.org/3.4/library/shelve.html

Issue #3346, opened 2016-07-29
Beat error with redis


jamesob:
Hi, 

I'm hitting stacktraces when running my celery worker.

I have Celery configured like so:
```
In [1]: import celery

In [2]: celery.__version__
Out[2]: u'4.0.0rc3'

In [3]: from limsana import celery

In [4]: celery.app.__dict__
Out[4]:
{'_conf': Settings({'task_serializer': 'json', 'result_serializer': 'json', 'task_always_eager': False, 'accept_content': ['json'], 'timezone': 'US/Pacific', u'broker_url': u'redis://xx.usw2.cache.amazonaws.com/0', u'result_backend': u'redis://xx.usw2.cache.amazonaws.com/0'}),
 '_config_source': None,
 '_finalize_mutex': <thread.lock at 0x7f02128df650>,
 '_fixups': [None],
 '_pending': deque([]),
 '_pending_defaults': deque([]),
 '_pending_periodic_tasks': deque([]),
 '_preconf': {'accept_content': ['json'],
  u'broker_url': u'redis://xx.usw2.cache.amazonaws.com/0',
  u'result_backend': u'redis://xx.usw2.cache.amazonaws.com/0',
  'result_serializer': 'json',
  'task_always_eager': False,
  'task_serializer': 'json',
  'timezone': 'US/Pacific'},
 '_preconf_set_by_auto': {u'broker_url', u'result_backend'},
 '_tasks': {},
 '_using_v1_reduce': None,
 'amqp_cls': u'celery.app.amqp:AMQP',
 'autofinalize': True,
 'clock': <LamportClock: 0>,
 'configured': False,
 'control_cls': u'celery.app.control:Control',
 'events_cls': u'celery.events:Events',
 'finalized': False,
 'fixups': {u'celery.fixups.django:fixup'},
 'loader_cls': u'celery.loaders.app:AppLoader',
 'log_cls': u'celery.app.log:Logging',
 'main': 'celery',
 'namespace': None,
 'on_after_configure': <Signal: Signal>,
 'on_after_finalize': <Signal: Signal>,
 'on_after_fork': <Signal: Signal>,
 'on_configure': <Signal: Signal>,
 'registry_cls': celery.app.registry.TaskRegistry,
 'set_as_current': True,
 'steps': defaultdict(set, {}),
 'task_cls': u'celery.app.task:Task',
 'user_options': defaultdict(set, {})}

In [7]: celery.app.conf.beat_schedule
Out[7]:
[...]

PR #3481, opened 2016-09-29
Fix KeyError when starting celery beat

alzeih:
The shelve module does not accept byte literals as keys on Py3, and
does not accept unicode literals as keys on Py2, so instead the
"informal" string is used for both.

There is a description about informal strings here:
https://docs.python.org/3/library/stdtypes.html#str


auvipy:
looks good. could you plz add changelog too?

alzeih:
I've rebased with name added to contributors.

I've not added to the changelog as the error doesn't exist in celery 3.1.23.

fridex:
This does not look correct to me as it will reintroduce https://github.com/celery/celery/issues/3346 I have already submitted patch for this https://github.com/celery/celery/pull/3464

PR #3525, opened 2016-10-27
Use the existing connection in the consumer Events step

JelteF:
## Description

This makes sure only a single connection attempt is performed by a worker. The previous behaviour caused a worker to need twice the BROKER_TIMEOUT before failing over to a different broker. This is also broken in celery 3.1 (in which I had the problem)

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -38,6 +38,7 @@ install:
   - "powershell extra\\appveyor\\install.ps1"
   - "%PYTHON%/Scripts/pip.exe install -U setuptools"
   - "%PYTHON%/Scripts/pip.exe install -r requirements/dev.txt"
+  - "%PYTHON%/Scripts/pip.exe install -r requirements/extras/eventlet.txt"
 
 build: off
 
--- a/celery/worker/consumer/events.py
+++ b/celery/worker/consumer/events.py
@@ -33,7 +33,7 @@ class Events(bootsteps.StartStopStep):
         # flush events sent while connection was down.
         prev = self._close(c)
         dis = c.event_dispatcher = c.app.events.Dispatcher(
-            c.connect(), hostname=c.hostname,
+            c.connection, hostname=c.hostname,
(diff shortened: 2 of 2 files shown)


JelteF:
For some reason the windows builds are failing, but I doubt it's because of my changes. I even disabled the changes for windows and it is still failing. I don't have windows to test on locally.

JelteF:
I fixed the issues with appveyor

ask:
Hey!

It's actually supposed to use a different connection for sending events, and at the very least it should respect the `app.connection_for_read()`, `app.connection_for_write()` distinction.

AMQP has the concept of flow-control (https://www.rabbitmq.com/flow-control.html), which means
that the broker may block producing connections under low disk-space or low memory conditions.

If the same connection is used for both consuming and producing, both will be blocked without the ability to recover from the situation.

If the connections are separate then the consumer may continue consuming, while the producer is blocked - this way RabbitMQ will heal as the consumer handles messages, freeing up disk space/memory.

ask:
The behavior is wrong now, instead of `c.connect()` it should use `app.connection_for_write()`.

It could definitely bypass `celery.worker.consumer.consumer.Consumer.connect`, but I'm not sure how we can get away using only one connection.

The worker may also use threads (Windows, transports other than amqp/redis/sqs), or gevent/eventlet  in which case
the connection cannot be reused for publishing at all.

JelteF:
Sorry for not responding earlier. I now understand the reasoning behind it and understand that having one connection is not possible. 

However, I'm quite sure there's now a race condition possible (I haven't tested it yet). If the first connection will fail over and then the connection restores the second connection might connect to a different broker, probably really screwing stuff up.

thedrow:
Can we close this PR? Have we reached to conclusion that these changes are not required?

thedrow:
I'm going to close this PR for now. If @ask or @JelteF has any suggestions on how to handle this correctly feel free to issue a new PR or rebase this one.

Issue #3457, opened 2016-09-21
AttributeError("'bytes' object has no attribute 'encode'",)


astronouth7303:
Running Celery 4.0.0rc4 on Python 3.4.3 on AWS EC2:

```
Traceback (most recent call last):
  File "/opt/python/run/venv/local/lib/python3.4/site-packages/kombu/utils/objects.py", line 40, in __get__
    return obj.__dict__[self.__name__]
KeyError: 'scheduler'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/lib64/python3.4/shelve.py", line 111, in __getitem__
    value = self.cache[key]
KeyError: b'entries'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/opt/python/run/venv/local/lib/python3.4/site-packages/celery/apps/beat.py", line 107, in start_scheduler
    service.start()
  File "/opt/python/run/venv/local/lib/python3.4/site-packages/celery/beat.py", line 522, in start
    humanize_seconds(self.scheduler.max_interval))
  File "/opt/python/run/venv/local/lib/python3.4/site-packages/kombu/utils/objects.py", line 42, in __get__
    value = obj.__dict__[self.__name__] = self.__get(obj)
  File "/opt/python/run/venv/local/lib/python3.4/site-packages/celery/beat.py", line 566, in scheduler
    return self.get_scheduler()
  File "/opt/python/run/venv/local/lib/python3.4/site-packages/celery/beat.py", line 561, in get_scheduler
    lazy=lazy,
  File "/opt/python/run/venv/local/lib/python3.4/site-packages/celery/beat.py", line 406, in __init__
    Scheduler.__init__(self, *args, **kwargs)
  File "/opt/python/run/venv/local/lib/python3.4/site-packages/celery/beat.py", line 205, in __init__
    self.setup_schedule()
  File "/opt/python/run/venv/local/lib/python3.4/site-packages/celery/beat.py", line 436, in setup_schedule
    self._store[b'entries']
  File "/usr/lib64/python3.4/shelve.py", line 113, in __getitem__
    f = BytesIO(self.dict[key.encode(self.keyencoding)])
AttributeError: 'bytes' object has no attribute 'encode'
```

astronouth7303:
My applicable settings:

```
CELERY_RESULT_BACKEND = 'django-cache'
CELERY_BROKER_URL = 'sqs://'
[...]

PR #3542, opened 2016-11-04
use simpler version to make a str type in python2 and python3

alzeih:
Addition to #3481 

https://github.com/celery/celery/issues/3457#issuecomment-258330774

### DIFF
--- a/celery/beat.py
+++ b/celery/beat.py
@@ -435,24 +435,24 @@ class PersistentScheduler(Scheduler):
         self._create_schedule()
 
         tz = self.app.conf.timezone
-        stored_tz = self._store.get(str(b'tz'))
+        stored_tz = self._store.get(str('tz'))
         if stored_tz is not None and stored_tz != tz:
             warning('Reset: Timezone changed from %r to %r', stored_tz, tz)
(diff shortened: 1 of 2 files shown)


ask:
Why don't we just use `b'x'`?

alzeih:
python3 shelve behaves differently to python2

`python3 -c "import shelve; s = shelve.open('file'); s[b'bytes']"`

```
  File "/usr/lib/python3.5/shelve.py", line 111, in __getitem__
    value = self.cache[key]
KeyError: b'bytes'

  File "<string>", line 1, in <module>
  File "/usr/lib/python3.5/shelve.py", line 113, in __getitem__
    f = BytesIO(self.dict[key.encode(self.keyencoding)])
AttributeError: 'bytes' object has no attribute 'encode'
```

Python3 shelve implementation will try and .encode() all keys, and bytes doesn't have an .encode() method.  https://hg.python.org/cpython/file/3.5/Lib/shelve.py#l125

This patch doesn't change the result type, as `str(b'')` and `str('')` are both type str with the keys used here, i.e. `'scheduler'`. It just removes the unnecessary `b`.

When python2 support is dropped, the explicit str('') can be removed.

PR #3565, opened 2016-11-08
Documentation: typo and clarification in beat scheduler setting

browniebroke:
Correct a typo PersistentScheduer -> PersistentScheduler 
Add an example of other possible values for `beat_scheduler` setting.


browniebroke:
Opened #3566 as I realised this was against the master branch

PR #3567, opened 2016-11-08
Doc: update installation for django_celery_beat

browniebroke:
With the current instructions, the application’s verbose name isn’t found and the label in the admin is DJANGO_CELERY_BEAT.

Updating installation instructions to fix that problem and follow best practices for [application users](https://docs.djangoproject.com/en/1.10/ref/applications/#for-application-users).

Feel free to discard this if you'd rather go down [the other route](https://docs.djangoproject.com/en/1.10/ref/applications/#for-application-authors) and define `default_app_config` in the django_celery_beat app.

### DIFF
--- a/docs/userguide/periodic-tasks.rst
+++ b/docs/userguide/periodic-tasks.rst
@@ -431,12 +431,12 @@ To install and use this extension:
 
         $ pip install django-celery-beat
 
-#. Add the ``django_celery_beat`` module to ``INSTALLED_APPS`` in your
+#. Add ``django_celery_beat.apps.BeatConfig`` to ``INSTALLED_APPS`` in your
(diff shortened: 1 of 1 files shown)


ask:
I think we should probably do whatever makes it easier to the user :)

I added the default app config to django_celery_results and django_celery_beat, so then we may not need this patch?

Just need to make a new version update for these packages.

browniebroke:
Ok sounds good to me.

vstoykov:
@ask I see that these packages are not touched for the last  two months. Is new version will be available for them or there is something pending?

I understand that this is not with high priority and probably there are more important tasks. But I just wanted to know if it is not forgotten.

Regards!

PR #3568, opened 2016-11-08
Doc: ReST format in installation

tnir:
Fix ReST format breaking in [Installation](http://docs.celeryproject.org/en/latest/getting-started/introduction.html#installation) docs, which appeared in 7db0791, aee706a.

Signed-off-by: Takuya Noguchi <tak.noguchi.iridge@gmail.com>


ask:
Thanks!

PR #3573, opened 2016-11-10
Chord counter to use result_expires and is touched

tayfun:
## Description

Chord counter keeps track of chord tasks that have finished. This needs
to be compatible with result_expires so that if a chord task result is
still in cache, chord will be able finish and join (unlock). Otherwise
we see that the chord counter has expired while the chord task result
still in cache. Incrementing an expired chord counter returns None and
comparing that to dependent task number throws an error.

Counter's timeout also needs to be refreshed on every chord part return.
Memcached backend didn't implement `expire` using client's `touch`, so
this is implemented here as well.

### DIFF
--- a/celery/backends/base.py
+++ b/celery/backends/base.py
@@ -750,7 +750,7 @@ class BaseKeyValueStoreBackend(Backend):
                 deps.delete()
                 self.client.delete(key)
         else:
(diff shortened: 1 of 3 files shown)


ask:
Thanks Tayfun!  Merged into master

PR #3587, opened 2016-11-13
typo in 4.0 release notes, .on_commit is the old Django API

Psycojoker:
## Description

Hello,

Fix a typo that reference to the old Django transaction API wihle the example reference the new one (`transaction.atomic`)

Kind regards,


pelme:
I think this PR should be reverted.

`transaction.on_commit` is the new API as of Django 1.9 that adds the ability to run code after transaction commit.

The example in question properly shows `transaction.on_commit(partial(...))`.

Issue #3586, opened 2016-11-12
Celery does not respect exceptions types when using a serializer different than pickle.


jcsaaddupuy:
## Checklist
```
~ : celery -A analystick report
software -> celery:4.0.0 (latentcall) kombu:4.0.0 py:3.5.2
            billiard:3.5.0.2 redis:2.10.5
platform -> system:Darwin arch:64bit imp:CPython
loader   -> celery.loaders.app.AppLoader
settings -> transport:redis results:redis://127.0.0.1:6379/1
```

## Steps to reproduce
(See example code below)

## Expected behavior

__When using a result serializer different than pickle__, exceptions types should be the same as the raised exception.

## Actual behavior
Celery does not respect the exception type but create _a new type_ instead.

The main problem is that instead of using the actual type of the exception, celery will [reconstruct a type](https://github.com/celery/celery/blob/0f87321df385c5f3dca717ec2a4a9c0d25f88054/celery/utils/serialization.py#L43-L45) on the fly, but without respecting the original exception module.

For example, using the `yaml` result serializer (I believe it will be the same for `json`): 
- if a task raises a `ValueError`, the caller will receive a `celery.backends.base.ValueError`
- if a task raises a `custom.module.CustomError`, the caller will receive a `celery.backends.base.CustomError`

This ends with wrongs behaviour when raising a exception from a task and trying to catch it from the caller.

### Minimal reproductible test
As an example, I've setup a minimal reproductible test, using a redis backend :

celery config (I can provide a full config if needed):
```python
CELERY_TASK_SERIALIZER = 'yaml'
CELERY_RESULT_SERIALIZER='yaml'
```

Tasks : 

```python
# module myapp.tasks
from myapp import celery_app
@celery_app.task
def raises_valueerror():
    raise ValueError('Builtin exception')

class CustomError(Exception):
    pass

@celery_app.task
def raises_customerror():
    raise CustomError('Custom exception', {'a':1})
```

Unittest : 
```python
from myapp import tasks
from myapp.tasks import CustomError

def test_builtin_exception():
    t = tasks.raises_valueerror.s()
    r = t.apply_async()

    exc = None
    try:
        r.get(propagate=True)
    except Exception as e:
        exc = e

    # with celery bug, the actual class of exc will be  `celery.backends.base.ValueError` instead of builtin ValueError
    assert isinstance(exc, ValueError), "Actual class %s}" % (exc.__class__)

def test_custom_exception():
    t = tasks.raises_customerror.s()
[...]

PR #3592, opened 2016-11-14
fixes exception deserialization when not using pickle

jcsaaddupuy:
fix for #3586 

This PR try to fix how exceptions are deserialized when using a serializer different than pickle.

This avoid to [create new types](https://github.com/jcsaaddupuy/celery/blob/8d4e613e24f6561fdaafd4e6ede582ceac882804/celery/utils/serialization.py#L46) for exceptions, by doing 2 things : 

- store the exception module in [celery/celery/backends/base.py::Backend::prepare_exception](https://github.com/jcsaaddupuy/celery/blob/8d4e613e24f6561fdaafd4e6ede582ceac882804/celery/backends/base.py#L243-L245) and reuse it in [celery/celery/backends/base.py::Backend::prepare_exception](https://github.com/jcsaaddupuy/celery/blob/8d4e613e24f6561fdaafd4e6ede582ceac882804/celery/backends/base.py#L249-L258) instead of using raw `__name__` as module name

- in [ celery/celery/utils/serialization.py::create_exception_cls](https://github.com/jcsaaddupuy/celery/blob/8d4e613e24f6561fdaafd4e6ede582ceac882804/celery/utils/serialization.py#L86-L98), try to find the exception class either from  `__builtins__`, or from the excetion module. Fallback on current behavior (which may still be wrong)

Also, it uses `exc.args` in [celery/celery/backends/base.py::Backend::prepare_exception](https://github.com/jcsaaddupuy/celery/blob/8d4e613e24f6561fdaafd4e6ede582ceac882804/celery/backends/base.py#L247) and pass it to the exception constructor in [celery/celery/backends/base.py::Backend::exception_to_python](https://github.com/jcsaaddupuy/celery/blob/8d4e613e24f6561fdaafd4e6ede582ceac882804/celery/backends/base.py#L255), instead of using `str(exc.message)` which could lead to unwanted behavior

This won't work in every cases. If a class is defined locally in a function, this code won't be able to import the exception class using `import_module` and the old (wrong) behavior will still be used.

### DIFF
--- a/celery/utils/serialization.py
+++ b/celery/utils/serialization.py
@@ -81,9 +81,9 @@ def itermro(cls, stop):
 
 def create_exception_cls(name, module, parent=None):
     """Dynamically create an exception class."""
-
     # handle builtin exceptions
-    if name in __builtins__ and isinstance(__builtins__[name], type(BaseException)):
+    if name in __builtins__ \
+            and isinstance(__builtins__[name], type(BaseException)):
         return __builtins__[name]
     # exception is not builtin, try to find it from its module
     exc = None
--- a/celery/utils/serialization.py
+++ b/celery/utils/serialization.py
@@ -81,21 +81,17 @@ def itermro(cls, stop):
 
 def create_exception_cls(name, module, parent=None):
     """Dynamically create an exception class."""
-    # handle builtin exceptions
-    if name in __builtins__ \
-            and isinstance(__builtins__[name], type(BaseException)):
-        return __builtins__[name]
-    # exception is not builtin, try to find it from its module
-    exc = None
     try:
+        # also works for 'builtins' module
         mod = import_module(module)
+        # should we raise if getattr fails ?
+        exc_cls = getattr(mod, name, None)
+        if exc_cls and isinstance(exc_cls, type(BaseException)):
+            return exc_cls
     except ImportError:
+        # module not found. We may want to raise instead of
+        # calling subclass_exception
         pass
-    else:
-        # should we raise ?
-        exc = getattr(mod, name, None)
-    if exc:
-        return exc
 
     # we could not find the exception, fallback and create a type.
     if not parent:
--- a/celery/utils/serialization.py
+++ b/celery/utils/serialization.py
@@ -10,6 +10,7 @@ from base64 import b64encode as base64encode, b64decode as base64decode
 from functools import partial
 from inspect import getmro
 from itertools import takewhile
+from importlib import import_module
 
 from kombu.utils.encoding import bytes_to_str, str_to_bytes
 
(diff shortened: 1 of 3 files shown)


jcsaaddupuy:
note : tests are failing in current statebecause of [unrelated pypy issue](https://travis-ci.org/celery/celery/jobs/175707878) (toxenv `pypy3-unit`)  , affecting other [PR builds](https://travis-ci.org/celery/celery/builds/175410691).

ask:
Thanks!  One thing that worries me about this is that it may not be safe to do, considering that it's basically the same thing pickle and the unsafe pyyaml methods does?  I know we are limiting it to exceptions by using the isinstance check, but there could be side effects from this that we are not considering.

jcsaaddupuy:
Hi, thanks for the reply!

What bother me here is that raising an exception in a task and catching it in the caller code only works with `pickle`.
Due to the call to `type()`, it does not works as expected with `yaml` or `json` when not using eager mode.

When not using `pickle`, an exception will indeed be propagated, but it won't be of the same type as the raised exception. 
This is not consistent and really misleading (for ex, a builtin `ValueError` will be propagated as a `celery.backends.base.ValueError`, see https://github.com/celery/celery/issues/3586)
As the type is different,  a `except ValueError` in the caller code will just not work.

As a developper and a celery user, I would prefer either :
- to catch a `ValueError` (or any other custom exception) when a `ValueError` (or any other custom exception)  is raised, using any serializer :)
- having `celery` raising a `not supported error` instead of creating a shadow type under the hood thats _looks like_ my exception type 

If that's not possible, the doc should explicitly state that support for propagated exceptions is not supported when using any other serializer than `pickle`.
This also raise a concern as `celery` is using  `json` serializer by default since `4.0.0`.

From my point of view, this seems pretty critical, and while I don't see them right now, I understand that there may be safety issues.

Anyway, thanks for the good work, please let me know if you see a better way to adress this issue !

celery/utils/serialization.py
```diff
@@ -80,6 +81,19 @@ def itermro(cls, stop):
 
 def create_exception_cls(name, module, parent=None):
     """Dynamically create an exception class."""
+    try:
+        # also works for 'builtins' module
+        mod = import_module(module)
```
thedrow: What happens when the module has side effects? You'll get a remote code execution vulnerability if you can inject a result somehow.

thedrow:
I'd like to see the build passing before we consider to merge this.
I have to agree that we need to figure this out.
Maybe we can merge this with a feature flag that is disabled by default.
It should be documented that this may or may not be unsafe to do.

celery/utils/serialization.py
johnarnold: If the module is being used in a task, it's already inside a trust boundary.  So some trust is assumed.

My head goes towards sandboxing the Exception inspection.  Like, run it in a locked down subprocess, just to get the signature.  I need to look at the code closer though.

johnarnold:
@thedrow  The build break doesn't have a build log -- looks like travisci issue, not code issue. Can someone kick it off again?

georgepsarakis:
@johnarnold I restarted the Travis CI job.

lpsinger:
This PR might not be accurately named, because you also get serialization errors if you *are* using pickle.

FWIW, I can reproduce this in Python 2, but not on Python 3.

Any chance of getting this rebased and fixed soon?

auvipy:
it would be great if you can take this and rebase

johnarnold:
@jcsaaddupuy  Do you want to rebase this?

cscetbon:
@lpsinger do you know why you can't reproduce it with Python 3 ? Has the module been rewritten and has a fix in Python 3 ?

auvipy:
cleaner version https://github.com/celery/celery/pull/5074

PATAPOsha:
I still have same issue with celery 4.3.0. Why is it closed?

PR #3601, opened 2016-11-17
MySQL server has gone away Exception... service retry

raintype:

## Description

that version series.


thedrow:
This is the wrong fix for this problem.
A connection problem should be addressed in the appropriate results backend.

PR #3602, opened 2016-11-17
DOC: fix typo in remote shutdown using control broadcast example

moutai:
DOC: fix typo in remote shutdown using control broadcast example


ask:
Thanks!

PR #3626, opened 2016-11-25
Be able to use customized test Celery init parameters.

sallner:
## Description

Up to this point, only the config was configurable in the pytest fixtures (`celery_config`). Some parameters for the init of the Celery test app, like task_cls, could not be changed in that way. Therefore I propose an additional fixture to change the `__init__` parameters (`celery_parameters`).


icemac:
I'd like to have this PR merged, too.

icemac:
Thanks a lot!

PR #3664, opened 2016-12-08
Fix code examples in routing documentation

icemac:
## Description

There are code examples in the documentation which are not valid python code: duplicate argument name `kwargs` in function definition (F831 as reported by flake8).


ask:
Thanks! Merged into master and 4.0 branches

PR #3669, opened 2016-12-09
Fixed #3651 - add `app` argument to `GroupResult.restore`.

### LINKED ISSUE #3651: Specifying app/backend in GroupResult.restore is slightly awkward
In Celery 3.x I used `GroupResult.restore('some-id')` and it somehow found my app and figured out the backend (not sure how, but it worked).

In Celery 4.x this does not work (it does not find the backend and crashes). That makes sense to me (I like the idea of explicit apps a lot!). `GroupResult.restore()` does not accept a `app` argument which would be convenient. It does however support a `backend` argument, so I changed my code to use `GroupResult.restore('my-id', backend=app.backend)` and everything is fine.

However, to me this is not that super-intuitive, my suggestion is that we should add an `app` kwarg to `restore()` that can be used like:

```py
GroupResult.restore('someid', app=app)
```

Do you think this makes sense? Then I'd be happy to work on a PR to add the argument!

pelme:
This makes the restore method behave the same way as the `GroupResult` constructor.

See #3651 for discussion.

### DIFF
--- a/celery/result.py
+++ b/celery/result.py
@@ -867,10 +867,13 @@ class GroupResult(ResultSet):
         return self.results
 
     @classmethod
-    def restore(cls, id, backend=None):
+    def restore(cls, id, backend=None, app=None):
         """Restore previously saved group result."""
+
+        app = app or cls.app
+
         return (
-            backend or (cls.app.backend if cls.app else current_app.backend)
+            backend or (app.backend if app else current_app.backend)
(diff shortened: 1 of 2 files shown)


ask:
Thanks! Merged into master.  Please add your name to https://github.com/celery/celery/edit/master/CONTRIBUTORS.txt

PR #3684, opened 2016-12-14
Refer worker request info to absolute time

regisb:
Previously, the time_start attribute of worker request objects refered
to a timestamp relative to the monotonic time value. This caused
time_start attributes to be at a time far in the past. We fix this by
calling the on_accepted callback with an absolute time_accepted
attribute.

Note that the current commit does not change the
celery.concurrency.base API, although it would probably make sense to
rename the "monotonic" named argument to "time".

This fixes the problem described in http://stackoverflow.com/questions/20091505/celery-task-with-a-time-start-attribute-in-1970/

Note that there are many different ways to solve this problem; in the proposed implementation I choose to modify the default value of a keyword argument that is AFAIK never used.

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -255,3 +255,4 @@ Arpan Shah, 2017/09/12
 Tobias 'rixx' Kunze, 2017/08/20
 Mikhail Wolfson, 2017/12/11
 Alex Garel, 2018/01/04
+Régis Behmo 2018/01/20
--- a/celery/worker/request.py
+++ b/celery/worker/request.py
@@ -9,6 +9,7 @@ from __future__ import absolute_import, unicode_literals
 import logging
 import sys
 from datetime import datetime
+from time import time
 from weakref import ref
 
 from billiard.common import TERM_SIGNAME
@@ -20,7 +21,7 @@ from celery.app.trace import trace_task, trace_task_ret
 from celery.exceptions import (Ignore, InvalidTaskError, Reject, Retry,
                                TaskRevokedError, Terminated,
                                TimeLimitExceeded, WorkerLostError)
-from celery.five import python_2_unicode_compatible, string
+from celery.five import monotonic, python_2_unicode_compatible, string
 from celery.platforms import signals as _signals
 from celery.utils.functional import maybe, noop
 from celery.utils.log import get_logger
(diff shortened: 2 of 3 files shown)


ask:
This changes the time_start of the base pool, but this is not used by e.g. the prefork pool
where billiard.pool is used instead, and it's using `monotonic()` too.

I'm not sure if time.time() will give us enough precision to do the timeout calculations that we do,
and it's also affected by events such as leap seconds and DST transitions.

Maybe a better way would be to convert the information to a timestamp in the `inspect active` etc information, like `time_elapsed` or reporting time_start as: `time() - monotonic() + monotonic_time_start` to get an approximate conversion to UNIX timestamp.

regisb:
If I'm not mistaken, the `billiard.pool.Pool.apply_async` method does not define a `monotonic` argument, which means that the `time_start` attribute of a worker request will not be defined. Is my interpretation correct?

AFAIK `time` is the most accurate function available in Python for time measurement: https://docs.python.org/3.5/library/time.html

Anyway, I think your suggestion not to modify `time_start` at such a fundamental level makes sense. Simply adding a new piece of information in the `info` method is safer (i.e: not going to break anything). However, if we keep the `time_start` info as-is, external tools such as Flower will continue to monitor the `time_start` attribute and report a date in 1970. I guess I'll have to patch them too :)

I suggest the following:

1. Rename the `time_start` info reported by `inspect.active` in `monotonic_time_start`
2. Add a `time_start` info equal to `time() + monotonic_time_start - monotic()`

thedrow:
We can still merge this pretty easily.
@regisb Any chance you'll ever finish this?

thedrow:
@regisb Ping?

regisb:
@thedrow Oops! Terribly sorry about this. I'm late on my OSS PRs, but I should be able to work on this starting from next week. Please don't close this. Thanks!

regisb:
Thanks for waiting! This is a new draft for reporting the absolute time in the 'inspect' commands. I haven't even tried to write new tests. Can you please tell me first if this fix goes in the right direction?

(ho and merry :santa:)

auvipy:
you could try writing test here to verify the change https://github.com/celery/celery/blob/master/t/unit/worker/test_request.py

regisb:
@auvipy Yes, I definitely will! I'd just like to be sure that the chosen implementation goes in the right direction. @ask what do you think? (take your time, this can wait after ~~you've sobered up~~ the holidays.

auvipy:
@ask don't actively work on celery project....

regisb:
Ok! So I'll start working on tests.

regisb:
Just added a unit test to the PR. I think this could now be ready to merge.

PR #3695, opened 2016-12-16
make missing args/kwargs in v1 message to return empty value in v2 message

### LINKED ISSUE #3687: Celery fails to convert tasks w/o kwargs from protocol v1 to v2
InvalidTaskError is being raised when there's no kwargs present in the message.

There's actually a test (https://github.com/celery/celery/blob/master/t/unit/worker/test_strategy.py#L33-L36) ensuring that an`InvalidTaskError` is raised in case there's no kwargs in the task.

This is not what used to happen in v1. Docs also say the following for kwargs:
```
Dictionary of keyword arguments. Will be an empty dictionary if not provided.
```
as you can see in
 http://docs.celeryproject.org/en/latest/internals/protocol.html#id8 

I can open a PR if you guys think this rule should really be removed.

Thanks!

rsichny:
fixes #3687

### DIFF
--- a/celery/worker/strategy.py
+++ b/celery/worker/strategy.py
@@ -30,7 +30,7 @@ def proto1_to_proto2(message, body):
         Tuple: of ``(body, headers, already_decoded_status, utc)``
     """
     try:
-        args, kwargs = body['args'], body['kwargs']
+        args, kwargs = body.get('args', ()), body.get('kwargs', {})
         kwargs.items  # pylint: disable=pointless-statement
     except KeyError:
         raise InvalidTaskError('Message does not have args/kwargs')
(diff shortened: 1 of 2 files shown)


celery/worker/strategy.py
```diff
@@ -30,7 +30,7 @@ def proto1_to_proto2(message, body):
         Tuple: of ``(body, headers, already_decoded_status, utc)``
     """
     try:
-        args, kwargs = body['args'], body['kwargs']
+        args, kwargs = body.get('args', ()), body.get('kwargs', {})
```
gabrielpjordao: Minor (probably will have the same effect), but the default is an empty list, rather than a tuple.

rsichny: args are always a tuple, i don't think that returning different type in case of empty args will be correct.

Issue #3707, opened 2016-12-20
TypeError in app/amqp.py when sending deprecated signal


michael-k:
## Checklist

- [x] I have included the output of ``celery -A proj report`` in the issue.
      (if you are not able to do this, then at least specify the Celery
       version affected).
- [x] I have verified that the issue exists against the `master` branch of Celery as well as versions `4.0.0` and `4.0.2`.

## Steps to reproduce
- Send a task from Django with `CELERY_TASK_PROTOCOL = 2`.

## Expected behavior
- Task is sent.

## Actual behavior
`TypeError: tuple indices must be integers or slices, not str` because `body` in the following snippet is a `tuple`, not a `dict`:
```python
if sent_receivers:  # XXX deprecated
    send_task_sent(sender=name, task_id=body['id'], task=name,
                   args=body['args'], kwargs=body['kwargs'],
                   eta=body['eta'], taskset=body['taskset'])
```
Source: https://github.com/celery/celery/blob/v4.0.2/celery/app/amqp.py#L564-L566

- Removing this block and our test suite passes.
- Using protocol 1 works too.

michael-k:
root@6bd634c2cddb:/opt/code/project# celery -A 'project' report
 
software -> celery:4.0.2 (latentcall) kombu:4.0.2 py:3.5.2
            billiard:3.5.0.2 py-amqp:2.1.4
platform -> system:Linux arch:64bit, ELF imp:CPython
loader   -> celery.loaders.app.AppLoader
settings -> transport:amqp results:disabled

DATE_FORMAT: 'N j, Y'
CSRF_COOKIE_DOMAIN: None
FILE_UPLOAD_DIRECTORY_PERMISSIONS: None
LOGGING: {
    'disable_existing_loggers': False,
    'filters': {   'require_debug_false': {   '()': 'django.utils.log.RequireDebugFalse'}},
    'handlers': {   'console': {   'class': 'logging.StreamHandler',
                                   'level': 'DEBUG'},
                    'mail_admins': {   'class': 'django.utils.log.AdminEmailHandler',
                                       'filters': ['require_debug_false'],
                                       'level': 'ERROR'}},
    'loggers': {   'django.request': {   'handlers': ['mail_admins'],
                                         'level': 'ERROR',
                                         'propagate': True},
                   'statsd': {   'handlers': ['console'],
                                 'level': 'DEBUG',
                                 'propagate': False}},
    'version': 1}
DISALLOWED_USER_AGENTS: []
[...]

PR #3714, opened 2016-12-21
Fixed TypeError due to change of task protocol

michael-k:
Fixes celery/celery#3707

I'm not sure if `isinstance(body, tuple)` is the best way to differentiate between protocol versions 1 and 2, but I haven't found a better one.  I'm happy to change that if someone can point me in the right direction.


michael-k:
I've added `test_send_task_message__with_receivers` which is a clone of `test_send_task_message__no_kwargs` (L237) but with `sent_receivers` in `celery/app/amqp.py` evaluating to `True`.

It only fails if any exception occurs (meaning it only tests syntax but no semantic) and only tests the protocol-version-2 part.  The whole `send_task_message` function looks untested for v1 to me and I don't feel familiar enough with celery's tests to change that.

Issue #3459, opened 2016-09-22
celery multi is working strange with systemd


gdubicki:
Hi,

As I had problems with running celery using systemd on Centos 7 I did some testing to find out what exactly is happening. I used the "tasks.py" app from the tutorial (http://docs.celeryproject.org/en/latest/getting-started/first-steps-with-celery.html#application).

Here are my findings.

**1. You HAVE TO use PIDFile systemd service option for running a single worker**

Example of a working service file with a single worker, but with 4 concurrent processes:

```
[Unit]
Description=Celery Service
After=network.target

[Service]
Type=forking
User=vagrant
Group=vagrant

# directory with tasks.py
WorkingDirectory=/home/vagrant/celery_example

# this is REQUIRED!
# you will still get a warning "PID file /var/run/celery/single.pid not readable (yet?) after start." from systemd but service will in fact be starting, stopping and restarting properly
PIDFile=/var/run/celery/single.pid

ExecStart=/home/vagrant/celery_example/venv/bin/celery multi start single-worker -A tasks --pidfile=/var/run/celery/single.pid --logfile=/var/log/celery/single.log "-c 4 -Q celery -l INFO"

ExecStop=/home/vagrant/celery_example/venv/bin/celery multi stopwait single-worker --pidfile=/var/run/celery/single.pid --logfile=/var/log/celery/single.log

ExecReload=/home/vagrant/celery_example/venv/bin/celery multi restart single-worker --pidfile=/var/run/celery/single.pid --logfile=/var/log/celery/single.log

[Install]
WantedBy=multi-user.target
```

**2. You CAN'T use `celery multi` --pidfile or --logfile options when running multiple workers**

Example of a working service file with 3 workers, 4 concurrent processes each:

```
[Unit]
Description=Celery Service
After=network.target

[Service]
Type=forking
User=vagrant
Group=vagrant

# directory with tasks.py
WorkingDirectory=/home/vagrant/celery_example

# don't set --pidfile or --logfile or it won't work!
ExecStart=/home/vagrant/celery_example/venv/bin/celery multi start 3 -A tasks "-c 4 -Q celery -l INFO"

ExecStop=/home/vagrant/celery_example/venv/bin/celery multi stopwait 3

ExecReload=/home/vagrant/celery_example/venv/bin/celery multi restart 3

[Install]
WantedBy=multi-user.target
```

When using pidfile or logfile only one worker is actually starting.

Can you explain some of this?

[...]

PR #3715, opened 2016-12-22
Modify the systemd script to make it work

chehsunliu:
The original systemd script use `celery multi start`, but actually systemd is unable to get all the parent PIDs. For now, `celery worker` is the only solution. Consider it as a workaround for #3459.

### DIFF
--- a/extra/systemd/README.rst
+++ b/extra/systemd/README.rst
@@ -1,0 +1,26 @@
+Usage
+=====
+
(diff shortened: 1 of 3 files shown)


rsichny:
in the description of #3459 `ExecStart=celery multi start` is used instead of using `ExecStart=/bin/sh -c 'celery multi ...'`. The second command works correctly, i don't see a reason to change the script.

PR #3716, opened 2016-12-22
Fix request context for blocking task apply (added hostname)

decaz:
## Description

I have CELERY_TASK_ALWAYS_EAGER = True in my project test settings and inside one of my tasks I am using self.request.hostname attribute, but it equals None.

So we should assign 'hostname' attribute inside the request context in blocking task apply method.


thedrow:
Feel free to add yourself to the contributors list.

PR #3721, opened 2016-12-23
Provide a transparent method to update the Scheduler heap

aleperno:
In previous versions, we were able to modify dynamically the `schedule`
dict and it will be seen reflected in the scheduled tasks since `tick`
method directly iterated over it.

In 4.0 version, a heap is used and set after the init (when `_heap` is
None), this breaks our ability to dynamically change the `schedule` and
see it reflected without directly modifying the `_heap` attribute thus
breaking the abstraction.

Since there are already methods than allows to modify the `schedule`
without going straight into inner implementation, providing a
method to modify the `_heap` without knowing of its internal makes
sense.

## Description

that version series.

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -225,3 +225,4 @@ Rick Wargo, 2016/12/02
 zhengxiaowai, 2016/12/07
 Michael Howitz, 2016/12/08
 Andreas Pelme, 2016/12/13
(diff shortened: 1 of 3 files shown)


aleperno:
@auvipy Hi, is there any estimate on when will be 4.0.3 tagged?
Thanks in advance

auvipy:
@ask handle release

PR #3731, opened 2016-12-30
Fix #3725: Task replaced with group does not complete

### LINKED ISSUE #3725: Replaced task does not complete when replacement sig is a group (4.0.2)
## Checklist

```
software -> celery:4.0.2 (latentcall) kombu:4.0.2 py:2.7.11
settings -> transport:amqp results:redis://localhost:6379/0

result_backend: u'redis://localhost:6379/0'
```
(The above output is against `master`.)

Also:
```
(foo) $ pip freeze | grep redis
redis==2.10.5
```

## Steps to reproduce
*foo.py:*
```python
from celery import Celery, group
app = Celery('foo', broker='amqp://guest:guest@localhost:5672//', backend='redis://localhost:6379/0')

@app.task
def multiply(a, b):
    return a * b

@app.task(bind=True)
def multiply_by_two(self, num):
    self.replace(multiply.s(num, 2))

@app.task(bind=True)
def multiply_all_by_two(self, nums):
    subtasks = [multiply_by_two.s(n) for n in nums]
    raise self.replace(group(*subtasks))
```

*bar.py:*
```python
from foo import multiply_by_two, multiply_all_by_two
if __name__ == '__main__':
    print multiply_by_two.delay(3).get()  # completes in 4.0.1 and 4.0.2
    print multiply_all_by_two.delay([1, 2, 3]).get()  # completes in 4.0.1; does not complete in 4.0.2
```

## Expected behavior
A replaced task should complete when the replacement sig is a group (and does in 4.0.1).

## Actual behavior
A replaced task does not complete when the replacement sig is a group (in 4.0.2).

morganiq:
### Summary ###
* Updates `chord.__length_hint__()` and `chord._traverse_tasks()` to correctly handle the case when `self.tasks` is a group.
* Updates integration test of complex chaining to reproduce the issue and pass.

### Details ###
Before `group.__iter__()` was removed in 8c7ac5d, passing either a list or a group through `list()` would yield the same thing: the list of tasks. After the removal of `group.__iter__()`, `list(g)` yields the same thing as `g.keys()`, so we need to handle the group case explicitly now.

Using `len(g.keys())` for `__length_hint__` will only happen to be the correct chord size sometimes; the rest of the time it will result in chords that never get unlocked due to a mismatch between the computed chord size and the size of the complete set of results returned.

Fixes #3725 - "Replaced task does not complete when replacement sig is a group"

### DIFF
--- a/t/integration/test_canvas.py
+++ b/t/integration/test_canvas.py
@@ -4,7 +4,7 @@ from celery import chain, chord, group
 from celery.exceptions import TimeoutError
 from celery.result import AsyncResult, GroupResult
 from .conftest import flaky
-from .tasks import add, collect_ids, ids
+from .tasks import add, collect_ids, ids, add_to_all
 
 TIMEOUT = 120
 
@@ -22,10 +22,11 @@ class test_chain:
             add.s(2, 2) | (
                 add.s(4) | add.s(8) | add.s(16)
             ) |
-            group(add.s(i) for i in range(4))
+            group(add.s(i) for i in range(4)) |
+            add_to_all.s(32)
         )
         res = c()
-        assert res.get(timeout=TIMEOUT) == [32, 33, 34, 35]
+        assert res.get(timeout=TIMEOUT) == [64, 65, 66, 67]
 
     @flaky
     def test_parent_ids(self, manager, num=10):
--- a/celery/canvas.py
+++ b/celery/canvas.py
@@ -1253,7 +1253,7 @@ class chord(Signature):
         )
 
     def _traverse_tasks(self, tasks, value=None):
-        stack = deque(list(tasks))
+        stack = deque(tasks)
         while stack:
             task = stack.popleft()
             if isinstance(task, group):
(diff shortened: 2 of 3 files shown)


t/integration/test_canvas.py
```diff
@@ -89,6 +89,18 @@ def assert_ids(r, expected_value, expected_root_id, expected_parent_id):
 class test_chord:
 
     @flaky
+    def test_group_chain(self, manager):
```
thedrow: Is this test still flaky?

t/integration/test_canvas.py
```diff
@@ -89,6 +89,18 @@ def assert_ids(r, expected_value, expected_root_id, expected_parent_id):
...
+            raise pytest.skip('Requires redis result backend.')
+        c = (
+            add.s(2, 2) |
+            group(add.s(i) for i in range(4)) |
+            add_to_all.s(8)
+        )
+        res = c()
+        assert res.get(timeout=TIMEOUT) == [12, 13, 14, 15]
+
+    @flaky
```
thedrow: Is this test still flaky?

morganiq:
FYI, rebased to 9d2566e and squashed some commits.

t/integration/test_canvas.py
morganiq: I don't know; the `@flaky` decorator looked to be indiscriminately applied to all tests in `t/integration`, so I included it for consistency. Someone else would have to answer about whether or where it is still needed.

thedrow:
Thanks!

PR #3736, opened 2017-01-04
AWS DynamoDB result backend

georgepsarakis:
## Description

The changes in this pull request add the ability to define [AWS DynamoDB](http://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Introduction.html) as the result backend.

DynamoDB is a NoSQL store that has both document-oriented and key-value semantics.

### Motivation

* Adheres to the [IAM permission system](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/iam-roles-for-amazon-ec2.html).
* Scalability with provisioned read/write throughput.
* No storage restrictions.
* Combined with [SQS](http://docs.celeryproject.org/en/latest/getting-started/brokers/sqs.html), can work seamlessly in an EC2 environment.

### Notes
- The module structure and test cases was for the most part, modeled after the Riak result backend.
- The table is automatically created if not exists. After creation, the code polls until the table is available (state `ACTIVE`).
- Perhaps more configuration/customization capabilities should be added, especially regarding the table schema.
- Currently can only be configured via URL (documentation pending).

### Table Schema

| Attribute Name  | Data Type | Description |
| ------------- | ------------- | ------------- | 
| `id` | `String` | The task result identifier |
| `result` | `Bytes` | The base64-encoded serialized task result |
| `timestamp` | `Number` | The current timestamp as a floating-point number |


georgepsarakis:
This is still a work in progress (documentation pending mainly), but any feedback will be greatly appreciated. Thanks!

georgepsarakis:
For anyone willing to test these changes, I used the following sample code:

### Create the task definitions module `tasks.py`

```python
from celery import Celery

# Assumes that AWS credentials will be available as environment variables
# or in ~/.aws/credentials 
app = Celery(
    'tasks',
    broker='redis://localhost',
    backend='dynamodb://@us-east-1'
)


@app.task
def add(x, y):
    return x + y
```

### Start the worker

```bash
celery -A tasks worker --loglevel=info
```

### Add task

```python
from tasks import add
task = add.delay(4, 4)
print task.get() # 8
```

georgepsarakis:
Documentation added and CI checks passed.

georgepsarakis:
@thedrow I don't have a problem with implementing the tests. For the time being I used mocking over the `boto3` function calls. I will take a look at `moto` in the meanwhile. Thanks for the feedback so far!

georgepsarakis:
@thedrow I discovered that AWS offers DynamoDB as a [downloadable Java application](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/DynamoDBLocal.html).

I have enabled the integration tests and I added the dependencies in Travis CI.

Let me know what you think!

georgepsarakis:
@thedrow I have enabled the integration tests with DynamoDB as a result backend.

If you think it is necessary, I will also add functional tests with `moto`, but could you please share some thoughts on what the format of these tests should be?

Thanks again for taking the time to review this.

georgepsarakis:
@auvipy I have rebased but I am trying to fix Travis CI issues related to #3750 and #3749 .

georgepsarakis:
I followed this [approach](https://docs.travis-ci.com/user/customizing-the-build#Explicitly-Including-Jobs) to specify the interpreter correctly.

auvipy:
wait for @ask

georgepsarakis:
@auvipy I finally resolved the Travis CI issues. I added Python 3.6 to the build matrix. Perhaps this solution is of interest for #3750.

georgepsarakis:
Unit tests [fail](https://travis-ci.org/celery/celery/jobs/190365340) with Python 3.6, but this is out-of-scope for this pull request. All integration tests seem to pass with 3.6 though.

ask:
Looks really good to me!  Very good work, I think this can be merged :)

georgepsarakis:
Thanks @ask :) !

georgepsarakis:
I think I found out what is causing the failure in unit tests with Python 3.6. Triggering the `RuntimeError` in [test_recursive](https://github.com/celery/celery/blob/b2027b9c6c3d71e96e4f5f7515385fc99a970abe/t/unit/app/test_log.py#L88), results in an endless loop.

Changing this [line](https://github.com/celery/celery/blob/b2027b9c6c3d71e96e4f5f7515385fc99a970abe/celery/utils/log.py#L85) to:


```python
'Logger {0!r} parents recursive'.format(l.name)
```

resolves the error.

The reason for this is that in order to construct the [representation](https://hg.python.org/cpython/file/3.6/Lib/logging/__init__.py#l1570) of the logger `l`, while generating the exception message, the [getEffectiveLevel](https://hg.python.org/cpython/file/3.6/Lib/logging/__init__.py#l1536) function never returns, since the logger's parent is the logger itself.

georgepsarakis:
Some fixes:
- [x] Resolved unit tests issue with endless loop for Python 3.6.
- [x] Unit tests were being skipped due to missing dependency boto3, added to *requirements/test-ci-default.txt*.

Unfortunately, the [documentation build job](https://travis-ci.org/celery/celery/jobs/194023718) is failing, but I don't think that it is related because it was passing before. I also found [this](https://github.com/sphinx-doc/sphinx/issues/2324) relevant Sphinx issue.

georgepsarakis:
Issue with Sphinx documentation build resolved using explicit hash seed value in tox environment.

thedrow:
Thank you very much!

PR #3757, opened 2017-01-12
Add supporting for task execution working with tornado coroutine

justdoit0823:
With this future, we can use tornado coroutine in celery app task when pool implementation is gevent. The main idea here is using a standalone tornado ioloop thread to execute coroutine task. If a task is a coroutine, the executor will register a callback on the tornado ioloop and switch to the related gevent hub. Here using a callback means that the coroutine will be totally executed inside the tornado ioloop thread. When the coroutine is finished, the executor will spawn a new greenlet which will switch to the previous task greenlet with the coroutine's result. Then task greenlet returns the result to the outer function inside the same greenlet. 

We can write code in celery task as following:

```python
from celery import app
from tornado import gen

@app.task
@gen.coroutine
def task_A():

    process_1()
    res = yield get_something()
    do_something_with_res()
    return res

@gen.coroutine
def get_something():

    res = {}
    return res
```

This is interesting when we use the tornado framework to write a project, we can easily divide partial logic into a celery task without any modifications. And in tornado 5.0, which will support Python 3 native coroutine, this future can connect more things together. I think people will like this.


celery/concurrency/base.py
```diff
@@ -44,6 +45,9 @@ def apply_target(target, args=(), kwargs={}, callback=None,
     else:
         callback(ret)
 
+    # Switch to the tornado ioloop greenlet
+    future_executor.switch_to_ioloop()
```
thedrow: Pretty sure this doesn't belong here since tornado workers pool should be optional.

justdoit0823: Yeah, recent works give me a idea about how to implement a more elegant interface between greenlet and tornado coroutine. I will start a daemon tornado ioloop thread and receive tornado coroutine in greenlet. When coroutine finishs, tornado ioloop can send result to related greenlet.

thedrow: So this PR is no longer relevant?

celery/app/trace.py
```diff
@@ -364,10 +364,15 @@ def trace_task(uuid, args, kwargs, request=None):
 
                 # -*- TRACE -*-
                 try:
-                    from celery.concurrency.future import future_executor
-                    R = retval = future_executor.apply_future(
-                        fun, *args, **kwargs)
-                    state = SUCCESS
+                    try:
+                        from celery.concurrency.future import get_future_executor
```
justdoit0823: In this case, the default future executor will do the right job as before. If the task is a coroutine function, which will be partially executed in the tornado ioloop. But the result of coroutine function can't immediately return.

thedrow:
The unit tests now hang or fail. Can you reproduce the results?

justdoit0823:
I have fixed the test case issue. In the `test_eventlet` module, the code `del sys.modules` can lead importing related modules of eventlet package again in the following test cases. And functions in the eventlet package still refer to the old module global variables such as `_threadlocal`. So there are two hubs in the same thread, which causes switching failed and hanging.

justdoit0823:
@thedrow Now, it's ready to review and test this future.

justdoit0823:
@thedrow @auvipy Sorry for the late reply. The problem here is neither writing the related document nor rebasing the repo. 

At the beginning of this PR, I'm satisified with this feature which supports the consistent program style in the `tornado` web program with `celery`, although the implementation may be not perfect enough in some situation. Now there are a piece of asynchronous task execution frameworks, which are not compatible with each other. I think we need a unified io mode pattern in our programs rather a lot of different wheel patterns. So I doubt whether it is a feature or a pitfall.

Yougoss:
In tornado IOLoop, sometimes it uses signal to awake select/epoll blocking.
I think use signal in thread not main perharps cause some problems.

auvipy:
closing for now

PR #3764, opened 2017-01-13
Add support for soft time limits in gevent pool

brechin:
## Description

Adds support for soft_time_limit usage with gevent worker pool.

### DIFF
--- a/celery/concurrency/gevent.py
+++ b/celery/concurrency/gevent.py
@@ -18,8 +18,8 @@ __all__ = ['TaskPool']
 
 
 def apply_timeout(target, args=(), kwargs={}, callback=None,
-                  accept_callback=None, pid=None, timeout=None, soft_timeout=None,
-                  timeout_callback=None, Timeout=Timeout,
+                  accept_callback=None, pid=None, timeout=None,
+                  soft_timeout=None, timeout_callback=None, Timeout=Timeout,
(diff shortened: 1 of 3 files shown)


celery/concurrency/gevent.py
```diff
@@ -17,15 +19,17 @@
...
                   apply_target=base.apply_target, **rest):
+    soft = None
     try:
-        with Timeout(timeout):
-            return apply_target(target, args, kwargs, callback,
-                                accept_callback, pid,
-                                propagate=(Timeout,), **rest)
-    except Timeout:
-        return timeout_callback(False, timeout)
+        with Timeout(soft_timeout, SoftTimeLimitExceeded) as soft:
```
georgepsarakis: Can you please explain something here? Since `soft_timeout` is lower than `timeout`:

```python
with gevent.Timeout(5, SoftTimeLimitExceeded):
      with gevent.Timeout(6):
            gevent.sleep(5.2)
```

The above code block won't always raise `SoftTimeLimitExceeded`?

brechin: @georgepsarakis Yes, I think you're right. I'll swap the nesting. Thanks!

brechin:
I think I need to do some more extensive testing and resubmit with more substantial test cases included. My apologies for the hasty PR.

PR #3777, opened 2017-01-20
Update tasks.rst to mention django's `on_commit` function.

orf:
## Description

The example in the documentation about launching a task after a successful commit is a bit outdated, using the `django.db.transaction.on_commit` function to launch the task is a better, and often the only feasible way in anything more complex than basic applications.

Also if you're using nested transactions it's possible to have a parent transaction roll back, despite the one that the task was launched in from succeeding. The `on_commit` function handles this properly, by queuing all callbacks until all transactions are properly committed.

### DIFF
--- a/docs/userguide/tasks.rst
+++ b/docs/userguide/tasks.rst
@@ -1739,21 +1739,15 @@ There's a race condition if the task starts executing
 before the transaction has been committed; The database object doesn't exist
 yet!
 
-The solution is to *always commit transactions before sending tasks
-depending on state from the current transaction*:
+The solution is to use the ``on_commit`` callback to launch your celery task 
+once all transactions have been committed successfully.
 
 .. code-block:: python
-
-    @transaction.commit_manually
+    from django.db.transaction import on_commit
+    
     def create_article(request):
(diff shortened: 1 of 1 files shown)


thedrow:
Thanks! Feel free to add yourself to the AUTHORS file.

michael-k:
:+1: for `on_commit`, but keep in mind that this was introduced in Django 1.9 and the current LTS is 1.8.  So maybe the old solution should not be dropped completely.

https://docs.djangoproject.com/en/1.10/topics/db/transactions/#performing-actions-after-commit

thedrow:
@orf Do you mind editing the documentation again or do you want me to pick it up?

orf:
I can add the old docs back in with a caveat, but I think a better solution would be to add a link to this project, which provided on_commit was merged into 1.9: https://github.com/carljm/django-transaction-hooks

As I mentioned before the old code has a few problems, and keeping it in there verbatim isn't a great idea in my opinion.

thedrow:
I agree. Linking to django-transaction-hooks is the way to go here.

Issue #3771, opened 2017-01-17
Bad behavior when a chord chains with a group


ryanhiebert:
## Checklist

✓ I have included the output of ``celery -A proj report`` in the issue.
✓ I have verified that the issue exists against the `master` branch of Celery.

Output of `celery -A tasks report`:

```
software -> celery:4.0.2 (latentcall) kombu:4.0.2 py:3.6.0
            billiard:3.5.0.2 py-amqp:2.1.4
platform -> system:Darwin arch:64bit imp:CPython
loader   -> celery.loaders.app.AppLoader
settings -> transport:amqp results:redis:///

result_backend: 'redis:///'
```

## Steps to reproduce

Create a `tasks.py` file with the following:

```python
from celery import Celery, group
app = Celery('tasks', backend='redis://')

@app.task
def echo(string):
    print(string)

def get_sig():
    before = group(echo.si('before {}'.format(i)) for i in range(3))
    connect = echo.si('connect')
    after = group(echo.si('after {}'.format(i)) for i in range(2))
    return before | connect | after
```

Then at the python prompt run:

```pycon
Python 3.6.0 (default, Jan  5 2017, 23:49:18)
[GCC 4.2.1 Compatible Apple LLVM 8.0.0 (clang-800.0.42.1)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import tasks; tasks.get_sig().delay()
```

Then run a celery worker with `celery -A tasks worker`, and see what output it gives.

## Expected behavior

I expected to see output that indicated the three `before` tasks ran, then the `connect` task, then the two `after` tasks.

## Actual behavior

This is the output that I saw from the celery worker:

```
celery@Ryans-MacBook-Air.local v4.0.2 (latentcall)

Darwin-16.3.0-x86_64-i386-64bit 2017-01-16 23:03:45

[config]
.> app:         tasks:0x1061ec5f8
.> transport:   amqp://guest:**@localhost:5672//
.> results:     redis://
.> concurrency: 4 (prefork)
.> task events: OFF (enable -E to monitor tasks in this worker)

[queues]
.> celery           exchange=celery(direct) key=celery


[2017-01-16 23:03:47,275: WARNING/PoolWorker-4] before 1
[2017-01-16 23:03:47,276: WARNING/PoolWorker-1] before 2
[2017-01-16 23:03:47,276: WARNING/PoolWorker-2] before 0
[2017-01-16 23:03:47,370: WARNING/PoolWorker-3] after 0
[...]

PR #3782, opened 2017-01-23
Added a test to ensure #3771 does not regress.

thedrow:
#3779 does not include the integration test required to ensure this bug will not regress.
I'm adding it to our integration suite.

### DIFF
--- a/t/integration/test_canvas.py
+++ b/t/integration/test_canvas.py
@@ -34,7 +34,7 @@ class test_chain:
         after = group(echo.si('after {}'.format(i)) for i in range(2))
 
         res = (before | connect | after).delay()
-        assert res.get(timeout=TIMEOUT)
+        assert res.get(timeout=TIMEOUT) == ['before 0', 'before 1', 'before 2', 'connect', 'after 0', 'after 1']
 
     @flaky
     def test_parent_ids(self, manager, num=10):
--- a/t/integration/test_canvas.py
+++ b/t/integration/test_canvas.py
@@ -4,7 +4,7 @@ from celery import chain, chord, group
 from celery.exceptions import TimeoutError
 from celery.result import AsyncResult, GroupResult
 from .conftest import flaky
-from .tasks import add, add_replaced, add_to_all, collect_ids, ids, echo
+from .tasks import add, add_replaced, add_to_all, collect_ids, ids, redis_echo
 
 TIMEOUT = 120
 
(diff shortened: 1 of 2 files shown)


ryanhiebert:
Thanks, I wasn't comfortable enough with the test suite to write it myself. I figured if worst came to worst, you'd end up just merging it, since it's a bug fix.

ryanhiebert:
@thedrow : What needs to happen to get this merged?

thedrow:
I need to find a way to reproduce the failure in such a way that allows me to verify how many times the task was run.
I haven't found a way yet.

georgepsarakis:
@thedrow how about using Redis to verify that (since it is already required for the chain/chord):

```python
@shared_task
def redis_echo(message):
    from redis import StrictRedis
    
    redis_connection = StrictRedis()
    redis_connection.rpush('redis-echo', message) 
    return message
```

Then in the test, you can do something like:

```python
def test_group_chord_group_chain(self, manager):
     from redis import StrictRedis   
     redis_connection = StrictRedis()
     ...
     assert redis_connection.lrange('redis-echo', 0, -1) == [<EXPECTED_SEQUENCE_OF_MESSAGES>]
```

thedrow:
Does this result in something that make sense? I'm seeing a lot more tasks than the number of results I have in redis.

t/integration/test_canvas.py
```diff
@@ -28,6 +28,21 @@ def test_complex_chain(self, manager):
         assert res.get(timeout=TIMEOUT) == [64, 65, 66, 67]
 
     @flaky
+    def test_group_chord_group_chain(self, manager):
+        if not manager.app.conf.result_backend.startswith('redis'):
+            raise pytest.skip('Requires redis result backend.')
+        from redis import StrictRedis
+        redis_connection = StrictRedis()
+
```
georgepsarakis: Perhaps it would be best if you cleared the Redis list, just in case the test is repeated:

```python
redis_connection.delete('redis-echo')
```

georgepsarakis:
@thedrow what do you mean? If I understand correctly the failure of the [build job](https://travis-ci.org/celery/celery/jobs/197630353) is consistent with the logs in #3771 :

```
[2017-01-16 23:03:47,275: WARNING/PoolWorker-4] before 1
[2017-01-16 23:03:47,276: WARNING/PoolWorker-1] before 2
[2017-01-16 23:03:47,276: WARNING/PoolWorker-2] before 0
[2017-01-16 23:03:47,370: WARNING/PoolWorker-3] after 0
```

There is no `connect` message after the `before` group tasks are completed.

Also, perhaps you should be clearing the Redis list contents before starting the test, see my comment above.

Also, the sequence will not be deterministic, since task groups can be executed in arbitrary order, correct? Which means that you will need a more complex assertion, splitting the list of results and asserting with sets probably:

```python
redis_messages = redis_connection.lrange('redis-echo', 0, -1)
assert set(['before 0', 'before 1', 'before 2']) == set(redis_messages[:3])
assert redis_messages[3] == 'connect'
assert set(redis_messages[4:]) ==set(['after 0', 'after 1'])
```

I guess, that now that you have a failed test, you can apply the patch to see if it fixes the issue.

georgepsarakis:
I think the larger number of results reported in the build job failure is explained, since the [flaky](https://github.com/celery/celery/blob/master/t/integration/conftest.py#L11) decorator retries the test on failure. If you clear the Redis list, I think it will be fine.

georgepsarakis:
@thedrow can I perhaps help with this PR? We can also merge #3779 if we manage to merge this :smiley: !

thedrow:
Sure thing. Take it from here.

thedrow:
Superseded by #3963.

PR #3783, opened 2017-01-23
Run unit tests in parallel

thedrow:
The build is painfully slow right now. I'm checking if we can speed up the unit tests at the very least.

### DIFF
--- a/requirements/test-ci-base.txt
+++ b/requirements/test-ci-base.txt
@@ -1,4 +1,5 @@
 pytest-cov
+pytest-xdist
 codecov
 -r extras/redis.txt
 -r extras/sqlalchemy.txt
--- a/t/unit/conftest.py
+++ b/t/unit/conftest.py
@@ -45,24 +45,25 @@ CASE_LOG_HANDLER_EFFECT = 'Test {0} modified handlers for the root logger'
 
 
 @pytest.fixture(scope='session')
-def celery_config():
+def celery_config(request):
+    xdist_suffix = getattr(request.config, 'slaveinput', {}).get('slaveid', '')
     return {
         'broker_url': 'memory://',
         'result_backend': 'cache+memory://',
 
-        'task_default_queue': 'testcelery',
-        'task_default_exchange': 'testcelery',
+        'task_default_queue': 'testcelery' + xdist_suffix,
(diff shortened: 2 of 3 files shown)


orf:
I would set `-n` to `auto`, rather than hard-code it to `2`. `auto` will detect the number of cpus and use that number :+1:

thedrow:
Not really since Travis uses a virtualized environment that doesn't expose the number of CPUs correctly.

orf:
Also perhaps adding `--durations=N` as an option is a good idea, pytest will print out the slowest `N` tests.

thedrow:
Right I remember that one of these tests failed.
The question is why though.

orf:
I'm going to take a stab in the dark and say that the multiple test runners are all using the same queue settings: https://github.com/celery/celery/blob/master/t/unit/conftest.py#L48

I think they should be prefixed with the 'worker id', i.e ` 'task_default_queue': 'testcelery_' + WORKER_ID,` or some such.

thedrow:
I don't think so since the broker is set to memory. I'll parametrize it anyway just to be on the safe side.

georgepsarakis:
@thedrow , after the merge of #3785, the builds are significantly faster. If you check the specifications of the virtualization environments offered by [Travis](https://docs.travis-ci.com/user/ci-environment/#Virtualization-environments), you can find that the current environment is `SUDO-ENABLED`, since  `sudo: required` is specified in `.travis.yml`.

This virtualization environment uses a **maximum** of 2 CPUs, so running tests in parallel won't surely  give much performance gain. Also, the boot time ranges from 20s to 52s, which is comparable to the time that tests consume.

I think it would be more beneficial if the project reverted to container-based environment, if possible, rather than adding complexity for running tests in parallel.

georgepsarakis:
Tried what I suggested above in #3829, the results were disappointing.

georgepsarakis:
@thedrow I added some optimizations with #3867 .

thedrow:
@georgepsarakis I think we need to resolve this somehow anyway. This failure indicates that one of our tests depends on another test.

thedrow:
I just tried running with xdist again and everything seems to be in order. I'll create a new PR shortly.

thedrow:
Ok so this is very strange. With 2 workers the test suite fails. With 8, it passes.

PR #3784, opened 2017-01-23
Add myself to the authors file

orf:
Just adding myself to the authors file for #3777. Should have done it as part of that MR rather than a separate one, sorry!


thedrow:
No problems :+1:

PR #3785, opened 2017-01-23
Cache pip installs

orf:
Perhaps caching pip installs would be a good speed up, it was set to `false` back in 2015 with the [following comment](https://github.com/celery/celery/commit/fe33f16e014611ee7267081b3380b5a0003faf78#commitcomment-17886249):

`The cache will probably make it faster, but caching in pip often cause trouble. Sadly I don't remember exactly why, maybe it refused to upgrade some version or similar?`

I've personally not seen this problem, let's see if it speeds the tests up at all.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: python
 sudo: required
-cache: false
+cache: pip
 python:
   - '2.7'
   - '3.4'
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: python
 sudo: required
 cache: pip
-
 python:
   - '2.7'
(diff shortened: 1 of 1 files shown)


georgepsarakis:
@orf seems like a considerable improvement, each job takes approximately 30% of the time without the caching enabled.

orf:
Indeed @georgepsarakis it seems to save a lot of time, I just triggered a new build. Could be a quick and easy way to speed up the tests @thedrow

thedrow:
I'm +1. If @ask won't object, I'll merge this.

thedrow:
No objection heard so I'm merging this.

PR #3788, opened 2017-01-24
Fix CELERYBEAT_LOG_FILE replacement option typo

tamers:
The CELERYBEAT_LOG_FILE was set to be replaced with `celery beat --loglevel`. It should be `celery beat --logfile`


thedrow:
Thanks. Feel free to add yourself to the authors file if you are not there already.

PR #3799, opened 2017-01-30
Add a link to django-transaction-hooks

orf:
Remove the somewhat unneeded warning and replace it with a link to the django-transaction-hooks library for Django LTS releases.

Ref #3799


thedrow:
Thanks!

PR #3815, opened 2017-02-07
Chord does not clean up subscribed channels in redis. Fixes #3812

### LINKED ISSUE #3812: Chord does not clean up subscribed channels in redis
## Checklist

       version affected).

```
settings -> transport:pyamqp results:redis://127.0.0.1/

result_backend: 'redis://127.0.0.1/'
```
## Steps to reproduce
1. Create chord on celery with eventlet on redis backend.
2. Run chord
3. Run `pubsub channels`
4. Run chord again
5. Run `pubsub channels`

Code example: https://gist.github.com/gugu/680452f754507ec6bea144a2b257c398
## Expected behavior
On step 3 and 5 we need to have 0 channels
## Actual behavior
On step 3 we have 4 open channels
On step 5 we have 8 open channels
On our live application in 2 hours (between restarts) we have 3 200 000 open channels, which generally slows everything and hell gates open with tenth of different issues.

PS: will try to return with pull request later

gugu:
This pull requests fixes issue, when chord gets result for subtasks, but does not do `UNSUBSCRIBE` command for them.

**What does this patch fix**:

Without this patch `chord(task() for i in range(50))` creates 51 subscriptions. After patch it will create 1

**What this patch does not fix**:

Every task, which result is not consumed, creates redis subscription. Even if `ignore_result` is specified. I plan to submit separate patch for this issue

I think, that changing old slow redis behavior to new fast and broken was not a good idea, but as soon as choice made, we need to make it usable

### DIFF
--- a/celery/backends/redis.py
+++ b/celery/backends/redis.py
@@ -286,6 +286,10 @@ class RedisBackend(base.BaseKeyValueStoreBackend, async.AsyncBackendMixin):
                         callback,
                         ChordError('Callback error: {0!r}'.format(exc)),
                     )
+                finally:
+                    task_ids = [decode(tup)[1] for tup in resl]
+                    for task_id in task_ids:
+                        self.result_consumer.cancel_for(task_id)
         except ChordError as exc:
             logger.exception('Chord %r raised: %r', request.group, exc)
             return self.chord_error_from_stack(callback, exc)
--- a/t/integration/tasks.py
+++ b/t/integration/tasks.py
@@ -13,8 +13,11 @@ def add(x, y):
     return x + y
 
 @shared_task
-def sum_(numbers):
+def delayed_sum(numbers, pause_time=1):
     """Sum the iterable of numbers."""
+    # Allow the task to be in STARTED state for
+    # a limited period of time.
+    sleep(pause_time)
     return sum(numbers)
 
 @shared_task(bind=True)
--- a/t/integration/tasks.py
+++ b/t/integration/tasks.py
@@ -12,6 +12,10 @@ def add(x, y):
     """Add two numbers."""
     return x + y
 
+@shared_task
+def sum_(numbers):
+    """Sum the iterable of numbers."""
+    return sum(numbers)
 
 @shared_task(bind=True)
 def add_replaced(self, x, y):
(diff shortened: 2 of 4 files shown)


georgepsarakis:
@thedrow I will see if I can add the integration test here.

thedrow:
@georgepsarakis Any chance you'll get to that integration test? I want to merge this before we release 4.2.

georgepsarakis:
@thedrow I believe something like the following will do:

### `t/integration/tasks.py`
```
@shared_task
def sum_(numbers):
    """
    Sum the iterable of numbers.
    :param list[int, float] numbers:
    """
    return sum(numbers)
```

### `t/integration/test_canvas.py`

```
def test_redis_subscribed_channels_leak(self, manager):
        if not manager.app.conf.result_backend.startswith('redis'):
            raise pytest.skip('Requires redis result backend.')
        redis_client = StrictRedis()
        result = chord([add.s(5, 6), add.s(6, 7)])(sum_.s())
        sleep(1)
        channels_before = \
            len(redis_client.execute_command('PUBSUB CHANNELS'))
        result.get(timeout=TIMEOUT)
        channels_after = \
            len(redis_client.execute_command('PUBSUB CHANNELS'))
        assert channels_after < channels_before
```

Unfortunately I won't have time until the weekend to test this though.

thedrow:
@georgepsarakis Ping? Would love to get this merged for 4.2

georgepsarakis:
@thedrow I have suggested a possible test in my previous comment. Should I try and add it to this PR's branch? @gugu what do you think?

thedrow:
Yes, please do.
Though this is contributed from his master so I'm not sure you can push.
I'll get to it if you won't :)

georgepsarakis:
@thedrow the test is failing, the number of channels does not decrease. If you can please cross-check the logic of the test.

thedrow:
The test looks 100% ok.
@gugu Any chance you'll take a look? It seems like the bug still exists despite the passing unit test that says otherwise.

thedrow:
@georgepsarakis So what should we do with this PR?

georgepsarakis:
@thedrow I don't think this change is required anymore. We can resume efforts here later, if for some reason the issue is still not resolved.

PR #3827, opened 2017-02-11
fix reduce

### LINKED ISSUE #3826: celery.schedules.crontab.__reduce__ does not return kwargs.
## Checklist

       version affected).

```

timezone: 'UTC'
beat_schedule: {
    'test task': {   'args': ('hello',),
                     'schedule': <crontab: 35 18 * * * (m/h/d/dM/MY)>,
                     'task': 'test.test'},
    'test task2': {   'args': ('world',),
                      'schedule': <crontab: 35 15 * * * (m/h/d/dM/MY)>,
                      'task': 'test.test'}}
accept_content: ['json']
task_serializer: 'json'
result_serializer: 'json'
```

## Steps to reproduce

Set a nowfun.  The nowfun will not be restored.  When starting up celery beat, I never get a working nowfun back.

## Expected behavior

relevant kwargs should be returned instead of `None` here: https://github.com/celery/celery/blob/master/celery/schedules.py#L532.  An appropriate `__setstate__` may also be necessary to cope with this (as `app` doesn't get stored in `__dict__` as `app`, but rather `_app`).

## Actual behavior

Only the constructor elements are returned from `__reduce__`,  dropping all kwargs passed to the original `crontab.__init__` when pickling.

Taywee:
Fixes #3826, allowing pickled crontab schedules to restore properly (and schedules are pickled and unpickled on ordinary beat startup.

### DIFF
--- a/celery/schedules.py
+++ b/celery/schedules.py
@@ -101,6 +101,11 @@ class BaseSchedule(object):
             return timezone.to_local_fallback(dt)
         return dt
 
+    def __eq__(self, other):
+        if isinstance(other, BaseSchedule):
+            return other.nowfun == self.nowfun
+        return NotImplemented
+
 
 @python_2_unicode_compatible
 class schedule(BaseSchedule):
@@ -398,6 +403,7 @@ class crontab(BaseSchedule):
         self._orig_day_of_week = cronfield(day_of_week)
         self._orig_day_of_month = cronfield(day_of_month)
         self._orig_month_of_year = cronfield(month_of_year)
+        self._orig_kwargs = kwargs
         self.hour = self._expand_cronspec(hour, 24)
         self.minute = self._expand_cronspec(minute, 60)
         self.day_of_week = self._expand_cronspec(day_of_week, 7)
@@ -529,7 +535,12 @@ class crontab(BaseSchedule):
                                  self._orig_hour,
                                  self._orig_day_of_week,
                                  self._orig_day_of_month,
-                                 self._orig_month_of_year), None)
+                                 self._orig_month_of_year), self._orig_kwargs)
+
+    def __setstate__(self, state):
+        # Calling super's init because the kwargs aren't necessarily passed in
+        # the same form as they are stored by the superclass
+        super(crontab, self).__init__(**state)
 
     def remaining_delta(self, last_run_at, tz=None, ffwd=ffwd):
         # pylint: disable=redefined-outer-name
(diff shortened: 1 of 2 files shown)


georgepsarakis:
@Taywee I think you need to add a test case [here](https://github.com/celery/celery/blob/master/t/unit/app/test_schedules.py#L99), to verify that the bug is fixed and to avoid regression. What do you think?

Taywee:
@georgepsarakis Oh yes, you're absolutely right.  I'll get that wrapped in as early as I can.  Thank you for the note.

Taywee:
@georgepsarakis I've made the changes and squashed the changes to a single commit.  Please let me know if there's anything else that needs changing.

celery/schedules.py
```diff
@@ -101,6 +101,11 @@ def to_local(self, dt):
             return timezone.to_local_fallback(dt)
         return dt
 
+    def __eq__(self, other):
+        if isinstance(other, BaseSchedule):
+            return other.nowfun == self.nowfun
```
georgepsarakis: I think you need `is` instead of equality check here.

Taywee: I figured an equality check was better in case you pass in a callable (such as a "timezone now" class which stores a timezone attribute and returns the current time with `__call__` based on its stored timezone).  `is` will fail after such an object is pickled and unpickled, even if it evaluates as equal.

georgepsarakis: You are right, if it is re-instantiated by another process, it will not be the same object. However, how will equality check work for functions in general? I am not aware of any concrete method of performing this check. See also [here](http://stackoverflow.com/questions/20059011/check-if-two-python-functions-are-equal).

Taywee: I'm not too worried about code equality for functions, more so that the nowfun will evaluate equivalent and operate the same after being pickled and unpickled.  It wouldn't work for lambdas.  General functions will pickle and unpickle in a way that preserves equality.

Note that celery beat, on startup, pickles and unpickles all schedules, so before this, nowfun is entirely nonfunctional in general.  I couldn't find a way that a passed in nowfun could be preserved and used after celery beat's startup.

georgepsarakis: I think there is a way, you can specify the full import path as a string and perform a dynamic import. This project performs lazy imports: https://github.com/Russell91/pythonpy/blob/master/pythonpy/__main__.py

This way you will have consistency on the function objects.

Taywee: What I meant is that before this pull request, in celery beat, nowfun does not work at all.

Taywee:
So what needs to be changed to get this merged?  I don't think there's a whole lot of extra work we should be doing to nowfun.  If the passed-in nowfun needs special considerations to make it pickleable, it should be done on the application side, with the understanding that it needs to be able to be pickled.

thedrow:
I don't see anything wrong with this PR. @georgepsarakis LGTM can we merge?

georgepsarakis:
@thedrow yes, sure. @Taywee sorry for the delay.

PR #3829, opened 2017-02-11
Switch Travis CI to container-based environment

georgepsarakis:
## Description

Use container-based infrastructure as a [Travis CI virtualization environment](https://docs.travis-ci.com/user/ci-environment/#Virtualization-environments), as `sudo` is not required in build jobs.

This change will result in decreased total completion time for build jobs (faster boot time) and potentially more CPU (2x).


georgepsarakis:
Builds are much slower, closing.

PR #3831, opened 2017-02-11
Add SSL option for redis backends

### LINKED ISSUE #3830: Redis backend does not support SSL?
From looking at the code and documentation, it appears that SSL is only supported for redis as a broker, but not as a result backend. Is that right?

Currently we do some yucky monkey-patching of the code in our tasks file:

```python
import celery.backends.redis
import redis
import ssl

original = celery.backends.redis.RedisBackend._params_from_url
def patched(*args, **kwargs):
    result = original(*args, **kwargs)
    result.update({
        'connection_class': redis.SSLConnection,
        'ssl_cert_reqs': ssl.CERT_NONE,
    })
    return result
celery.backends.redis.RedisBackend._params_from_url = patched
```

The code change in celery to support SSL for redis backends would be small. Would you accept a patch for that?

It should probably use a different config name than `broker_use_ssl`. Maybe `backend_use_ssl`?

chriskuehl:
Fixes #3830.

This adds a setting `redis_backend_use_ssl`, identical to `broker_use_ssl`, but used by the backend to determine whether it should use SSL.

It also updates the `broker_use_ssl` documentation which was previously only correct for pyamqp and not redis.

### DIFF
--- a/docs/userguide/configuration.rst
+++ b/docs/userguide/configuration.rst
@@ -878,6 +878,16 @@ The fields of the URL are defined as follows:
     Database number to use. Default is 0.
     The db can include an optional leading slash.
 
+.. setting:: redis_backend_use_ssl
+
+``redis_backend_use_ssl``
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Default: Disabled.
+
+The Redis backend supports SSL. The valid values of this options are the same
+as :setting:`broker_use_ssl`.
+
 .. setting:: redis_max_connections
 
 ``redis_max_connections``
--- a/docs/userguide/configuration.rst
+++ b/docs/userguide/configuration.rst
@@ -1817,7 +1817,7 @@ The setting must be a dict the keys:
 *  ``ssl_cert_reqs`` (required): one of the ``SSLContext.verify_mode`` values:
     * ``ssl.CERT_NONE``
     * ``ssl.CERT_OPTIONAL``
-    * ``ssl.CERT_REQURIED``
+    * ``ssl.CERT_REQUIRED``
 *  ``ssl_ca_certs`` (optional): path to the CA certificate
 *  ``ssl_certfile`` (optional): path to the client certificate
 *  ``ssl_keyfile`` (optional): path to the client key
--- a/docs/userguide/configuration.rst
+++ b/docs/userguide/configuration.rst
@@ -1766,6 +1766,11 @@ Default: Disabled.
 
 Toggles SSL usage on broker connection and SSL settings.
 
+The valid values for this option vary by transport.
+
+``pyamqp``
+__________
+
 If ``True`` the connection will use SSL with default SSL settings.
 If set to a dict, will configure SSL connection according to the specified
 policy. The format used is Python's :func:`ssl.wrap_socket` options.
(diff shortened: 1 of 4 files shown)


chriskuehl:
For context, my patch was mostly based off of what kombu does: https://github.com/celery/kombu/blob/9c1044ef3a55128ed25053bd75981454b4ae8f3c/kombu/transport/redis.py#L887-L894

chriskuehl:
I'm not sure why landscape is saying there were new errors. When I click the link, it says:

    The following messages were introduced in this pull request.
    Hurrah! No new errors were introduced!

ask:
This looks good! 

configcheck says "redis_backend_use_ssl" is not documented, maybe either document it, or add it to the ignore list in docs/conf.py

chriskuehl:
@ask I did add documentation for that config option. I'm not sure why it's not being detected—it does have `.. setting:: redis_backend_use_ssl` right above. Any idea why it's still failing configcheck?

docs/userguide/configuration.rst
```diff
@@ -1793,6 +1808,21 @@ certificate authority:
...
+``redis``
+_________
+
+
+The setting must be a dict the keys:
+
+*  ``ssl_cert_reqs`` (required): one of the ``SSLContext.verify_mode`` values:
+    * ``ssl.CERT_NONE``
+    * ``ssl.CERT_OPTIONAL``
+    * ``ssl.CERT_REQURIED``
```
georgepsarakis: Possible typo here: `CERT_REQUIRED`

chriskuehl:
oops, thanks @thedrow and @georgepsarakis!

thedrow:
I have no idea why configcheck is failing. @georgepsarakis Do you know how does it work?
Should we just add it to the ignored checks?

georgepsarakis:
@thedrow I found the following:

- The Sphinx ConfigCheckBuilder can be found [here](https://github.com/celery/sphinx_celery/blob/master/sphinx_celery/configcheck.py).
- All project settings are recovered from [celery.app.defaults](https://github.com/celery/celery/blob/master/celery/app/defaults.py), this can be seen inf the Sphinx configuration file [here](https://github.com/celery/celery/blob/master/docs/conf.py#L88).
- The documented options are being read from [docs/userguide/configuration.rst](https://github.com/celery/celery/blob/master/docs/userguide/configuration.rst).
- The **symmetric difference** between documented and all project options is computed. This means a warning will be displayed if either of these sets has different elements from the other, excluding of course the ignored options.

@chriskuehl The solution is to modify the Redis namespace in `celery.app.defaults` and include something like:
```
backend_use_ssl=Option(type='bool', default=False)
```

chriskuehl:
Thanks! Trying that now.

chriskuehl:
Looks like that worked -- thanks for the help!

georgepsarakis:
Great :smile: ! @chriskuehl feel free to add yourself to the CONTRIBUTORS list.

JonPeel:
How does / will this work in Celery 5 exactly? With it's absence in 4.0.2 I forked Celery just so I could throw this change in for 4.x... but found ensuring we consistently use an SSL connection pool pretty confusing.

Firstly - if using a URL to configure the back end for redis, the SSL settings are over-written:
https://github.com/celery/celery/blob/master/celery/backends/redis.py#L144

PR for that if wanted: https://github.com/celery/celery/pull/4119

..then I had to set both `redis_backend_use_ssl` set as well as `broker_use_ssl` to ensure the 'ssl' property was still set on the connection info, presumably for kombu: https://github.com/celery/kombu/blob/master/kombu/transport/redis.py#L887

...is that still required in 5.0?

thedrow:
Yes it is still required. The results backend and the broker are completely different things.
You may want to use SSL with RabbitMQ and have your results stored in Redis without SSL since Redis doesn't support it and would require you to use a proxy to terminate the SSL connection.
If you use Redis both as a queue and as a results backend we'd like to provide the flexibility as well.
Your Redis queue cluster may reside on one datacenter and the Redis results backend cluster may reside on a different datacenter, requiring SSL to ensure no one hijacks your data (or vice versa).

JonPeel:
Thanks - PR for updating connparams after parsing URLs already open; that fixed, I think the rest is answered by documentation / your reply.

PR #3834, opened 2017-02-13
Handle ignore_result

gugu:

## Description

Currently (according to my checks and code grep) ignore_result option is a stub, it does nothing. This patch skips backend calls for tasks with `ignore_result=True`

This is needed to minimize effect of broken redis backend support

### DIFF
--- a/celery/app/base.py
+++ b/celery/app/base.py
@@ -690,7 +690,7 @@ class Celery(object):
                   add_to_parent=True, group_id=None, retries=0, chord=None,
                   reply_to=None, time_limit=None, soft_time_limit=None,
                   root_id=None, parent_id=None, route_name=None,
-                  shadow=None, chain=None, task_type=None, **options):
+                  shadow=None, chain=None, task_type=None, ignore_result=False, **options):
         """Send task by name.
 
         Supports the same arguments as :meth:`@-Task.apply_async`.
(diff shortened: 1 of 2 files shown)


gugu:
Anyone here?

auvipy:
yes

celery/app/task.py
```diff
@@ -532,7 +532,7 @@ def apply_async(self, args=None, kwargs=None, task_id=None, producer=None,
         return app.send_task(
             self.name, args, kwargs, task_id=task_id, producer=producer,
             link=link, link_error=link_error, result_cls=self.AsyncResult,
-            shadow=shadow, task_type=self,
+            shadow=shadow, task_type=self, ignore_result=getattr(self, 'ignore_result', False),
```
georgepsarakis: I think this is a `Task` class attribute so you do not need `getattr`. Also, shouldn't this change be applied to `apply`?

celery/app/base.py
```diff
@@ -733,7 +733,8 @@ def send_task(self, name, args=None, kwargs=None, countdown=None,
             producer = amqp.Producer(connection, auto_declare=False)
         with self.producer_or_acquire(producer) as P:
             with P.connection._reraise_as_library_errors():
-                self.backend.on_task_call(P, task_id)
+                if not ignore_result:
+                    self.backend.on_task_call(P, task_id)
```
georgepsarakis: Interesting. So this only affects backends that implement the `on_task_call` method?

jenstroeger:
Any progress on this issue?

PR #3847, opened 2017-02-17
Bugfix: IndexError raised when chord has an empty header

marcgibbons:
Given that a chord has an empty list for a header, an unhandled IndexError is raised as the `_get_app` tries to access element `[0]` of an empty list.

To support / gracefully handle chords without any signatures in the header, check that there are in fact elements in the list before accessing; otherwise, use the body task to find the app.

### DIFF
--- a/t/unit/tasks/test_canvas.py
+++ b/t/unit/tasks/test_canvas.py
@@ -607,6 +607,10 @@ class test_chord(CanvasCase):
         x = chord([t1], body=t2)
         assert x.app is t2._app
 
+    def test_app_when_header_is_empty(self):
+        x = chord([], self.add.s(4, 4))
+        assert x.app is self.add.app
+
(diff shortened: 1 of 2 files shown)


t/unit/tasks/test_canvas.py
```diff
@@ -607,6 +607,10 @@ def test_app_when_app_in_task(self):
         x = chord([t1], body=t2)
         assert x.app is t2._app
 
+    def test_app_when_header_is_empty(self):
+        x = chord([], self.add.s(4, 4))
+        assert x.app is self.add.app
```
georgepsarakis: Indeed the test you added seems to verify the failure case (tried it with `master`).

georgepsarakis:
@marcgibbons thanks for your contribution :smiley: !

marcgibbons:
@georgepsarakis Many thanks!

PR #3850, opened 2017-02-17
fix maybe_make_aware

### LINKED ISSUE #3849: celery.utils.time.maybe_make_aware docstring incorrect
The docstring for [celery.utils.time.maybe_make_aware](https://github.com/celery/celery/blob/master/celery/utils/time.py#L321) claims that it doesn't do anything if dt is already timezone-aware, but this is incorrect, as it will try to convert to the localtime or utc instead.  This doesn't look like correct behavior, and doesn't allow timezones to work correctly in the crontab scheduler.

Taywee:
Fixes #3849

### DIFF
--- a/celery/utils/time.py
+++ b/celery/utils/time.py
@@ -322,9 +322,10 @@ def maybe_make_aware(dt, tz=None):
     """Convert dt to aware datetime, do nothing if dt is already aware."""
     if is_naive(dt):
         dt = to_utc(dt)
-    return localize(
-        dt, timezone.utc if tz is None else timezone.tz_or_local(tz),
-    )
+        return localize(
+            dt, timezone.utc if tz is None else timezone.tz_or_local(tz),
+        )
+    return dt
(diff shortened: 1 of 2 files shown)


georgepsarakis:
Seems accurate. If you don't mind a suggestion, could you please add a test case [here](https://github.com/celery/celery/blob/master/t/unit/utils/test_time.py#L169):

```
from pytz import timezone

utc_now = datetime.utcnow()
dt = maybe_make_aware(utc_now)
assert dt.tzinfo == timezone('UTC')
assert dt.timetuple()[:-1] == utc_now.timetuple()[:-1]

# Does not modify anything if already aware
dt = maybe_make_aware(dt, tz='US/Eastern')
assert dt.tzinfo == timezone('UTC')
```

Taywee:
Sure, I've added those asserts now.

thedrow:
Thanks!

PR #3852, opened 2017-02-17
celery.beat.Scheduler: fix _when to make tz-aware (#3851)

### LINKED ISSUE #3851: celery.beat.Scheduler._when wrong result for timezone-aware datetimes
[celery.beat.Scheduler._when](https://github.com/celery/celery/blob/master/celery/beat.py#L235) reduces into a timetuple ignoring tzinfo.  If you have multiple schedules with different tzinfos, the ones in later-shifted timezones will possibly never run, because they will almost always be at the bottom of the priority queue.

Taywee:
Fixes #3851

### DIFF
--- a/celery/beat.py
+++ b/celery/beat.py
@@ -28,7 +28,7 @@ from .five import (
 )
 from .schedules import maybe_schedule, crontab
 from .utils.imports import load_extension_class_names, symbol_by_name
-from .utils.time import humanize_seconds
+from .utils.time import humanize_seconds, maybe_make_aware
 from .utils.log import get_logger, iter_open_logger_fds
 
 __all__ = [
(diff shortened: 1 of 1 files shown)


thedrow:
It feels like some test coverage is required in order to ensure this won't regress. Can you add a test that verifies that celery beat works with multiple time zones?

thedrow:
Ping @Taywee

Taywee:
I agree that some tests are probably necessary.  I'm not sure how those tests should look, however.  `_when` on its own would be easy enough to test, but unit tests on private methods are usually frowned upon.

thedrow:
You should test `populate_heap()` and verify that the schedule has been adjusted to the timezone.

auvipy:
It seems it is fixed, correct me if I am wrong.

auvipy (changes requested):
please rebase. if you do not have time i will take this over after a week

auvipy (commented):
closing as it seems to be implemented in main branch with other relevant changes. https://github.com/celery/celery/blob/main/celery/beat.py#L299

PR #3861, opened 2017-02-22
Fixed a bug in Python 3.5: Serialization of CouchDB key in set method.

fmind:
Hello,

In Python 3.5, the couchdb backend 4.0.2 has in bug in its set method.

The code works in Python 2.

This is what happens when I use the backend for the "First steps with Celery" tutorial:
TypeError: b'celery-task-meta-c50dad53-29b5-4cd7-9a64-1e6089e71c81' is not JSON serializable

In backends/couchdb.py, the key parameter is a bytes instead of a string. Thus, it cannot be serialize.

Decoding the key with ASCII codec fixed the bug.

Since I don't know the reason to why key is a bytes, please double check that this is indeed the right fix.

### DIFF
--- a/celery/backends/couchdb.py
+++ b/celery/backends/couchdb.py
@@ -82,7 +82,7 @@ class CouchBackend(KeyValueStoreBackend):
             return None
 
     def set(self, key, value):
-        data = {'_id': key, 'value': value}
+        data = {'_id': key.decode('ascii'), 'value': value}
         try:
             self.connection.save(data)
         except pycouchdb.exceptions.Conflict:


celery/backends/couchdb.py
```diff
@@ -82,7 +82,7 @@ def get(self, key):
             return None
 
     def set(self, key, value):
-        data = {'_id': key, 'value': value}
+        data = {'_id': key.decode('ascii'), 'value': value}
```
georgepsarakis: In order to maintain Python2/3 compatibility, I believe it is generally advisable to use the compatibility functions in [five](https://github.com/celery/celery/blob/master/celery/five.py) (which in its turn uses [vine.five](https://github.com/celery/vine/blob/master/vine/five.py)).

So, for consistency perhaps it would be better to use `string`:

```
from celery.five import string

data = {'_id': string(key), 'value': value}
```

georgepsarakis:
@fmind any update on this? If you need further help please feel free to notify me.

thedrow:
Superseded by #4166.

PR #3867, opened 2017-02-27
Fix task ETA issues when timezone is defined in configuration

### LINKED ISSUE #3753: default_retry_delay does not work when enable_utc is False
## Checklist

       version affected).
```
software -> celery:4.0.2 (latentcall) kombu:4.0.2 py:3.4.0
            billiard:3.5.0.2 redis:2.10.5
settings -> transport:redis results:redis://localhost:6379/0

result_backend: 'redis://localhost:6379/0'
broker_url: 'redis://localhost:6379/0'
```

## Steps to reproduce
```
from celery import Celery
app = Celery('vanilla', broker='redis://localhost:6379/0', backend='redis://localhost:6379/0')

app.conf.enable_utc = False

@app.task(bind=True, max_retries=3, default_retry_delay=60)
def Fail(self):
    try:
        raise AssertionError('Oh My!')
    except Exception as e:
        self.retry(exc=e)
```
## Expected behavior
A delay of around 60 seconds should happen between retries.

## Actual behavior
The retry delay does not take any effect. This problem only occurs when enable_utc is set to False, as in the example above. The retry delay is however respected when enable_utc is left at its default value of True.

```
[2017-01-10 03:50:20,724: INFO/MainProcess] Received task: vanilla.Fail[8e3928ed-5a6a-4811-b2b6-c4b25ee51c74]  
[2017-01-10 03:50:20,749: INFO/MainProcess] Received task: vanilla.Fail[8e3928ed-5a6a-4811-b2b6-c4b25ee51c74]  ETA:[2017-01-10 03:51:20.727154+00:00] 
[2017-01-10 03:50:20,752: INFO/PoolWorker-4] Task vanilla.Fail[8e3928ed-5a6a-4811-b2b6-c4b25ee51c74] retry: Retry in 60s: AssertionError('Oh My!',)
[2017-01-10 03:50:20,778: INFO/MainProcess] Received task: vanilla.Fail[8e3928ed-5a6a-4811-b2b6-c4b25ee51c74]  ETA:[2017-01-10 03:51:20.755246+00:00] 
[2017-01-10 03:50:20,779: INFO/PoolWorker-5] Task vanilla.Fail[8e3928ed-5a6a-4811-b2b6-c4b25ee51c74] retry: Retry in 60s: AssertionError('Oh My!',)
[2017-01-10 03:50:20,809: INFO/PoolWorker-2] Task vanilla.Fail[8e3928ed-5a6a-4811-b2b6-c4b25ee51c74] retry: Retry in 60s: AssertionError('Oh My!',)
[2017-01-10 03:50:20,809: INFO/MainProcess] Received task: vanilla.Fail[8e3928ed-5a6a-4811-b2b6-c4b25ee51c74]  ETA:[2017-01-10 03:51:20.783516+00:00] 
[2017-01-10 03:50:20,812: ERROR/PoolWorker-4] Task vanilla.Fail[8e3928ed-5a6a-4811-b2b6-c4b25ee51c74] raised unexpected: AssertionError('Oh My!',)
  File "/auto/firex-ott/moe_4.0.2/lib/python3.4/site-packages/celery/app/trace.py", line 367, in trace_task
  File "/auto/firex-ott/moe_4.0.2/lib/python3.4/site-packages/celery/app/trace.py", line 622, in __protected_call__
    return self.run(*args, **kwargs)
  File "/auto/firex-ott/moe_4.0.2/vanilla.py", line 9, in Fail
    self.retry(exc=e)
  File "/auto/firex-ott/moe_4.0.2/lib/python3.4/site-packages/celery/app/task.py", line 668, in retry
    raise_with_context(exc)
  File "/auto/firex-ott/moe_4.0.2/vanilla.py", line 7, in Fail
    raise AssertionError('Oh My!')
AssertionError: Oh My!
```

georgepsarakis:
## Description

- [x] Modified `App.now` to correctly return the  timezone-aware current date/time, according to the configured timezone or UTC.
- [x] Added unit tests for the changed/added methods.
- [x] Optimize tests by reducing wait time for some test cases due to default polling interval value.

Fixes #3753.

### DIFF
--- a/celery/app/base.py
+++ b/celery/app/base.py
@@ -1233,8 +1233,7 @@ class Celery(object):
         return self.amqp.producer_pool
 
     def uses_utc_timezone(self):
-        """Check if the application uses the UTC timezone.
-        """
+        """Check if the application uses the UTC timezone."""
         return self.conf.timezone == 'UTC' or self.conf.timezone is None
 
     @cached_property
--- a/celery/app/base.py
+++ b/celery/app/base.py
@@ -870,7 +870,8 @@ class Celery(object):
 
     def now(self):
         """Return the current time and date as a datetime."""
-        return self.loader.now(utc=self.conf.enable_utc)
+        from datetime import datetime
+        return datetime.utcnow().replace(tzinfo=self.timezone)
 
     def select_queues(self, queues=None):
         """Select subset of queues.
@@ -1231,6 +1232,11 @@ class Celery(object):
     def producer_pool(self):
         return self.amqp.producer_pool
 
+    def uses_utc_timezone(self):
+        """Check if the application uses the UTC timezone.
+        """
+        return self.conf.timezone == 'UTC' or self.conf.timezone is None
+
     @cached_property
     def timezone(self):
         """Current timezone for this app.
@@ -1239,9 +1245,12 @@ class Celery(object):
         :setting:`timezone` setting.
         """
         conf = self.conf
-        tz = conf.timezone
+        tz = conf.timezone or 'UTC'
         if not tz:
-            return (timezone.get_timezone('UTC') if conf.enable_utc
-                    else timezone.local)
-        return timezone.get_timezone(conf.timezone)
+            if conf.enable_utc:
+                return timezone.get_timezone('UTC')
+            else:
+                if not conf.timezone:
+                    return timezone.local
+        return timezone.get_timezone(tz)
 App = Celery  # noqa: E305 XXX compat
--- a/celery/app/amqp.py
+++ b/celery/app/amqp.py
@@ -21,7 +21,7 @@ from celery.local import try_import
 from celery.utils.nodenames import anon_nodename
 from celery.utils.saferepr import saferepr
 from celery.utils.text import indent as textindent
-from celery.utils.time import maybe_make_aware, to_utc
+from celery.utils.time import maybe_make_aware
 
 from . import routes as _routes
 
--- a/celery/app/amqp.py
+++ b/celery/app/amqp.py
@@ -407,17 +407,11 @@ class AMQP(object):
         if countdown:  # convert countdown to ETA
             self._verify_seconds(countdown, 'countdown')
             now = now or self.app.now()
-            timezone = timezone or self.app.timezone
             eta = now + timedelta(seconds=countdown)
-            if utc:
-                eta = to_utc(eta).astimezone(timezone)
         if isinstance(expires, numbers.Real):
             self._verify_seconds(expires, 'expires')
             now = now or self.app.now()
-            timezone = timezone or self.app.timezone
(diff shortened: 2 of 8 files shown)


celery/app/base.py
```diff
@@ -870,7 +870,8 @@ def prepare_config(self, c):
 
     def now(self):
         """Return the current time and date as a datetime."""
-        return self.loader.now(utc=self.conf.enable_utc)
+        from datetime import datetime
+        return datetime.utcnow().replace(tzinfo=self.timezone)
```
Chronial: I think this is wrong. This will not return the current time in `self.timezone`. Let's say for example that `self.timezone` is +8, at it is currently `12:00 UTC`. This method will return `12:00 +8`, which is the same point in time as `04:00 UTC` – not `12:00 UTC`.
Note that this current implementation might even return nonsensical times, like `2017-03-25 02:30` for a Europe/Berlin timezone – a time that doesn't exist.

The correct way to get the current time in a local timezone is `self.timezone.localize(datetime.utcnow().replace(tzinfo=pytz.utc))`. See the [same functionality in django](https://github.com/django/django/blob/master/django/utils/timezone.py#L191) and the [pytz docs](http://pytz.sourceforge.net/).

But you can't just make a function that used to return utc times return times from timezones that observe DST. If the caller code is not aware of that, things might break. Example: You return `2017-03-24 03:30 Europe/Berlin`. The caller adds 24h to this and then checks if the time is before `04:00`. Adding 24h crosses a DST boundary, so the above +24h is actually `2017-03-25 04:30`. But if you add 24h to a `2017-03-24 03:30 Europe/Berlin` datetime object, you actually get `2017-03-25 03:30 Europe/Berlin [NON-DST]`. You need to call `timezone.localize` on that to turn that into a sensible datetime object – see the [pytz docs](http://pytz.sourceforge.net/#localized-times-and-date-arithmetic).

Hope this helps.

auvipy: @Chronial would you mind sending a pr with your suggestions?

Chronial: Providing a correct implementation using timezones is not easy. Since I moved away from using celery (its broken timzone handling beeing one of the reasons ^^), I unfortunately do not have the time to provide a full PR for this.

Just wanted to report this bug, since I stumbled upon it while browsing.

georgepsarakis: @Chronial you are right, this has already been corrected by other contributors in [master/latest release](https://github.com/celery/celery/blob/master/celery/app/base.py#L885).

Chronial: Ah, sorry for digging up this old PR. I came here via the mention on #4006 and didn't realize how old this PR was.

georgepsarakis: No problem @Chronial thanks for the feedback.

PR #3842, opened 2017-02-15
Docs: Use a wording a bit closer to English

bittner:
The sentence to explain the first code sample is a bit bumpy. This change should make it a bit easier to digest, hopefully.

PR #3882, opened 2017-03-02
Add Peter Bittner to authors

bittner:
As suggested in #3842 by @thedrow


bittner:
Ooops, wrong file! I'll wait until #3881 is merged, then I retry with the right one!  :smile:

PR #3887, opened 2017-03-03
Fix grammar, update .gitignore

staticfox:
## Description
Just a couple things I noticed :) Fixed a grammar error that I found here http://docs.celeryproject.org/en/latest/django/first-steps-with-django.html which lead to finding another typo.
Also noticed `.eggs/` wasn't in `.gitignore` so I added it. (It was generated when I ran `make test`, running version `Python 3.6.0 (default, Jan 16 2017, 12:12:55) `

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@ dist/
 *.egg-info
 *.egg
 *.egg/
+*.eggs/
 build/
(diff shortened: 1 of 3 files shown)


thedrow:

PR #3791, opened 2017-01-26
Populate heap when periodic tasks are changed

wzywno:
## Description

Checks if the heap is valid. If there are changes in scheduled tasks the heap is populated again.

Fixes [7](https://github.com/celery/django-celery-beat/issues/7)

PR #3890, opened 2017-03-03
Populate heap when periodic tasks are changed - fixed PEP8

gitaarik:
I improved @wzywno's pullrequest:
https://github.com/celery/celery/pull/3791

I fixed the PEP8 issues, hope it can be merged soon.

I also squashed the commits together in one, so there's a clean history.

### DIFF
--- a/celery/beat.py
+++ b/celery/beat.py
@@ -197,6 +197,7 @@ class Scheduler(object):
                              self.max_interval)
         self.Producer = Producer or app.amqp.Producer
         self._heap = None
+        self.old_schedulers = None
         self.sync_every_tasks = (
             app.conf.beat_sync_every if sync_every_tasks is None
             else sync_every_tasks)
@@ -257,7 +258,9 @@ class Scheduler(object):
         adjust = self.adjust
         max_interval = self.max_interval
 
-        if self._heap is None:
+        if (self._heap is None or
+                not self.schedules_equal(self.old_schedulers, self.schedule)):
+            self.old_schedulers = self.schedule
             self.populate_heap()
 
         H = self._heap
@@ -280,6 +283,18 @@ class Scheduler(object):
                 heappush(H, verify)
                 return min(verify[0], max_interval)
         return min(adjust(next_time_to_run) or max_interval, max_interval)
+
+    def schedules_equal(self, a, b):
+        if a.keys() != b.keys():
+            return False
+        for name, model in a.items():
+            b_model = b.get(name)
+            if not b_model:
+                return False
+            if (hasattr(model.schedule, '__repr__') and
+                    model.schedule.__repr__() != b_model.schedule.__repr__()):
+                return False
+        return True
 
     def should_sync(self):
         return (


AJRenold:
Question on this solution - it seems a bit inefficient to check if the schedule has changed every time `tick` is called. Why not add a method which is called whenever the schedule changes? I've only worked with the [django-celery-beat ](https://github.com/celery/django-celery-beat) beat command, but that is the change which I made to resolve this issue - https://github.com/celery/django-celery-beat/blob/master/django_celery_beat/schedulers.py#L289 - setting Schedule._heap to None was not the cleanest way to do this because _heap is 'private' to the parent class, but it was the only way I saw to invalidate the heap when the schedule was updated

brianmay:
This doesn't seem to resolve the problem for me.

brianmay:
My initial test was flawed, this PR does solve celery/django-celery-beat#7.

brianmay:
@AJRenold I have tried the master branch, which includes your change - I checked to make sure, and it doesn't resolve the issue.

AJRenold:
Hey, interesting! Do you think we can write a failing test on django-celery-beat? What exactly are you changing in your PeriodicTask object, is it one of IntervalSchedule, CrontabSchedule or SolarSchedule?

brianmay:
I think I was changing the PeriodicTask by pointing it to a different IntervalSchedule.

Yes, new test cases is always a good idea.

celery/beat.py
```diff
@@ -257,7 +258,9 @@ def tick(self, event_t=event_t, min=min, heappop=heapq.heappop,
         adjust = self.adjust
         max_interval = self.max_interval
 
-        if self._heap is None:
+        if (self._heap is None or
+                not self.schedules_equal(self.old_schedulers, self.schedule)):
```
georgepsarakis: The [schedule](https://github.com/celery/celery/blob/master/celery/schedules.py#L171) class has defined the `_eq__` method, so I think you can compare directly the dictionaries `self.old_schedulers` and `self.schedule`, you should not need an additional function to achieve this.

celery/beat.py
```diff
@@ -257,7 +258,9 @@ def tick(self, event_t=event_t, min=min, heappop=heapq.heappop,
         adjust = self.adjust
         max_interval = self.max_interval
 
-        if self._heap is None:
+        if (self._heap is None or
```
georgepsarakis: Could you please add a test case verifying the flow for this logical branch?

gitaarik:
To be honest I don't know about the logic in these changes, I just made it PEP8. @wzywno, do you care to take a closer look?

brianmay:
Just noticed this change only works if you change the periodic task. If you change the interval object, the new value is not reloaded. Will retest again against the master version, make sure this wasn't confusing me.

I also have a branch based on this one that uses ``!=`` instead of ``schedules_equal``, which appears to work fine.

brianmay:
Ok, reconfirmed that master of django-celery-beat by itself does not fix this.

celery/beat.py
brianmay: See https://github.com/celery/django-celery-beat/issues/7#issuecomment-267451710 for why this apparently ugly code is required.

brianmay:
Just deleted a comment I just made because it was wrong/misleading.

I created a pull request to address the feedback here, but found that it broke. As documented here https://github.com/celery/django-celery-beat/issues/7#issuecomment-267451710

With the simplified code, changes the the interval table are not detected.

celery/beat.py
```diff
@@ -281,6 +284,18 @@ def tick(self, event_t=event_t, min=min, heappop=heapq.heappop,
                 return min(verify[0], max_interval)
         return min(adjust(next_time_to_run) or max_interval, max_interval)
 
+    def schedules_equal(self, a, b):
+        if a.keys() != b.keys():
```
brianmay: Yes, that does appear wrong. Will try to understand how this code works.

brianmay: ``set(a.keys()) != set(b.keys())`` works for me.

celery/beat.py
```diff
@@ -281,6 +284,18 @@ def tick(self, event_t=event_t, min=min, heappop=heapq.heappop,
...
+            return False
+        for name, model in a.items():
+            b_model = b.get(name)
+            if not b_model:
+                return False
+            if (hasattr(model.schedule, '__repr__') and
+                    model.schedule.__repr__() != b_model.schedule.__repr__()):
```
brianmay: I believe this if comparison can be simplified to:

```
if model.schedule != b_model.schedule:
```

brianmay:
celery/celery#3958 works for me.

celery/beat.py
brianmay: `model` and `b_model` are of type `django_celery_beat.schedulers.ModelEntry`.

So this code is django_celery_beat specific. Not convinced this is appropriate code for celery :-(

celery/beat.py
brianmay: These `self.old_schedulers` and `self.schedule` are of type `dict`, so the `==` or `!=` is only a shallow comparison.

celery/beat.py
brianmay: Oh, wait, my objection to my own pull request might be unfounded.

`django_celery_beat.schedulers.ModelEntry` is actually inherited from `celery.beat ScheduleEntry` - so I am guessing is actually a standard celery object with some extra bits added.

subha-py:
Can we have a working example and celery beat/ celery link? I am not able to run scheduled tasks with django 1.11

brianmay:
@subha-py Django 1.11 support works for me... Maybe open another bug report with full details?

brianmay:
I think this pull request can now be closed, it was merged in #3958.

PR #3891, opened 2017-03-04
Fixes LimitedSets by eliminating hashing ambiguity; refs #3879

caronc:
## Description
I presume this bug is already fixed up stream but can't test it since I'm not using the 4.x or 5.x branch in production. __LimitedSet()__ appears to have ambiguity issues on faster systems that are very easy to reproduce.  You just need to run the test suite.

This patch just ensures that all entries are unique and sequentially stored based on time by adding 1 micro-second to the key (time) entry before storing it in the heap if it shares the same value as the previously inserted one.

This issue Fixes #3879

PR #3892, opened 2017-03-04
LimitedSets support on faster systems refs #3879 and #3891

### LINKED ISSUE #3879: LimitedSet does not perform as expected - tests fail
## Checklist

- Celery Version v3.1.25
- [ X] I have verified that the issue exists against the `master` branch of Celery.

## Steps to reproduce
Run unit tests

## Expected behavior
That they'd pass :wink:

## Actual behavior
Why they all pass except the ones related to LimitedSet

The results appear to be the absolute invert of what is expected; have a look; here is the output:
```
__________________________________________________ test_LimitedSet.test_add ___________________________________________________

self = <celery.tests.utils.test_datastructures.test_LimitedSet testMethod=test_add>

    def test_add(self):
        if sys.platform == 'win32':
            raise SkipTest('Not working properly on Windows')
        s = LimitedSet(maxlen=2)
        s.add('foo')
        s.add('bar')
        for n in 'foo', 'bar':
            self.assertIn(n, s)
        s.add('baz')
        for n in 'bar', 'baz':
>           self.assertIn(n, s)
E           AssertionError: 'bar' not found in LimitedSet(2)

celery/tests/utils/test_datastructures.py:185: AssertionError
__________________________________________________ test_LimitedSet.test_iter __________________________________________________

self = <celery.tests.utils.test_datastructures.test_LimitedSet testMethod=test_iter>

    def test_iter(self):
        if sys.platform == 'win32':
            raise SkipTest('Not working on Windows')
        s = LimitedSet(maxlen=3)
        items = ['foo', 'bar', 'baz', 'xaz']
        for item in items:
            s.add(item)
        l = list(iter(s))
        for item in items[1:]:
>           self.assertIn(item, l)
E           AssertionError: 'bar' not found in ['baz', 'foo', 'xaz']

celery/tests/utils/test_datastructures.py:242: AssertionError
_________________________________________________ test_LimitedSet.test_update _________________________________________________

self = <celery.tests.utils.test_datastructures.test_LimitedSet testMethod=test_update>

    def test_update(self):
        s1 = LimitedSet(maxlen=2)
        s1.add('foo')
        s1.add('bar')
    
        s2 = LimitedSet(maxlen=2)
        s2.update(s1)
        self.assertItemsEqual(list(s2), ['foo', 'bar'])
    
        s2.update(['bla'])
>       self.assertItemsEqual(list(s2), ['bla', 'bar'])

celery/tests/utils/test_datastructures.py:280: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
celery/tests/case.py:367: in assertItemsEqual
    return self.assertSequenceEqual(expected, actual, msg=msg)
E   AssertionError: Sequences differ: ['bla', 'foo'] != ['bar', 'bla']
E   
E   First differing element 0:
E   bla
E   bar
E   
E   - ['bla', 'foo']
E   + ['bar', 'bla']
```
Just to give you more details on my environment (in case it's related):

* I'm using CentOS 7.x putting my on Python 2.7.5
* billiard v3.3.0.20
* kombu v3.0.37
* pyparsing v1.5.6
* pytest v3.0.6
* dateutil v2.2-1
* anyjson v0.3.3
* amqp v2.1.4
* pytz v2016.10
* uuid v1.30

caronc:
## Description
 __LimitedSet()__ appears to have ambiguity issues on faster systems that are very easy to reproduce.  You just need to run the test suite. I originally just patched the version from the branch I'm using (see #3891) This patch is just for the master as well.

This patch just ensures that all entries are unique and sequentially stored based on time by adding 1 micro-second to the key (time) entry before storing it in the heap if it shares the same value as the previously inserted one.

This issue Fixes #3879

### DIFF
--- a/celery/utils/collections.py
+++ b/celery/utils/collections.py
@@ -575,12 +575,14 @@ class LimitedSet(object):
     def add(self, item, now=None):
         # type: (Any, float) -> None
         """Add a new item, or reset the expiry time of an existing item."""
-        now = now or time.time()
         if item in self._data:
             self.discard(item)
-        if now <= self.last_added:
-            # Force uniqueness
-            now = self.last_added + 1e-6
+        if now is None:
+            # Get time
+            now = time.time()
+            if now <= self.last_added:
+                # Force uniqueness (because we're not a call from update())
+                now = self.last_added + 1e-6
         entry = (now, item)
         self._data[item] = entry
         heappush(self._heap, entry)
--- a/celery/utils/collections.py
+++ b/celery/utils/collections.py
@@ -576,12 +576,6 @@ class LimitedSet(object):
         now = now or monotonic()
         if item in self._data:
             self.discard(item)
-        if now is None:
-            # Get time
-            now = monotonic()
-            if now <= self.last_added:
-                # Force uniqueness (because we're not a call from update())
-                now = self.last_added + 1e-6
         entry = (now, item)
         self._data[item] = entry
         heappush(self._heap, entry)
--- a/celery/utils/collections.py
+++ b/celery/utils/collections.py
@@ -526,7 +526,7 @@ class LimitedSet(object):
         False
         >>> len(s)  # maxlen is reached
         50000
-        >>> s.purge(now=time.time() + 7200)  # clock + 2 hours
+        >>> s.purge(now=monotonic() + 7200)  # clock + 2 hours
         >>> len(s)  # now only minlen items are cached
         4000
         >>>> 57000 in s  # even this item is gone now
--- a/celery/utils/collections.py
+++ b/celery/utils/collections.py
@@ -542,6 +542,8 @@ class LimitedSet(object):
         self.expires = 0 if expires is None else expires
         self._data = {}
         self._heap = []
+        # Ensures inserts do not become ambiguous
+        self.last_added = 0.0
 
         if data:
             # import items from data
@@ -576,9 +578,14 @@ class LimitedSet(object):
         now = now or time.time()
         if item in self._data:
             self.discard(item)
+        if inserted <= self.last_added:
+            # Force uniqueness
+            inserted = self.last_added + 1e-6
         entry = (now, item)
         self._data[item] = entry
         heappush(self._heap, entry)
+        # Update our last-inserted time for future reference
+        self.last_added = inserted
         if self.maxlen and len(self._data) >= self.maxlen:
             self.purge()
 
--- a/celery/utils/collections.py
+++ b/celery/utils/collections.py
@@ -3,7 +3,7 @@
 from __future__ import absolute_import, unicode_literals
 
 import sys
-import time
+from celery.five import monotonic
 
 from collections import (
     Callable, Mapping, MutableMapping, MutableSet, Sequence,
@@ -579,7 +579,7 @@ class LimitedSet(object):
             self.discard(item)
         if now is None:
             # Get time
-            now = time.time()
+            now = monotonic()
             if now <= self.last_added:
                 # Force uniqueness (because we're not a call from update())
                 now = self.last_added + 1e-6
@@ -633,7 +633,7 @@ class LimitedSet(object):
             now (float): Time of purging -- by default right now.
                 This can be useful for unit testing.
         """
-        now = now or time.time()
+        now = now or monotonic()
         now = now() if isinstance(now, Callable) else now
         if self.maxlen:
             while len(self._data) > self.maxlen:
--- a/celery/utils/collections.py
+++ b/celery/utils/collections.py
@@ -578,14 +578,14 @@ class LimitedSet(object):
         now = now or time.time()
         if item in self._data:
             self.discard(item)
-        if inserted <= self.last_added:
+        if now <= self.last_added:
             # Force uniqueness
-            inserted = self.last_added + 1e-6
+            now = self.last_added + 1e-6
         entry = (now, item)
         self._data[item] = entry
         heappush(self._heap, entry)
         # Update our last-inserted time for future reference
-        self.last_added = inserted
+        self.last_added = now
         if self.maxlen and len(self._data) >= self.maxlen:
             self.purge()
 
--- a/t/unit/utils/test_collections.py
+++ b/t/unit/utils/test_collections.py
@@ -5,7 +5,7 @@ import pytest
 
 from collections import Mapping
 from itertools import count
-from time import time
+from celery.five import monotonic
 
 from case import skip
 from billiard.einfo import ExceptionInfo
@@ -198,21 +198,21 @@ class test_LimitedSet:
         s = LimitedSet(maxlen=10, expires=1)
         [s.add(i) for i in range(10)]
         s.maxlen = 2
-        s.purge(now=time() + 100)
+        s.purge(now=monotonic() + 100)
         assert len(s) == 0
 
         # not expired
         s = LimitedSet(maxlen=None, expires=1)
         [s.add(i) for i in range(10)]
         s.maxlen = 2
-        s.purge(now=lambda: time() - 100)
+        s.purge(now=lambda: monotonic() - 100)
         assert len(s) == 2
 
         # expired -> minsize
         s = LimitedSet(maxlen=10, minlen=10, expires=1)
         [s.add(i) for i in range(20)]
         s.minlen = 3
-        s.purge(now=time() + 3)
+        s.purge(now=monotonic() + 3)
         assert s.minlen == len(s)
         assert len(s._heap) <= s.maxlen * (
             100. + s.max_heap_percent_overload) / 100
@@ -293,7 +293,7 @@ class test_LimitedSet:
 
     def test_iterable_and_ordering(self):
         s = LimitedSet(maxlen=35, expires=None)
-        # we use a custom clock here, as time.time() does not have enough
+        # we use a custom clock here, as monotonic() does not have enough
         # precision when called quickly (can return the same value twice).
         clock = count(1)
         for i in reversed(range(15)):
--- a/t/unit/utils/test_collections.py
+++ b/t/unit/utils/test_collections.py
@@ -293,8 +293,6 @@ class test_LimitedSet:
 
     def test_iterable_and_ordering(self):
         s = LimitedSet(maxlen=35, expires=None)
-        # we use a custom clock here, as monotonic() does not have enough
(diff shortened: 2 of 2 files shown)


thedrow:
@caronc Will you be able to provide a unit test proving this behaviour works as expected?

caronc:
My problem was that your current test suite was failing for me, and this patch allowed it to pass.  [Here is the synopsis of the issue](https://github.com/celery/celery/issues/3879) (originally fixed on the version i use - 3.1).

Basically LimitedSet objects store newly added entries using time() which doesn't have a very good resolution (especially on faster systems). Entries are stored up to their mili-second when processors these days can handle stuff in the nano-seconds.  It's possible for 2 actions to take place on the same millisecond (thus skewing the insert and pop order later on); Consider this piece of code:
```python
#!/usr/bin/env python
# -*- encoding: utf-8 -*-

from time import time as now

if __name__ == "__main__":
    for x in range(1, 1000):
        a = now()
        b = now()
        if a == b:
            print 'dupe'

    exit(0)
```
In a perfect world, if you ran the code above, you should 'never' see the word _dupe_.  But the problem is: you will see the word, and you'll see it several times; At least I do.  Systems will vary based on their performance.  Your test servers you guys are using are certainly so overworked, your bug isn't exploited; but it's there! :wink:

So the added check just guarantee's a unique resolution... The fix itself is: if for whatever reason 2 entries get entered in the hash table with the exact same time, the last one is just bumped into the future by one micro second (the smallest possible increment using the current choice of time resolution).

**TL;DR:** I guess what I'm getting at is:  the tests you have in place are all you need to test this change (no new test is needed).

thedrow:
With the code snippet you provided I can reproduce the problem. Well done :)

I'm not 100% certain that we need all this branching.
Python 3 has introduced a [monotonic clock](https://docs.python.org/3/library/time.html#time.monotonic) which doesn't reproduce the same issue with your provided test case (I extended the loop iterations to 10 million just to ensure).
It has a [backport](https://pypi.python.org/pypi/monotonic) to Python 2 which we can use.
Do you think we can use it to resolve the same issue?

caronc:
I like your idea better TBH. So maybe this fix just requires updating the dependencies and changing the __time.time__ reference to be __time.monotonic__ instead?

Are you using the time field at all at the end of the day? I mean, if not, another (more light weight) route would be to just create a generator function that returns _yield n + 1_? Or no? Thoughts?

thedrow:
We need a time function because of https://github.com/caronc/celery/blob/e2d2c328175bec5330aaf4292c2426392b44fc17/celery/utils/collections.py#L645
Note that "The reference point of the returned value is undefined, so that only the difference between the results of consecutive calls is valid." so we need to initialize a value to diff from in the LimitedSet constructor.

caronc:
Gotcha... that makes sense!  I think your idea is the best solution then. Sometime his weekend (if you don't mind waiting), i'll haul in the _time.monotonic()_ backport and just do some testing on my end (as i'm using Python v2.7). I'll update this pull request when i'm happy for your feedback.  I'll make sure it's compatible with Python v3.x (i may need to wrap the import in a try/catch).

Your method is safer since on a really busy system, my approach could potentially continue to skew the actual insert time significantly if there was never a break in processing.

celery/utils/collections.py
```diff
@@ -3,7 +3,13 @@
 from __future__ import absolute_import, unicode_literals
 
 import sys
-import time
+
+try:
```
georgepsarakis: Implementation and Python2/3 compatibility for the monotonic time function is already included in https://github.com/celery/vine/blob/1.0/vine/five.py#L63 , so perhaps it would be better to reuse this?

caronc: Couldn't agree more! 😄 - standby

caronc:
Looks like vine.five.monotonic and monotonic (separately maintained python package) just don't work with some types of systems (same unit-test failures in both cases).  Unless I'm overlooking something here? ☹️

requirements/default.txt
```diff
@@ -1,3 +1,4 @@
 pytz>dev
 billiard>=3.5.0.2,<3.6.0
 kombu>=4.0.2,<5.0
+monotonic>=1.3
```
thedrow: This dependency is not needed in Python 3.x.

celery/utils/collections.py
```diff
@@ -573,9 +573,15 @@ def clear(self):
...
         # type: (Any, float) -> None
         """Add a new item, or reset the expiry time of an existing item."""
-        now = now or time.time()
+        now = now or monotonic()
         if item in self._data:
             self.discard(item)
+        if now is None:
+            # Get time
+            now = monotonic()
+            if now <= self.last_added:
```
thedrow: You don't need this code anymore.

celery/utils/collections.py
```diff
@@ -573,9 +573,15 @@ def clear(self):
     def add(self, item, now=None):
         # type: (Any, float) -> None
         """Add a new item, or reset the expiry time of an existing item."""
-        now = now or time.time()
+        now = now or monotonic()
```
thedrow: `now` should be equal to the diff between the time the LimitedSet was initialized and the current value.

georgepsarakis: @thedrow why do you need it to be relative?

caronc:
There is something unique about the **MATRIX_TOXENV** set to _unit that_ just doesn't support the monotonic() type.

I addressed the missed code clean-up i had forgotten (sorry about that - had some git troubles). Perhaps i should set it back to the original pull request?

georgepsarakis:
@caronc some of the [test cases](https://github.com/celery/celery/blob/29df527147514b3eab7fbc3dbf41ff181bf1dcd9/t/unit/utils/test_collections.py#L201) for `LimitedSet.purge` still pass the result of `time.time()`. This makes me wonder if there are cases inside the codebase where `purge` is called using this type of value. @caronc @thedrow we need to check how the `LimitedSet` methods that expect `now` as a parameter are being called in the codebase.

caronc:
maybe purge() should only take a integer parameter which is automatically applied to the monotonic() clock? I can't see a use case where you'd want to specify a future time and then apply math to it.  You're always going to be deleting things relative to ..'now' (+/- the accepted variable as input).

This might significantly break things for people who've adopted this class otherwise though; so i'm going out on a limb with this suggestion.

caronc:
@georgepsarakis thanks for pointing that out! I searched through for other references to **LimitedSet()** in the code; and although it is referenced, it doesn't specifically go into setting the **now=** value or exclusively using the purge function.

Tests updated and now we're back to all green lights which is a good thing 😄

celery/utils/collections.py
thedrow: From the documentation:
> Return the value (in fractional seconds) of a monotonic clock, i.e. a clock that cannot go backwards. The clock is not affected by system clock updates. The reference point of the returned value is undefined, so that only the difference between the results of consecutive calls is valid.

t/unit/utils/test_collections.py
```diff
@@ -293,7 +293,7 @@ def test_update(self):
 
     def test_iterable_and_ordering(self):
         s = LimitedSet(maxlen=35, expires=None)
-        # we use a custom clock here, as time.time() does not have enough
+        # we use a custom clock here, as monotonic() does not have enough
```
thedrow: @caronc Have you read the comment about this test?
This is no longer the case with the monotonic clock.
We probably need to adjust this test to use the real clock.

caronc: Almost seems like this test isn't necessary at all any more at all.  You can't get any closer to the real time other than monotonic().  I could sub in the **s.add(i, now=next(clock))** to just **s.add(i, now=monotonic())** ([reference](https://github.com/celery/celery/blob/2394e738a7d0c0c736f5d689de4d32325ba54f48/t/unit/utils/test_collections.py#L298-L300)), but that would just be re-iterating what is under the hood anyway.

Or are you suggesting that if we detect two (specified) times that are exactly the same, we add **1e9** to it (instead of **1e6**)? It would be a bit like merging the original pull request with this one.  This test function you're referencing here could then pass in **time()**? Might be overkill at this point though; thoughts?

thedrow:
@caronc Are you able to address the code review?

t/unit/utils/test_collections.py
thedrow: So after re-reading the test I think it is necessary to ensure ordering.
You can just remove the comment.

caronc:
Before you commit this, i just found another problem. The below code works perfect:
```python
#!/usr/bin/env python
# -*- encoding: utf-8 -*-

# Python v2.7
from monotonic import monotonic

if __name__ == "__main__":
    for x in range(1, 1000):
        a = now()
        b = now()
        if a == b:
            print 'dupe'
```

However the below code fails with tons of dupes:
```python
#!/usr/bin/env python
# -*- encoding: utf-8 -*-

from celery.five import monotonic as now

if __name__ == "__main__":
    for x in range(1, 1000):
        a = now()
        b = now()
        if a == b:
            print 'dupe'
```

caronc:
I ran into a new issue that isn't part of celery but worth sharing on this pull request in case anyone else is monitoring it.  Monotonic functionality is currently broken in Red Hat 7.x (& CentOS 7.x) distributions (at the glib library level).  Since Oracle Linux 7.x is just copy/paste the the same OS, I have to presume it's broken for them too.  I created [this bug report](https://bugzilla.redhat.com/show_bug.cgi?id=1499480) with some sample C code on how to reproduce the problem.  Hopefully it will get fixed soon (as this new pull request is still broken for the likes of me).

t/unit/utils/test_collections.py
caronc: done!

celery/utils/collections.py
thedrow: We need to address this before merging

georgepsarakis: @thedrow isn't generating a unique value here the requirement? `monotonic` achieves just that.

georgepsarakis:
@caronc I tested your code sample (Ubuntu 14.04), running 1 million iterations (with `xrange`) and no duplicate was found.

thedrow:
Thanks! Sorry it took so long to merge this :)

PR #3855, opened 2017-02-20
bugfix

Acey9:
In class ElasticsearchBackend
The function set(self, key) assign the document id to the "key”, But the function _index does not set the document id to the "key".
in the function get(self, key) assign the document id is the "key”, because the backend no document id is the "key" ,so I can not get the task result

PR #3924, opened 2017-03-19
elasticsearch: Fix serializing keys

staticfox:
## Description
elasticsearch requires keys to be a string, not bytes, so make sure we convert the key to a string. Otherwise, the elasticsearch library raises a TypeError:

```
SerializationError({b'celery-task-meta-1ce4d5c5-8efc-49a8-80fe-7d9ae5e94706': '{"status": "SUCCESS", "result": null, "traceback": null, "children": [], "task_id": "1ce4d5c5-8efc-49a8-80fe-7d9ae5e94706"}', '@timestamp': '2017-03-19T20:26:03.078Z'}, TypeError('keys must be a string',)):
  File "/home/static/flask/.venv/lib/python3.6/site-packages/elasticsearch/serializer.py", line 48, in 
dumps
    return json.dumps(data, default=self.default, ensure_ascii=False)
  File "/usr/lib64/python3.6/json/__init__.py", line 238, in dumps
    **kw).encode(obj)
  File "/usr/lib64/python3.6/json/encoder.py", line 199, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/usr/lib64/python3.6/json/encoder.py", line 257, in iterencode
    return _iterencode(o, 0)
TypeError: keys must be a string
```

As an aside, this looks like the same issue as https://github.com/celery/celery/pull/3861 so I went ahead and made sure to use your compat layer's `string()` function.

Also made sure this works on python2.7 as well:

```
[04:38:14 PM] static@homefox# python2.7
Python 2.7.13 (default, Dec 21 2016, 07:16:46) 
[GCC 6.2.1 20160830] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> body = {b'a': {'some1': 1, 'some2': 2}}
>>> dict((k.decode('ascii'), v) for k, v in body.items())
{u'a': {'some1': 1, 'some2': 2}}
>>> 
```

### DIFF
--- a/celery/backends/elasticsearch.py
+++ b/celery/backends/elasticsearch.py
@@ -105,6 +105,7 @@ class ElasticsearchBackend(KeyValueStoreBackend):
             self._index(key, data, refresh=True)
 
     def _index(self, id, body, **kwargs):
+        body = dict((string(k), v) for k, v in body.items())
         return self.server.index(
             id=string(id),
             index=self.index,
--- a/celery/backends/elasticsearch.py
+++ b/celery/backends/elasticsearch.py
@@ -3,8 +3,9 @@
 from __future__ import absolute_import, unicode_literals
 from datetime import datetime
 from kombu.utils.url import _parse_url
+from kombu.utils.encoding import bytes_to_str
 from celery.exceptions import ImproperlyConfigured
-from celery.five import string, items
+from celery.five import items
 from .base import KeyValueStoreBackend
 try:
     import elasticsearch
@@ -105,9 +106,9 @@ class ElasticsearchBackend(KeyValueStoreBackend):
             self._index(key, data, refresh=True)
 
     def _index(self, id, body, **kwargs):
-        body = {string(k): v for k, v in items(body)}
+        body = {bytes_to_str(k): v for k, v in items(body)}
         return self.server.index(
-            id=string(id),
+            id=bytes_to_str(id),
             index=self.index,
             doc_type=self.doc_type,
             body=body,
--- a/celery/backends/elasticsearch.py
+++ b/celery/backends/elasticsearch.py
@@ -105,7 +105,7 @@ class ElasticsearchBackend(KeyValueStoreBackend):
             self._index(key, data, refresh=True)
 
     def _index(self, id, body, **kwargs):
-        body = dict((string(k), v) for k, v in body.items())
+        body = {string(k): v for k, v in five.items(body)}
         return self.server.index(
             id=string(id),
             index=self.index,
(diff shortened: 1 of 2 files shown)


celery/backends/elasticsearch.py
```diff
@@ -88,6 +89,7 @@ def set(self, key, value):
             self._index(key, data, refresh=True)
 
     def _index(self, id, body, **kwargs):
+        body = dict((string(k), v) for k, v in body.items())
```
georgepsarakis: This line can be simplified by using a dictionary comprehension (Celery 4 requires Python 2.7 and above). Also, you can avoid creating a new list in Python 2 by using the compatibility function `items` from `five`:

```
from celery.five import items

body = {string(key): value for key, value in items(body)}
```

What do you think?

celery/backends/elasticsearch.py
```diff
@@ -88,6 +89,7 @@ def set(self, key, value):
...
         return self.server.index(
```
georgepsarakis: Totally unrelated but since you are using this backend perhaps you can help. Why isn't the `id` argument value passed to `index` here? Shouldn't this be:

```
return self.server.index(
    id=id,
    index=self.index,
    doc_type=self.doc_type,
    body=body,
    **kwargs
)
```

celery/backends/elasticsearch.py
staticfox: Sure, that's fine

celery/backends/elasticsearch.py
georgepsarakis: I think this is resolved in #3855.

thedrow: Yes, that is resolved there.

georgepsarakis:
@staticfox there were some changes merged for this particular backend, so please rebase when possible.

celery/backends/elasticsearch.py
thedrow: Also, why isn't this a dictionary comprehension? We don't support Python 2.6 anymore.

thedrow:
@georgepsarakis I fixed the minor styling problems myself. Is this ready for merge or do we need tests as well?

thedrow:
crap. something went really wrong here.

thedrow:
Ok so it turns out `five.string` doesn't encode bytes to unicode on Python 3.
Do we want to fix this on vine or merge this as is?

georgepsarakis:
@thedrow I added some changes. Tell me what you think.

thedrow:
Great! Another bug solved :)

PR #3940, opened 2017-03-28
fix serializing keys  and supports elasticsearch backend options settings

Acey9:
1. 

fix the build errors of [#3855](https://github.com/celery/celery/pull/3855) 

2. supports elasticsearch backend options settings
such as:

elasticsearch_retry_on_timeout
Default: False
Should timeout trigger a retry on different node?

elasticsearch_max_retries
Default: 3.
Maximum number of retries before an exception is propagated.

elasticsearch_timeout
Default: 10.0 seconds.
Global timeout,used by the elasticsearch result backend.

Can you merge both?

### DIFF
--- a/celery/backends/elasticsearch.py
+++ b/celery/backends/elasticsearch.py
@@ -32,8 +32,8 @@ class ElasticsearchBackend(KeyValueStoreBackend):
     scheme = 'http'
     host = 'localhost'
     port = 9200
-    es_retry_on_timeout=False
-    es_timeout=10
+    es_retry_on_timeout = False
+    es_timeout = 10
     es_max_retries = 3
 
     def __init__(self, url=None, *args, **kwargs):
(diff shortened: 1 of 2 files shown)


celery/backends/elasticsearch.py
```diff
@@ -31,10 +32,14 @@ class ElasticsearchBackend(KeyValueStoreBackend):
     scheme = 'http'
     host = 'localhost'
     port = 9200
+    es_retry_on_timeout=False
```
georgepsarakis: For uniformity and readability I think it is preferrable to surround `=` with spaces:

```
es_retry_on_timeout = False
```

docs/userguide/configuration.rst
```diff
@@ -1057,6 +1057,33 @@ Example configuration
 
 .. _conf-riak-result-backend:
 
+.. setting:: elasticsearch_retry_on_timeout
```
georgepsarakis: This documentation change belongs to line 1057 (above `.. conf-riak-result-backend:`).

celery/backends/elasticsearch.py
```diff
@@ -53,6 +58,17 @@ def __init__(self, url=None, *args, **kwargs):
...
 
+        self.es_retry_on_timeout = (
+                _get('elasticsearch_retry_on_timeout') or self.es_retry_on_timeout
+                )
+        self.es_timeout = (
+                _get('elasticsearch_timeout') or self.es_timeout
+                )
+
+        self.es_max_retries = (
+                _get('elasticsearch_max_retries') or self.es_max_retries
```
georgepsarakis: If someone sets maximum retries to 0, the setting value will revert to the default. So, perhaps you need to check explicitly for `None`?

celery/backends/elasticsearch.py
AlmightyOatmeal: It is also PEP 8 friendly.

Acey9:
sorry, i will fix these and push again

Acey9:
i will create a new pr for u

georgepsarakis:
@Acey9 I think there was no need to open new Pull Request. Opening new Pull Requests does not allow someone to easily review previous comments.

Acey9:
@georgepsarakis Thanks for your reply，but i have submitted a new one, can you tell me what to do next? please

PR #3962, opened 2017-04-06
return a boolean value according to the doc

netspyer:
## Description

According to the function's [docstring](https://github.com/celery/celery/blob/master/celery/platforms.py#L633):

`"""Return true value if signal by ``name`` exists on this platform."""`

it makes much more sense to return a **boolean** value .


thedrow:

PR #3963, opened 2017-04-07
Add regression test for chain duplication in chords (#3771)

georgepsarakis:
## Description

Continuation of #3782, that introduces an integration test to be combined with the fix provided in #3779.


georgepsarakis:
@ryanhiebert I tested with your solution in #3779 and the test is passing, if it is applied.

ryanhiebert:
Awesome, thank you!

PR #3969, opened 2017-04-10
Set the SO_REUSEADDR option on the socket

tbodt:
If the server exits before the client does, the port will appear to be in use for 90 seconds unless you set this option. TCP is annoying.

### DIFF
--- a/celery/contrib/rdb.py
+++ b/celery/contrib/rdb.py
@@ -124,6 +124,7 @@ class Rdb(Pdb):
         this_port = None
         for i in range(search_limit):
             _sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+            _sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
             this_port = port + skew + i
             try:
                 _sock.bind((host, this_port))


thedrow:
@auvipy I think that a failing test case is quite hard to produce here.

celery/contrib/rdb.py
```diff
@@ -124,6 +124,7 @@ def get_avail_port(self, host, port, search_limit=100, skew=+0):
         this_port = None
         for i in range(search_limit):
             _sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+            _sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
```
thedrow: What happens if SO_REUSEADDR is not supported on the system you are using?
SO_REUSEADDR was added in Linux 3.9 which is fairly recent. Also, it might not be supported on other systems as well.

tbodt: `SO_REUSEADDR` has been around since time immemorial. You're probably looking at `SO_REUSEPORT`, which was added in 3.9.

tbodt: Actually, according to MSDN, it's available since Windows 95. So there's no problem there. https://msdn.microsoft.com/en-us/library/windows/desktop/ms740621(v=vs.85).aspx

thedrow: This is what happens if you just copy/paste without reading... :blush:

PR #3977, opened 2017-04-13
Allow to have a custom request

mvaled:
Allowing a custom Request eases the task of handling timeouts (even hard timeouts).

### Rationale

Some (poorly written) bits of code catch exceptions quite broadly:

```
  try:
      ...
  except:
      ...
```

This hurts tasks when a SoftTimeLimitError is raised inside such blocks of
code.  Rewriting those smelly bits of code can take a lot of effort, and
sometimes, the code belongs to a third-party library which makes the task even
harder.

Using custom a request allows to catch hard time limits.

Your app can be customized like:

```python
   from celery import Task as BaseTask
   from celery.worker.request import Request as BaseRequest

   class Request(BaseRequest):
       def on_timeout(self, soft, timeout):
          super(Request, self).on_timeout(soft, timeout)
          if not soft:
	     print('Something hard hit me!')

    class MyTask(BaseTask):
        Request = Request

    @app.task(base=MyTask, bind=True)
    def sometask(self):
        import time
        while True:
            try:
                time.sleep(100)
            except:
                pass
```

PR #3983, opened 2017-04-16
SQS integration tests

georgepsarakis:
## Description

Enables integration tests with the SQS broker.

- [x] Update `.travis.yml` and `tox.ini` files.
- [x] Add CI script that starts [SQS moto server](https://github.com/spulec/moto).
- [x] Skip Manager operations that would not work with SQS.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -56,23 +56,33 @@ basepython =
 
 [testenv:2.7-integration-sqs]
 commands =
+    pip uninstall -y kombu
     pip install git+https://github.com/celery/kombu
+    py.test -xsv t/integration
 
 [testenv:3.4-integration-sqs]
 commands =
+    pip uninstall -y kombu
     pip install git+https://github.com/celery/kombu
+    py.test -xsv t/integration
 
 [testenv:3.5-integration-sqs]
 commands =
+    pip uninstall -y kombu
     pip install git+https://github.com/celery/kombu
+    py.test -xsv t/integration
 
 [testenv:3.6-integration-sqs]
 commands =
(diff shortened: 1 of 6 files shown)


georgepsarakis:
@auvipy @thedrow although this generally is working, I am seeing some failures intermittently and I do not understand what may be wrong. If you could offer some help I would appreciate it.

tox.ini
```diff
@@ -49,6 +54,36 @@ basepython =
     pypy: pypy
     flake8,flakeplus,apicheck,linkcheck,configcheck,pydocstyle: python2.7
 
+[testenv:2.7-integration-sqs]
```
georgepsarakis: If anyone knows a way to rewrite these `tox` sections in a less repetitive way, please let me know.

thedrow:
You might have encountered a bug in the SQS broker code or maybe the moto SQS server isn't responding fast enough.

thedrow:
@georgepsarakis strange... check the test failures here: https://github.com/celery/celery/pull/3977

georgepsarakis:
@thedrow yes it is strange. Do you think perhaps it is a combination of Canvas and the Redis result backend?

thedrow:
It might be a combination of the implementation of the "half-async" brokers such as Redis and SQS. They are async only when consuming and not when publishing.

thedrow:
@georgepsarakis Are you still working on this or should we turn this to an issue and close the PR?

georgepsarakis:
@thedrow I will resume efforts at a later time, unfortunately I could not achieve much.

PR #3987, opened 2017-04-19
send properties to before_publish handler

bbgwilbur:

## Description

The full properites should be sent to the before_publish handler instead of just kwargs so that the handler can act on any properties that are part of the message.

that version series.


kotfic:
+1

bbgwilbur:
This change was applied in https://github.com/celery/celery/commit/069ed5434ed86b9af7485c8632673a2bfbc2d771

PR #3994, opened 2017-04-24
Make id, name always accessible from logging.Formatter via extra

yoichi:
In case of task error, logging.Formatter can refer task id and task name via extra.
But in case of success, it cannot do so (extra is not set).
This patch set extra and improve the traceability.


celery/app/trace.py
```diff
@@ -623,3 +622,7 @@ def __protected_call__(self, *args, **kwargs):
             return orig(self, *args, **kwargs)
         BaseTask.__call__ = __protected_call__
         BaseTask._stackprotected = True
+
+
+def _info(fmt, context):
```
georgepsarakis: I would suggest to move this function at the top of the file and retain the name `info`, thus keeping the changes to a minimum.
Also, I think you need to add documentation in the function docstring in order to resolve the Travis CI errors.

yoichi: Thank you for the suggestion. I've added a docstring including the reason for adding extra.

yoichi:
The pydocstyle error in Travis CI is not caused by my change.

```
/home/travis/build/celery/celery/celery/app/trace.py:547 in public function `setup_worker_optimizations`:
        D401: First line should be in imperative mood; try rephrasing (found 'Setup')
```

georgepsarakis:
Thanks! This will probably be merged after version 4.0.3 is released. Feel free to add yourself to the list of [CONTRIBUTORS](https://github.com/celery/celery/blob/master/CONTRIBUTORS.txt).

thedrow:
Thanks!

thedrow:
@georgepsarakis Since we're releasing 4.1 I took the liberty of merging this. We can still revert and restore this PR if needed. Just let me know.

PR #3995, opened 2017-04-24
4.0

anilaratna2:

## Description

Use 3.1.x option CELERY_SEND_EVENTS in celery 4.0 instead of CELERYD_SEND_EVENTS for compatibility

that version series.


anilaratna2:
wrong commit

PR #4035, opened 2017-05-18
Signal before_task_publish not doing what documented

txomon:
## Description

Signal before_task_publish is sending kwargs instead of properties, which is at least a superset. Also, the docs announce it can be modified, but kwargs is not being passed.

### DIFF
--- a/celery/app/amqp.py
+++ b/celery/app/amqp.py
@@ -540,7 +540,7 @@ class AMQP(object):
                     sender=name, body=body,
                     exchange=exchange, routing_key=routing_key,
                     declare=declare, headers=headers2,
-                    properties=kwargs, retry_policy=retry_policy,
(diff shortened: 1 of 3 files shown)


txomon:
The build is failing for something totally unrelated to this PR

georgepsarakis:
@txomon could you please rebase?

txomon:
@georgepsarakis I understand you updated it, but it's still failing? Anyway, just ping me if you need anything else

georgepsarakis:
@txomon yes I did so this resolved 2 of the build job failures. The fix for the final one is already in `master` so you just need to rebase your branch :smile: .

georgepsarakis:
@txomon please ignore my previous comment, `master` does not contain this fix.

PR #2533, opened 2015-03-11
Support Redis Sentinel as result backend

tristan0x:
Hi there,

We (at Cogniteev) recently switched to Redis as result backend that provides a very efficient implementation of chords. Yet we missed high-availability to use it on production.

Here is a patch in Redis result backend that provide support to Sentinel.

PR #4053, opened 2017-05-25
At the present time celery backend will not support sentinel, but the…

strawposter:
… support of sentinel is realized in kombu.

A basic example that shows how this can be done is here: https://github.com/dealertrack/celery-redis-sentinel

I tried to do something similar, but I'm sure that my code can be made more elegant.

I will be grateful for any help.

### DIFF
--- a/celery/backends/redis.py
+++ b/celery/backends/redis.py
@@ -342,7 +342,10 @@ class RedisBackend(base.BaseKeyValueStoreBackend, async.AsyncBackendMixin):
     def password(self):
         return self.connparams['password']
 
+
 class SentinelBackend(RedisBackend):
+    """Sentinel task result store."""
+
     def __init__(self, *args, **kwargs):
         super(SentinelBackend, self).__init__(*args, **kwargs)
 
--- a/celery/backends/redis.py
+++ b/celery/backends/redis.py
@@ -4,6 +4,7 @@ from __future__ import absolute_import, unicode_literals
 
 from functools import partial
 
+from kombu import Connection
 from kombu.utils.functional import retry_over_time
 from kombu.utils.objects import cached_property
 from kombu.utils.url import _parse_url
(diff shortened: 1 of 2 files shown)


strawposter:
unlikely

auvipy:
did you test this localy or in dev server?

strawposter:
Yes, everything works. That would not violate the logic of the library, it is necessary to use the kombu.
But to get the redis client you need to use the construction: connection.connect()._avail_channels[0].client
It seems too cumbersome to me, but another way to get the redis client through kombu I did not find.

georgepsarakis:
@auvipy This [Pull Request](https://github.com/celery/celery/pull/2533/) should be related.

celery/backends/redis.py
```diff
@@ -340,3 +341,22 @@ def db(self):
...
+
+class SentinelBackend(RedisBackend):
+    """Sentinel task result store."""
+
+    def __init__(self, *args, **kwargs):
+        super(SentinelBackend, self).__init__(*args, **kwargs)
+
+        _get = self.app.conf.get
+
+        self.broker_url = _get('BROKER_URL') or {}
```
georgepsarakis: This module contains the implementation of the result backend, so my question is why are the settings for the broker used here?

strawposter: Hi, George
This code sample is intended when BROKER_URL == BACKEND_URL.
I will agree that this is not flexible, but it is fixable. It does not matter.
This can be redesigned to receive exactly BACKEND_URL.
The important thing is that the celery bakend still does not support the sentinel.
I also wrote that I would be grateful if someone would rewrite this code more correctly, but so far I have not seen a significant contribution and therefore I am confident,
I also wrote that I would be grateful if someone would rewrite this code more correctly, but so far I have not seen a significant contribution.

strawposter: This code demonstrates an approach that is based on the following principles:
- it is necessary to use a a child class that inherits from the RadishBakend;
- tt is necessary to use a kombu.Connection, because the whole backend uses kombu.

thedrow:
Thanks for the contribution. #4144 supersedes this PR.

PR #4061, opened 2017-05-31
Fix task eager Django setting names

rpatterson:
## Description

Under celery 4.0.2, I've tried the settings in this document before this change, variations with and without the `*_TASK_*` element, and a few other variations, and the two names in this change are the only two that work for me.


auvipy:
could you also check travis fails?

rpatterson:
@auvipy Is that question for me?  No, I don't know anything about the build process and honestly I won't learn it.  ;-)  I just figured a PR fixing the doc bug was better than an issue reporting it.  I have a hard time imagining my very simple doc change broke anything but if you can't accept this or address the build errors yourself, at least this is here for others to find.

PR #4074, opened 2017-06-06
Handle possibility there are no workers in `celery graph workers`

donkopotamus:
Fixes issue where an exception is raised if eg `celery graph workers` is invoked when there are no workers at all.


thedrow:
This is a utility. I vote for merging as it is.
We can open an issue about covering it with tests later.
@auvipy What do you say?

auvipy:
thanks for your input @thedrow we can do so

PR #4077, opened 2017-06-07
Catch NotRegistered exception for errback

bbgwilbur:
If the errback task is not registered in the currently running worker, the arity_greater check will fail. We can just assume its going to work as an old-style signature and deal with the possible error if it doesn't later.


thedrow:
This is too old to merge and I'm working on this.
Closing.

PR #4078, opened 2017-06-08
Fix for crash on Windows

SPKorhonen:
Function **trace.setup_worker_optimizations** is not called if process is not forked thus leaving trace module partially uninitialized, __localized_ attribute in particular. This causes celery to crash on Windows. The whole **if**-clause seems spurious as the same initialization step are also performed by the else branch. Removing the if clause and calling fixes the problem on Windows.

### DIFF
--- a/celery/concurrency/prefork.py
+++ b/celery/concurrency/prefork.py
@@ -63,14 +63,7 @@ def process_initializer(app, hostname):
                   bool(os.environ.get('CELERY_LOG_REDIRECT', False)),
                   str(os.environ.get('CELERY_LOG_REDIRECT_LEVEL')),
                   hostname=hostname)
-    if os.environ.get('FORKED_BY_MULTIPROCESSING'):
-        # pool did execv after fork
-        trace.setup_worker_optimizations(app, hostname)
-    else:
-        app.set_current()
-        set_default_app(app)
-        app.finalize()
-        trace._tasks = app._tasks  # enables fast_trace_task optimization.
+    trace.setup_worker_optimizations(app, hostname)
     # rebuild execution handler for all tasks.
     from celery.app.trace import build_tracer
     for name, task in items(app.tasks):


SPKorhonen:
Would a test under /t/integration/test_tasks be ok? As this would be the simplest way to do it. Basically just run the minimal "add" task on ensure that correct result is generated. If unit test is needed I need to see where to implement the test as it spans the functionality of several modules (app, trace, prefork)

auvipy:
go for the integration test first

SPKorhonen:
When I try to run the integration test(s) with pytest I get error message about missing fixtures ("fixture 'celery_app' not found"). Am I missing something or is there a special way to start these tests? 

A quick comparison with the /t/unit/conftest.py for unit tests reveals that there are several includes from celery.contrib.pytest missing from /t/integration/conftest.py.

thedrow:
@SPKorhonen I'm not sure I follow. What do you need celery_app for?

floqqi:
@SPKorhonen @auvipy Any progress here? This issue is blocking my upgrade to celery 4 until now :(

georgepsarakis:
@SPKorhonen you can see how the tests are being ran if you see the commands that are being executed in the [Travis CI](https://github.com/celery/celery/blob/master/.travis.yml) file.

drhagen:
For what its worth, I got the error described in #4081 in Windows 10 with Celery 4.1.0, Django, and RabbitMQ. I manually applied the patch described in this pull request to the celery package, and the error went away. Celery now works after changing that one line.

equinoxel:
IMHO you don't need to change the code. it's enough to define `FORKED_BY_MULTIPROCESSING=1` environment variable for the worker instance.

auvipy:
it's enough to define FORKED_BY_MULTIPROCESSING=1 environment variable for the worker instance.

In that case pointing to docs about defining FORKED_BY_MULTIPROCESSING=1 environment variable for the worker instance.

zany1:
@equinoxel @auvipy can you point me to the docs about setting the  FORKED_BY_MULTIPROCESSING=1 , I'm new to python, couldn't find it on google.

illgitthat:
@zany1 On windows, you can set the environment variable in a command prompt window using `set FORKED_BY_MULTIPROCESSING=1`

This will be a temporary environment variable. To store it permanently, see [setting environment variables](https://www.computerhope.com/issues/ch000549.htm)

illgitthat:
@auvipy saw this issue closed a few times. Ran into this today on Windows and didn't see any mention on the environment variable in the docs.

zany1:
@illgitthat I fixed this issue long ago, don't remember how, but I would highly recommend you stop using Django or celery on windows, too many bugs and I believe they have indefinitely stopped updating for windows. Just use VMware to virtualize any distro of Linux.

W7PEA:
I hit this today... Added a little more specific direction to use the installation on Windows guide for RabbitMQ. 
PR: https://github.com/celery/celery/pull/4802

thesunlover:
there is a good suggestion how to start the latest version of celery: [https://stackoverflow.com/a/47331438/2848256](with the help of eventlet)

ChaitanyaVankadaru:
https://github.com/celery/celery/issues/4178

Use eventlet as below:

pip install eventlet
celery -A <module> worker -l info -P eventlet
It works for me on Windows 10 + celery 4.2 + python 3.7

matteius:
I ran into this tonight -- doesn't seem like we should have to use eventlet on Windows.   What was wrong with this patch?  It just never got the unit test coverage requested?

Galsor:
It seems eventlet is no longer a solution for windows as they use `os.register_at_fork` which is only available in Unix and throws exceptions on Windows.
[see](https://github.com/eventlet/eventlet/issues/644)

with `Windows10`, `Celery 4.4.7`,` Eventlet 0.27.0`
I could only make it work using `celery -A worker -l info -P solo`

exzzx:
> It seems eventlet is no longer a solution for windows as they use `os.register_at_fork` which is only available in Unix and throws exceptions on Windows.
> [see Eventlet Issue](https://github.com/eventlet/eventlet/issues/644)
> 
> with `Windows10`, `Celery 4.4.7`,` Eventlet 0.27.0`
> I could only make it work using `celery -A worker -l info -P solo`

From: https://stackoverflow.com/questions/63811316/celery-worker-throws-attributeerror-module-os-has-no-attribute-register-at-f, you can specify the version of eventlet. pip install eventlet==0.26

It works fine for me.

paulon-zervoulakus:
FORKED_BY_MULTIPROCESSING=1 works for me.. But I wonder my celery is ver. 5.1.2 for windows and still the issue is alive.

PR #4085, opened 2017-06-13
Using Exception.args to serialize/deserialize exceptions instead of str(Exception)

frostoov:
...cause its more generic


thedrow:
This looks good. I'm not going to merge this for the next fix release but it will definitely hit 4.1.

thedrow:
Thanks for your contribution.

PR #4093, opened 2017-06-16
Add link to Version 2

alukach:
To the best of my knowledge, the `before_task_publish` can have V2 task messages (like `after_task_publish`, as documented).


alukach:
Oops, already submitted...

Issue #3772, opened 2017-01-18
Some Old Style Configuration Variable Names are Incorrect in Docs


pnovotnak:
## Checklist

- [✓] I have included the output of ``celery -A proj report`` in the issue.
      (if you are not able to do this, then at least specify the Celery
       version affected).
- [✓] I have verified that the issue exists against the `master` branch of Celery.

## Steps to reproduce

Follow the Django tutorial, then visit the configuration / defaults documentation page and attempt to use some of the old style configuration settings.

Using the old style setting `CELERY_DEFAULT_QUEUE` has no effect while `CELERY_TASK_DEFAULT_QUEUE` works as intended. I suppose this is probably the same for many or all of the `task_.*` variables, though I have not tested them.

```python
# proj/proj/celery.py
...
# Using a string here means the worker don't have to serialize
# the configuration object to child processes.
# - namespace='CELERY' means all celery-related configuration keys
#   should have a `CELERY_` prefix.
app.config_from_object('django.conf:settings', namespace='CELERY')
...
```

## Expected behavior

One should be able to configure the app by referring to the documentation.

## Actual behavior

Configuration docs are broken.

pnovotnak:
Forgot to actually include `celery -A proj report` ¯\_(ツ)_/¯ [celery-report.txt](https://github.com/celery/celery/files/717747/celery-report.txt)

pnovotnak:
Also broken;

* CELERYD_PREFETCH_MULTIPLIER -> CELERY_WORKER_PREFETCH_MULTIPLIER

thedrow:
@auvipy Is this fixed?

marcomayer:
@thedrow @auvipy :This is still wrong on http://docs.celeryproject.org/en/latest/userguide/configuration.html#std:setting-task_default_queue and it just cost me half a day of work to figure it out. Is there any documentation or maybe a README or some other way to actually find out the correct configuration values that work?

thedrow:
@marcomayer That document should be kept up to date.
If you find anything wrong, please issue a new pull request fixing the problem.

ckm2k1:
@thedrow The prefixes for settings in the docs don't seem to correspond to what Celery 4.1.0 expects. 
Ex: 
Bad: `CELERY_TASK_ALWAYS_EAGER = True`
Good: `CELERY_ALWAYS_EAGER = True`

[...]

PR #4094, opened 2017-06-16
Fix CELERY_TASK_ settings docs

alukach:
I've noticed that settings `CELERY_QUEUES` or `CELERY_DEFAULT_QUEUE` don't seem to apply to the configuration as described in the documentation.

I don't fully understand the configuration logic, but it appears that the logic for loading settings via `app.config_from_object` is something along the lines of `conf[setting_name] = getattr(config_obj, namespace + '_' + setting_name.upper(), defaults[setting_name])`.  This means that if a setting is named `task_queues`, the relevant old-style setting should be `CELERY_TASK_QUEUES`, not `CELERY_QUEUES` as written.  I believe that #4061 attempted to resolve this but stopped a bit short.

Here are some quick tests that confirm this for just two of the settings:

```python
from celery import Celery
from kombu import Queue
app = Celery()

class ConfigGood:
    CELERY_TASK_DEFAULT_QUEUE = 'my-default-queue'
    CELERY_TASK_QUEUES = [Queue('a'), Queue('b')]

app.config_from_object(ConfigGood, namespace='CELERY')
assert app.conf.task_default_queue == 'my-default-queue'
assert list(app.amqp.queues.keys()) == ['a', 'b']

app = Celery()

class ConfigBad:
    CELERY_DEFAULT_QUEUE = 'my-default-queue'
    CELERY_QUEUES = [Queue('a'), Queue('b')]

app.config_from_object(ConfigBad, namespace='CELERY')
try:
    assert app.conf.task_default_queue == 'my-default-queue'
except AssertionError:
    print("Misconfigured")
try:
    assert list(app.amqp.queues.keys()) == ['a', 'b']
except AssertionError:
    print("Misconfigured")
```

I've taken a bit of a leap in assuming that all other `task_*` settings follow the same logic. I did not attempt to update any other settings so I may be mistaken here.  I would appreciate another set of eyes/tests on these changes.  Perhaps more config validations should be rolled into the unit tests.

It's possible that I'm misinterpretting the documentation by assuming that any of the "Setting name" values mentioned in the documentation should work with the `app.config_from_object` method.  If so, then I will close this PR and attempt to clarify the docs.


alukach:
Should resolve #3772

georgepsarakis:
@alukach nice work. I believe the option names along with their default values can be found [here](https://github.com/celery/celery/blob/master/celery/app/defaults.py#L76), so it would be a good place to cross-check everything. I will have a look later too.

georgepsarakis:
@auvipy this contains documentation changes only, can I merge? Are we still waiting for a release or can I merge approved pull requests?

auvipy:
sorry I was on my vacations

PR #4101, opened 2017-06-21
Add options for exponential backoff with task autoretry

singingwolfboy:
Celery's [task autoretry support](http://docs.celeryproject.org/en/master/userguide/tasks.html#automatic-retry-for-known-exceptions) is very handy, but it's missing an important feature: [exponential backoff](https://en.wikipedia.org/wiki/Exponential_backoff). When you're writing a large Python application that has the potential to overwhelm another service it depends on, this is an intelligent, respectful way to allow the service to process requests as fast as it can.

I've added documentation for this code, but I'm not sure what is the best way to write automated tests. I have one test that is failing, and I don't understand why -- I don't even know if it's the right sort of test to write. Can someone help me out with writing tests for this change?

### DIFF
--- a/celery/app/base.py
+++ b/celery/app/base.py
@@ -37,7 +37,7 @@ from celery.utils import abstract
 from celery.utils.collections import AttributeDictMixin
 from celery.utils.dispatch import Signal
 from celery.utils.functional import first, maybe_list, head_from_fun
-from celery.utils.time import timezone
+from celery.utils.time import timezone, get_exponential_backoff_interval
 from celery.utils.imports import gen_task_name, instantiate, symbol_by_name
 from celery.utils.log import get_logger
 from celery.utils.objects import FallbackContext, mro_lookup
@@ -464,10 +464,8 @@ class Celery(object):
 
             autoretry_for = tuple(options.get('autoretry_for', ()))
             retry_kwargs = options.get('retry_kwargs', {})
-            retry_backoff = options.get('retry_backoff', False)
-            if retry_backoff is True:
-                retry_backoff = 1
-            retry_backoff_max = options.get('retry_backoff_max', 600)
+            retry_backoff = int(options.get('retry_backoff', False))
+            retry_backoff_max = int(options.get('retry_backoff_max', 600))
             retry_jitter = options.get('retry_jitter', True)
 
             if autoretry_for and not hasattr(task, '_orig_run'):
@@ -478,17 +476,13 @@ class Celery(object):
                         return task._orig_run(*args, **kwargs)
                     except autoretry_for as exc:
                         if retry_backoff:
-                            countdown = (
-                                retry_backoff * (2 ** task.request.retries)
-                            )
-                            if retry_jitter is True:
-                                countdown = random.randrange(countdown + 1)
-                            elif isinstance(retry_jitter, int):
-                                jitter = random.randrange(retry_jitter + 1)
-                                countdown += jitter * random.choice((1, -1))
-                            if retry_backoff_max:
-                                countdown = min(countdown, retry_backoff_max)
-                            retry_kwargs['countdown'] = countdown
+                            retry_kwargs['countdown'] = \
+                                get_exponential_backoff_interval(
+                                    factor=retry_backoff,
+                                    retries=task.request.retries,
+                                    maximum=retry_backoff_max,
+                                    full_jitter=retry_jitter
+                                )
                         raise task.retry(exc=exc, **retry_kwargs)
 
                 task._orig_run, task.run = task.run, run
--- a/celery/app/base.py
+++ b/celery/app/base.py
@@ -472,22 +472,23 @@ class Celery(object):
 
             if autoretry_for and not hasattr(task, '_orig_run'):
 
-                if retry_backoff:
-                    countdown = retry_backoff * (2 ** task.request.retries)
-                    if retry_jitter is True:
-                        countdown = random.randrange(countdown)
-                    elif isinstance(retry_jitter, int):
-                        jitter = random.randrange(retry_jitter)
-                        countdown += jitter * random.choice((1, -1))
-                    if retry_backoff_max:
-                        countdown = min(countdown, retry_backoff_max)
-                    retry_kwargs['countdown'] = countdown
-
                 @wraps(task.run)
                 def run(*args, **kwargs):
                     try:
                         return task._orig_run(*args, **kwargs)
                     except autoretry_for as exc:
+                        if retry_backoff:
+                            countdown = (
+                                retry_backoff * (2 ** task.request.retries)
+                            )
+                            if retry_jitter is True:
+                                countdown = random.randrange(countdown)
+                            elif isinstance(retry_jitter, int):
+                                jitter = random.randrange(retry_jitter)
+                                countdown += jitter * random.choice((1, -1))
+                            if retry_backoff_max:
+                                countdown = min(countdown, retry_backoff_max)
+                            retry_kwargs['countdown'] = countdown
                         raise task.retry(exc=exc, **retry_kwargs)
 
                 task._orig_run, task.run = task.run, run
--- a/celery/app/base.py
+++ b/celery/app/base.py
@@ -473,7 +473,7 @@ class Celery(object):
             if autoretry_for and not hasattr(task, '_orig_run'):
 
                 if retry_backoff:
-                    countdown = retry_backoff * (2 ** task.retries)
+                    countdown = retry_backoff * (2 ** run.retries)
                     if retry_jitter is True:
                         countdown = random.randrange(countdown)
                     elif isinstance(retry_jitter, int):
--- a/celery/app/base.py
+++ b/celery/app/base.py
@@ -480,8 +480,7 @@ class Celery(object):
                                     factor=retry_backoff,
                                     retries=task.request.retries,
                                     maximum=retry_backoff_max,
-                                    full_jitter=retry_jitter
-                                )
+                                    full_jitter=retry_jitter)
                         raise task.retry(exc=exc, **retry_kwargs)
 
                 task._orig_run, task.run = task.run, run
--- a/celery/app/base.py
+++ b/celery/app/base.py
@@ -473,7 +473,7 @@ class Celery(object):
             if autoretry_for and not hasattr(task, '_orig_run'):
 
                 if retry_backoff:
-                    countdown = retry_backoff * (2 ** run.retries)
+                    countdown = retry_backoff * (2 ** task.request.retries)
                     if retry_jitter is True:
                         countdown = random.randrange(countdown)
                     elif isinstance(retry_jitter, int):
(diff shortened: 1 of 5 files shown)


georgepsarakis:
@singingwolfboy a good feature indeed. I have a suggestion though, do you think perhaps that you could take advantage of [tenacity](https://github.com/jd/tenacity) API? Exponential backoff and jitter are already implemented and well-tested in that package. I assume that we could eventually replace the autoretry-on-exception logic with `tenacity` altogether?

pachewise:
Your test is failing because `_task_from_fun(...)` is only called once when the function is decorated. You'll probably want to move your backoff logic into `task.retry`, and pass in stuff like `retry_backoff` as part of `retry_kwargs`.

singingwolfboy:
@georgepsarakis: [tenacity](https://github.com/jd/tenacity) looks like a useful library, but replacing the existing retry logic is outside the scope of this pull request. I'm not going to introduce a new dependency for something that can be coded up so easily. Maybe you can make your own pull request that refactors the existing retry logic to use it?

@pachewise Thanks for pointing that out! I moved the logic around, and the test is now passing.

What else do I need to do, to get this PR ready to merge?

t/unit/tasks/test_tasks.py
```diff
@@ -251,6 +268,16 @@ def test_autoretry(self):
         self.autoretry_task.apply((1, 0))
         assert self.autoretry_task.iterations == 6
 
+    @patch('random.randrange')
+    def test_autoretry_backoff(self, randrange):
```
georgepsarakis: If you could add another test case that verifies the jitter flow it would be nice!

celery/app/base.py
```diff
@@ -471,6 +477,18 @@ def run(*args, **kwargs):
                     try:
                         return task._orig_run(*args, **kwargs)
                     except autoretry_for as exc:
+                        if retry_backoff:
```
georgepsarakis: The more compact form could potentially be (I think the `if` block is not required):
```
countdown = retry_backoff * (2 ** task.request.retries) # will be zero if retry_backoff = 0
countdown += random.randrange(retry_jitter + 1) * random.choice((1, -1)) # if retry_jitter is 0, the result is always zero too
# Adjust according to maximum backoff
countdown = min(retry_backoff_max, countdown)
```

- `random.randrange` excludes the endpoint that is why I added `1`. This was also excluding defining `1` as a possible jitter value.
- If `retry_jitter` is 0, `countdown` remains the same.
- If `retry_jitter` is 2, and delay is set to 8, then, possible random values are `[0, 1, 2]`, and `countdown` will be in the range `[6, 10]`.

singingwolfboy:
Sorry it took me so long to get back to this, but I believe I've addressed your review comments!

The one thing that I wanted to discuss was the question of mixed types. I can understand why you think it might be a bad idea: most type systems are very strict about a value being only a single type. In this case, though, I think it's a nice enhancement that strikes a good balance between making simple things simple and making complex things possible.

In the simple case, a developer writing a task can opt-in to exponential backoff by using a boolean type. It's very easy and clear to read `retry_backoff=True` and understand that the developer defining the task wants to enable backoff, and probably doesn't care about the implementation details of _how_ the backoff works.

However, things get a little tricker in the complex case, where a developer _does_ care about the implementation details of how the backoff works. If we do strict typing, boolean _or_ integer, then the developer needs to set two arguments: perhaps `retry_backoff=True` to turn on the feature, and `retry_backoff_factor=3` to set a multiplicative factor of 3 in calculating backoff. But if you're configuring the backoff factor, that _implies_ that you want to use backoff -- so why should you have to set two arguments, if one is implied?

Now, we could say that `retry_backoff` must always a be an integer and never a boolean. If the integer is 1 or higher, then backoff is on; if the integer is 0, then backoff is off. However, that makes the simple case more complex than it needs to be. An unfamiliar developer reading the code sees `retry_backoff=1`, and has to figure out why it's an integer rather than a boolean. The function call is being forced to set implementation details unnecessarily! Much simpler to treat it as a boolean on/off switch, if you don't care about the implementation details.

What do you think?

t/unit/tasks/test_tasks.py
singingwolfboy: Added!

celery/app/base.py
singingwolfboy: > `random.randrange` excludes the endpoint that is why I added `1.` This was also excluding defining `1` as a possible jitter value.

Good catch! Thanks, I fixed that.

As for making the code more compact, I'd like to do that, but I'm bumping up against the "max characters per line" style restriction. So, I'll keep things the way they currently are in the pull request.

georgepsarakis: Thanks for the changes! Do you mind if I try this change on this branch then?

t/unit/tasks/test_tasks.py
```diff
@@ -251,6 +279,37 @@ def test_autoretry(self):
         self.autoretry_task.apply((1, 0))
         assert self.autoretry_task.iterations == 6
 
+    @patch('random.randrange', side_effect=lambda i: i-1)
```
thedrow: flake8 is complaining about the lack of spaces between `-`

t/unit/tasks/test_tasks.py
```diff
@@ -251,6 +279,37 @@ def test_autoretry(self):
...
+        with patch.object(task, 'retry', wraps=task.retry) as fake_retry:
+            task.apply(("http://httpbin.org/error",))
+
+        assert task.iterations == 4
+        retry_call_countdowns = [
+            call[1]['countdown'] for call in fake_retry.call_args_list
+        ]
+        assert retry_call_countdowns == [1, 2, 4, 8]
+
+    @patch('random.randrange', side_effect=lambda i: i-1)
```
thedrow: flake8 is complaining about the lack of spaces between `-`

t/unit/tasks/test_tasks.py
singingwolfboy: Oops! Fixed.

celery/app/base.py
singingwolfboy: You're certainly welcome to check out my code from this branch and manually test it, if that's what you mean. If you're asking to change the code in this branch, I'd prefer if you made a new branch off of this one, make whatever changes you're suggesting there, and make a pull request back to the `exponential-backoff` branch of my fork. That way, I can see exactly what you're suggesting, without you modifying my work unilaterally.

georgepsarakis:
@singingwolfboy thanks for your feedback! I have created #4153 as a temporary pull request, in order to display my suggested changes. Please let me know what you think.

By the way, this seems to be a popular feature request so thanks again for your efforts :smile: .

singingwolfboy:
@georgepsarakis I pulled in your proposed changes from #4153, and I updated the documentation. Is this ready to merge?

docs/userguide/tasks.rst
```diff
@@ -746,6 +746,69 @@ If you want to automatically retry on any error, simply use:
...
+    delay factor. For example, if this option is set to ``3``, the first retry
+    will delay 3 seconds, the second will delay 6 seconds, the third will
+    delay 12 seconds, the fourth will delay 24 seconds, and so on. By default,
+    this option is set to ``False``, and autoretries will not be delayed.
+
+.. attribute:: Task.retry_backoff_max
+
+    A number. If ``retry_backoff`` is enabled, this option will set a maximum
+    delay between task autoretries. By default, this option is set to ``600``,
+    which is 10 minutes. To allow unlimited delay between autoretries,
```
georgepsarakis: The unlimited delay is not supported now, so could you please remove it? I don't believe that this would be desired probably.

t/unit/utils/test_time.py
```diff
@@ -253,3 +254,41 @@ def test_utcoffset(self, patching):
         assert utcoffset(time=_time) is not None
         _time.daylight = False
         assert utcoffset(time=_time) is not None
+
+
+class test_get_exponential_backoff_interval:
+
+    @patch('random.randrange', lambda n: n - 2)
+    @patch('random.choice', lambda sequence: sequence[0])
```
georgepsarakis: Sorry this is actually unused. Could you please remove it? Or I can do it if you allow me changes.

t/unit/utils/test_time.py
```diff
@@ -253,3 +254,41 @@ def test_utcoffset(self, patching):
...
+    def test_bound_by_maximum(self):
+        maximum_boundary = 100
+        assert get_exponential_backoff_interval(
+            factor=40,
+            retries=3,
+            maximum=maximum_boundary
+        ) == maximum_boundary
+
+    @patch('random.randrange', lambda n: n - 1)
+    @patch('random.choice', lambda sequence: -1)
```
georgepsarakis: Same as above, the `random.choice` patching is not required.

singingwolfboy:
@georgepsarakis go ahead and make those changes, they sound good to me 👍

georgepsarakis:
@singingwolfboy many thanks for your contribution!

singingwolfboy:
✨ 💖 ✨

codingjoe:
@geoffreybauduin any chance you are going to release this any time soon? I found it in the documentation and only realised that this feature isn't tagged in any version yet.

codingjoe:
Never mind, just found the v4.2 milestone.

geoffreybauduin:
@condingjoe I think you meant to mention @georgepsarakis ?

codingjoe:
ahah yes, sorry :/
very quick response tho, impressive 👍

jonashaag:
I like how the backoff is not merely a boolean but configurable. With the current implementation it's always power-of-2 backoff however, with a configurable fixed scaling factor. I.e. your series will always be `r * 2 ** n`. I'm not sure if a fixed scaling factor is useful for an exponential series in the first place? It will quickly vanish when the series progresses.

My proposal is to make the backoff factor be the exponential's base, so your series will be `r ** n` with configurable `r` (default 2). What do you think?

georgepsarakis:
@jonashaag I have found these resources useful, for determining the algorithm:
- https://www.awsarchitectureblog.com/2015/03/backoff.html
- https://github.com/jd/tenacity/blob/dd0c50a540785fd67daf726eec84cfe373b26a0c/tenacity/wait.py#L172

If you think there is an error or possible improvement for the current implementation, feel free to comment.

jonashaag:
Interesting read. I still think if only the scale factor or the base are configurable, it should be the base (default 2).

georgepsarakis:
Scale factor is necessary in order to determine the correct magnitude of retry interval. With these options, the user has much more control over the interval progression, e.g.:

```
print [ get_exponential_backoff_interval(0.001, retry, 5) for retry in range(12) ]
[0.001, 0.002, 0.004, 0.008, 0.016, 0.032, 0.064, 0.128, 0.256, 0.512, 1.024, 2.048]
```

I don't really understand how with your suggestion we can configure millisecond intervals.

jonashaag:
I agree, maybe both should be configurable. Or the unit should be milliseconds.

singingwolfboy:
There is a lot of value in simplicity. If users need to configure backoff settings to such fine-grained detail, it would be better for them to _not_ use Celery’s built-in autoretry feature, and instead implement retry and backoff in their tasks. That way, the user can specify their exact needs to whatever degree of precision they want, without worrying about Celery’s settings.

Issue #4102, opened 2017-06-22
Solar Scheduling broken


DDevine:
## Checklist
software -> celery:4.0.2 (latentcall) kombu:4.0.2 py:3.5.3
            billiard:3.5.0.2 redis:2.10.5
platform -> system:Linux arch:64bit imp:CPython

## Steps to reproduce
This is related to the problem I saw in https://github.com/celery/django-celery-beat/issues/55
After sorting out django_celery_beat issues I discovered a problem in `celery.schedules.solar.remaining_estimate()`:
> [2017-06-22 16:19:54,442: WARNING/MainProcess] File "/opt/myapp-env/lib/python3.5/site-packages/celery/schedules.py", line 777, in remaining_estimate
> [2017-06-22 16:19:54,442: WARNING/MainProcess] start=last_run_at_utc, use_center=self.use_center,
> [2017-06-22 16:19:54,442: WARNING/MainProcess] TypeError
> [2017-06-22 16:19:54,442: WARNING/MainProcess] :
> [2017-06-22 16:19:54,443: WARNING/MainProcess] next_transit() got an unexpected keyword argument 'use_center'

I managed to stop `celery beat` from crashing by removing the `use_center` argument, but I suspect this is probably the wrong thing to do.

The exact event I was using was solar_noon.

## Expected behaviour
Solar tasks should work.

DDevine:
I have found a lot more issues that I have discussed over in https://github.com/celery/django-celery-beat/issues/57 - however apart from one small tweak to django-celery-beat, the issues are actually in the `celery.schedules.solar` class.

There are problems with converting local time (in LMT) to UTC time and also scheduling solar and crontab tasks at the same time does not seem possible, likely because `last_run_at` is interdependent between the two schedulers.

PR #4103, opened 2017-06-22
Fixed exception caused by next_transit receiving an unexpected argument.

DDevine:
This fixes https://github.com/celery/celery/issues/4102 
The problem was that arguments were being passed to methods with signatures that could not take them, causing an exception.

### DIFF
--- a/celery/schedules.py
+++ b/celery/schedules.py
@@ -728,7 +728,7 @@ class solar(BaseSchedule):
         'dawn_nautical': True,
         'dawn_civil': True,
         'sunrise': False,
-        'solar_noon': True,
+        'solar_noon': False,
         'sunset': False,
         'dusk_civil': True,
         'dusk_nautical': True,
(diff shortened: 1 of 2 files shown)


celery/schedules.py
```diff
@@ -783,10 +783,16 @@ def remaining_estimate(self, last_run_at):
...
-                start=last_run_at_utc, use_center=self.use_center,
-            )
+            if self.use_center:
+                next_utc = getattr(self.cal, self.method)(
+                	self.ephem.Sun(),
+                	start=last_run_at_utc, use_center=self.use_center,
+                )
+            else:
+                next_utc = getattr(self.cal, self.method)(
+                        self.ephem.Sun(),start=last_run_at_utc
```
georgepsarakis: This is probably a PEP8 error: `, start=`.

celery/schedules.py
```diff
@@ -783,10 +783,16 @@ def remaining_estimate(self, last_run_at):
         last_run_at_utc = localize(last_run_at, timezone.utc)
         self.cal.date = last_run_at_utc
         try:
-            next_utc = getattr(self.cal, self.method)(
-                self.ephem.Sun(),
-                start=last_run_at_utc, use_center=self.use_center,
-            )
+            if self.use_center:
+                next_utc = getattr(self.cal, self.method)(
```
georgepsarakis: What are the alternatives for `self.method` here?

DDevine: I don't see a problem with using self.method - am I missing something?

celery/schedules.py
DDevine: Good catch. Oops!

thedrow:
@georgepsarakis Should we merge this before the release? What's wrong here and how do we fix it?

DDevine:
The failing tests appears to be because of a HTTP 500 exception - not my code. 

I read that there has been a merge freeze - so my pull requests have missed the boat somehow. That's quite unfortunate.

thedrow:
@DDevine Ignore the landscape failure. We're gonna have to deactivate it again.
What @auvipy means is that test coverage is missing for this feature.
If you'll add it, I'll merge this for the next bug fix release.

PR #4119, opened 2017-07-03
Re-order use of redis SSL params vs configuration from URL

JonPeel:
Allow redis_backend_use_ssl to be used without being trumped by URL configuration

### DIFF
--- a/celery/backends/redis.py
+++ b/celery/backends/redis.py
@@ -133,9 +133,10 @@ class RedisBackend(base.BaseKeyValueStoreBackend, async.AsyncBackendMixin):
                 socket_connect_timeout and float(socket_connect_timeout),
         }
 
-
         if url:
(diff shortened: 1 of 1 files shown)


celery/backends/redis.py
```diff
@@ -141,10 +146,6 @@ def __init__(self, host=None, port=None, db=None, password=None,
             self.connparams.update(ssl)
             self.connparams['connection_class'] = redis.SSLConnection
 
-        if url:
```
georgepsarakis: From what I have tested with `_params_from_url` no replacement will occur here, because the SSL options are passed as the `defaults` dictionary. Would you mind explaining why this is required?

JonPeel:
Ah. Ignore - seems I was fixing a non-issue and, when it started working, thought it was this, not the settings change made.

PR #4120, opened 2017-07-04
Add note about RPC backend not support chords

alukach:
Chords are not supported with the RPC backend (https://github.com/celery/celery/commit/a883f6871b54c1292920e68cf8c01d2efc464f3f), however this does not appear to be documented anywhere. I've put a note under the Chord docs on the Canvas page, however I would also be willing to place this not on the [RPC Backend Settings docs](http://docs.celeryproject.org/en/latest/userguide/configuration.html#rpc-backend-settings) as well if that's desired.

### DIFF
--- a/docs/userguide/canvas.rst
+++ b/docs/userguide/canvas.rst
@@ -766,7 +766,8 @@ Chords
 
     Tasks used within a chord must *not* ignore their results. If the result
     backend is disabled for *any* task (header or body) in your chord you
-    should read ":ref:`chord-important-notes`."
+    should read ":ref:`chord-important-notes`." Chords are not currently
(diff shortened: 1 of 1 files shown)


alukach:
@georgepsarakis not sure why it failed, looks like a bug in the build process, would you mind restarting this job: https://travis-ci.org/celery/celery/jobs/250121385?  It's the only failed job in the build. I don't think I have permissions to restart it, however I'm guessing/hoping that a simple "try again" will fix the build.

georgepsarakis:
@alukach the build error after the base branch change is unrelated, I restarted the job.

thedrow:
Thanks for your contribution.

Issue #3436, opened 2016-09-08
Memory leak with redis broker


Wyzzcow:
celery leaks when redis becomes unavailable.
This is easily reproducible by stopping redis while celery is working.

```
from celery import Celery
from celery.task import periodic_task
from datetime import timedelta
app = Celery('tasks', broker='redis://:@localhost:6379/13')

@periodic_task(run_every=timedelta(seconds=.5))
def print_hello():
    print "hello"
```

celery -A tasks worker
celery -A tasks beat

The following screenshot shows the celery beat process when redis is stopped shortly after celery beat is started, and then restarted (memory is never freed).

![image](https://cloud.githubusercontent.com/assets/8915667/18351089/c12b4128-75a6-11e6-99a1-40452cb5d854.png)

```
amqp (1.4.9)
billiard (3.3.0.23)
celery (3.1.23)
kombu (3.0.35)
redis (2.10.5)
```

mkuchen:
+1 on this - I have...
```
billiard (3.3.0.23)
celery (3.1.23)
kombu (3.0.35)
redis (2.10.3)
```

...and I am currently experiencing the same problem with my app; a linear growth in memory usage by both my web workers and my async workers (on Heroku).  I've made a workaround by cycling the dynos every 3 hours, but this is less than optimal obviously.

Here's some compelling evidence that it's a leak in Celery delay/apply_async:
```
>>> import objgraph
>>> from my_app.tasks import add
>>> 
>>> add(1,2)
3
>>> objgraph.show_growth()
>>> add(1,2)
3
>>> objgraph.show_growth()
>>> import gc; gc.collect()
0
>>> objgraph.show_growth()
>>> add.delay(2,3)
<AsyncResult: 36bd1893-8540-40e3-a1a1-2d1a93e44b14>
>>> import gc; gc.collect()
20
>>> objgraph.show_growth()
function             30012      +371
dict                 40348      +120
tuple                23707       +93
list                 38877       +55
weakref               6820       +51
type                  4605       +46
getset_descriptor     3065       +39
property              1549       +17
cached_property        129       +15
module                1649       +13
```

@ask - do you have any troubleshooting suggestions for this?  I'm contemplating setting `CELERY_MAX_CACHED_RESULTS`, but it says in the documentation that there's a default value of 100 for that, so it shouldn't be causing the problem.

costasgambit:
There's a ref cycle in AsyncResult. This line, specifically:

[...]

PR #4131, opened 2017-07-11
Removing weak-references in promises to bound methods (Fixes #3813)

### LINKED ISSUE #3813: Redis Backend has a memory leak with AsyncResult.get()
## Checklist
       version affected).
```
            billiard:3.5.0.2 redis:2.10.5
settings -> transport:redis results:redis://[redacted]
```
### Celery Settings
```
BROKER_BACKEND = 'redis'
BROKER_URL = environ.get('REDIS_URL')
CELERY_RESULT_BACKEND = environ.get('REDIS_URL')
CELERYD_CONCURRENCY = 1
BROKER_POOL_LIMIT = 0
CELERY_TASK_RESULT_EXPIRES = 60
CELERY_MAX_CACHED_RESULTS = 1
CELERY_DISABLE_RATE_LIMITS = True
CELERY_ACCEPT_CONTENT = ['json']
CELERY_TASK_SERIALIZER = 'json'
CELERY_RESULT_SERIALIZER = 'json'
```

## Steps to reproduce
1) Create celery app with Redis as transport and Backend (same cache for both)
2) Create a task within the celery app that returns a result
``` python
@CELERY_APP.task
def add():
    return 1 + 2
```
3) Execute `add.apply_async().get()` somehow (I have a Django app on Heroku so I made an endpoint for it)
4) Use `objgraph.show_growth()` to see that `AsyncResult` objects are leaking

## Expected behavior
`AsyncResult` objects get released after the result is received after the `AsyncResult.get()` call.  The `AsyncResult` is properly released when getting the result from `AsyncResult.result`.

## Actual behavior
`AsyncResult` objects accumulate in the `AsyncResult.backend._pending_results.concrete` dict.  If the `AsyncResult` is popped from the dict, it is released as expected.
Note that the leak does not occur when executing tasks eagerly.

## Code sample
``` python
@CELERY_APP.task
def add():
    return 1 + 2

def add_request():
    response = add.apply_async().get()

    import objgraph
    objgraph.show_growth()
```

vinodc:
`AsyncResult._on_fulfilled` is never called to remove the pending result because the weakref to the function is dead-on-arrival since `_on_fulfilled` is a bound method.


vinodc:
@auvipy Where would you recommend the test case be included?

thedrow:
I suspect @ask added the weakrefs for a reason. Removing them might uncover another bug.

A test case verifying that the memory leak is gone can be written using the gc module. You can check if the gc still tracks the promise after it should have been released.

thedrow:
@vinodc I'd be more comfortable to merge this if you'll describe what's wrong and how this PR fixes it.
We do need tests that cover this or at least a test case that demonstrates the problem.

vinodc:
@thedrow More information is available here: https://github.com/celery/celery/issues/3813#issuecomment-314549985. Although we've verified the changes work great and memory leaks no longer occur, I haven't had time to determine where and how to include a unit test for this at the moment.

thedrow:
@vinodc Reproduction of this bug probably belongs to https://github.com/celery/cyanide
What I'd like to see is a reproduction of the bug you are describing and reproduction of the bug @costasgambit found as well to ensue this fix doesn't regress any of those.

@costasgambit Can you pitch in your thoughts about this change? We're not sure we understand your comment in https://github.com/celery/celery/issues/3436#issuecomment-246003104

costasgambit:
@thedrow Yes, I did find a ref cycle in #3436. My second comment was to clarify that it existed only in the master branch at the time (a Celery 4 RC) instead of the 3.X release that the OP initially complained about. 

We've stuck with Celery 3 in production, so I can't say for sure if it's fixed in the current release. FWIW, I've tried running a test instance of our latest code with Celery 4.1 and I didn't notice any leakage. I've also tried it with the patch that removes the weakrefs and I still didn't notice any leakage.

thedrow:
I'm so confused. @ask Got some time to take a look at this?

iafilatov:
I can confirm this issue on 4.1.0. Is the fix confirmed for 4.2? Anything else needed, tests etc., before it can be merged? Actually, it would be fantastic to have 4.1.1 if possible because this is a 100% reproducible when you call `AsyncResult.get()` on a running task.

georgepsarakis:
@iafilatov it seems that the issue conditions and impact need further description. At this point it would be best if you could possibly display how to reproduce through a test case.

iafilatov:
Well, first of all the change made in https://github.com/celery/celery/commit/b52d44f1a2c916ccfebb0d144589ff74ac0079b7 mean that `AsyncResult.on_ready` started using a weakref to a bound method. These are _always_ dead on arrival. [Since a bound method is ephemeral, a standard weak reference cannot keep hold of it.](https://docs.python.org/3/library/weakref.html#weakref.WeakMethod)

```python
class A(object):
    def m():
        pass

a = A()

from weakref import ref

assert ref(a)() is a
assert ref(a.m)() is None
```

As a consequence, `AsyncResult.on_fulfilled` is never called and `backend.remove_pending_result` is also never called, so if any result gets into  `backend._pending_results`, it can never get out. How does it get it there? When `AsyncResult.get` is called before the task has completed.

```python
class AsyncResult(ResultBase):

    def __init__(self, id, backend=None,
        self.backend = backend or self.app.backend
        self.on_ready = promise(self._on_fulfilled, weak=True)  # has a dead weakref to self._on_fulfilled

    def _on_fulfilled(self, result):                            # this is never called
        self.backend.remove_pending_result(self)                # <-X- backend._pending_results
        return result

    def get(self, ...):
        if self._cache:                                         # if already done
            if propagate:
            return self.result

        self.backend.add_pending_result(self)                   # ---> backend._pending_results
```

I'm not sure how to write a test (for cyanide?) at this point but I've put together a trivial app that demonstrates the problem:
```python
import gc
import os

from celery import Celery
from celery.result import AsyncResult
from gevent import sleep


celery_app = Celery('cml', backend='redis://redis/',
                           broker='amqp://guest:guest@rabbitmq/')


@celery_app.task(queue='cml')
def i_produce_large_results(s, n):
    sleep(1)
    return s * n


def print_my_mem():
    for line in open('/proc/{}/status'.format(os.getpid())):
        if line.startswith('VmRSS'):
            print(line.strip())


def print_async_results():
    num = sum(1 for o in gc.get_objects()
              if isinstance(o, AsyncResult) and o.ready())
    print('AsyncResult #: {}'.format(num))


if __name__ == '__main__':
    while True:
        res = i_produce_large_results.delay('a', 10 ** 6)
        res.get()
        assert res.ready()
        del res
        while gc.collect():
            pass
        print_async_results()
        print_my_mem()
```
```shell
# celery worker -A cml -P gevent -Q cml -l info

$  python cml.py 
AsyncResult #: 1
VmRSS:	   34364 kB
AsyncResult #: 2
VmRSS:	   37152 kB
AsyncResult #: 3
VmRSS:	   38208 kB
AsyncResult #: 4
VmRSS:	   39264 kB
AsyncResult #: 5
VmRSS:	   40056 kB
...
AsyncResult #: 97
VmRSS:	  131228 kB
AsyncResult #: 98
VmRSS:	  132020 kB
AsyncResult #: 99
VmRSS:	  133076 kB
AsyncResult #: 100
VmRSS:	  134132 kB
AsyncResult #: 101
VmRSS:	  135188 kB
...
```
The same app with this pr applied:
```
$ python cml.py 
AsyncResult #: 0
VmRSS:	   33488 kB
AsyncResult #: 0
VmRSS:	   36276 kB
AsyncResult #: 0
VmRSS:	   36376 kB
AsyncResult #: 0
VmRSS:	   36376 kB
AsyncResult #: 0
VmRSS:	   36376 kB
...
AsyncResult #: 0
VmRSS:	   36320 kB
AsyncResult #: 0
VmRSS:	   36320 kB
AsyncResult #: 0
VmRSS:	   36320 kB
AsyncResult #: 0
VmRSS:	   36320 kB
...
```

This isn't some edge case deadlock in a 1000th greenlet during Harvest Moon, this a plain coding mistake. Just like having a mutable object in function defaults, but a lot more subtle and easy to make. And it needs to be corrected. Yes, a concrete ref in `on_ready` would create a ref cycle, but we have gc for that. It the caller releases the result when it's no longer needed, everything should be fine.

thedrow:
Thank you for the analysis. I'll dedicate some time to add this to our test suite and I greentree that we'll merge this before releasing a new version.

iafilatov:
Great news, thanks!

thedrow:
Thank you very much for your patience on this PR.
This is now merged :)

iafilatov:
Thanks!

donkopotamus:
One of the commits in this merge does not look quite right.  In particular this change:

            if self._on_full:
    -            self._on_full.then(promise(self.on_ready))
    +            self._on_full.then(promise(self._on_ready))

This causes the following method to be called:

    def _on_ready(self):
        self.backend.remove_pending_result(self)
        if self.backend.is_async:
            self.on_ready()

Which will raise an exception.  The `backend.remove_pending_result` expects the value to have an `.id` and a `ResultSet` does not have an `id`.

      File "...\site-packages\celery\backends\async.py", line 175, in remove_pending_result
        self._remove_pending_result(result.id)
    AttributeError: 'ResultSet' object has no attribute 'id'

Issue #4134, opened 2017-07-13
Use of non-standard string representation prevents reproducible builds


pkmoore:
## Checklist

- [ Yes ] I have verified that the issue exists against the `master` branch of Celery.

## Steps to reproduce
Build documentation using "make docs"

## Expected behavior
Documentation is produced with non-deterministic content (i.e. memory addresses) stripped

## Actual behavior
Documentation is produced that includes memory addresses that prevent reproducible builds

I am attempting to address an issue identified by the "reproducible builds" effort (https://wiki.debian.org/ReproducibleBuilds).  The documentation generated when building Celery (via Sphinx) contains memory addresses that prevent Celery builds from being reproducible.  

Sphinx has included code to strip memory addresses in a form similar to "<__main__.A at 0x7f68cb685710>" which has fixed memory address related reproducibility issues in many cases.

Celery specifies a non-standard format here: https://github.com/celery/celery/blob/master/celery/app/utils.py#L73 .  This format results in output similar to <Celery default:0x101510d10> (with a ":" rather than " at ").  I propose change making the format used by this code match the more standard style so that Sphinx is able to strip out these memory addresses.  If this is an acceptable solution I would be glad to put together a patch to make the required change.

Thanks!

georgepsarakis:
Sounds reasonable. Could you provide an example of a documentation excerpt that this issue appears for reference? Thanks!

pkmoore:
One example is Documentation/reference/celery.app.html which contains content like:
```
<code class="descclassname">celery.app.</code>
<code class="descname">default_app</code>
<em class="property"> = &lt;Celery default:0x7ffff021e9d0&gt;</em>
<a class="headerlink" href="#celery.app.default_app" title="Permalink to this definition">¶</a>
```

The fix I propose would change the third line from the above to: 

```
<em class="property"> = &lt;Celery default at 0x7ffff021e9d0&gt;</em>
```
by adjusting the format of:
```
 return '{0}:{1:#x}'.format(app.main or '__main__', id(app))
```
to
```
 return '{0} at {1:#x}'.format(app.main or '__main__', id(app))
```
which would get picked up by existing sphinx code and modified to:

```
<em class="property"> = &lt;Celery default&gt;</em>
```
Alternatively, I could just change the format to not include the the result of id(app), i.e. something like:

```
 return '{0}'.format(app.main or '__main__')
```
and we can skip the Sphix code all together.

Which is preferred?

Thanks!

[...]

PR #4139, opened 2017-07-17
Make appstr use standard format (#4134)

pkmoore:
## Description

Fixes issue #4134 

Make appstr use a standard format so that Sphinx can correctly detect and strip memory addresses from the output documentation.  These memory addresses prevent builds from being reproducible.


georgepsarakis:
@pkmoore yes.

PR #4148, opened 2017-07-20
Broker connection uses the heartbeat setting from app config unless set otherwise

mperice:
As the commit message says, the broker connection should use the heartbeat setting from app config file unless overwritten by the keyword argument of the _connection method. This also effects connections for gossiping and mingling, where previously connections stayed opened in case of a client server crash.

### DIFF
--- a/celery/app/base.py
+++ b/celery/app/base.py
@@ -816,7 +816,7 @@ class Celery(object):
             port or conf.broker_port,
             transport=transport or conf.broker_transport,
             ssl=self.either('broker_use_ssl', ssl),
-            heartbeat=self.either('broker_heartbeat',heartbeat),
+            heartbeat=heartbeat or self.conf.broker_heartbeat,
             login_method=login_method or conf.broker_login_method,
             failover_strategy=(
                 failover_strategy or conf.broker_failover_strategy
--- a/celery/app/base.py
+++ b/celery/app/base.py
@@ -816,7 +816,7 @@ class Celery(object):
             port or conf.broker_port,
             transport=transport or conf.broker_transport,
             ssl=self.either('broker_use_ssl', ssl),
-            heartbeat=heartbeat,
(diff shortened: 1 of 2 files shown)


thedrow:
Can you please include test coverage for both cases?
The test suite currently passes which is a sign we haven't covered this correctly yet.

celery/app/base.py
```diff
@@ -816,7 +816,7 @@ def _connection(self, url, userid=None, password=None,
             port or conf.broker_port,
             transport=transport or conf.broker_transport,
             ssl=self.either('broker_use_ssl', ssl),
-            heartbeat=heartbeat,
+            heartbeat=self.either('broker_heartbeat',heartbeat),
```
thedrow: flake8 is complaining about a missing space after the `,`.

mperice:
Fixed @thedrow's suggestion and added a unit test for it.

t/unit/app/test_app.py
```diff
@@ -786,6 +786,17 @@ def my_failover_strategy(it):
...
+    def test_amqp_heartbeat_settings(self):
+        # Test default broker_heartbeat value
+        assert self.app.connection('amqp:////value').heartbeat == 0
+
+        # Test passing heartbeat through app configuration
+        self.app.conf.broker_heartbeat = 60
+        assert self.app.connection('amqp:////value').heartbeat == 60
+
+        # Test passing heartbeat as connection argument
+        assert self.app.connection('amqp:////value', heartbeat=30).heartbeat == 30
```
thedrow: flake8 is complaining again because of line length.
Please extract the heartbeat result to a variable.

thedrow:
We're currently on merge freeze for the upcoming release.
This will hit the next bugfix release.

thedrow:
Thanks! This will hit GA soon :)

PR #4153, opened 2017-07-22
Exponential backoff for autoretry

georgepsarakis:
## Description

**Note:** This is a temporary pull request for displaying proposed changes in #4101.

- [x] Convert backoff interval calculations to a function in `celery.utils.time` and add unit tests.
- [x] Simplify jitter parameters and revert to Full Jitter as used in [tenacity](https://github.com/jd/tenacity/blob/dd0c50a540785fd67daf726eec84cfe373b26a0c/tenacity/wait.py#L172) and explained in https://www.awsarchitectureblog.com/2015/03/backoff.html

### DIFF
--- a/celery/utils/time.py
+++ b/celery/utils/time.py
@@ -4,6 +4,7 @@ from __future__ import absolute_import, print_function, unicode_literals
 
 import numbers
 import os
+import random
 import sys
 import time as _time
 
@@ -27,6 +28,7 @@ __all__ = [
     'humanize_seconds', 'maybe_iso8601', 'is_naive',
     'make_aware', 'localize', 'to_utc', 'maybe_make_aware',
     'ffwd', 'utcoffset', 'adjust_timestamp',
+    'get_exponential_backoff_interval',
 ]
 
 PY3 = sys.version_info[0] == 3
@@ -383,3 +385,20 @@ def utcoffset(time=_time, localtime=_time.localtime):
 def adjust_timestamp(ts, offset, here=utcoffset):
     """Adjust timestamp based on provided utcoffset."""
     return ts - (offset - here()) * 3600
+
+
+def get_exponential_backoff_interval(
+    factor,
+    retries,
+    maximum,
+    full_jitter=False
+):
+    """Calculate the exponential backoff wait time."""
+    # Will be zero if factor equals 0
+    countdown = factor * (2 ** retries)
+    # Full jitter according to
+    # https://www.awsarchitectureblog.com/2015/03/backoff.html
+    if full_jitter:
+        countdown = random.randrange(countdown + 1)
+    # Adjust according to maximum wait time and account for negative values.
+    return max(0, min(maximum, countdown))
--- a/t/unit/utils/test_time.py
+++ b/t/unit/utils/test_time.py
@@ -3,7 +3,7 @@ import pytest
 import pytz
 from datetime import datetime, timedelta, tzinfo
 from pytz import AmbiguousTimeError
-from case import Mock
+from case import Mock, patch
 from celery.utils.time import (
     delta_resolution,
     humanize_seconds,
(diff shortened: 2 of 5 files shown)


t/unit/utils/test_time.py
```diff
@@ -253,3 +254,41 @@ def test_utcoffset(self, patching):
         assert utcoffset(time=_time) is not None
         _time.daylight = False
         assert utcoffset(time=_time) is not None
+
+
+class test_get_exponential_backoff_interval:
+
+    @patch('random.randrange', lambda n: n - 2)
```
singingwolfboy: Is there a reason why you're mocking `randrange` as `lambda n: n - 2` here, but mocking it as `lambda n: n - 1` in the `test_negative_values` test? It's not _wrong_, but it's confusing.

celery/utils/time.py
```diff
@@ -383,3 +385,20 @@ def utcoffset(time=_time, localtime=_time.localtime):
 def adjust_timestamp(ts, offset, here=utcoffset):
     """Adjust timestamp based on provided utcoffset."""
     return ts - (offset - here()) * 3600
+
+
+def get_exponential_backoff_interval(
+    factor,
+    retries,
+    maximum,
+    full_jitter=False
```
singingwolfboy: Looks like you removed the feature I described in the documentation, where you can pass `retry_jitter` an integer and it will treat it as a maximum deviation from the calculated countdown value. Is that intentional? I imagine that some applications will only want a small amount of jitter applied to their calculated value, which is why I included that feature.

t/unit/utils/test_time.py
georgepsarakis: By intuition alone, it seemed more appropriate to not use edge case values (the endpoint) here, as I would like to display that jitter **does** affect the result.

celery/utils/time.py
georgepsarakis: When I was writing this function, it seemed kind of odd that it should handle both full jitter and custom deviation. After looking at `tenacity` source code and reading the AWS blog post, I thought that perhaps retaining the simple Full Jitter algorithm, was the best choice here.

For example if you set a jitter deviation of 1:
- `factor=2, retries=1` yields a countdown of 4 , resulting in a delay in the range `[3, 5]`
- `factor=2, retries=4` yields a countdown of 32 , resulting in a delay in the range `[31, 33]`

So in the fourth retry, the jitter deviation is proportionately negligible and you may be actually increasing the possibility of e.g. HTTP calls to a service happening almost at the same time, whereas using a uniform distribution of the delay time of 32, and have let's say 10 concurrent tasks, there is a greater chance that they will be retried at a more timely manner.

```
# Full jitter
>>> sorted([random.randrange(32) for _ in range(10)])
>>> [7, 15, 16, 20, 22, 23, 25, 27, 29, 31]
# Custom deviation
>>> sorted([32 + random.randrange(2) for _ in range(10)])
>>> [32, 32, 32, 32, 32, 32, 33, 33, 33, 33]
```

I thought that it would be better to use the simple version for now, and get some user feedback. If you have an actual use case of a service that implements a different algorithm, please let me know.

singingwolfboy: No, I don't have an actual use case. I included that version because, when I first started thinking about the concept of jitter and making a pull request to Celery, that was the algorithm that I thought of. But if research shows that the "full jitter" algorithm is superior, then I'm fine with dropping my naive algorithm.

However, we'll need to update the documentation, so that it no longer refers to a feature that doesn't exist!

georgepsarakis: Yes, the documentation should be updated, I didn't make any changes until we finalized on the structure and settings. If you agree with the changes then we should somehow transfer these to your PR. Please tell me how you would like to proceed.

georgepsarakis:
Closing since #4101 was merged.

PR #4157, opened 2017-07-24
Update Celery/Python/Pypy versions in the README

samueldg:
## Description

In the `README`, updated some version numbers that were invalidated by 4.1.0 release:

- Celery bumped to 4.1;
- Python 3.6 now supported;
- Pypy bumped to 5.8, which is used for CI.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -40,10 +40,10 @@ in such a way that the client enqueues an URL to be requested by a worker.
 What do I need?
 ===============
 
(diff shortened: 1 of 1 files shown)


samueldg:
Wasn't too sure about removing 5.4 and 5.5 from the Pypy versions, but I figured it was safe since they are not included in CI anymore, and so effectively not tested against.

PR #4166, opened 2017-07-27
Fix couchdb keys

thedrow:
## Description
Addresses code review for #3861

### DIFF
--- a/t/unit/backends/test_couchdb.py
+++ b/t/unit/backends/test_couchdb.py
@@ -43,15 +43,45 @@ class test_CouchBackend:
 
         CouchBackend.get should return  and take two params
         db conn to couchdb is mocked.
-        TODO Should test on key not exists
-
         """
         x = CouchBackend(app=self.app)
         x._connection = Mock()
         get = x._connection.get = MagicMock()
-        # should return None
         assert x.get('1f3fab') == get.return_value['value']
         x._connection.get.assert_called_once_with('1f3fab')
+
+    def test_get_non_existent_key(self):
+        x = CouchBackend(app=self.app)
+        x._connection = Mock()
+        get = x._connection.get = MagicMock()
+        get.side_effect = pycouchdb.exceptions.NotFound
+        assert x.get('1f3fab') is None
+        x._connection.get.assert_called_once_with('1f3fab')
+
+    @pytest.mark.parametrize("key", ['1f3fab', b'1f3fab'])
+    def test_set(self, key):
+        x = CouchBackend(app=self.app)
+        x._connection = Mock()
+
+        x.set(key, 'value')
+
+        x._connection.save.assert_called_once_with({'_id': '1f3fab',
+                                                    'value': 'value'})
+
+    @pytest.mark.parametrize("key", ['1f3fab', b'1f3fab'])
+    def test_set_with_conflict(self, key):
+        x = CouchBackend(app=self.app)
+        x._connection = Mock()
+        x._connection.save.side_effect = (pycouchdb.exceptions.Conflict, None)
+        get = x._connection.get = MagicMock()
+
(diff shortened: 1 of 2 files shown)


thedrow:
Please hold off merging until tests are written.

georgepsarakis:
For reference, the tests should be equivalent with https://github.com/celery/celery/pull/3924 .

thedrow:
Finally got around to doing this. @georgepsarakis I'd appreciate a code review.

t/unit/backends/test_couchdb.py
```diff
@@ -30,26 +30,77 @@ def test_init_no_pycouchdb(self):
...
+    def test_set_with_conflict(self):
+        x = CouchBackend(app=self.app)
+        x._connection = Mock()
+        x._connection.save.side_effect = (pycouchdb.exceptions.Conflict, None)
+        get = x._connection.get = MagicMock()
+
+        x.set('1f3fab', 'value')
+
+        x._connection.get.assert_called_once_with('1f3fab')
+        x._connection.get('1f3fab').__setitem__.assert_called_once_with(
```
georgepsarakis: @thedrow I found this part a bit obscure. Just a suggestion then:

```
key = '1f3fab'
already_stored_data = {'_id': key, 'value': 'old_value'}
get = x._connection.get = MagicMock(return_value=already_stored_data)
updated_data = {'_id': key, 'value': 'new_value'}
x.set(key, updated_data['value'])

x._connection.get.assert_called_once_with(key)
# Dictionary should be mutated now, but I don't think we need this, because of the next assertion
assert get.return_value == updated_data
x._connection.save.assert_called_with(updated_data)
```

t/unit/backends/test_couchdb.py
```diff
@@ -30,26 +30,77 @@ def test_init_no_pycouchdb(self):
...
+        x = CouchBackend(app=self.app)
+        x._connection = Mock()
+
+        x.set(b'1f3fab', 'value')
+
+        x._connection.save.assert_called_once_with({'_id': '1f3fab',
+                                                    'value': 'value'})
+
+    def test_set_bytes_key_with_conflict(self):
+        x = CouchBackend(app=self.app)
```
georgepsarakis: These tests are identical, except for testing for the different types of key (which of course is the scope of this PR). Perhaps for the sake of DRY, we could just merge/group these cases using a loop, so instead of 4 we have 2? I think it will be less work for future readers and contributors. What do you think?

thedrow: We can parameterize the test using pytest. Good idea.

t/unit/backends/test_couchdb.py
thedrow: I think I'll keep this test as it is. There's no reason to change a test that already proves that the code works.
Feel free to issue a new PR about it if you feel strongly about it :)

ghedb:
Setting the values is now fixed but it looks like there is still an issue with getting the values from couchdb using python 3.

I don't know how to go about getting a pull request started but I believe this would fix the issue the same way that the set was fixed ( "works for me" ).

```
diff --git a/celery/backends/couchdb.py b/celery/backends/couchdb.py
index 49d26564c..1e8924d4a 100644
--- a/celery/backends/couchdb.py
+++ b/celery/backends/couchdb.py
@@ -82,7 +82,7 @@ class CouchBackend(KeyValueStoreBackend):
 
     def get(self, key):
         try:
-            return self.connection.get(key)['value']
+            return self.connection.get(bytes_to_str(key))['value']
         except pycouchdb.exceptions.NotFound:
             return None
```

georgepsarakis:
@ghedb nice finding. Would you mind opening a separate issue and just add a reference to this PR? Thanks!

Issue #4169, opened 2017-07-28
Crontab schedule delta incorrect when self.now() uses different timezone to last_run_at


DDevine:
When a different timezone is given from self.now() - which could be provided to the schedule via the `nowfun()` argument to what last_run_at has, the delta is offset by the difference between the timezones. This means that the task never executes because it is perpetually offset by a number of hours.

You may come across this issue if you are using django_celery_beat which provides a `nowfun()` with the Django timezone.

thedrow:
Closing as #4173 fixed this issue.

PR #4170, opened 2017-07-28
Make sure that last_run_at and self.now() comparable

DDevine:
Localize last_run_at for crontab schedule so that the delta between last_run_at and now as potentially given via self.nowfun are compatible.

A fix for https://github.com/celery/celery/issues/4169

### DIFF
--- a/celery/schedules.py
+++ b/celery/schedules.py
@@ -548,7 +548,7 @@ class crontab(BaseSchedule):
         tz = tz or self.tz
         last_run_at = localize(
             self.maybe_make_aware(last_run_at),
-            self.now().tzinfo)
+            self.now().tzinfo or timezone.utc)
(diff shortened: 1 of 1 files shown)


DDevine:
I think I understand the test failed...

georgepsarakis:
@DDevine can you please check #4173 ? It may be related with this issue.

DDevine:
My brain is currently melted, so I might be very wrong, but your fix might also resolve this issue. 
The logic behind my way was the idea that if nowfun was provided like in the case of django-celery-beat then the timezone from that should be used for localising the datetimes. 
Perhaps both of our fixes should be applied?

DDevine:
Testing with the latest celery master it seems intervals have been broken by #4173 - with or without my changes. I'll have to dive deep and work out where it is going wrong again.

Issue #4199, opened 2017-08-14
Celery beat keeps running after supervisor stopped it


martialp:
## Steps to reproduce
Use the `extra/supervisord/celerybeat.conf` file to daemonize celery beat
 
## Expected behavior
Supervisor should be able to start and stop celery beat

## Actual behavior
Supervisor can start beat, but when the stop signal is sent, beat keeps running.

Adding `stopasgroup=true` to celerybeat.conf fix the problem. I also noticed that the celeryd.conf file already include `killasgroup=true`. It should probably be `stopasgroup=true` too, because the latter seems to allow not only the SIGKILL, but other more graceful signals (see [supervisor documentation](http://supervisord.org/configuration.html)).

Note that `stopasgroup` didn't exist when killasgroup [has been added](https://github.com/celery/celery/commit/637b4adf7a129fcf0c0a4cdac57946e63ccf9ec3) in the celeryd.conf file. Here is the supervisor [changelog](http://supervisord.org/changes.html#b1-2012-09-10).

georgepsarakis:
Good catch! It sounds to me that perhaps both boolean options should be defined for both graceful and immediate shutdown. Would you like to open a pull request for this?

martialp:
In fact, `stopasgroup` implies `killasgroup` is true ([http://supervisord.org/configuration.html](http://supervisord.org/configuration.html)). So, `stopasgroup` alone should be fine. I will open a pull request for this!

PR #4200, opened 2017-08-14
Adds stopasgroup to the supervisor scripts

martialp:
## Description
Add `stopasgroup` config to the supervisor scripts to fix issue #4199.

### DIFF
--- a/extra/supervisord/celeryd.conf
+++ b/extra/supervisord/celeryd.conf
@@ -24,7 +24,7 @@ startsecs=10
 ; Increase this if you have very long running tasks.
 stopwaitsecs = 600
 
-; Causes supervisor to send the stop signal to the whole process group.
+; Causes supervisor to send the termination signal (SIGTERM) to the whole process group.
 stopasgroup=true
 
 ; Set Celery priority higher than default (999)
--- a/extra/supervisord/celeryd.conf
+++ b/extra/supervisord/celeryd.conf
@@ -24,10 +24,8 @@ startsecs=10
 ; Increase this if you have very long running tasks.
 stopwaitsecs = 600
 
-; When resorting to send SIGKILL to the program to terminate it
-; send SIGKILL to its whole process group instead,
-; taking care of its children as well.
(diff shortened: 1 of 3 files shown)


GabLeRoux:
:shipit: Looks good to me :)

extra/supervisord/celeryd.conf
```diff
@@ -24,10 +24,8 @@ startsecs=10
 ; Increase this if you have very long running tasks.
 stopwaitsecs = 600
 
-; When resorting to send SIGKILL to the program to terminate it
-; send SIGKILL to its whole process group instead,
-; taking care of its children as well.
-killasgroup=true
+; Causes supervisor to send the stop signal to the whole process group.
```
martialp: It is not in the documentation, but if we take a look at the `test_stopasgroup` function in the supervisor [sources](https://github.com/Supervisor/supervisor/blob/master/supervisor/tests/test_process.py), we can see that SIGTERM signal is used.

georgepsarakis: Thanks for looking into this. Do you mind updating this comment to use `send the termination` or `send the SIGTERM`?

georgepsarakis:
This is ready for merge. If you would like please add yourself to the list of [contributors](https://github.com/celery/celery/blob/master/CONTRIBUTORS.txt).

cadl:
When `stopasgroup` enabled, supervisord will send TERM signal to the worker and it's forks. The forks will raise a "WorkerLostError: Worker exited prematurely: signal 15 (SIGTERM)"  and lost the task(if `reject_on_worker_lost` is false).

 Maybe the default should be false?

georgepsarakis:
@cadl As the documentation on [task_reject_on_worker_lost](http://docs.celeryproject.org/en/latest/userguide/configuration.html#std:setting-task_reject_on_worker_lost) states, this is an issue with terminating the worker in general, I don't think it is caused by stopping the forked processes, if that is what you mean.

PR #4204, opened 2017-08-15
Remove unused code from _create_app contextmanager

rpkilby:
Hello. As far as I can tell, `app_contained` is unused and undocumented, so the corresponding `is_not_contained` check should be unnecessary/unreachable. 

I did a little bit of digging, and it looks like this is a remnant of the old test case class, which had a [`self.contained`](https://github.com/celery/celery/blob/3eb34cbf2cd938375b3c6711d5d44e965543c71b/celery/tests/case.py#L170-L171) check. The conversion to py.test in 29df527147514b3eab7fbc3dbf41ff181bf1dcd9 renamed  `contained` to `app_contained`, but it has otherwise been unused.

### DIFF
--- a/celery/contrib/pytest.py
+++ b/celery/contrib/pytest.py
@@ -13,8 +13,7 @@ NO_WORKER = os.environ.get('NO_WORKER')
 
 
 @contextmanager
-def _create_app(request,
-                enable_logging=False,
+def _create_app(enable_logging=False,
                 use_trap=False,
                 parameters={},
                 **config):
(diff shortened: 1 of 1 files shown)


rpkilby:
@auvipy - I'm happy to add one, but I'm not entirely sure what that test would look like.

rpkilby:
The difficulty here is that this would test unused code that is then removed by this PR. Once the code is removed, the test will no longer be relevant. Ultimately, the question is whether the conditional's [inner block](https://github.com/celery/celery/blob/v4.1.0/celery/contrib/pytest.py#L39-L40) affects the integration tests. The best that I can do is show that it is not relevant.
- This PR shows that the tests pass *without* the inner block executing. 
- I'll create a separate PR that demonstrates the tests passing *with* the inner block executing.

rpkilby:
Just ran #4211 successfully, so as far as I can tell `test_app.set_current()` has no effect here.

berkerpeksag:
Thanks!

PR #4205, opened 2017-08-16
Fixes #4106 - change GroupResult:as_tuple() to include parent

### LINKED ISSUE #4106: GroupResult.as_tuple forget serialize group parent
```
celery -A ct report

software -> celery:4.0.2 (latentcall) kombu:4.0.2 py:2.7.9
settings -> transport:amqp results:cache+memcached://127.0.0.1:11211/

result_backend: u'cache+memcached://127.0.0.1:11211/'

```

**Steps to reproduce**

```
# ct.py

# ct.py
from celery import Celery

app = Celery(
    'ct',
    broker='amqp://guest@localhost//',
    backend='cache+memcached://127.0.0.1:11211/'
)

@app.task
def nothing(*args):
   return args

if __name__ == '__main__':
    app.start()
```

Start worker

`python ct.py worker -l INFO`

In `python` console

```
>>> from celery import chain, group
>>> from celery.result import result_from_tuple
>>> from ct import nothing
>>> c = nothing.s() | group(nothing.s(), nothing.s())
>>> r = c.delay()
>>> r
<GroupResult: f46d7bcc-3140-40e8-9e21-d2f74eef6f28 [4854af6e-8e4a-48d8-98e9-d5f0eefcce20, 5c8d4ec1-84ff-4845-a9b2-9f07fd9bed19]>
>>> r.parent
<AsyncResult: a359ae60-ed18-4876-a67b-fc8cd83759a2>
>>> t = r.as_tuple()
>>> r2 = result_from_tuple(t)
>>> r2
<GroupResult: f46d7bcc-3140-40e8-9e21-d2f74eef6f28 [4854af6e-8e4a-48d8-98e9-d5f0eefcce20, 5c8d4ec1-84ff-4845-a9b2-9f07fd9bed19]>
>>> r2.parent

>>> r2.parent is None
True
```

**Expected behavior**

```
>>> r2 = result_from_tuple(t)
>>> r2.parent
<AsyncResult: a359ae60-ed18-4876-a67b-fc8cd83759a2>
>>>
```

pachewise:
## Description

`GroupResult` doesn't include the parent in `as_tuple()`, though sometimes it does have a parent (see issue #4106 for an example).

### DIFF
--- a/celery/result.py
+++ b/celery/result.py
@@ -815,6 +815,7 @@ class GroupResult(ResultSet):
     Arguments:
         id (str): The id of the group.
         results (Sequence[AsyncResult]): List of result instances.
+        parent (ResultBase): Parent result of this group.
     """
 
     #: The UUID of the group.
@@ -823,8 +824,12 @@ class GroupResult(ResultSet):
     #: List/iterator of results in the group
     results = None
 
-    def __init__(self, id=None, results=None, **kwargs):
+    #: Parent Result of the group, if any
+    parent = None
+
+    def __init__(self, id=None, results=None, parent=None, **kwargs):
         self.id = id
+        self.parent = parent
         ResultSet.__init__(self, results, **kwargs)
 
     def save(self, backend=None):
@@ -865,7 +870,7 @@ class GroupResult(ResultSet):
                                          ', '.join(r.id for r in self.results))
 
     def as_tuple(self):
-        return self.id, [r.as_tuple() for r in self.results]
+        return (self.id, self.parent), [r.as_tuple() for r in self.results]
 
     @property
     def children(self):
@@ -968,14 +973,14 @@ def result_from_tuple(r, app=None):
     app = app_or_default(app)
     Result = app.AsyncResult
     if not isinstance(r, ResultBase):
-        res, nodes = r
-        if nodes:
-            return app.GroupResult(
-                res, [result_from_tuple(child, app) for child in nodes],
-            )
-        # previously didn't include parent
-        id, parent = res if isinstance(res, (list, tuple)) else (res, None)
+        (id, parent), nodes = r
         if parent:
             parent = result_from_tuple(parent, app)
+
+        if nodes:
+            return app.GroupResult(
+                id, [result_from_tuple(child, app) for child in nodes], parent=parent,
+            )
+
         return Result(id, parent=parent)
     return r
--- a/celery/result.py
+++ b/celery/result.py
@@ -824,9 +824,6 @@ class GroupResult(ResultSet):
     #: List/iterator of results in the group
     results = None
 
-    #: Parent Result of the group, if any
-    parent = None
-
     def __init__(self, id=None, results=None, parent=None, **kwargs):
         self.id = id
         self.parent = parent
@@ -858,7 +855,11 @@ class GroupResult(ResultSet):
 
     def __eq__(self, other):
         if isinstance(other, GroupResult):
-            return other.id == self.id and other.results == self.results
+            return (
+                other.id == self.id and
+                other.results == self.results and
+                other.parent == self.parent
+            )
         return NotImplemented
 
     def __ne__(self, other):
--- a/celery/result.py
+++ b/celery/result.py
@@ -980,7 +980,8 @@ def result_from_tuple(r, app=None):
 
         if nodes:
             return app.GroupResult(
-                id, [result_from_tuple(child, app) for child in nodes], parent=parent,
+                id, [result_from_tuple(child, app) for child in nodes],
+                parent=parent,
             )
 
         return Result(id, parent=parent)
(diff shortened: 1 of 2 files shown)


pachewise:
FYI, gonna squash these commits before I merge

t/unit/tasks/test_result.py
```diff
@@ -892,3 +892,12 @@ def test_GroupResult(self):
...
         assert x, result_from_tuple(x == self.app)
+
+    def test_GroupResult_with_parent(self):
+        parent = self.app.AsyncResult(uuid())
+        result = self.app.GroupResult(
+            uuid(), [self.app.AsyncResult(uuid()) for _ in range(10)],
+            parent
+        )
+        second_result = result_from_tuple(result.as_tuple(), self.app)
+        assert second_result.parent == parent
```
georgepsarakis: I would suggest to add an assertion on the task ID as well.

celery/result.py
```diff
@@ -969,13 +974,15 @@ def result_from_tuple(r, app=None):
...
-            return app.GroupResult(
-                res, [result_from_tuple(child, app) for child in nodes],
-            )
-        # previously didn't include parent
         id, parent = res if isinstance(res, (list, tuple)) else (res, None)
         if parent:
             parent = result_from_tuple(parent, app)
+
+        if nodes:
+            return app.GroupResult(
```
georgepsarakis: Did you happen to locate other instantiations of `GroupResult`? Perhaps the `parent` argument should be added in more calls.

georgepsarakis:
@pachewise we can squash them upon merge, so don't worry about this.

t/unit/tasks/test_result.py
pachewise: taskID of the parent? or the second_result?

georgepsarakis: Sorry, the equality checks for the ID as well. How about this assertion then:

```
assert second_result == result
```

pachewise: GroupResult's == is implemented as:

````python
    def __eq__(self, other):
        if isinstance(other, GroupResult):
            return other.id == self.id and other.results == self.results
        return NotImplemented
````

Wondering if we should include the parents in the comparisons, too?

File: celery/result.py (hunk 2 above)
Comment (pachewise) [inline celery/result.py]: screen shot 2017-08-16 at 7 31 50 pm

first one and last four might need to be changed.

File: celery/result.py [hunk 3]
```diff
@@ -823,8 +824,12 @@ class GroupResult(ResultSet):
     #: List/iterator of results in the group
     results = None
 
-    def __init__(self, id=None, results=None, **kwargs):
+    #: Parent Result of the group, if any
+    parent = None
```
pachewise: ResultBase already defines `parent = None`. Wondering if we should get rid of L818/L828.

t/unit/tasks/test_result.py
georgepsarakis: Sounds right.

celery/result.py
georgepsarakis: Yes, it seems to be a class variable in `ResultBase`, so I would keep 818, but remove 828 as it is redundant.

celery/result.py
georgepsarakis: I have done some checking on `as_tuple` and `result_from_tuple` calls as well, I believe the changes do not affect these.

celery/result.py
```diff
@@ -853,7 +855,11 @@ def __bool__(self):
 
     def __eq__(self, other):
         if isinstance(other, GroupResult):
-            return other.id == self.id and other.results == self.results
+            return (
```
georgepsarakis: Perhaps add two simple test cases: one for `__eq__` and one for `as_tuple`. What do you think?

pachewise:
Would the `__eq__` case be creating two GroupResults with the same args, and comparing them with `==`, and the as_tuple() one be what I have now?

pachewise:
@georgepsarakis added a test for `__eq__`; let me know if you had something different in mind.

georgepsarakis:
@pachewise looks great! I was thinking something very simple for testing `as_tuple` like the following (avoid regressions):

```
def test_GroupResult_as_tuple(self):
      parent = self.app.AsyncResult(uuid())
      result = self.app.GroupResult(
            'group-result-1',            
             [self.app.AsyncResult('async-result-{}'.format(index) for index in range(2)],
             parent
             )
      (result_id, parent), group_results = result.as_tuple()
      assert result_id == result.id
      assert parent is parent
      assert isinstance(group_results, list)
      assert group_results == [(('async-result-0', None), None), (('async-result-1', None), None)]
```

Tell me what you think.

pachewise:
@georgepsarakis looks good, will add it

pachewise:
@georgepsarakis any other feedback?

georgepsarakis:
@pachewise thanks for the contribution. Please feel free to add yourself in the [contributors list](https://github.com/celery/celery/blob/master/CONTRIBUTORS.txt).

tbazadaykin:
When do you plan to add it to the release?

PR #4217, opened 2017-08-20
Prefer dict literal syntax over dict()

atombrella:
https://doughellmann.com/blog/2012/11/12/the-performance-impact-of-using-dict-instead-of-in-cpython-2-7-2/

It's mostly in things that appear to be called few times, but still seems like a good idea.


georgepsarakis:
@atombrella once you are finished, notify me to have a quick second look. Thanks!

atombrella:
@georgepsarakis I'm finished :) I checked with flake8-comprehensions, and replaced all of them.

thedrow:
This is awesome. Thank you!

PR #4224, opened 2017-08-22
add cassandra_options

scttcper:
## Description
Adds back cassandra_options that was available in celery v3.1
https://github.com/celery/celery/blob/d08b1057234bb7774623108fa343af7a1c658e7a/celery/backends/cassandra.py#L81-L82  
There's a bunch of kwargs that can be passed on cluster creation, currently only defaults can be used.
http://datastax.github.io/python-driver/api/cassandra/cluster.html


thedrow:
Thanks for the patch :)

PR #4227, opened 2017-08-25
fix crontab description of month_of_year behaviour

erebus1:
## Description
In [docs](http://docs.celeryproject.org/en/latest/userguide/periodic-tasks.html#crontab-schedules) said:

> crontab(0, 0, month_of_year='*/3') -> Execute on the first month of every quarter.

Which is a bit confused, because, in fact it will run task each day on the first month of every quarter.


docs/userguide/periodic-tasks.rst
```diff
@@ -273,8 +273,8 @@ Some examples:
 | ``crontab(0, 0, day_of_month='11',``    | Execute on the eleventh of May every year. |
 |          ``month_of_year='5')``         |                                            |
 +-----------------------------------------+--------------------------------------------+
-| ``crontab(0, 0,``                       | Execute on the first month of every        |
-|         ``month_of_year='*/3')``        | quarter.                                   |
+| ``crontab(0, 0,``                       | Execute each day(!) on the first month     |
```
thedrow: Is there a need for the exclamation mark?

georgepsarakis: I think `each` should be replaced by `every`.

georgepsarakis:
@erebus1 can you please address the review comments?

erebus1:
@georgepsarakis sorry for so long time waiting, I missed this comments
I've updated PR

PR #4228, opened 2017-08-25
log WARN on missed crontab parameter

erebus1:
## Description

When we initialise crontab in such manner: crontab(hour=5), we do not have any warnings, but as a result we have really confused behaviour: task will run each minute from 5 till 6 o'clock

This logging will prevent incorrect crontab configuration

### DIFF
--- a/celery/schedules.py
+++ b/celery/schedules.py
@@ -430,8 +430,7 @@ class crontab(BaseSchedule):
             if unit in defined_units:
                 last_defined_unit = unit
             elif last_defined_unit:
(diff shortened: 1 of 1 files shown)


thedrow:
I agree it is surprising behavior.
Crontab behaves this way so I guess that's why it was emulated here.

thedrow:
There should be a way to disable this warning though.
Some people will want this behavior.

thedrow:
I agree. Thanks for the contribution anyway :)

erebus1:
Ok, thanks for take time for this PR

PR #4235, opened 2017-08-30
Beat: support scheduler entries without a schedule

mkai:
I'm using custom scheduler entries with an overridden ``is_due()`` method that do not require a schedule (``entry.schedule``) to be set. Judging from the ``entry._default_now()`` method, this seems to be supported by the scheduler in Celery 4.1, however that method is not used everywhere, leading to an AttributeError in my use case.

This change just uses ``_default_now()`` everywhere which fixes the issue. It shouldn't have any impact on the default implementation.

### DIFF
--- a/celery/beat.py
+++ b/celery/beat.py
@@ -97,17 +97,17 @@ class ScheduleEntry(object):
         self.kwargs = kwargs
         self.options = options
         self.schedule = maybe_schedule(schedule, relative, app=self.app)
-        self.last_run_at = last_run_at or self._default_now()
+        self.last_run_at = last_run_at or self.default_now()
         self.total_run_count = total_run_count or 0
 
-    def _default_now(self):
+    def default_now(self):
         return self.schedule.now() if self.schedule else self.app.now()
 
     def _next_instance(self, last_run_at=None):
         """Return new instance, with date and count fields updated."""
         return self.__class__(**dict(
             self,
-            last_run_at=last_run_at or self._default_now(),
+            last_run_at=last_run_at or self.default_now(),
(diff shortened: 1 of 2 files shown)


celery/beat.py
```diff
@@ -237,7 +237,7 @@ def is_due(self, entry):
     def _when(self, entry, next_time_to_run, mktime=time.mktime):
         adjust = self.adjust
 
-        return (mktime(entry.schedule.now().timetuple()) +
+        return (mktime(entry._default_now().timetuple()) +
```
georgepsarakis: This is a protected method. You should probably create instead a public method or getter in `ScheduleEntry`.

mkai: Agreed. But a public method or getter would just duplicate (or call) ``_default_now``. So how about just making that public, i.e. merely removing the underscore and slightly refactoring?

celery/beat.py
```diff
@@ -97,17 +97,17 @@ def __init__(self, name=None, task=None, last_run_at=None,
         self.kwargs = kwargs
         self.options = options
         self.schedule = maybe_schedule(schedule, relative, app=self.app)
-        self.last_run_at = last_run_at or self._default_now()
+        self.last_run_at = last_run_at or self.default_now()
         self.total_run_count = total_run_count or 0
 
-    def _default_now(self):
+    def default_now(self):
```
georgepsarakis: Since this was a protected method, other people may have already been using it in sub-classes. This was the main reason I suggested that you add a new method or property getter.

georgepsarakis:
@mkai thanks for your contribution!

Issue #2275, opened 2014-09-23
Attempting to get task data from backend when `CELERY_ALWAYS_EAGER=True` should raise an exception


turtlemonvh:
## Short description

Getting the result of a task according to [the FAQ instructions](http://docs.celeryproject.org/en/v2.2.5/faq.html#how-do-i-get-the-result-of-a-task-if-i-have-the-id-that-points-there) fails in a non-obvious way when `CELERY_ALWAYS_EAGER` is `True`.

## Full description

In `celery.app.task.Task.apply_async`, if `CELERY_ALWAYS_EAGER` is `True`, `Task.apply` is called instead of `app.base.Celery.send_task`.  `send_task` hits the backend store (the queue) but `apply` just runs the task locally.

*This is fine* (but could be improved upon - see "Solution Option 3", below).  It makes sense that retrieving the status of a task from the backend would not be possible if the backend is never touched.

Unfortunately, functions for checking state fail in a non-obvious way that can be confusing to users.

The base function (see `celery.backends.database.DatabaseBackend._get_task_meta_for`) defaults to returning a new `Task` object with `status = states.PENDING` and `result = None` if nothing is found in the backend. This function is called by `BaseBackend.get_task_meta`, which is called by `BaseBackend.get_status`, which is the function that is used in the `AsyncResult.state` property to get task state, and also in [`AsyncResult.ready()`](http://celery.readthedocs.org/en/latest/reference/celery.result.html#celery.result.AsyncResult.ready).

Thus, for tasks which the system does not know about, the result of a query for their state is not an `Exception`, but a new object with a valid state and empty results field.

### Solution Option 1

I think the most straightforward solution is to be explicit and raise an error if the user attempt to check the state of a task via the backend when `CELERY_ALWAYS_EAGER` is `True`.

This could be fixed by raising an exception (possibly `QueueNotFound`, maybe a new exception) in `celery.backends.base.BaseBackend.get_task_meta` if `CELERY_ALWAYS_EAGER` is `True`.  For example:

    if self.conf.CELERY_ALWAYS_EAGER:
        raise QueueNotFound("Cannot access backend when CELERY_ALWAYS_EAGER is True")

Without this change, [this FAQ response](http://docs.celeryproject.org/en/v2.2.5/faq.html#how-do-i-get-the-result-of-a-task-if-i-have-the-id-that-points-there) can be a bit confusing:

> How do I get the result of a task if I have the ID that points there?
> 
> Answer: Use Task.AsyncResult:

[...]

PR #4239, opened 2017-08-31
Changed raise RuntimeError to RuntimeWarning when task_always_eager is True

harukaeru:
## Description
This PR changes RuntimeError to RuntimeWarning when `task_always_eager` is True.

Sometimes I tested what is related to celery task using `task_always_eager` in celery v3.
(The tests are not only celery task but codes using celery task and other codes)

I know the config cannot be completely tested async task but it's useful when doing rush works.
I never use them in the production code but in the test, I use.

I read the issue (https://github.com/celery/celery/issues/2275) was produced and the commit (https://github.com/celery/celery/commit/c71cd08fc72742efbfc846a81020939aa3692501) resolved the above.

I almost agree with them but people who want to test perfectly only don't turn on `task_always_eager`.
Others who want to test synchronously also want to use `task_always_eager`.

### DIFF
--- a/celery/backends/base.py
+++ b/celery/backends/base.py
@@ -10,6 +10,7 @@ from __future__ import absolute_import, unicode_literals
 
 import sys
 import time
+import warnings
 
 from collections import namedtuple
 from datetime import timedelta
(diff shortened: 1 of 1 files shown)


auvipy:
could you check for relevant travis failures?

PR #4240, opened 2017-08-31
Care about chain's link_error tasks

### LINKED ISSUE #4232: Joining to chains loses link_error of least chain
https://github.com/celery/celery/blob/cbbf481801079f0e2cfbfe464c9ecfe3ccc7a067/celery/canvas.py#L408-L414

Should be something like
```python
    link_error_sigs = other._with_list_option('link_error')
    sig.tasks.extend(
        reduce(
            lambda t, s: t.on_error(s), link_error_sigs, t.clone())
        for t in other.tasks)
```

agladkov:
Fixes #4232
Chain could have link_error signatures for error processing.
Joining chains copies tasks from other chain to original one.
Thus copying loses other chain's link_error signatures.
Assigning chain's link_error signatures to each task could have the same effect.
Each task from other chain are cloned to leave original ones as is.
In appending task to chain clone full chain's state and append task to chain's tasks.

### DIFF
--- a/celery/canvas.py
+++ b/celery/canvas.py
@@ -404,13 +404,24 @@ class Signature(dict):
         if not isinstance(self, _chain) and isinstance(other, _chain):
             # task | chain -> chain
             return _chain(
-                seq_concat_seq((self,), other.tasks), app=self._app)
+                seq_concat_seq((self,), [
+                    reduce(
+                        lambda t, s: t.on_error(s),
+                        other._with_list_option('link_error'),
+                        t.clone())
+                    for t in other.tasks]),
+                    app=self._app)
         elif isinstance(other, _chain):
             # chain | chain -> chain
             sig = self.clone()
             if isinstance(sig.tasks, tuple):
                 sig.tasks = list(sig.tasks)
-            sig.tasks.extend(other.tasks)
+            # assign chain's link_error sugnatures to each chain's task
+            sig.tasks.extend(reduce(
+                    lambda t, s: t.on_error(s),
+                    other._with_list_option('link_error'),
+                    t.clone())
+                for t in other.tasks)
             return sig
         elif isinstance(self, chord):
             # chord(ONE, body) | other -> ONE | body | other
--- a/celery/canvas.py
+++ b/celery/canvas.py
@@ -402,7 +402,7 @@ class Signature(dict):
                             self._with_list_option('link_error'),
                             t.clone())
                         for t in self.tasks], other),
-                        app=self._app)
+                    app=self._app)
             # task | group() -> chain
             return _chain(self, other, app=self.app)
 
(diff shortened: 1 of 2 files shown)


celery/canvas.py
```diff
@@ -395,23 +395,41 @@ def __or__(self, other):
             other = maybe_unroll_group(other)
             if isinstance(self, _chain):
                 # chain | group() -> chain
-                sig = self.clone()
-                sig.tasks.append(other)
-                return sig
+                return _chain(
+                    seq_concat_item([
+                        reduce(
```
georgepsarakis: Can you please explain the logic here? I cannot figure out the use of `reduce`. It seems that the concept is to have the next task signature become the error callback?

agladkov: With `reduce` I'm assigning tasks from `link_error` of current chain to individual tasks with respect to their order.
This code equals to:

    new_tasks = []
    for task in self.tasks:
      new_task = task.clone()
      new_tasks.append(new_task)
      for error_handler in self._with_list_option('link_error'):
        new_task.on_error(error_handler)
    return _chain(seq_concat_item(new_tasks, other), app=self._app)`

georgepsarakis: I would go with this version since it is far more readable. Also, I think that using `_with_list_option` is not required.

```
tasks = [task.clone() for task in self.tasks]
error_handler = self.options.get('link_error')
if error_handler is not None:   
    for task in tasks:
        task.on_error(error_handler)
```

What do you think?

agladkov: OK, I'll update my branch.
I've used `_with_list_option` because `link_error` is a list of tasks and there are some logic on setting it in case of `MutableSequence`.
But here we just reading it, so I can use `.options.get('link_error', [])`

celery/canvas.py
```diff
@@ -395,23 +395,37 @@ def __or__(self, other):
             other = maybe_unroll_group(other)
             if isinstance(self, _chain):
                 # chain | group() -> chain
-                sig = self.clone()
-                sig.tasks.append(other)
-                return sig
+                tasks = [t.clone() for t in self.tasks]
+                link_error = self.options.get('link_error', [])
```
georgepsarakis: Glad you agreed with the suggestions :smile: ! One question here, is `link_error` a list of functions?

agladkov: Yes, `link_error` is list of signatures.
I've wrote it on your code suggestion ;)

celery/canvas.py
```diff
@@ -395,23 +395,37 @@ def __or__(self, other):
...
-                sig.tasks = list(sig.tasks)
-            sig.tasks.extend(other.tasks)
-            return sig
+            # assign chain's link_error sugnatures to each chain's task
+            tasks = [t.clone() for t in self.tasks]
+            link_error = self.options.get('link_error', [])
+            for sig in link_error:
+                for task in tasks:
+                    task.on_error(sig)
+            other_tasks = [t.clone() for t in other.tasks]
```
georgepsarakis: Seems this is a repeated pattern, the more DRY approach would be to use a function for these 5 occurrences.

agladkov: I've though about it when applying your suggestion :)
Wait for one more commit.

thedrow:
Thanks!

Issue #4242, opened 2017-08-31
Failed test execution due to inconsistent boto3 version requirements


agladkov:
Double requirement given: boto3==1.4.3 (from -r /home/travis/build/celery/celery/requirements/extras/dynamodb.txt (line 1)) (already in boto3>=1.4.6 (from -r /home/travis/build/celery/celery/requirements/extras/sqs.txt (line 1)), name='boto3')

thedrow:
This is fixed in #4243.

PR #4237, opened 2017-08-30
Upgrade boto3 version for DynamoDB to 1.4.6

georgepsarakis:
- [x] Fix Travis CI with multiple versions of `boto3`.

PR #4243, opened 2017-08-31
Fix required boto3 version

agladkov:
Fixes issue #4242

Update boto3 version for dynamodb requirements


georgepsarakis:
I had already opened #4237 but this is a much simpler solution, thanks!

PR #4251, opened 2017-09-06
Add new function to handle etas and limits together

### LINKED ISSUE #3888: Rate limit ignored when ETA specified
While I accept this is a limitation, it's not a very clear one from the documentation. It would be good to add it as a note/warning somewhere to clarify that ETA and rate limit are incompatible.

http://docs.celeryproject.org/en/latest/internals/worker.html#id1
http://docs.celeryproject.org/en/latest/userguide/tasks.html?highlight=delay#Task.rate_limit

## Checklist

## Steps to reproduce

This is explained [here](http://stackoverflow.com/questions/30804857/celery-rate-limit-not-respected-when-scheduling-via-eta-option).

## Expected behavior

I was expecting that the task would be delayed as per the ETA, and then be executed sometimes after that, respecting the rate limit.

## Actual behavior

The rate limit is ignored entirely by Celery.

arpanshah29:
Fixes #3888

Allows rate limit to be specified along with ETA.

**Behavior**

Task would be delayed as per the ETA, and then be executed sometimes after that, respecting the rate limit.

### DIFF
--- a/celery/worker/consumer/consumer.py
+++ b/celery/worker/consumer/consumer.py
@@ -305,6 +305,12 @@ class Consumer(object):
             return bucket.add(request)
         return self._schedule_bucket_request(request, bucket, tokens)
 
+    def _limit_post_eta(self, request, bucket, tokens):
+        self.qos.decrement_eventually()
+        if bucket.contents:
+            return bucket.add(request)
+        return self._schedule_bucket_request(request, bucket, tokens)
+
     def start(self):
         blueprint = self.blueprint
         while blueprint.state not in STOP_CONDITIONS:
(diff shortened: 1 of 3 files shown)


ask:
I think the build is failing for some unrelated issue.

@thedrow I looked at this, and it looks good to me.  We tested it manually, and it seems to work. Merge?

thedrow:
Manual testing is fine and all but we need automated tests to ensure this doesn't break.
@ask I'm OK with merging this as long as we have an issue indicating that this requires unit/integration tests.
@arpanshah29 Is it possible for you to provide unit tests and integration tests for this PR?

arpanshah29:
@thedrow I added a unit test to cover this case. I am holding off on integration tests after discussing with @ask it seems that it is unclear how to precisely predict what the deterministic observable behavior would be in a test.

Does this seem good to merge?

PR #4262, opened 2017-09-12
Disable backend for beat

### LINKED ISSUE #4261: celery beat subscribes to task meta
## Checklist

       version affected).

```
software -> celery:4.1.0 (latentcall) kombu:4.1.0 py:3.6.1
            billiard:3.5.0.3 py-amqp:2.2.1
settings -> transport:amqp results:redis://redis:6379/
```

## Steps to reproduce

1. Configure a result_backend and set up a beat_schedule that includes a task
1. Run `celery beat`

## Expected behavior

1. Tasks are started and there is no traffic between `beat` and the result backend

`beat` shouldn't need to communicate with the backend because it doesn't do anything with the results

## Actual behavior

1. `beat` subscribes to the meta on the backend server for every task that it creates. Using `tcpdump`, I can actually watch `SUBSCRIBE` messages being sent for every task, regardless of the `ignore_result` setting.

This is actually a larger issue due to #3812 - the `beat` process does not unsubscribe ever, and at a certain point begins to send massive subscription requests to the Redis server. We saw giant spikes in outbound traffic from beat which seemed fishy, and traced it to this issue.

## Workaround

The way we have worked around this in the meantime is to have a separate config for `beat` that doesn't have a `result_backend` set.

jmaroeder:
## Description

When using `beat` (at least in its standalone form), it is not necessary to subscribe to events on the result backend. These subscriptions happen in the `on_task_call` method of the backend. This PR ensures that no `SUBSCRIBE` messages are sent.

Fixes #4261

### DIFF
--- a/celery/bin/beat.py
+++ b/celery/bin/beat.py
@@ -66,10 +66,10 @@
 """
 from __future__ import absolute_import, unicode_literals
 from functools import partial
-from types import MethodType
+
+from celery.backends.base import DisabledBackend
 from celery.platforms import detached, maybe_drop_privileges
 from celery.bin.base import Command, daemon_options
-from celery.utils import noop
 
 __all__ = ['beat']
 
@@ -101,8 +101,8 @@ class beat(Command):
         kwargs.pop('app', None)
         beat = partial(self.app.Beat,
                        logfile=logfile, pidfile=pidfile, **kwargs)
-        # standalone beat does not care about results, so replace on_task_call
-        self.app.backend.on_task_call = MethodType(noop, self.app.backend)
+        # standalone beat does not care about results, so disable backend
+        self.app.backend = DisabledBackend(self.app)
 
         if detach:
             with detached(logfile, pidfile, uid, gid, umask, workdir):
--- a/celery/bin/beat.py
+++ b/celery/bin/beat.py
@@ -66,8 +66,10 @@
 """
 from __future__ import absolute_import, unicode_literals
 from functools import partial
+from types import MethodType
 from celery.platforms import detached, maybe_drop_privileges
 from celery.bin.base import Command, daemon_options
+from celery.utils import noop
 
 __all__ = ['beat']
 
@@ -99,6 +101,8 @@ class beat(Command):
         kwargs.pop('app', None)
         beat = partial(self.app.Beat,
                        logfile=logfile, pidfile=pidfile, **kwargs)
+        # standalone beat does not care about results, so replace on_task_call
+        self.app.backend.on_task_call = MethodType(noop, self.app.backend)
 
         if detach:
             with detached(logfile, pidfile, uid, gid, umask, workdir):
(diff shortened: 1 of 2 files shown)


celery/bin/beat.py
```diff
@@ -99,6 +101,8 @@ def run(self, detach=False, logfile=None, pidfile=None, uid=None,
         kwargs.pop('app', None)
         beat = partial(self.app.Beat,
                        logfile=logfile, pidfile=pidfile, **kwargs)
+        # standalone beat does not care about results, so replace on_task_call
+        self.app.backend.on_task_call = MethodType(noop, self.app.backend)
```
georgepsarakis: Would it be more straightforward to disable the `backend` altogether perhaps? See below:

```
from celery.backends.base import DisabledBackend

self.app.backend = DisabledBackend()
```

thedrow:
This might be a problem for people who do want to save results in a database but they don't consume them in celery beat.
Also, what happens if a canvas is scheduled? You'd need results.

I'm not 100% sure this PR is valid.
Can we think of the possible use-cases here?

celery/bin/beat.py
jamesmallen: I didn't realize there was that - that might be a less hacky way to resolve this issue instead of monkeypatching the `on_task_call` method

jamesmallen:
@thedrow celery beat (when run as a standalone) is unable to do anything with the results - I believe the result backend is a setting on the worker instance(s).

If you schedule a canvas, I believe that everything necessary to determine the subsequent flow is embedded in in the task message (http://docs.celeryproject.org/en/latest/internals/protocol.html) - in protocol v2, this shows up in the `callbacks`, `errbacks`, `chain`, and `chord` attributes.

The reason for this PR is that a lot of chatter is created that serves no purpose - `beat` doesn't do anything with the result messages other than subscribe to them, putting unnecessary strain on the backend server and the beat instance.

celery/bin/beat.py
jamesmallen: I've implemented the suggested change and added a test to ensure the backend has been replaced

georgepsarakis:
@thedrow @auvipy any thoughts on this? I think it is valid.

thedrow:
Not sure, I've been meaning to call @ask about it. I'll try to catch him this week.

auvipy:
I am also not quite sure about the change right at the moment..

georgepsarakis:
@jamesmallen we believe we have resolved certain issues concerning the Redis result backend:
- https://github.com/celery/celery/pull/4709
- https://github.com/celery/celery/pull/4666
The fixes have been included in the latest release.

As you pointed out in #4261 this was the major issue, so perhaps this PR is no longer required.
What do you think?

auvipy:
one reasoning to get this could be, separating the concern of beat and result back end in stand alone mode, but what if not in stand alone mode?

auvipy:
This can be done on projects need.

PR #4280, opened 2017-09-23
Allow to create group with single task (fixes issue #4255)

### LINKED ISSUE #4255: [BUG] Exception on single task in group.
## Checklist
```
// I have included the output of ``celery -A proj report`` in the issue.
//      (if you are not able to do this, then at least specify the Celery
//       version affected).
//
// Doesn't make sense as it client-side issue.
```

## Steps to reproduce
On client send a group with single task in it
```python3.6
def fta_add_to_pool(applications_id: List[int]):
    return group(
        *[
            app.signature('fta_tasks.add_to_pool', args=[application_id])
            for application_id in applications_id
        ]
    ).apply_async().get()
```

## Expected behavior
I expected group to execute with a single task like there was no group.

## Actual behavior
Exception:
```
  File "***/src/celery/celery/canvas.py", line 976, in apply_async
    app = self.app
  File "***/src/celery/celery/canvas.py", line 1140, in app
    app = self.tasks[0].app
  File "***/src/celery/celery/utils/functional.py", line 213, in __getitem__
    self.__consumed.append(next(self.__it))
TypeError: 'Signature' object is not an iterator```
```

## Comment
It should be easy to fix but I'm not sure about celery typing and etc, so I decided to make the issue.

https://github.com/celery/celery/blob/39d86e7dbe7f3a0a43e973910be880146de96fb7/celery/canvas.py#L948-L958

We just need to check if this single task is Task and not something else.

agladkov:
Fixes #4255

### DIFF
--- a/celery/canvas.py
+++ b/celery/canvas.py
@@ -950,6 +950,8 @@ class group(Signature):
             tasks = tasks[0]
             if isinstance(tasks, group):
                 tasks = tasks.tasks
+            if isinstance(tasks, abstract.CallableSignature):
+                tasks = [tasks.clone()]
             if not isinstance(tasks, _regen):
                 tasks = regen(tasks)
         Signature.__init__(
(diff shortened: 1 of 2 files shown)


thedrow:
Thanks!

Issue #4284, opened 2017-09-25
Environment variable for Broker URL is prioritized over app settings


Checkroth:
## Checklist

- [x] I have included the output of ``celery -A proj report`` in the issue.
      (if you are not able to do this, then at least specify the Celery
       version affected).
- [x] I have verified that the issue exists against the `master` branch of Celery.

## Steps to reproduce
Celery version: 4.0.2 (latentcall)
Issue is about the celery python app, which we are currently using with django (not django-celery).
Issue can be seen here: https://github.com/celery/celery/blob/master/celery/app/utils.py#L85-L106

My team had a misunderstanding about Celery settings and environment variables (we thought the celery app would set the config from environment variables automatically using `config_from_object`). We had CELERY_RESULT_BACKEND and CELERY_BROKER_URL set in our environment variables. This worked perfectly for a while, until we started doing operations that relied on the result backend.

It took a while to figure out _why_ the broker url was working fine but the backend was not. After some investigation, I realized that the broker url wasn't set at all, and that the call to `broker_url` was actually using the environment variable we had set instead of the app settings.

## Expected behavior
Celery broker settings should take priority. Currently, even if we had properly configured celery, a stray environment variable could break our application.

Ideally, the broker url would not be read from environment variables at all, at least not at the level in which it is currently read, as this only causes confusion. I am not familiar with why it is implemented this way, but it looks like it is probably a bit of legacy code.

## Actual behavior
Celery broker url is read from environment variables before app settings.

Checkroth:
I have added a test to demonstrate the issue:

https://github.com/celery/celery/pull/4285

auvipy:
this should exist in master too?

ant31:
I don't see an issue here, many software prioritizes environment variable over conffile when both exist. 

Things to eventually improve: 
 - document  what's the precedence order
 - log from where the configuration is taken

thomas-riccardi:
This is not the current issue: there is no `conffile` involved in the scenario.
The issue is precedence of env vars over explicit configuration passed as `Celery` constructor or otherwise edited by explicit code (dev controlled), not a user controlled conf file vs env var.

Checkroth:
@thomas-riccardi is correct.

My main concern would be that it _seems_ like your settings are being taken when you explicitly set them, but Celery is "magically" picking up environment variables that you might not even be aware that you have set.

[...]

PR #4285, opened 2017-09-25
Add failing test for broker url priority ref issue #4284

Checkroth:
## Description
This PR is to assist in showing the issue described here: https://github.com/celery/celery/issues/4284

All this PR does is add a failing test that _should_ be passing, if the issue above is resolved.

This PR does not solve the issue mentioned. I leave that up to the discretion of a more knowledgeable party, if they decide that it is indeed an issue at all. I do not expect this PR to be merged unless the issue is resolved and you want this test to remain in the repository.

### DIFF
--- a/t/unit/app/test_utils.py
+++ b/t/unit/app/test_utils.py
@@ -26,6 +26,13 @@ class test_Settings:
         assert self.app.conf.find_value_for_key(
             'always_eager') is False
 
+    def test_broker_priority(self, monkeypatch):
+        prepatch_broker_url = 'memory://thisshouldbethebroker'
+        self.app.conf.broker_url = prepatch_broker_url
+        monkeypatch.setenv('CELERY_BROKER_URL',
+                           'memory://thisshouldntbethebroker')
+        assert self.app.conf.broker_url == prepatch_broker_url
+
     def test_table(self):
         assert self.app.conf.table(with_defaults=True)
         assert self.app.conf.table(with_defaults=False)
--- a/t/unit/app/test_utils.py
+++ b/t/unit/app/test_utils.py
@@ -33,6 +33,11 @@ class test_Settings:
                            'memory://thisshouldntbethebroker')
         assert self.app.conf.broker_url == prepatch_broker_url
 
+    def test_broker_dynamic_setting(self):
+        prepatch_broker_url = 'memory://thisshouldbethebroker'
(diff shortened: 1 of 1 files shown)


t/unit/app/test_utils.py
```diff
@@ -26,6 +26,13 @@ def test_find_value_for_key(self):
         assert self.app.conf.find_value_for_key(
             'always_eager') is False
 
+    def test_broker_priority(self, monkeypatch):
+        prepatch_broker_url = 'memory://thisshouldbethebroker'
+        self.app.conf.broker_url = prepatch_broker_url
+        monkeypatch.setenv('CELERY_BROKER_URL',
+                           'memory://thisshouldntbethebroker')
+        assert self.app.conf.broker_url == prepatch_broker_url
```
georgepsarakis: The application configuration has already been resolved at this stage, so I think this test case does not actually prove the issue. Configuration cannot be changed dynamically.

Checkroth: I got that feeling when I wrote this, however calling `self.app.conf.broker_url` after setting it to `prepatch_broker_url`, but before setting `CELERY_BROKER_URL` in the environment, it does return the `prepatch_broker_url` string

georgepsarakis:
@Checkroth @auvipy I think the configuration works as expected. Please consider the following:

```
import os
from celery.app.utils import Settings  # sub-class of celery.utils.collections.ConfigurationView
configuration = Settings(dict(broker_url='amqp://hello.world'))
print configuration.broker_url  # amqp://hello.world
os.environ['CELERY_BROKER_URL'] = 'amqp://hello.world.from.environment.variable'
print configuration.broker_url  # amqp://hello.world.from.environment.variable
```

I hope the priority is clearer now.

However your test seems to raise another question regarding the behavior of `ConfigurationView` and properties in sub-classes such as `Settings.broker_url`; an exception should have been raised when you try to set a value for `broker_url` and this does not happen. Perhaps `AttributeDictMixin` is responsible for this.

Checkroth:
@georgepsarakis Thank you for clarifying.

My initial issue was that, when explicitly setting your Celery config, the broker url would read from your environment variables rather than your explicit setting if `CELERY_BROKER_URL` was set.

It appears that this is by design, so this pull request adds no value to Celery. I will write my thoughts on the issue over at #4284 , because I still feel that there is something to be done to avoid this confusion in the future.

PR #4292, opened 2017-09-27
Celery doesn't re-use DB connections with Django (fixes #4116)

### LINKED ISSUE #4116: Django celery fixup doesn't respect Django settings for PostgreSQL connections

## Checklist

       version affected).

## Steps to reproduce

## Expected behavior

## Actual behavior

## Potential Resolution
``` Python

def _close_database(self):
    try:
    except AttributeError:
            try:
                conn.close()
            except self.interface_errors:
                pass
                str_exc = str(exc)
                    raise

```

Schweigi:
Fixes #4116 

Celery doesn't re-use DB connections when the setting `CONN_MAX_AGE` is set in Django. Instead connections are closed immediately. This leads to a high load on the database when there are a lot of Celery workers. It also affects query performance negatively.

This PR uses the Django `close_if_unusable_or_obsolete()` method to close a DB connection instead of `close()`.

`close_if_unusable_or_obsolete()` is available at least since Django 1.8 (see https://github.com/django/django/commit/28308078f397d1de36fd0da417ac7da2544ba12d) which is the minimum supported version for Celery 4.0.

Let me know if you need anything else to merge this.

### DIFF
--- a/celery/fixups/django.py
+++ b/celery/fixups/django.py
@@ -183,7 +183,7 @@ class DjangoWorkerFixup(object):
     def _close_database(self):
         for conn in self._db.connections.all():
             try:
-                conn.close()
+                conn.close_if_unusable_or_obsolete()
             except self.interface_errors:
                 pass
             except self.DatabaseError as exc:
--- a/t/unit/fixups/test_django.py
+++ b/t/unit/fixups/test_django.py
@@ -216,11 +216,12 @@ class test_DjangoWorkerFixup(FixupCase):
             f._db.connections.all.side_effect = lambda: conns
 
             f._close_database()
-            conns[0].close.assert_called_with()
-            conns[1].close.assert_called_with()
-            conns[2].close.assert_called_with()
+            conns[0].close_if_unusable_or_obsolete.assert_called_with()
+            conns[1].close_if_unusable_or_obsolete.assert_called_with()
+            conns[2].close_if_unusable_or_obsolete.assert_called_with()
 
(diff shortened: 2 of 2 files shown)


Schweigi:
I would like to add that we use this fix successfully in production and hand no issues so far.

thedrow:
Thank you very much!

ograycode:
Are there any plans for a patch fix (4.1.x) that will include this?

thedrow:
Yes but we still have to merge a few more PRs in order to release 4.1.

dralley:
@thedrow Is a 4.1.1 coming soon?  Are there any PRs blocking it?

auvipy:
4.2 is coming one blocker remain. check the milestone

auvipy:
this change is creating regression as per reports, should we revert this?

thedrow:
If people have problems with one method and others have a problem with another we should introduce a configuration option to use either.
The default should be to use the method in this PR in my opinion.
Please file an issue about it.

thedrow:
@georgepsarakis What do you think?

georgepsarakis:
@thedrow I read the issues linking here, I think that they mostly concern finding a correct value for `CONN_MAX_AGE`. I don't we should revert, but I agree on the setting or better documentation.

Chronial:
This PR actually broke celery's DB handling. I would suggest a revert.

thedrow:
@Schweigi Care to weight in?

auvipy:
celery 4.x doesn't support Django-celery package

PR #4302, opened 2017-09-28
Ignore celery.exception.Ignore on autoretry

zpl:
Autoretry for task should ignore celery.exception.Ignore, which is generated by self.replace()

otherwise, it goes to infinite loop.

```python
@app.task(autoretry_for=(Exception,), default_retry_delay=1,
          max_retries=None,
          bind=True,acks_late=True)
def TaskA(self):
    raise self.replace(TaskB.s()) # Always retrying because of replace

```

### DIFF
--- a/celery/app/base.py
+++ b/celery/app/base.py
@@ -476,7 +476,7 @@ class Celery(object):
                     try:
                         return task._orig_run(*args, **kwargs)
                     except Ignore:
-                        raise
+                        pass
                     except autoretry_for as exc:
                         if retry_backoff:
                             retry_kwargs['countdown'] = \
--- a/celery/app/base.py
+++ b/celery/app/base.py
@@ -27,7 +27,7 @@ from celery._state import (
     get_current_worker_task, connect_on_app_finalize,
     _announce_app_finalized,
 )
-from celery.exceptions import AlwaysEagerIgnored, ImproperlyConfigured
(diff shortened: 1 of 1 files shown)


thedrow:
@zpl Does TaskB returns Ignore?

thedrow:
My unit test raises `RuntimeError: maximum recursion depth exceeded` so this fix doesn't do the trick it seems.

thedrow:
@zpl Ping? See comment above ^

zpl:
@thedrow , could you please share unit test?

In my case 'raise self.replace(TaskB.s()) ' never happens because it retries on Ignore.


// just got my laptop from service

celery/app/base.py
```diff
@@ -475,6 +475,8 @@ def _task_from_fun(self, fun, name=None, base=None, bind=False, **options):
                 def run(*args, **kwargs):
                     try:
                         return task._orig_run(*args, **kwargs)
+                    except Ignore:
+                        raise
```
georgepsarakis: @thedrow I think re-raising `Ignore` causes an maximum recursion condition. Perhaps simply `pass` here?
Also, if you can share the unit test, I might be able to take a look.

thedrow: I'll do so tomorrow.

thedrow, line 479:
```suggestion
                        pass
```

PR #4312, opened 2017-10-04
Fix equal schedules

mariia-zelenova:
Currently celery beat updates scheduled tasks only if new task is added/deleted/enabled/disabled or schedule/crontab changes. But beat should update tasks also in case of arguments or options change.

Probably fixes https://github.com/celery/django-celery-beat/issues/7#issuecomment-307661658

### DIFF
--- a/celery/beat.py
+++ b/celery/beat.py
@@ -156,6 +156,24 @@ class ScheduleEntry(object):
             return id(self) < id(other)
         return NotImplemented
 
+    def editable_fields_equal(self, other):
+        for attr in ['task', 'args', 'kwargs', 'options', 'schedule']:
+            if getattr(self, attr) != getattr(other, attr):
+                return False
+        return True
+
+    def __eq__(self, other):
+        """Will only compare "editable" fields:
+        ``task``, ``schedule``, ``args``, ``kwargs``, ``options``.
+        """
+        return self.editable_fields_equal(other)
+
+    def __ne__(self, other):
+        """Will only compare "editable" fields:
+        ``task``, ``schedule``, ``args``, ``kwargs``, ``options``.
+        """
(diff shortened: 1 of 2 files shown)


mariia-zelenova:
> The proposed patch, while valid, does not provide a generic fix to the problem.
If we overrode the __eq__ method instead, we could compare schedules anywhere without fear.

I'm not sure that it is fair to override `__eq__` method for my purposes. Class `ScheduleEntry` has attributes which are not needed to compare here like `last_run_at`, `total_run_count` and `app`, and overriding `__eq__` with this logic will shadow changes in these fields. If it looks good for you - i'll do it.

mariia-zelenova:
@thedrow What do you think about https://github.com/celery/celery/pull/4312#issuecomment-334593406 ?

georgepsarakis:
@mariia-zelenova currently the `ScheduleEntry.__eq__` method is not implemented. You can perform this change by initially using the subset of attributes that you comparing right now in your code. Does that make sense?

mariia-zelenova:
Fixed. But I still think that we are making a hidden bomb because we redefine a standart operator which intuitively should compare all the arguments.

mariia-zelenova:
flake8 check seems to be broken in master.

celery/beat.py
```diff
@@ -156,6 +156,24 @@ def __lt__(self, other):
             return id(self) < id(other)
         return NotImplemented
 
+    def editable_fields_equal(self, other):
+        for attr in ['task', 'args', 'kwargs', 'options', 'schedule']:
```
thedrow: Use a tuple instead of a list since this is a constant.

thedrow:
Thanks!

PR #4313, opened 2017-10-06
Add couchbase dependency to CI + Python 3/PyPy fixes

thedrow:
This PR ensures we run the unit tests for our couchbase results backend.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -78,7 +78,7 @@ before_install:
               curl localhost:8000
           fi
     - |
-          wget -qO - http://packages.couchbase.com/ubuntu/couchbase.key | sudo apt-key add -
+          wget -qO - https://packages.couchbase.com/ubuntu/couchbase.key | sudo apt-key add -
           sudo apt-add-repository -y 'deb http://packages.couchbase.com/ubuntu trusty trusty/main'
           sudo apt-get update && sudo apt-get install -y libcouchbase-dev
 after_success:
--- a/.travis.yml
+++ b/.travis.yml
@@ -77,6 +77,10 @@ before_install:
               sleep 10
               curl localhost:8000
           fi
+    - |
+          wget -qO - http://packages.couchbase.com/ubuntu/couchbase.key | sudo apt-key add -
+          sudo apt-add-repository -y 'deb http://packages.couchbase.com/ubuntu trusty trusty/main'
+          sudo apt-get update && sudo apt-get install -y libcouchbase-dev
 after_success:
   - .tox/$TRAVIS_PYTHON_VERSION/bin/coverage xml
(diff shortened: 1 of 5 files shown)


thedrow:
Build failures are unrelated. We need to resolve the lint errors in a separate PR.

t/unit/backends/test_couchbase.py
```diff
@@ -106,15 +106,3 @@ def test_backend_params_by_url(self):
             assert x.username == 'johndoe'
             assert x.password == 'mysecret'
             assert x.port == 123
-
```
auvipy: just curious why the test is being dropped? or am I missing something?

thedrow: The method is not being used by the couchbase results backend at all.
Also, it no longer behaves as expected.
A future refactor of that method should ensure that if we're converting every key to string we'll also decode byte strings to UTF-8.

.travis.yml
```diff
@@ -66,6 +66,10 @@ before_install:
               sleep 10
               curl localhost:8000
           fi
+    - |
+          wget -qO - http://packages.couchbase.com/ubuntu/couchbase.key | sudo apt-key add -
```
georgepsarakis: You can use HTTPS here.

requirements/test-ci-default.txt
```diff
@@ -14,6 +14,7 @@
 -r extras/gevent.txt
 -r extras/elasticsearch.txt
 -r extras/couchdb.txt
+-r extras/couchbase.txt
```
georgepsarakis: I think we also need to start setting an exact value of the dependencies in the [requirements](https://github.com/celery/celery/blob/master/requirements/extras/couchbase.txt) file and also specify the couchbase_ffi for Pypy?

celery/backends/couchbase.py
```diff
@@ -7,6 +7,10 @@
 from celery.exceptions import ImproperlyConfigured
 from .base import KeyValueStoreBackend
 try:
+    import couchbase_ffi # noqa
```
georgepsarakis: Shouldn't this [dependency](https://github.com/couchbaselabs/couchbase-python-cffi) be installed on the PyPy Travis CI jobs as well or is it not required?

thedrow: We don't need this dependency for the tests. The client is mocked. It is required to install the default client because otherwise we won't be able to instantiate the couchbase results backend at all.

requirements/test-ci-default.txt
thedrow: You are correct.

.travis.yml
thedrow: You are right. It was copy/pasted from their installation instructions.

Issue #2131, opened 2014-07-08
Systemd service file is invalid


moggers87:
Output of `systemctl status celery` after installing the service file and reloading systemd's configs:
```
Jul 08 21:55:05 localhost.localdomain systemd[1]: [/etc/systemd/system/celery.service:11] Executable path is not absolute, ignoring: ${CELERY_BIN} multi start $CELERYD_NODES      -A $CELERY_APP --pidfile=${CELERYD_PID_FILE}      --logfile=${CELERYD_LO..."      $CELERYD_OPTS
Jul 08 21:55:05 localhost.localdomain systemd[1]: [/etc/systemd/system/celery.service:12] Executable path is not absolute, ignoring: ${CELERY_BIN} multi stopwait $CELERYD_NODES      --pidfile=${CELERYD_PID_FILE}
Jul 08 21:55:05 localhost.localdomain systemd[1]: [/etc/systemd/system/celery.service:13] Executable path is not absolute, ignoring: ${CELERY_BIN} multi restart $CELERYD_NODES      -A ${CELERY_APP} --pidfile=${CELERYD_PID_FILE}      --logfile=${CELERY..."      $CELERYD_OPTS
Jul 08 21:55:05 localhost.localdomain systemd[1]: celery.service lacks ExecStart setting. Refusing.
```

I've tried using this service file on both CentOS 7 and Fedora 20 with the same results.

jxstanford:
A little late to the party, but here's what I did to get workers and beat going under CentOS 7.  It seems to be working fine:

/etc/sysconfig/celery:
```
DJANGO_SETTINGS_MODULE="goldstone.settings.production"
CELERY_APP="goldstone"

# Worker settings
CELERYD_NODES="w1"
CELERYD_OPTS="--time-limit=300 --concurrency=8"
CELERYD_LOG_FILE="/var/log/goldstone/celery-%N.log"
CELERYD_STATE_DIR="/var/run/celery"
CELERYD_PID_FILE="/var/run/celery/goldstone-%N.pid"
CELERYD_LOG_LEVEL="INFO"

# Beat settings

CELERYBEAT_LOG_FILE="/var/log/goldstone/celerybeat.log"
CELERYBEAT_STATE_DIR="/var/run/celerybeat"
CELERYBEAT_PID_FILE="/var/run/celerybeat/goldstone-celerybeat.pid"
CELERYBEAT_LOG_LEVEL="INFO"
[...]

PR #4330, opened 2017-10-14
Typo in the example systemd service

palewire:


lafrech:
I believe the dash/minus sign here is intended. It means "ignore errors if file does not exist".

See https://lists.freedesktop.org/archives/systemd-devel/2013-February/009227.html for instance.

palewire:
Fair, though in the context of the page this example recommends the creation of that file. Without it there, the script would fail unless the environment variables were set via another means. No other means are suggested in the page.

My view is that the example deployment options should be tightly packaged together.

lafrech:
> Fair, though in the context of the page this example recommends the creation of the environment variables file. Without it there, the systemd script would fail unless the variables are set by another means. No other means are suggested in that section.

Yes, you're right. I think the fact that you called it a "typo" made me suspect you might not know about that specific syntax and then I didn't think any further. But you're right that it is not consistent.

> My view is that the example deployment options should be packaged together as tightly as possible. This is beyond the scope of this ticket, but I also think this section of the documentation could be improved by helping readers understand it's important their code be present in working directory path and that the recommended /var/run/celery and /var/log/celery directories will likely need to be created with the proper permissions.

As someone who just spent days trying to deamonize his celery/flask application, I can only confirm this. I wish I could be more constructive by providing helpful comments and ideally a PR, but people before me with much greater knowledge about systemd have been arguing about those sample files already and here we are (search systemd in issues, or just look at this one: https://github.com/celery/celery/issues/2131). I'm afraid there are many different cases and it is difficult to provide a plug and play example.

georgepsarakis:
@lafrech thanks a lot for sharing that information! I thought it might have been intentional, but I couldn't find  proper documentation.

georgepsarakis:
@palewire I am closing this, as from our discussion so far, it seems that it is better to silently ignore the absence of the `/etc/conf.d/` settings file.

lafrech:
@palewire has a good point and since those variables are not defined anywhere else, the service will crash if `/etc/conf.d/celery` does not exist so we'd rather stop with an explicit error as soon as this file is found missing. In other words, this PR should be merged IMHO: it is slightly better with than without it. Not a big deal anyway.

georgepsarakis:
@lafrech you are right.

PR #4357, opened 2017-11-01
Fixes #4356: Modify order of protocol upgrade detection.

freakboy3742:
## Description

If a task is executed under Celery 3.1.25, and is retried under Celery 4 node, and then retried again under a Celery 3.1.25 node, the Celery 3.1.25 node will crash, resulting in message loss.

When the Celery 3.1.25 task is executed on a Celery 4, the task message is upgraded to Protocol 2. However, the upgrade results in a hybrid message that complies with both formats, and when the task fails and is retried on the Celery 3.1.25 worker, the "hybrid" message is mis-identified as a Protocol 1

By changing the detection order, the "hybrid" is correctly identified as being Protocol 2 compliant.

Unfortunately, this change must be made against the 3.1 branch, as the error lies in the "forward-compatibility" code introduced in that version.

Issue #4356, opened 2017-11-01
Task loss on retry when using a hybrid/staged Celery 3->4 deployment


freakboy3742:
If you have a Celery 3.1.25 deployment involving many workers, and you want to upgrade to Celery 4, you may wish to do "canary" testing of a limited subset of workers to validate that the upgrade won't introduce any problems, prior to upgrading your entire worker fleet to Celery4. This "canary" mode  involves having both Celery 3.1.25 and Celery 4 workers running at the same time.

However, if you do this, and you have tasks that retry, you experience problems if a task is attempted on a Celery 3.1.25 node, then a Celery 4 node, and then a Celery 3.1.25 node. 

When the Celery 3.1.25 task is executed on a Celery 4, the task message is upgraded to Protocol 2. However, the upgrade results in a hybrid message that complies with *both* formats, and when the task fails and is retried on the Celery 3.1.25 worker, the "hybrid" message is mis-identified as a Protocol 1 message, resulting in a hard crash and message loss.

## Checklist

- [X] I have included the output of ``celery -A proj report`` in the issue.
- [X] I have verified that the issue exists against the `master` branch of Celery.

## Steps to reproduce

A full reproduction case can be found in this gist:

https://gist.github.com/ewdurbin/ddf4b0f0c0a4b190251a4a23859dd13c

In local testing, the following two versions were used:

###Celery 3.1.25:
```
software -> celery:3.1.25 (Cipater) kombu:3.0.37 py:2.7.13
            billiard:3.3.0.23 redis:2.10.6
platform -> system:Darwin arch:64bit imp:CPython
loader   -> celery.loaders.app.AppLoader
settings -> transport:redis results:disabled

BROKER_URL: 'redis://localhost:6379//'
CELERY_ENABLE_UTC: True
CELERY_RESULT_SERIALIZER: 'json'
CELERY_ACCEPT_CONTENT: ['json']
CELERY_TIMEZONE: 'UTC'
CELERY_TASK_SERIALIZER: 'json'
```

###Celery 4.1.0:
```
software -> celery:4.1.0 (latentcall) kombu:4.1.0 py:2.7.13
            billiard:3.5.0.3 redis:2.10.6
platform -> system:Darwin arch:64bit imp:CPython
loader   -> celery.loaders.app.AppLoader
settings -> transport:redis results:disabled

task_serializer: 'json'
result_serializer: 'json'
CELERY_ENABLE_UTC: True
accept_content: ['json']
enable_utc: True
[...]

PR #4358, opened 2017-11-01
Refs #4356: Handle "hybrid" messages that have moved between Celery versions

freakboy3742:
The patch in #4357 addresses a problem with forward compatibility in message handling. This 

If you apply the patch from #4357, then modify [the test case reported in #4356](https://gist.github.com/ewdurbin/ddf4b0f0c0a4b190251a4a23859dd13c) to cycle from Celery 3.1 -> 4.1 -> 3.1 -> 4.1, you get a similar error, but on the Celery 4.1 worker:
```
  File "/Users/rkm/zapier/celery/celery4-venv/lib/python2.7/site-packages/billiard/pool.py", line 1747, in safe_apply_callback
    fun(*args, **kwargs)
  File "/Users/rkm/zapier/celery/celery4-venv/lib/python2.7/site-packages/celery/worker/request.py", line 366, in on_failure
    self.id, exc, request=self, store_result=self.store_errors,
  File "/Users/rkm/zapier/celery/celery4-venv/lib/python2.7/site-packages/celery/backends/base.py", line 165, in mark_as_failure
    if request.chord:
  File "/Users/rkm/zapier/celery/celery4-venv/lib/python2.7/site-packages/kombu/utils/objects.py", line 44, in __get__
    value = obj.__dict__[self.__name__] = self.__get(obj)
  File "/Users/rkm/zapier/celery/celery4-venv/lib/python2.7/site-packages/celery/worker/request.py", line 485, in chord
    _, _, embed = self._payload
ValueError: too many values to unpack
```

This patch handles the "hybrid" syntax on the Celery 4.1 worker.

### DIFF
--- a/celery/worker/strategy.py
+++ b/celery/worker/strategy.py
@@ -64,7 +64,7 @@ def clean_hybrid_protocol(message, body):
     return (args, kwargs, embed), headers, True, body.get('utc', True)
 
 
-def proto1_to_proto2(message, body, hybrid=False):
+def proto1_to_proto2(message, body):
     """Convert Task message protocol 1 arguments to protocol 2.
 
     Returns:
--- a/celery/worker/strategy.py
+++ b/celery/worker/strategy.py
@@ -24,7 +24,7 @@ logger = get_logger(__name__)
 
 
 def hybrid_to_proto2(message, body):
-    "Create a fresh protocol 2 message from a hybrid protocol 1/2 message."
+    """Create a fresh protocol 2 message from a hybrid protocol 1/2 message."""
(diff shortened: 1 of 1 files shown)


thedrow:
I think that in this case a unit test that verifies that we can move a message between protocol v1 and v2 after a retry is enough to validate this.
The unit test should just test the flows of translating a v1 message to v2 message and a v2 message to v1 message.

thedrow:
@freakboy3742 any chance you'll get to write that unit test?
We would like to have this fixed released in 4.2.

PR #4369, opened 2017-11-06
task_replace chord inside chord fix (fixes #4368)

### LINKED ISSUE #4368: task replace chord inside chord not working.
## Steps to reproduce

if we trying to replace task in a chord by another chord we have a problem - outer chord body not called

add_to_chord not working too

for example

```python

@app.task(bind=True, ignore_result=True)
def some_task(self, args=None, **kwargs):
    #add_to_chord not working too
    #self.add_to_chord(chord([echo.s(name='t1'), echo_s.s(name='t2'), echo.s(name='t3')], echo.s(name='inner_chord_end')))
    raise self.replace(chord([echo.s(name='t1'), echo.s(name='t2'), echo.s(name='t3')], echo.s(name='inner_chord_end')))

@app.task(bind=True, ignore_result=True)
def echo(self, args=None, **kwargs):
    print(kwargs['name'])
    return kwargs['name']


result = chord([some_task.s(), echo.s(name='task')], echo.s(name='total end')).apply_async()
```

## Expected behavior
       echo.s(name='total end') called

## Actual behavior
      echo.s(name='total end') SKIPPED

zpl:
Fixes #4368

Sorry, but I have no time to write unit tests, could someone help? Thanks.

### DIFF
--- a/celery/app/task.py
+++ b/celery/app/task.py
@@ -856,10 +856,8 @@ class Task(object):
             for t in reversed(self.request.chain):
                 sig |= signature(t, app=self.app)
 
-        sig.freeze(self.request.id,
-                   group_id=self.request.group,
-                   chord=chord,
-                   root_id=self.request.root_id)
+        sig = sig.set(chord=self.request.chord, group_id=self.request.group, root_id=self.request.root_id)
+        sig.freeze(self.request.id)
 
         sig.delay()
         raise Ignore('Replaced by new task')
@@ -878,9 +876,8 @@ class Task(object):
         """
         if not self.request.chord:
             raise ValueError('Current task is not member of any chord')
-        result = sig.freeze(group_id=self.request.group,
-                            chord=self.request.chord,
-                            root_id=self.request.root_id)
+        sig = sig.set(chord=self.request.chord, group_id=self.request.group, root_id=self.request.root_id)
+        result = sig.freeze()
         self.backend.add_to_chord(self.request.group, result)
         return sig.delay() if not lazy else sig
 
--- a/celery/app/task.py
+++ b/celery/app/task.py
@@ -11,7 +11,8 @@ from kombu.utils.uuid import uuid
 from celery import current_app, group, states
 from celery._state import _task_stack
 from celery.canvas import signature
-from celery.exceptions import Ignore, MaxRetriesExceededError, Reject, Retry
+from celery.exceptions import (Ignore, ImproperlyConfigured,
+                               MaxRetriesExceededError, Reject, Retry)
 from celery.five import items, python_2_unicode_compatible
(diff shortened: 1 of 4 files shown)


thedrow:
What's the difference between the old code and the new code?

zpl:
The problem is that  sig.freeze doesn't set group_id and chord for new task.

that's we need to call sig.set before

thedrow:
I don't understand why the call to`freeze()` was removed in that case.

thedrow:
@zpl ping?

zpl:
@thedrow, plz double check. it's NOT removed. it's replaced to


sig.freeze(self.request.id)


so it's replaced to sig.set + sig.freeze

because sig.freeze don't store state for sig

AlexHill:
Looking at the code for `chord.freeze()` this makes sense. [`Signature.freeze()`](https://github.com/celery/celery/blob/master/celery/canvas.py#L254) sets some keys of `self.options` from its arguments, whereas [`chord.freeze()`](https://github.com/celery/celery/blob/master/celery/canvas.py#L1194) doesn't.

In my testing the case where a nested chord's body is replaced still doesn't work with this patch. It would be nice to fix that case too. I can write tests for this.

thedrow:
Thank you very much. I'm assigning the issue to you. You should be able to push to this branch. :)

AlexHill:
I've rebased this, added tests and made sure the fix applies to replaced bodies as well as header tasks. @zpl can you test to make sure your case is still fixed please?

thedrow:
@AlexHill Can you please address the lint and test failures?

AlexHill:
Yep. I've just added some integration tests for `add_to_chord` since that's changed in this PR as well.

celery/app/task.py
```diff
@@ -846,20 +846,20 @@ def replace(self, sig):
```
AlexHill: The branch above is never entered during the test suite, i.e. `'chord' in sig.options` is never true. I think this makes sense - for it to be true, we'd have to be replacing a task in a chord with another task that is a part of a different chord. I'm not sure how that would work - it might be worth just raising an exception in this branch rather than having this undefined untested behaviour in there.

thedrow:
Also when tests are added to a PR please remove the Needs Test Coverage label.

zpl:
>@zpl can you test to make sure your case is still fixed please?

@AlexHill, Thanks for your work. I will test it at monday and write results. thanks.

thedrow:
@AlexHill @zpl This PR needs to be rebased.

@zpl Did you get the chance to test this?

AlexHill:
I've rebased and added the exception I described above, plus an accompanying test. Would be good if this could sneak into 4.2.

PR #4350, opened 2017-10-27
Adding 'shadow' property to as_task_v2

jmdacruz:
Attempt at fixing #4347

Issue #4347, opened 2017-10-27
apply_async's "shadow" parameter not working?


jmdacruz:
If I'm reading correctly, [apply_async](http://docs.celeryproject.org/en/latest/reference/celery.app.task.html#celery.app.task.Task.apply_async) has an argument called `shadow`, which overrides the task name on the logs and monitoring. This doesn't seem to be working, or I'm doing something wrong (quite possibly)

## Checklist

- [X] I have included the output of ``celery -A proj report`` in the issue.
      (if you are not able to do this, then at least specify the Celery
       version affected).
- [x] I have verified that the issue exists against the `master` branch of Celery.

### output of ``celery -A proj report``

```
software -> celery:4.1.0 (latentcall) kombu:4.1.0 py:2.7.12
            billiard:3.5.0.3 redis:2.10.6
platform -> system:Linux arch:64bit, ELF imp:CPython
loader   -> celery.loaders.app.AppLoader
settings -> transport:redis results:redis://localhost:6379/4

broker_url: u'redis://localhost:6379/4'
result_backend: u'redis://localhost:6379/4'
```

## Steps to reproduce

Create an application based on the example in [First Steps With Celery](http://docs.celeryproject.org/en/latest/getting-started/first-steps-with-celery.html)

* `tasks.py`

```python
from celery import Celery

app = Celery('tasks', broker='redis://localhost:6379/4', backend='redis://localhost:6379/4')

@app.task
def add(x, y):
    return x + y
```

* `main.py`

```python
from tasks import add

if __name__ == "__main__":

    t1 = add.apply_async(args=(2,2), shadow="foo")
    t2 = add.apply_async(args=(3,3), shadow="bar")

    r1 = t1.get()
    r2 = t2.get()
    print("R1 = {}, R2 = {}".format(r1, r2))
```

Execute the worker:

```
celery -A tasks worker --loglevel=info
```

Execute flower:

```
flower -A tasks
```

## Expected behavior

Flower's "tasks" dashboard shows the task names as "foo" and "bar".

## Actual behavior

Flower's "tasks" dashboard shows the task names as "tasks.add" in both cases. The worker's log also shows the original task name without modification:

```
[2017-10-26 20:29:22,153: INFO/MainProcess] Received task: tasks.add[351ff35e-a834-4ccf-aa25-28239367024e]
[2017-10-26 20:29:22,155: INFO/MainProcess] Received task: tasks.add[f6a411fe-d3c8-4c8a-9604-8b892d901088]
[...]

PR #4379, opened 2017-11-11
Allow shadow to override task name in trace and logging messages.

hclihn:
Attempt at fixing #4347 in addition to #4350. 

## Description
The single-line commit in `celery/app/amqp.py` from #4350 did not fix the issue entirely as it was pointed out in the comment of the PR. To fix this issue in the "received" log entries, we need to fix a few places in `celery/celery/app/trace.py` which will fix the "received" log/trace entries.

### DIFF
--- a/celery/app/trace.py
+++ b/celery/app/trace.py
@@ -141,6 +141,11 @@ def get_log_policy(task, einfo, exc):
             return log_policy_expected
         return log_policy_unexpected
 
+    
+def get_task_name(name, request):
+    """Use 'shadow' in request for the task name if applicable."""
+    return getattr(request, 'shadow', name)
+    
 
 class TraceInfo(object):
     """Information about task execution."""
@@ -151,10 +156,6 @@ class TraceInfo(object):
         self.state = state
         self.retval = retval
         
-    @staticmethod
-    def get_name(task, req):
-        return req.shadow if hasattr(req, 'shadow') else task.name
-
     def handle_error_state(self, task, req,
                            eager=False, call_errbacks=True):
         store_errors = not eager
@@ -190,7 +191,7 @@ class TraceInfo(object):
                                     reason=reason, einfo=einfo)
             info(LOG_RETRY, {
                 'id': req.id,
-                'name': self.get_name(task, req),
+                'name': get_task_name(task.name, req),
                 'exc': text_t(reason),
             })
             return einfo
@@ -238,7 +239,7 @@ class TraceInfo(object):
         context = {
             'hostname': req.hostname,
             'id': req.id,
-            'name': self.get_name(task, req),
+            'name': get_task_name(task.name, req),
             'exc': exception,
             'traceback': traceback,
             'args': sargs,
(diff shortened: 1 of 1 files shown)


celery/app/trace.py
```diff
@@ -356,6 +360,7 @@ def trace_task(uuid, args, kwargs, request=None):
             task_request = Context(request or {}, args=args,
                                    called_directly=False, kwargs=kwargs)
             root_id = task_request.root_id or uuid
+            the_name = request['shadow'] if 'shadow' in request else name
```
georgepsarakis: I might be wrong here but isn't this the equivalent of:

```
trace_name = task_request.shadow if hasattr('shadow', task_request) else name
```

celery/app/trace.py
```diff
@@ -150,6 +150,10 @@ class TraceInfo(object):
     def __init__(self, state, retval=None):
         self.state = state
         self.retval = retval
+        
+    @staticmethod
+    def get_name(task, req):
```
georgepsarakis: If you modify this method signature to receive the name directly you can use it in both occurrences:

```
def get_task_name(name, request):
    return getattr(request, 'shadow', name)
```

The use of `getattr` is also a simplification here :smile: !

hclihn: Thought about that earlier. But the problem is that `request` in `trace_task()` is a dictionary, while `req` in `TraceInfo` is an object (`Context` in `celery/celery/app/task.py`). Sure, we can write more tests in the common function `get_task_name()` for both data types, but whether that's the best approach depends on the coding preference. Just my 2 cents.

celery/app/trace.py
hclihn: Got it. Yes, if we use `task_request`, we can combine the 2 parts. :-)

hclihn:
Updates the code with suggested changes. :-)

celery/app/trace.py
```diff
@@ -141,6 +141,11 @@ def get_log_policy(task, einfo, exc):
             return log_policy_expected
         return log_policy_unexpected
 
+    
+def get_task_name(name, request):
```
georgepsarakis: Thanks for applying the changes! Sorry for not suggesting it correctly before, but the repetition of `name` in the function name - signature can possibly be omitted:

```
def get_task_name(request, default):
```

What do you think?

hclihn: It's fine with me.

hclihn:
Modified `get_task_name()` to use `default` (instead of `name`) per suggestion.

georgepsarakis:
@hclihn would it be possible to add a small test case for this [here](https://github.com/celery/celery/blob/master/t/unit/tasks/test_trace.py#L314)? I guess you can simply use `Mock` objects for `task` and `request` parameters.

hclihn:
There is a  bug in get_task_name(). Need to make sure that get_task_name() returns default if request.shadow is None or an empty string.

hclihn:
The unit test code for get_task_name() is in PR #4397.

Issue #4362, opened 2017-11-03
Documentation for celery with SQS as message broker should have pycurl and crypto mentioned as dependencies.


krdeepak:
If pycurl and crypto modules are not installed, worker dies just after starting without any debug information. This information should be included in the documentation.

georgepsarakis:
Have you installed the extension requirements for SQS support?

```
$ pip install celery[sqs]
```

ElAutoestopista:
@georgepsarakis that's ok, But the documentation doesn't says anything about this command:

http://docs.celeryproject.org/en/latest/getting-started/brokers/sqs.html

As, for instance, does for Redis:

http://docs.celeryproject.org/en/latest/getting-started/brokers/redis.html

The docs mention installing boto, but it's not sufficient. In fact, boto, boto3, pycurl and crypto are required. I think the docs should be updated with the command you specified above,

georgepsarakis:
@ElAutoestopista you are right, the `Installation` section must be updated. Would you perhaps like to open a Pull Request for that?

ElAutoestopista:
Sure, I'll check it!

ElAutoestopista:
@georgepsarakis https://github.com/celery/celery/pull/4382

thedrow:
Since this issue is fixed in #4382 I'm going to close this issue.

PR #4382, opened 2017-11-11
Updated documentation of celery with SQS support

ElAutoestopista:
## Description

Updated documentation of celery with SQS support as discussed on issue https://github.com/celery/celery/issues/4362

### DIFF
--- a/docs/getting-started/brokers/sqs.rst
+++ b/docs/getting-started/brokers/sqs.rst
@@ -14,7 +14,7 @@ library using :command:`pip`:
 
 .. code-block:: console
 
-    $ pip install -U boto3
+    $ pip install celery[sqs]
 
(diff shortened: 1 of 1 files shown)


docs/getting-started/brokers/sqs.rst
```diff
@@ -14,7 +14,7 @@ library using :command:`pip`:
 
 .. code-block:: console
 
-    $ pip install -U boto3
+    $ pip install celery[sqs]
```
georgepsarakis: Could you please update the preceding note about `boto` too? Thanks!

ElAutoestopista: Done, I've used the one for Redis as a template, because it also uses bundle.

thedrow:
Thanks!

PR #4403, opened 2017-11-26
Fix periodic tasks running twice the day after DST (Fix #1604)

### LINKED ISSUE #1604: Possible DST issue?
I just noticed a crontab task setup to run at 12:00 today for some reason ran at at 
both 11:00 and then again at 12:00. I've checked the server date and time and it does seem to have made the switch to winter time last night so i'm not sure if this was a fluke or really related to the DST but i don't see any other reason for it to run the task twice...

My crontab setup, for this particular task, looks like this:

````

CELERYBEAT_SCHEDULE = {
    'event_suggestions': {
        'task': 'events.event_suggestions',
        'schedule': crontab(hour=12, minute=0, day_of_week=1)
    }
}
````
Versions are:
celery==3.0.24
django-celery==3.0.23

vbarbaresi:
## Description

The issue was with celery running with UTC enabled, and a timezone with DST.
When computing the remaining time after DST change, `last_run_at` time
was localized in the previous time zone (before DST)
The resulting end date was then also localized in the previous time zone.
So the task was running two times (running an hour early, then on the correct time)

In case the DST was starting the night before, the task was running an hour late the day after.

### Proposed fix

Use the relevant time zone to compute the remaining time:
in case UTC offets are different between now and the previous run, replace tzinfo of the previous run with the current one. 

There could be undesirable side-effects: if we change timezone settings with the database scheduler. But this wasn't supported anyway and required a reset.

### DIFF
--- a/celery/utils/time.py
+++ b/celery/utils/time.py
@@ -209,6 +209,9 @@ def remaining(start, ends_in, now=None, relative=False):
         ~datetime.timedelta: Remaining time.
     """
     now = now or datetime.utcnow()
+    if now.utcoffset() != start.utcoffset():
+        # Timezone has changed, or DST started/ended
+        start = start.replace(tzinfo=now.tzinfo)
     end_date = start + ends_in
     if relative:
         end_date = delta_resolution(end_date, ends_in)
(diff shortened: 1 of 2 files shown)


thedrow:
Thank you!

PR #4404, opened 2017-11-26
Include license file in the generated wheel package

jdufresne:
The wheel package format supports including the license file. This is done using the `[metadata]` section in the `setup.cfg` file. For additional information on this feature, see:

https://wheel.readthedocs.io/en/stable/index.html#including-the-license-in-the-generated-wheel-file

Distributing a wheel now complies with license:

> Redistributions in binary form must reproduce the above copyright notice, this list of conditions and the following disclaimer in the documentation and/or other materials provided with the distribution.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -22,3 +22,6 @@ requires = pytz >= 2016.7
 
 [bdist_wheel]
 universal = 1
+
+[metadata]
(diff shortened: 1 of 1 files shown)


thedrow:
Thank you!

PR #4097, opened 2017-06-18
Fix for https://github.com/celery/celery/issues/4096

nuwang:
## Description

Fixes: https://github.com/celery/celery/issues/4096

Issue #4096, opened 2017-06-18
Consul Backend decodes key incorrectly


nuwang:
## Checklist
software -> celery:4.0.2 (latentcall) kombu:4.0.2 py:3.6.0
            billiard:3.5.0.2 py-amqp:2.1.4
platform -> system:Darwin arch:64bit imp:CPython
loader   -> celery.loaders.app.AppLoader
settings -> transport:amqp results:consul://localhost:8500/

broker_url: 'amqp://guest:********@localhost:5672//'
result_backend: 'consul://localhost:8500/'
result_serializer: 'json'

## Steps to reproduce

Using Consul as a result backend for Celery with Python 3 will raise an exception.

## Expected behavior

Should store values successfully.

## Actual behavior

The following exception is raised in Python 3:
```
[2017-06-18 14:16:47,346: ERROR/MainProcess] Pool callback raised exception: AttributeError("'bytes' object has no attribute 'encode'",)
Traceback (most recent call last):
  File "/Users/Nuwan/work/cloudmanv2/venv/lib/python3.6/site-packages/billiard/pool.py", line 1748, in safe_apply_callback
    fun(*args, **kwargs)
  File "/Users/Nuwan/work/cloudmanv2/venv/lib/python3.6/site-packages/celery/worker/request.py", line 366, in on_failure
    self.id, exc, request=self, store_result=self.store_errors,
  File "/Users/Nuwan/work/cloudmanv2/venv/lib/python3.6/site-packages/celery/backends/base.py", line 163, in mark_as_failure
    traceback=traceback, request=request)
  File "/Users/Nuwan/work/cloudmanv2/venv/lib/python3.6/site-packages/celery/backends/base.py", line 309, in store_result
    request=request, **kwargs)
  File "/Users/Nuwan/work/cloudmanv2/venv/lib/python3.6/site-packages/celery/backends/base.py", line 652, in _store_result
    self.set(self.get_key_for_task(task_id), self.encode(meta))
  File "/Users/Nuwan/work/cloudmanv2/venv/lib/python3.6/site-packages/celery/backends/consul.py", line 84, in set
    key = self._key_to_consul_key(key)
  File "/Users/Nuwan/work/cloudmanv2/venv/lib/python3.6/site-packages/celery/backends/consul.py", line 54, in _key_to_consul_key
    key = key.encode('utf-8')
AttributeError: 'bytes' object has no attribute 'encode'
```
## Reason

[...]

PR #4416, opened 2017-12-04
[backends]: Consul: Do not encode, but decode a byte() object

wido:
We want a string and it's already a byte() code array, so we should
decode it instead of encoding it.

Signed-off-by: Wido den Hollander <wido@widodh.nl>

## Description

In #4096 and #4097 this was mentioned, but closed again.

I verified this issue and tested with Ubuntu 16.04 and Python 3.5 and the issue exists indeed.

### DIFF
--- a/celery/backends/consul.py
+++ b/celery/backends/consul.py
@@ -6,9 +6,10 @@
 """
 from __future__ import absolute_import, unicode_literals
 
+from kombu.utils.encoding import bytes_to_str
 from kombu.utils.url import parse_url
 
-from celery.backends.base import PY3, KeyValueStoreBackend
+from celery.backends.base import KeyValueStoreBackend
 from celery.exceptions import ImproperlyConfigured
 from celery.utils.log import get_logger
 
@@ -53,8 +54,7 @@ class ConsulBackend(KeyValueStoreBackend):
                                     consistency=self.consistency)
 
     def _key_to_consul_key(self, key):
-        if PY3:
-            key = key.encode('utf-8')
+        key = bytes_to_str(key)
         return key if self.path is None else '{0}/{1}'.format(self.path, key)
 
     def get(self, key):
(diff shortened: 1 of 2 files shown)


wido:
Ping @nuwang for some input

nuwang:
@wido The last time I tried this - I ran into this exact issue - the string needed to be decoded instead of encoded. However, when I submitted the patch, it turned out that the travis build was failing, for precisely the opposite reason, which made no sense to me at all. I didn't have a lot of time to investigate it in depth at the time though and when I tried again several weeks later, with the latest version of celery, I could no longer recreate the issue. I'm running MacOS X with Python 3.6. I may have upgraded from 3.5->3.6, although that seems like a very unlikely reason for this issue, but I really can't recall what else could have changed.

wido:
@nuwang Yes, I tested this a lot about a year ago but now suddenly it's broken for me while testing under Ubuntu 16.04:

<pre>
root@alpha:~# lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 16.04.3 LTS
Release:	16.04
Codename:	xenial
root@alpha:~# python3 --version
Python 3.5.2
root@alpha:~# 
</pre>

By applying this patch all is resolved for me and it works just fine.

thedrow:
Thanks for the contribution. We actually use `from kombu.utils.encoding import bytes_to_str` to conditionally convert bytes to a string.
Could you get rid of the current Python 3 code branch and use that?
Also, could you add a test in lines of https://github.com/celery/celery/blob/master/t/unit/backends/test_elasticsearch.py#L111 to ensure this problem will not regress?

wido:
Done! I've pushed a new commit and tests are running now

wido:
@thedrow I did. Travis is happy now! Could you take a look?

thedrow:
Thank you! Feel free to add yourself to the AUTHORS file.

PR #4421, opened 2017-12-06
Strip `self` from sphinx documentation of bound tasks

lpsinger:
In Sphinx-generated documentation, the `self` argument should be hidden from the signature of a task that was created with `bind=True`.

(Sphinx does normally strip the `self` or `cls` argument from method signatures.)


auvipy:
could you look for the reason o build failure? https://travis-ci.org/celery/celery/jobs/312177364 ?

lpsinger:
> could you look for the reason o build failure? https://travis-ci.org/celery/celery/jobs/312177364 ?

It's a spurious job failure. It passes in my own build: https://travis-ci.org/lpsinger/celery/builds/312177267

Issue #2202, opened 2014-08-19
Empty group restores as AsyncResult.


schinckel:
When attempting to restore a group that was empty, we get an `AsyncResult` instead of a `GroupResult`:

    from celery import group
    from celery.result = GroupResult

    # This is actually a simplification: the list comprehension used here
    # may evaluate to empty.
    job = group([])
    result = job.apply_async()

    GroupResult.save(result)

Then, in another thread, possibly:

    task = GroupResult.restore(result.id)

IMHO: this should be a `GroupResult` with an empty list for `.results`, but it's coming back as an `AsyncResult`.

ask:
Closing this, as we don't have the resources to complete this task.

GroupResult.restore and .save should not be used anymore.  I've removed use within celery itself already, and it's a bad pattern to use since it's difficult to clean up from.

schinckel:
So how do we persist group results then? It's still documented as working, with no mention of deprecation at http://celery.readthedocs.io/en/latest/reference/celery.result.html#celery.result.GroupResult

ask:
You could save the list of results in a database yourself.  The existing methods are using pickle so it's not really safe as it works currently.

We have no resources to work on fixing problems with the implementation, and neither to find a better solution at this point.  It's not just this feature, I closed 200 issues today to try and get some future focus.  

I haven't made the change in the code yet, we cannot deprecate it in time for the 4.0 release, but I can remove the feature from documentation to avoid new users finding out about it.

I would like to have a solution for it, but this being open source we have no money to throw at it and no companies to help us.  Only a shit ton of issues that bring me down every day to the point of just choosing one to work on is difficult.  I have 24000 unread mails in my inbox, and all I can do is just look at them fly by.

schinckel:
No worries. Thanks for your time in actually responding to my further questions (and thanks for celery!)

I'll see if I can come up with a nice pattern for saving and supply a documentation PR at some point.

red-hood:
Hi,

I am currently using Celery 4.0.2 and seem to suffer from the same issue mentioned here. @ask wrote

red-hood:
Hi,

I am currently using Celery 4.0.2 and seem to suffer from the same issue mentioned here. @ask wrote:

> GroupResult.restore and .save should not be used anymore. I've removed > use within celery itself already, and it's a bad pattern to use since it's 
> difficult to clean up from.

[...]

PR #4427, opened 2017-12-09
Correctly restore an empty GroupResult (#2202)

AlexHill:
This fixes issue #2202, where a saved empty GroupResult is restored as an AsyncResult. The fix is a simple one-liner and I've added a test verifying the behaviour.

The bug: to return a `GroupResult`, a certain variable `nodes` must evaluate boolean true. In the case of an empty `GroupResult`, `nodes` is an empty list, which evaluates false. The fix is to tighten up the check by adding `is not None`.

### DIFF
--- a/celery/result.py
+++ b/celery/result.py
@@ -998,7 +998,7 @@ def result_from_tuple(r, app=None):
         if parent:
             parent = result_from_tuple(parent, app)
 
-        if nodes:
+        if nodes is not None:
             return app.GroupResult(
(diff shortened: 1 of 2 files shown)


thedrow:
@auvipy If you merge something please add it to the milestone :)

AlexHill:
Wow. Thanks for reviewing and merging this so quickly! 👍

thedrow:
Seems like there is a problem though. Have a look at #4428 if you have a few minutes.

PR #4428, opened 2017-12-09
Added an integration test for #4427

thedrow:
This PR adds an integration test for #4427 to complete the existing test coverage provided by the contributor.

### DIFF
--- a/t/integration/test_canvas.py
+++ b/t/integration/test_canvas.py
@@ -112,6 +112,15 @@ class test_chain:
 class test_group:
 
     @flaky
+    def test_empty_group_result(self):
+        task = group([])
+        result = task.apply_async()
+
+        GroupResult.save(result)
+        task = GroupResult.restore(result.id)
+        assert task.results == []
+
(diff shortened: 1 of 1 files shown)


AlexHill:
The problem here is that this test is using the base `GroupResult` instead of the `manager.app.GroupResult`, which overrides the `app` property with a valid `app` attr.

`GroupResult.restore()` could potentially fall back on using `current_app` in this case, but it doesn't currently.

This seems to do the trick:

```python
    @flaky
    def test_empty_group_result(self, manager):
        if not manager.app.conf.result_backend.startswith('redis'):
            raise pytest.skip('Requires redis result backend.')

        task = group([])
        result = task.apply_async()
        result.save()

        task = manager.app.GroupResult.restore(result.id)
        assert isinstance(task, GroupResult)
        assert task.results == []
```

This is how `GroupResult` is used in the existing unit tests.

thedrow:
Thanks. I'll fix this and merge :)

PR #4429, opened 2017-12-09
Introduced Build Stages to our build process

thedrow:
We will now run the linting tools before anything else.
This will save time on our build queue and reduce feedback cycles.
I've also added a minor improvement that makes flake8 run with two parallel jobs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,16 +29,22 @@ matrix:
     env: TOXENV=pypy-integration-dynamodb PYPY_VERSION="pypy2.7-5.8.0"
   - python: '3.5'
     env: TOXENV=flake8
+    stage: lint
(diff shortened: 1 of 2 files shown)


georgepsarakis:
@thedrow very nice thought! As a future enhancement, I would suggest to also separate unit and integration tests, also for better readability.

auvipy:
I agree with @georgepsarakis

thedrow:
@georgepsarakis This is currently not easily supported by Travis since we don't want to specify all build jobs manually.
I totally agree. Let's open an issue about it.

thedrow:
See https://github.com/celery/celery/issues/4434
Let's discuss the issue there :)

PR #4431, opened 2017-12-09
Fix current_app fallback in GroupResult.restore()

AlexHill:
See #4428 - when `result()` is called on the base `GroupResult` class, with no `app` or `backend` arguments, it fails to correctly fall back to using `current_app`.


thedrow:
Thank you very much!

PR #4432, opened 2017-12-09
Resolve TypeError on `.get` from nested groups

### LINKED ISSUE #4274: TypeError: get() got an unexpected keyword argument 'on_interval' when nested chain ends in group
## Checklist

       version affected).

## Steps to reproduce

Given this task…

```python
from myapp.celery import app

@app.task
def say_something(msg):
    print(msg)
```

…this moderately complex workflow…

```python
import celery

from myapp.tasks import say_something

workflow = celery.chain(
    say_something.si('outer chain start'),
    celery.group(
        say_something.si('outer group'),
        celery.chain(
            say_something.si('inner chain start'),
            celery.group(
                say_something.si('inner group a'),
                say_something.si('inner group b'),
            ),
            # say_something.si('inner chain end'),
        ),
    ),
    say_something.si('outer chain end'),
)

workflow.delay()
```

…causes this error…

    ERROR:celery.app.builtins:Chord '70cb9fbe-4843-49ba-879e-6ffd76d63226' raised: TypeError("get() got an unexpected keyword argument 'on_interval'",)
      File "/usr/local/lib/python3.6/site-packages/celery/app/builtins.py", line 80, in unlock_chord
        ret = j(timeout=3.0, propagate=True)
      File "/usr/local/lib/python3.6/site-packages/celery/result.py", line 698, in join
        interval=interval, no_ack=no_ack, on_interval=on_interval,
    TypeError: get() got an unexpected keyword argument 'on_interval'
    INFO:celery.app.trace:Task celery.chord_unlock[cac53aaf-c193-442c-b060-73577be77d0f] succeeded in 0.04921864200150594s: None

…and the `outer chain end` task is never executed.

If I un-comment the `inner chain end` task, the whole workflow executes as expected.

Below is a sample of the output of `celery -A proj report`:

    software -> celery:4.1.0 (latentcall) kombu:4.1.0 py:3.6.1
    settings -> transport:amqp results:django-db
    
    BROKER_URL: 'amqp://admin:********@rabbitmq:5672//'
    
    CELERY_ACCEPT_CONTENT: ['json']
    CELERY_RESULT_BACKEND: 'django-db'
    CELERY_TASK_SERIALIZER: 'json'
    CELERY_WORKER_HIJACK_ROOT_LOGGER: False

## Expected behaviour

The `outer chain end` should be executed once all other tasks have completed.

## Actual behavior

    ERROR:celery.app.builtins:Chord '70cb9fbe-4843-49ba-879e-6ffd76d63226' raised: TypeError("get() got an unexpected keyword argument 'on_interval'",)
      File "/usr/local/lib/python3.6/site-packages/celery/app/builtins.py", line 80, in unlock_chord
        ret = j(timeout=3.0, propagate=True)
      File "/usr/local/lib/python3.6/site-packages/celery/result.py", line 698, in join
        interval=interval, no_ack=no_ack, on_interval=on_interval,
    TypeError: get() got an unexpected keyword argument 'on_interval'
    INFO:celery.app.trace:Task celery.chord_unlock[cac53aaf-c193-442c-b060-73577be77d0f] succeeded in 0.04921864200150594s: None

myw:
## Description

#4274 is caused by the fact that when the result backend does not support native join, `ResultSet.get` calls `ResultSet.join`, which itself calls `ResultSet.get`. When that second `.get` call is made, `on_interval` is passed in. 

Currently, `ResultSet.get` does not take such an argument, so the call results in the `TypeError` described in the issue. However, since `ResultSet.join` _does_ take such an option (as does `.join_native`), it's safe to pass along down to the join methods and should be accepted by `.get`. 

This PR changes `ResultSet.get` to accept and pass along the `on_interval` argument to `ResultSet.join` (or `ResultSet.join_native`), which resolves the issue.

The PR also adds a unit test that prevents regression of this behavior. The unit test was verified to fail on `master` (at https://github.com/celery/celery/commit/28c2c09d380c62f9e17776811735a5c8c4ed8320), but passes on the current branch head. The unit test only tests the behavior when native join is not supported by the backend, as backends that support native join did not trigger this error. The unit test can be easily changed to cover both scenarios, just in case, if deemed necessary.

Finally, the PR adds an integration test that tests the scenario laid out in the unit test. As the integration test is run on the DynamoDB backend, which does not support native join, it should also prevent further regression of this behavior.

Fixes #4274

### DIFF
--- a/celery/result.py
+++ b/celery/result.py
@@ -628,7 +628,7 @@ class ResultSet(ResultBase):
 
     def get(self, timeout=None, propagate=True, interval=0.5,
             callback=None, no_ack=True, on_message=None,
-            disable_sync_subtasks=True):
+            disable_sync_subtasks=True, on_interval=None):
         """See :meth:`join`.
 
         This is here for API compatibility with :class:`AsyncResult`,
@@ -640,7 +640,8 @@ class ResultSet(ResultBase):
         return (self.join_native if self.supports_native_join else self.join)(
             timeout=timeout, propagate=propagate,
             interval=interval, callback=callback, no_ack=no_ack,
-            on_message=on_message, disable_sync_subtasks=disable_sync_subtasks
+            on_message=on_message, disable_sync_subtasks=disable_sync_subtasks,
+            on_interval=on_interval,
         )
 
     def join(self, timeout=None, propagate=True, interval=0.5,
--- a/t/integration/test_canvas.py
+++ b/t/integration/test_canvas.py
@@ -130,6 +130,33 @@ class test_group:
             assert parent_id == expected_parent_id
             assert value == i + 2
 
+    @flaky
+    def test_nested_group(self, manager):
+        assert manager.inspect().ping()
+
+        c = group(
+            add.si(1, 10),
+            group(
+                add.si(1, 100),
+                group(
+                    add.si(1, 1000),
+                    add.si(1, 2000),
+                ),
+            ),
+        )
+        res = c()
+
+        assert res.get(timeout=TIMEOUT) == [
+            11,
+            [
+                101,
+                [
+                    1001,
+                    2001,
+                ],
+            ]
+        ]
+
 
 def assert_ids(r, expected_value, expected_root_id, expected_parent_id):
     root_id, parent_id, value = r.get(timeout=TIMEOUT)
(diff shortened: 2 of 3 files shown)


thedrow:
This is a wonderful contribution with a detailed and clear description of the fix.
Thank you very much! :+1: 

We'd like to add an integration test for this, if you can, please use the example provided in #4274.
If you won't get to do it, I promise I will :)

myw:
I'll give it a shot—integration testing takes a bit more context to jump into, but from what I gather, your TOX config will run the integration tests with DynamoDB, which does not support `join_native`, so this would get tested. I can't really set this up on my own to verify that it fails without the patch, but hopefully it's a direct enough map to the original issue that it suffices.

myw:
Using the workflow described in #4274 failed in integration testing; it seemed like this was due to issues running chords in DynamoDB and getting them to finish before the timeout. In any case, it seems like Redis is the only backend used with chords on integration testing, and that would not test the use case. To get around this, I switched to a simpler example workflow that only uses groups. It's not as close to the original issue, but it mimics the GroupResult structure tested in the unit test, so it should suffice. Please let me know if you need anything else.

Thanks for making Celery, folks!

thedrow:
I've added the original test case as well and skipped it if the backend doesn't support native joins. I just prefer to be on the safe side :)
I'll deal with the test failures from here.
Thank you very much!

thedrow:
Thank you very much!

PR #4435, opened 2017-12-10
Parallel doc lints

thedrow:
This will enable us to run our doc lints faster. We do need to install the newest version of sphinx_celery though.
See https://github.com/celery/sphinx_celery/issues/4 & https://github.com/celery/sphinx_celery/pull/5.

### DIFF
--- a/celery/contrib/sphinx.py
+++ b/celery/contrib/sphinx.py
@@ -88,3 +88,7 @@ def setup(app):
     app.add_autodocumenter(TaskDocumenter)
     app.add_directive_to_domain('py', 'task', TaskDirective)
     app.add_config_value('celery_task_prefix', '(task)', True)
+    
+    return {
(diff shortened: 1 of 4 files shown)


auvipy:
should the windows 2.7 failure ignored?

thedrow:
It's a network error on their part.
I'm gonna wait for @georgepsarakis to approve this for a bit before merging this.

georgepsarakis:
@thedrow I am adding the Sphinx [documentation link](http://www.sphinx-doc.org/en/stable/extdev/#extension-metadata) as a reference. So, this only means that files are being parsed by Sphinx in parallel? I compared the completion times of build jobs with another PR and the speed improvement is in the order of 5-10% perhaps.

thedrow:
Yes, files are read in parallel.
5%-10% was worth the time I spent on this.

Issue #1014, opened 2012-09-30
Chain and chord exception handling


meowcoder:
For example this simple chain:

``` python
result = (add.s('1', 2) | mul.s(3))()
result.state # 'PENDING'
result.get() 
# Never returns or raises any exception on client-side.
# Expected result:
result.state # 'FAILURE' or something
result.get()
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
# skipped
TypeError: cannot concatenate 'str' and 'int' objects

# Here is a hacky workaround:
def check_exception(fn):
    def wrapper(*args, **kw):
        lst = list(args) + list(kw.values())
        exceptions = [x for x in lst if isinstance(x, Exception)]
        if exceptions:
            return exceptions[0]

        try:
            return fn(*args, **kw)
        except Exception, exc:
            return exc
    return wrapper

@celery.task(name='add')
@check_exception
def add(x, y):
    return x + y

# in this case return value is:
TypeError("unsupported operand type(s) for +: 'int' and 'str'")
# which is also acceptable, even though raise would be more appropriate
```

Well, actually get() blocks,  as for chain I don't know what's happening to it, looks like it just vanishes into the void, at least I don't see any scheduled or active tasks.

Am I doing something wrong or is this a bug?

**UPDATE**

``` python
@celery.task(name='test-chain')
def test_chain(x):
    finish = finished.s()

    a = add.s(x, 2)
    a.link_error(finish)

    b = mul.s(3)
    b.link_error(finish)

    return celery.chain(a, b, finish)
```

This code actually kinda works, I get to wait for finish subtask, instead of never executed mul subtask, and I get result every time.
Verbose and not pretty, but it's ok, I guess.

Now there is another issue:

``` python
@celery.task(name='filter_exceptions')
def filter_exceptions(args):
    for exception in filter(isexception, args):
        raise exception

    return args

exception_filter = filter_exceptions.s()
projects = celery.group([get_project.s(id) for id in projects])

header = celery.group(
    get_submission.s(submission.id),
    get_jira_issue.s(str(submission.sub_ticket))
)

task = (header | exception_filter | projects)
```

Is there any way to avoid exception_filter? Without it get_project subtasks might receive (Exception, issue) tuple as first argument, where as I want get_project subtasks not to execute at all.

jurrian:
[...]

PR #4442, opened 2017-12-12
Fix for get() follow_parents exception handling (#1014)

jurrian:
## Description

When working with chains, the situation might be that the first chain task has failed. Calling `x.get(propagate=True, follow_parents=True)` should cope with that since it checks if the parent tasks have raised exceptions.

However, in my experience, when `x.get(propagate=True, follow_parents=True)` is called the failed task is still pending at that moment and will become failed seconds later. This creates a bug in the behaviour of `follow_parents`.

This fix calls `get(propagate=True, follow_parents=False)` on each parent instead, which will cause it to raise directly when the parent task becomes failed. It looks like `maybe_throw()` can be safely replaced since it is called by `get()` at some other place.

In order to be more comprehensive, I extended the documentation for `follow_parents`.

I am not experienced enough in this project to see the whole picture, so please advise on possible problems that might arise with this fix for #1014 .


jurrian:
I am not yet sure if I can provide these tests. Maybe I'll have some time later to dive into it.
If someone urgently needs this fix, contributions are welcome!

PR #4443, opened 2017-12-12
Support chords with empty headers

### LINKED ISSUE #4333: chord does nothing when the header is an empty list
## Checklist

- [ x ] I have included the output of ``celery -A proj report`` in the issue.
       version affected). 
Affected version: 4.1.0

- [ x ] I have verified that the issue exists against the `master` branch of Celery.

## Steps to reproduce

Initiate the Celery configuration, and run the following code.
```py
@celery.task()
def add(a, b):
    return sum(a) + b

print(chord([], add.s(3)).apply_async().get())
```

## Expected behavior

Does not wait, executes the callback task, prints 3.

## Actual behavior

Does nothing, the `get` call is stuck forever. I do not know if it's a bug, or a non-feature, but it should be at least documented. The use case behind this is we generate a list dynamically using Python's list comprehension `[my_task.s(i) for i in range(n)]`, the list is potentially empty.

Another thing is the fact that the response sent to the callback is a list only if there is more than one element in the list is also counter intuitive and should be at best overridable.

http://docs.celeryproject.org/en/latest/userguide/canvas.html#chords

Thanks a lot in advance

AlexHill:
When a chord's header group contains no tasks, its body is never executed. The reason for this is pretty straightforward: execution of the body is triggered by completion of header tasks. If there are no header tasks to be completed, the body is never executed.

An empty `GroupResult` is considered completed and successful, so I think it's reasonable to expect a chord with an empty group header to complete.

The fix is to run the body "manually" in `chord.run()` when there are no header tasks.

In addition, I've pulled out the parts of the various `apply_chord()`s that actually call the header and changed its signature accordingly, instead calling the header in `chord.run()` before calling `apply_chord()` only if there are header tasks.

Fixes #4333

### DIFF
--- a/celery/backends/base.py
+++ b/celery/backends/base.py
@@ -412,23 +412,19 @@ class Backend(object):
     def on_chord_part_return(self, request, state, result, **kwargs):
         pass
 
-    def fallback_chord_unlock(self, group_id, body, result=None,
-                              countdown=1, **kwargs):
-        kwargs['result'] = [r.as_tuple() for r in result]
+    def fallback_chord_unlock(self, header_result, body, countdown=1,
+                              **kwargs):
+        kwargs['result'] = [r.as_tuple() for r in header_result]
         self.app.tasks['celery.chord_unlock'].apply_async(
-            (group_id, body,), kwargs, countdown=countdown,
+            (header_result.id, body,), kwargs, countdown=countdown,
         )
 
     def ensure_chords_allowed(self):
         pass
 
-    def apply_chord(self, header, partial_args, group_id, body,
-                    options={}, **kwargs):
+    def apply_chord(self, header_result, body, **kwargs):
         self.ensure_chords_allowed()
-        fixed_options = {k: v for k, v in items(options) if k != 'task_id'}
-        result = header(*partial_args, task_id=group_id, **fixed_options or {})
-        self.fallback_chord_unlock(group_id, body, **kwargs)
-        return result
+        self.fallback_chord_unlock(header_result, body, **kwargs)
 
     def current_task_children(self, request=None):
         request = request or getattr(get_current_task(), 'request', None)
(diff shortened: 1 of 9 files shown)


thedrow:
Quite the radical change we have here. I'd have to review this after coffee :)

AlexHill:
Yep, fair enough - the kernel of the changes is that the header is called in `chord.run` instead of in `apply_chord` - the rest of the changes flow on from there. Let me know if anything needs clarification :)

celery/backends/base.py
```diff
@@ -412,23 +412,19 @@ def add_to_chord(self, chord_id, result):
     def on_chord_part_return(self, request, state, result, **kwargs):
         pass
 
-    def fallback_chord_unlock(self, group_id, body, result=None,
-                              countdown=1, **kwargs):
-        kwargs['result'] = [r.as_tuple() for r in result]
+    def fallback_chord_unlock(self, header_result, body, countdown=1,
```
thedrow: Are we sure this won't break anyone's code?

celery/backends/redis.py
```diff
@@ -251,15 +251,13 @@ def _unpack_chord_result(self, tup, decode,
...
-                    result=None, options={}, **kwargs):
+    def apply_chord(self, header_result, body, **kwargs):
         # Overrides this to avoid calling GroupResult.save
         # pylint: disable=method-hidden
         # Note that KeyValueStoreBackend.__init__ sets self.apply_chord
         # if the implements_incr attr is set.  Redis backend doesn't set
         # this flag.
-        options['task_id'] = group_id
-        return header(*partial_args, **options or {})
+        pass
```
thedrow: Are we sure this applies for all cases? That code is there for a reason.

AlexHill: Yep - `apply_chord()` is only called from `chord.run()`, and we're unconditionally calling the header (with the replaced task ID) there now: https://github.com/celery/celery/pull/4443/files#diff-e27381cb5c64031dd28aa6855c78cfddR1277

AlexHill:
Yeah, I wasn't entirely sure about what's considered a public. If there are third-party backends that out there that inherit from Celery's base backend, this could certainly break them. Googling "celery third-party backends" doesn't yield anything I significant that I can see.

If we can get away with it's a nice simplification - we were calling the header (and replacing the group ID) in three separate places, and having to pass along the arguments everywhere and so on. Now we just call the header in the one place the same way, and the backends use `apply_chord` to do whatever setup is required to handle results.

That said if you want to be strict about maintaining these signatures (and the behaviour of calling the header inside each backend's `apply_chord`) then I get that, and can dial this back a little.

thedrow:
This fixes a real problem so I'm going to risk it.

PR #4445, opened 2017-12-13
Report coverage correctly

thedrow:
As it turns out this repository does not report coverage to codecov at all since the path to the executables has changed at some point.
This should fix the problem.


thedrow:
We're being pinged by codecov again.
I hope this works as expected :)

PR #4448, opened 2017-12-13
Run chord_unlock on same queue as chord body - fixes #4337

### LINKED ISSUE #4337: chord_unlock isn't using the same queue as chord
I've created a tasks_queue and priority_task_queue. in addition to that i've set the tasks_queue to be used as default queue. Now every time i would like a task to be perfrom right away i use priority_queue which is generally empty most of the time. now, the problem araise whenever i'm using chord. i noticed that chords are unable to complete their task because their chord_unlock counterpart always run on the default tasks queue(tasks_queue)

## Steps to reproduce

- Ubnutu 16 server
- celery==4.1.0
- kombu==4.1.0
- boto==2.48.0
- boto3==1.4.7
- botocore==1.7.31
- Using SQS for as broker.

## Expected behavior
I expected the chord_unlock to be performed in the same queue the chord was executed. This causes a major problem for me. Since SQS does not support priority tasks. I'm forced to use two queues to enable the effect of tasks priority. but since chord_unlock will always use the same default queue. priority_tasks using chords, will never be able to finish whenever the default queue is packed.

## Actual behavior
the chord was consumed from priority_task_queue and the chord_unlock got stuck in line on the regular tasks_queue

AlexHill:
See issue #4337. This passes a `queue` argument when calling `chord_unlock.apply_async()`. The value is taken from body.options['queue'], falling back to body.type.queue or None.

### DIFF
--- a/t/unit/backends/test_base.py
+++ b/t/unit/backends/test_base.py
@@ -63,6 +63,12 @@ class test_BaseBackend_interface:
     def setup(self):
         self.b = BaseBackend(self.app)
 
+        @self.app.task(shared=False)
+        def callback(result):
+            pass
+
+        self.callback = callback
+
     def test__forget(self):
         with pytest.raises(NotImplementedError):
             self.b._forget('SOMExx-N0Nex1stant-IDxx-')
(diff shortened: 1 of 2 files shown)


t/unit/backends/test_base.py
```diff
@@ -80,9 +86,33 @@ def test_apply_chord(self, unlock='celery.chord_unlock'):
             uuid(),
             [self.app.AsyncResult(x) for x in range(3)],
         )
-        self.b.apply_chord(header_result, None)
+        self.b.apply_chord(header_result, self.callback.s())
```
thedrow: Why did you change this test?

thedrow:
Please ignore the coverage failure. I fixed it today in master :)

t/unit/backends/test_base.py
AlexHill: `fallback_chord_unlock` now uses the `body` argument instead of simply passing it on to `chord_unlock`, so where `None` worked before, it needs to be real signature now.

PR #4456, opened 2017-12-16
Perform a serialization roundtrip on eager apply_async.

### LINKED ISSUE #4008: Eager mode hides serialization side-effects
Running a task in eager mode doesn't pass the arguments through serialization, and therefore hides some of its side-effects. For example, calling a celery task in eager mode with UUID arguments, will leave them as UUIDs, whereas passing them through serialization (with eager mode switched off) converts them to strings. This hid a bunch of bugs in our application code.

Do you agree it would be more reasonable to pass the arguments through the serialization backends, even in eager mode, to have a more consistently behaving API?

charettes:
Fixes #4008 

/cc @AlexHill

### DIFF
--- a/t/unit/app/test_builtins.py
+++ b/t/unit/app/test_builtins.py
@@ -94,7 +94,9 @@ class test_group(BuiltinsCase):
         self.maybe_signature = self.patching('celery.canvas.maybe_signature')
         self.maybe_signature.side_effect = pass1
         self.app.producer_or_acquire = Mock()
-        self.app.producer_or_acquire.attach_mock(ContextMock(), 'return_value')
+        self.app.producer_or_acquire.attach_mock(
+            ContextMock(serializer='json'), 'return_value'
+        )
         self.app.conf.task_always_eager = True
         self.task = builtins.add_group_task(self.app)
         BuiltinsCase.setup(self)
--- a/celery/app/task.py
+++ b/celery/app/task.py
@@ -5,6 +5,7 @@ from __future__ import absolute_import, unicode_literals
 import sys
 
 from billiard.einfo import ExceptionInfo
+from kombu import serialization
 from kombu.exceptions import OperationalError
 from kombu.utils.uuid import uuid
 
@@ -514,6 +515,17 @@ class Task(object):
 
         app = self._get_app()
         if app.conf.task_always_eager:
+            with app.producer_or_acquire(producer) as eager_producer:
+                serializer = options.get(
+                    'serializer', eager_producer.serializer
+                )
+                body = args, kwargs
+                content_type, content_encoding, data = serialization.dumps(
+                    body, serializer
+                )
+                args, kwargs = serialization.loads(
+                    data, content_type, content_encoding
+                )
             with denied_join_result():
                 return self.apply(args, kwargs, task_id=task_id or uuid(),
                                   link=link, link_error=link_error, **options)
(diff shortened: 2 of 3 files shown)


t/unit/app/test_builtins.py
```diff
@@ -104,6 +107,11 @@ def test_apply_async_eager(self):
         self.task.apply_async((1, 2, 3, 4, 5))
         self.task.apply.assert_called()
 
+    def test_apply_async_eager_serialization_failure(self):
+        self.task.apply = Mock(name='apply')
+        with pytest.raises(EncodeError):
+            self.task.apply_async((1, 2, 3, 4, {1}))
+
```
AlexHill: Is there somewhere better this can go than in the groups test case?

AlexHill:
Thanks @charettes! @thedrow what's your position on this change?

t/unit/app/test_builtins.py
charettes: Probably, I wasn't sure where exactly it should live. Any suggestions?

charettes:
> I think we'll need to look at Signature and all of its subclasses as well

Should I move the code the to `Signature.apply_async` instead?

t/unit/app/test_builtins.py
thedrow: The test case should live under test_tasks.py in my opinion.

AlexHill:
> Should I move the code the to Signature.apply_async instead?

I think having that code where it is now will cover most cases, but I'm not 100% sure, sorry. We'll need to write tests that check that this works consistently with direct invocation (i.e. `my_task.delay(*args)`), signatures (`my_task.s(*args).delay()`), but also chords, chains, groups, etc. I'll look deeper into this when I have more time.

georgepsarakis:
@AlexHill I think the change applies for `Signature` as well, since a call to `Signature.apply_async` essentially calls the task's `apply_async`.

@charettes are you interested in continuing this work? I think this is an important change, so it could be included in the next release. Please let me know how we can help.

charettes:
@georgepsarakis I rebased and addressed the remaining comment by moving the test to `test_tasks.py`. Let me know if you need anything from me to get this merged.

georgepsarakis:
@AlexHill do you have any further feedback on this? Otherwise it should be ready for merge I think.

curtisgr:
Hello, the inclusion of this line: https://github.com/celery/celery/pull/4456/files#diff-a4622e17daa3bfaf3dc55dbd6b35ced7R518 changes the behavior such that we now have to specify a broker when running our django test suite, whereas before we did not need to do so.

Is this a desired side effect of this change? Are we abusing celery in some way by not wanting to specify or use a broker during test runs?

alexkiro:
> Hello, the inclusion of this line: https://github.com/celery/celery/pull/4456/files#diff-a4622e17daa3bfaf3dc55dbd6b35ced7R518 changes the behavior such that we now have to specify a broker when running our django test suite, whereas before we did not need to do so.

FWIW, I also ran into a situation where this behavior was breaking my tests as well. As I was using various mocked objects that were being passed around through tasks in my tests, and while they were serialized I was losing references to them in the test case. 

To work around this, I disabled this new behavior by registering a `noop` serializer:

```
from kombu import serialization
serialization.register("noop", lambda x: x, lambda x: x, "x-python/noop")
```

And then configuring celery to use that. E.g. for django:

```
from unittest.mock import patch
patch("django.conf.settings.CELERY_TASK_SERIALIZER", "noop").start()
```

PR #4468, opened 2017-12-29
Integration test to verify PubSub unsubscriptions

georgepsarakis:
## Description

Introduce an integration test to verify that PubSub channels Redis Backend are not increasing with time. See also #3812.

- [x] Add an integration task with a sleep time, in order to detect tasks in `STARTED` state.
- [x] Count the number of PubSub channels (subscriptions) before and after the chord execution.

### DIFF
--- a/t/integration/tasks.py
+++ b/t/integration/tasks.py
@@ -15,6 +15,15 @@ def add(x, y):
     return x + y
 
 
+@shared_task
+def delayed_sum(numbers, pause_time=1):
+    """Sum the iterable of numbers."""
+    # Allow the task to be in STARTED state for
+    # a limited period of time.
+    sleep(pause_time)
+    return sum(numbers)
+
+
 @shared_task(bind=True)
 def add_replaced(self, x, y):
     """Add two numbers (via the add task)."""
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -87,4 +87,4 @@ commands =
     pydocstyle {toxinidir}/celery
 
 [testenv:isort-check]
-commands = isort --project celery --order-by-type -rc -c {toxinidir}/celery {toxinidir}/t
+commands = isort --project celery --diff --order-by-type -rc -c {toxinidir}/celery {toxinidir}/t
```
thedrow: Why this change? If needed we can discuss it in a separate PR.

georgepsarakis: @thedrow it is not directly related, but it was quite difficult to understand why the `isort-check` build job was failing. By providing the diff, there is actual feedback to the developer about the changes that `isort` would perform in order to pass.

I agree that however small the change, optimally we should apply it separately. Let me know what works best in your opinion.

thedrow: Please pull this change to a different PR as it is not crucial for us to merge in order to release 4.2.

georgepsarakis: Opened #4469 for this purpose.

georgepsarakis: @thedrow I removed the commit and rebased.

PR #4473, opened 2018-01-04
request on_timeout now ignores soft time limit exception (fixes #4412)

alexgarel:
## Description

When Request.on_timeout receive a **soft** timeout from billiard, it should just issue a warning, but let the task go on as it may catch the exception and return (this is what soft timeout are for).


alexgarel:
@auvipy here is the PR you suggest me to propose on issue #4412

PR #4474, opened 2018-01-04
Replace TaskFormatter with TaskFilter

robyoung:
Replacing `TaskFormatter` with a `TaskFilter` as that can be more easily reused when overriding the logging system.

I've left the `TaskFormatter` in but marked it as deprecated in case people are using it.


robyoung-digital:
The reason for deprecating the formatter is because it's not being used anymore and the logic has moved into the filter. I didn't want to keep the formatter around because there is a risk the two would diverge and leave a confusing situation where a user may be using the filter but logging is behaving differently to celery (because the filter isn't being used).

With regard to the apicheck error; it seems to failing to connect to ppa.launchpad.net. It is possible to re-run the tests?

auvipy:
i re run the tests

Issue #4430, opened 2017-12-09
Arguments fail to pass to chained members of a group


Chris7:
## Checklist

- [x] I have included the output of ``celery -A proj report`` in the issue.
```
software -> celery:4.1.0 (latentcall) kombu:4.1.0 py:3.6.3
            billiard:3.5.0.3 py-amqp:2.2.2
platform -> system:Linux arch:64bit, ELF imp:CPython
loader   -> celery.loaders.default.Loader
settings -> transport:amqp results:rpc:///

BROKER_URL: 'amqp://guest:********@rabbit:5672//'
CELERYD_PREFETCH_MULTIPLIER: 10
CELERY_ENABLE_UTC: True
CELERY_QUEUES: [<unbound Queue high-pri -> <unbound Exchange ''(direct)> -> high-pri.#>,
 <unbound Queue low-pri -> <unbound Exchange ''(direct)> -> low-pri.#>]
CELERY_RESULT_BACKEND: 'rpc:///'
CELERY_RESULT_SERIALIZER: 'json'
CELERY_TASK_SERIALIZER: 'json'
CELERY_TIMEZONE: None
CELERY_WORKER_DIRECT: True
Queue: <class 'kombu.entity.Queue'>
os: <module 'os' from '/usr/local/lib/python3.6/os.py'>
```
- [x] I have verified that the issue exists against the `master` branch of Celery.

## Steps to reproduce

Here is a minimal test to reproduce:
```
from celery import Celery, group
app = Celery('tasks')

@app.task
def pre():
  print('in pre')
  return 30

@app.task
def post1(arg1=None):
  print('arg1={}'.format(arg1))
  return arg1 + 1e1

@app.task
def post2(arg2=None):
  print('arg2={}'.format(arg2))
  return arg2 + 1e2

@app.task
def post3(arg3=None):
  print('arg3={}'.format(arg3))
  return arg3 + 1e3

@app.task
def post4(arg4=None):
  print('arg4={}'.format(arg4))
  return arg4 + 1e4

def wf():
  return pre.s() | group([post1.s(), post2.s() | group([post3.s(), post4.s()])])
```

Run `wf().apply_async()`

## Expected behavior

`post2` receives the arguments from `pre.s()`.

## Actual behavior

`post2` receives no arguments

p-himik:
I believe this is a duplicate of #3968

auvipy:
closing in favor of the older issue

Issue #3968, opened 2017-04-10
The sequence `chain(task.s(), group(chain(task.s())))` does not provide arguments for the second task


tbazadaykin:
In the [Documentations](http://docs.celeryproject.org/en/latest/userguide/canvas.html#the-primitives) says:

> Groups and chords accepts partial arguments too, so in a chain the return value of the previous task is forwarded to all tasks in the group...

but it doesn't if group contain chain.

**Steps to reproduce**

```
# ct.py
from __future__ import absolute_import, unicode_literals
from celery import Celery

app = Celery(
    'ct',
    broker='amqp://guest@localhost//',
    backend='cache+memcached://127.0.0.1:11211/'
)

@app.task
def inc(x):
   x += 1
   return x

@app.task
def nothing(x):
   return x

if __name__ == '__main__':
    app.start()
```

Start worker:
`python ct.py worker -l info`

In python console:

```
>>> from ct import inc, nothing
>>> from celery import group, chain
>>> t = chain(inc.s(),group(chain(nothing.s(), inc.s()), inc.s()))
>>> r = t.delay(0)
>>> 
```

In `celery` log:

```
[2017-04-10 12:43:36,126: INFO/MainProcess] Received task: ct.inc[5c1beeff-d785-4d81-af2c-35a5bba839ab]  
[2017-04-10 12:43:37,257: WARNING/PoolWorker-1] /home/vagrant/ct-env/local/lib/python2.7/site-packages/celery/app/trace.py:542: RuntimeWarning: Exception raised outside body: TypeError('nothing() takes exactly 1 argument (0 given)',):
Traceback (most recent call last):
  File "/home/vagrant/ct-env/local/lib/python2.7/site-packages/celery/app/trace.py", line 424, in trace_task
    parent_id=uuid, root_id=root_id,
  File "/home/vagrant/ct-env/local/lib/python2.7/site-packages/celery/canvas.py", line 987, in apply_async
    args=args, kwargs=kwargs, **options))
  File "/home/vagrant/ct-env/local/lib/python2.7/site-packages/celery/canvas.py", line 1063, in _apply_tasks
    **options)
  File "/home/vagrant/ct-env/local/lib/python2.7/site-packages/celery/canvas.py", line 567, in apply_async
    dict(self.options, **options) if options else self.options))
  File "/home/vagrant/ct-env/local/lib/python2.7/site-packages/celery/canvas.py", line 597, in run
    first_task.apply_async(**options)
[...]

PR #4481, opened 2018-01-11
Propagate arguments to chains inside groups

Chris7:
This addresses issue #4430 that may be a duplicate of #3968. 

Arguments were not propagated to chains within groups because in the `apply_async` method of `group`, `self._prepared` is called without partial arguments (as shown)
```
    options, group_id, root_id = self._freeze_gid(options)
    tasks = self._prepared(self.tasks, [], group_id, root_id, app)
    p = barrier()
    results = list(self._apply_tasks(tasks, producer, app, p,
                                         args=args, kwargs=kwargs, **options))
```

Consequentially, in `self._prepared`, when the task is frozen and no arguments have been passed, the task becomes frozen without any arguments.

```
    if partial_args and not task.immutable:
        task.args = tuple(partial_args) + tuple(task.args)
    yield task, task.freeze(group_id=group_id, root_id=root_id
```

Thus, a frozen chain has been created with no input arguments. It seems this is done because the subsequent call to `apply_tasks` is supposed to parameterize the tasks with arguments. However, in the `run` method of `_chain`, if a task is frozen, the arguments are not updated and it is run as follows:
```
        if self._frozen:
            tasks, results = self._frozen
        else:
            tasks, results = self.prepare_steps(
                args, self.tasks, root_id, parent_id, link_error, app,
                task_id, group_id, chord,
            )
```

Thus, because the frozen tasks and results were not provided the input arguments, they are ignored. To fix this, I simply removed the condition on `self._frozen`. 

I did this without knowing exactly the reason for frozen and hoping the unit tests would prove if this was a bad idea. This is a very important issue to me, so I'd love additional feedback if this is an unsuitable fix and how it should be implemented.

### DIFF
--- a/celery/canvas.py
+++ b/celery/canvas.py
@@ -570,13 +570,10 @@ class _chain(Signature):
         args = (tuple(args) + tuple(self.args)
                 if args and not self.immutable else self.args)
 
-        if self._frozen:
-            tasks, results = self._frozen
-        else:
-            tasks, results = self.prepare_steps(
-                args, self.tasks, root_id, parent_id, link_error, app,
-                task_id, group_id, chord,
-            )
+        tasks, results = self.prepare_steps(
+            args, self.tasks, root_id, parent_id, link_error, app,
+            task_id, group_id, chord,
+        )
 
         if results:
             if link:
--- a/t/integration/test_canvas.py
+++ b/t/integration/test_canvas.py
@@ -35,6 +35,21 @@ class test_chain:
         res = c()
         assert res.get(timeout=TIMEOUT) == [64, 65, 66, 67]
 
+    @flaky
+    def test_group_results_in_chain(self, manager):
+        # This adds in an explicit test for the special case added in commit
+        # 1e3fcaa969de6ad32b52a3ed8e74281e5e5360e6
+        c = (
+            group(
+                add.s(1, 2) | group(
+                    add.s(1), add.s(2)
+                )
+            )
+        )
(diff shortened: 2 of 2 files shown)


Chris7:
Another fix to this is passing args in `self._prepared` and not passing args in `self._apply_tasks`.

Chris7:
I think it might be a bad cache on travis, as it's passing locally. Going to close/reopen to force a rebuild.

Chris7:
Just needed to rebase, I'll check it out. Thanks for your quick response

auvipy:
looks like unit test is failing now, however I have restarted the build

auvipy:
@AlexHill do you have time to share some insight?

Chris7:
@auvipy It failed because of my second commit that removed the `_frozen` attribute, as I thought it was confusing to leave it around if it was unused. I see the test cases are referencing it though to extract tasks to perform. I'm curious why the cache is needed for _chain but not other methods.

Chris7:
Here is the commit introducing `_frozen`: https://github.com/celery/celery/commit/1e3fcaa969de6ad32b52a3ed8e74281e5e5360e6

Without _frozen, the special case of `group(A.s() | group(B.s(), C.s()))` is still supported. I added in an additional test to confirm this as there were no unit tests added in that commit.

auvipy:
thanks for digging it

Chris7:
Thanks for the quick turnaround!

Issue #4334, opened 2017-10-19
Improving the development process


vladcalin:
Hello celery community.

I have a proposition that I believe can improve the development process and allow new contributors to get involved easier in the development of celery:

I propose to have some easy to spin up development process using docker compose or vagrant that will build the development process in one shot without having to configure databases, brokers, the celery app instance, etc.

I am thinking about having a single script that will spin up a ready development with the required dependecies.

Some high-level example of what I am thinking:

```
$ setup_celery_dev python=3.6 os=ubuntu broker=redis backend=cassandra test_project=../weird-bug-example
```
And after running the command, the developer should have a VM with ubuntu, python3.6, celery and all dependencies installed with redis,cassandra extras, and a mounted volume with some source code that can reproduce some bug, and maybe some aliases for starting a worker, starting celerybeat, etc.

Also, a tag for "Easy picklings" (like django has) can allow new contributors to pick easier some issues to resolve as a way to learning to contribute to this project.

What do you think?

georgepsarakis:
I think what you are describing can be achieved quite well by providing a [Docker Compose](https://docs.docker.com/compose/) file.

georgepsarakis:
@vladcalin I added the tag `good first issue`, which is now [suggested](https://help.github.com/articles/helping-new-contributors-find-your-project-with-labels/) from Github.

auvipy:
there is also an issue to release official docker image and other modern deployment options like kubernete

PR #4482, opened 2018-01-11
Add docker-compose and base dockerfile for development

Chris7:
This adds a docker based development environment. This removes the need for users to install their own rabbit/redis/virtual environment/etc. to begin development of celery. I use this myself (since after moving to docker I have essentially nothing installed on my computer anymore) and saw interest in it from issue #4334 so thought a PR may be appropriate to share my setup.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -455,7 +455,8 @@ Developing and Testing with Docker
 
 Because of the many components of celery, such as a broker and backend,
 `Docker`_ and `docker-compose`_ can be utilized to greatly simplify the
-development and testing cycle.
+development and testing cycle. The docker configuration here requires a
+docker version of at least 17.09.
 
 The docker components can be found within the :file:`docker/` folder and the
 docker image can be built via:
@@ -483,6 +484,10 @@ Some useful commands to run:
 * ``make test``
 
     To run the test suite
+
+* ``tox``
+
+    To run tox and test against a variety of configurations
 
 By default, docker-compose will mount the celery and test folders in the docker
 container, allowing code changes and testing to be immediately visible inside
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -454,8 +454,8 @@ Developing and Testing with Docker
 ----------------------------------
 
 Because of the many components of celery, such as a broker and backend,
-`Docker<https://www.docker.com/>`_ and `docker-compose<https://docs.docker.com/compose/>`_
-can be utilized to greatly simplify the development and testing cycle.
+`Docker`_ and `docker-compose`_ can be utilized to greatly simplify the
+development and testing cycle.
 
 The docker components can be found within the :file:`docker/` folder and the
 docker image can be built via:
@@ -483,8 +483,10 @@ Some useful commands to run:
     To run the test suite
 
 By default, docker-compose will mount the celery and test folders in the docker
-container, allowing code changes and testing to be immediately visible inside
-the docker container.
+container, allowing code changes and testing to be immediately visible inside the docker container.
+
+.. _`Docker`: https://www.docker.com/
+.. _`docker-compose`: https://docs.docker.com/compose/
 
 .. _contributing-testing:
 
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -448,6 +448,44 @@ fetch and checkout a remote branch like this::
     https://notes.envato.com/developers/rebasing-merge-commits-in-git/
 .. _`Rebase`: https://help.github.com/rebase/
 
+.. _contributing-docker-development:
+
+Developing and Testing with Docker
+----------------------------------
+
+Because of the many components of celery, such as a broker and backend,
+`Docker<https://www.docker.com/>`_ and `docker-compose<https://docs.docker.com/compose/>`_
+can be utilized to greatly simplify the development and testing cycle.
+
+The docker components can be found within the :file:`docker/` folder and the
+docker image can be built via:
+
+.. code-block:: console
+
+    $ docker-compose build celery
+
+and run via:
+
+.. code-block:: console
+
+    $ docker-compose run --rm celery <command>
+
+where <command> is a command to execute in a docker container.
+
+Some useful commands to run:
+
+* ``bash``
+
+    To enter the docker container like a normal shell
+
+* ``make test``
+
+    To run the test suite
+
+By default, docker-compose will mount the celery and test folders in the docker
+container, allowing code changes and testing to be immediately visible inside
+the docker container.
+
 .. _contributing-testing:
 
 Running the unit test suite
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -470,7 +470,9 @@ and run via:
 
     $ docker-compose run --rm celery <command>
 
-where <command> is a command to execute in a docker container.
+where <command> is a command to execute in a docker container. The `--rm` flag
+indicates that the container should be removed after it is exited and is useful
+to prevent accumulation of unwanted containers.
 
 Some useful commands to run:
 
@@ -483,7 +485,9 @@ Some useful commands to run:
     To run the test suite
 
 By default, docker-compose will mount the celery and test folders in the docker
-container, allowing code changes and testing to be immediately visible inside the docker container.
+container, allowing code changes and testing to be immediately visible inside
+the docker container. Environment variables, such as the broker and backend to
+use are also defined in the :file:`docker/docker-compose.yml` file.
 
 .. _`Docker`: https://www.docker.com/
 .. _`docker-compose`: https://docs.docker.com/compose/
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -453,13 +453,13 @@ fetch and checkout a remote branch like this::
 Developing and Testing with Docker
 ----------------------------------
 
-Because of the many components of celery, such as a broker and backend,
+Because of the many components of Celery, such as a broker and backend,
 `Docker`_ and `docker-compose`_ can be utilized to greatly simplify the
-development and testing cycle. The docker configuration here requires a
-docker version of at least 17.09.
+development and testing cycle. The Docker configuration here requires a
+Docker version of at least 17.09.
 
-The docker components can be found within the :file:`docker/` folder and the
-docker image can be built via:
+The Docker components can be found within the :file:`docker/` folder and the
+Docker image can be built via:
 
 .. code-block:: console
 
@@ -471,7 +471,7 @@ and run via:
 
     $ docker-compose run --rm celery <command>
 
-where <command> is a command to execute in a docker container. The `--rm` flag
+where <command> is a command to execute in a Docker container. The `--rm` flag
 indicates that the container should be removed after it is exited and is useful
 to prevent accumulation of unwanted containers.
 
@@ -479,7 +479,7 @@ Some useful commands to run:
 
 * ``bash``
 
-    To enter the docker container like a normal shell
+    To enter the Docker container like a normal shell
 
 * ``make test``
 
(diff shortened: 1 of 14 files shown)


pachewise:
could we add documentation on how to setup and use the docker env?

Chris7:
Sure. I'm changing things up so all the integration/etc. tests can pass instead of having it just under python 3.6 so it might take a little bit until everything is figured out.

georgepsarakis:
@Chris7 thank you for your efforts on this, as it certainly has great value for development. Let me know if you need any help!

Chris7:
Thanks @georgepsarakis. I would appreciate any feedback on the documentation I just put up! The dockerfile itself seems to be about complete and I've found the areas that needed updates for non-localhost setups.

docker/Dockerfile
```diff
@@ -0,0 +1,78 @@
...
+    zlib1g-dev
+
+# Install couchbase
+RUN wget http://packages.couchbase.com/clients/c/libcouchbase-2.8.4_jessie_amd64.tar \
+    && tar -vxf libcouchbase-2.8.4_jessie_amd64.tar \
+    && dpkg -i libcouchbase-2.8.4_jessie_amd64/libcouchbase2-core_2.8.4-1_amd64.deb \
+    && dpkg -i libcouchbase-2.8.4_jessie_amd64/libcouchbase-dev_2.8.4-1_amd64.deb
+
+# Setup our home environment
+ENV CELERY_USER someone
```
georgepsarakis: Perhaps simply `celery` as the user/group name would be more intuitive.

docker/Dockerfile
```diff
@@ -0,0 +1,78 @@
...
+RUN addgroup --gid 1000 ${CELERY_USER}
+RUN adduser --system --disabled-password --uid 1000 --gid 1000 ${CELERY_USER}
+RUN usermod -aG sudo ${CELERY_USER}
```
georgepsarakis: Is it necessary to promote the user to sudoer?

docker/Dockerfile
```diff
@@ -0,0 +1,78 @@
...
+
+# Python versions to test against
+RUN VERSION_ALIAS="python2.7" pyenv install 2.7.14
+RUN VERSION_ALIAS="python3.4" pyenv install 3.4.8
+RUN VERSION_ALIAS="python3.5" pyenv install 3.5.5
+RUN VERSION_ALIAS="python3.6" pyenv install 3.6.4
+
+# Install celery
+
+COPY --chown=1000:1000 requirements ${HOME}/requirements
```
georgepsarakis: Is the `--chown` switch for `COPY` new? Perhaps we should state in the documentation which Docker versions are supported.

docker/docker-compose.yml
```diff
@@ -0,0 +1,33 @@
...
+      - ../celery:/home/someone/celery
+      # Because of the __pycache__ files, PYTHONDONTWRITEBYTECODE must be
+      # set on the host as well or py.test will throw configuration errors.
+      - ../t:/home/someone/t
+    depends_on:
+      - rabbit
+      - redis
+      - dynamodb
+
+  rabbit:
```
georgepsarakis: Generally it is recommended to use exact tags instead of latest images, as it improves consistency and we can always ask on bug reports which version of a database someone uses.

docker/entrypoint
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+export PATH="${HOME}/.pyenv/bin:$PATH"
```
georgepsarakis: I am not sure if this is required, as you already have this in an `ENV` instruction.

docker/entrypoint
```diff
@@ -0,0 +1,5 @@
...
+eval "$(pyenv init -)"
+eval "$(pyenv virtualenv-init -)"
+exec "$@"
```
georgepsarakis: :+1: for using `exec` to correctly propagate signals!

celery/contrib/testing/worker.py
```diff
@@ -102,7 +102,7 @@ def _start_worker_thread(app,
     setup_app_for_worker(app, loglevel, logfile)
     assert 'celery.ping' in app.tasks
     # Make sure we can connect to the broker
-    with app.connection() as conn:
+    with app.connection(hostname=os.environ.get('TEST_BROKER')) as conn:
```
georgepsarakis: Will passing `None` here revert to using `localhost`?

CONTRIBUTING.rst
```diff
@@ -448,6 +448,50 @@ fetch and checkout a remote branch like this::
...
+indicates that the container should be removed after it is exited and is useful
+to prevent accumulation of unwanted containers.
+
+Some useful commands to run:
+
+* ``bash``
+
+    To enter the docker container like a normal shell
+
+* ``make test``
```
georgepsarakis: I think this needs to be replaced with the proper `tox` command?

docker/Dockerfile
```diff
@@ -0,0 +1,78 @@
...
+  -r requirements/pkgutils.txt \
+  -r requirements/test.txt \
+  -r requirements/test-ci-base.txt \
+  -r requirements/test-integration.txt
+
+COPY --chown=1000:1000 MANIFEST.in Makefile setup.py setup.cfg tox.ini ${HOME}/
+COPY --chown=1000:1000 t ${HOME}/t
+COPY --chown=1000:1000 celery ${HOME}/celery
+
+RUN pyenv exec pip install -e .
```
georgepsarakis: The Dockerfile is quite extensive. Perhaps if it was divided in smaller scripts, depending on functionality, it would be better. 

Something like:

```
docker/scripts/create-linux-user.sh
docker/scripts/install-couchbase.sh
docker/scripts/install-pyenv.sh
...
```

What do you think?

CONTRIBUTING.rst
Chris7: I can add that to the section. `tox` takes a lot longer to run so I think it's useful to provide the simple `make test` command.

celery/contrib/testing/worker.py
Chris7: It will. The default hostname is `None` in which localhost is assumed

docker/Dockerfile
Chris7: This is up to you -- I made it `someone` because with `celery`, the directories become confusing, having `/home/celery/celery/celery` just looked odd.

docker/Dockerfile
Chris7: I'll remove this. I did this for debugging.

docker/Dockerfile
Chris7: It is new thankfully. It is supported in Docker 17.09+. I'll add a note to the contribution guide.

docker/Dockerfile
Chris7: I agree. I prefer developing this way because it's easier to debug separate layers.

docker/entrypoint
Chris7: I'll check and remove if it is unnecessary.

docker/docker-compose.yml
Chris7: Good point -- will do.

Chris7:
Thanks for the review @georgepsarakis. Let me know your opinions on the username. I added a `.env` file as well for sticking the username (and anything else we'd want to configure for the docker-compose) so it can be changed more easily.

Chris7:
I wanted this addition to add as little as possible until others are more comfortable with docker (and we iron out the inevitable docker bugs from OSX, I'm running linux), so adding more to the makefile and other reorganizations were kept to a minimum. However, one thing that is missing is a test suite that uses the different backend/brokers available because tox is basically a static setup. What are your thoughts on adding a separate make command that would rotate the environment variables for the broker/backend? Or do you have any other ideas on how those could be tested with minimal additions?

docker/Dockerfile
georgepsarakis: True. How about changing it to `celery-user` and on lines 68-70 use trailing slashes to copy only contents? That way you will end up having `/home/celery-user/celery`.

docker/Dockerfile
```diff
@@ -0,0 +1,78 @@
...
+
+# the compiled files from earlier steps will cause py.test to fail with
+# an ImportMismatchError
+RUN find . -name "*.pyc" -delete
```
georgepsarakis: I think the `Makefile` contains something like this: `clean-pyc`

Chris7: It does. :+1:

Chris7: AND it handles `__pycache__`! :champagne:

CONTRIBUTING.rst
```diff
@@ -448,6 +448,55 @@ fetch and checkout a remote branch like this::
     https://notes.envato.com/developers/rebasing-merge-commits-in-git/
 .. _`Rebase`: https://help.github.com/rebase/
 
+.. _contributing-docker-development:
+
+Developing and Testing with Docker
+----------------------------------
+
+Because of the many components of celery, such as a broker and backend,
```
georgepsarakis: Some really minor points; better use `Docker` instead of `docker` everywhere and `Celery` instead of `celery`.

celery/contrib/testing/worker.py
georgepsarakis: Great, thanks for verifying.

docker/Dockerfile
```diff
@@ -0,0 +1,72 @@
...
+    libssl-dev \
+    pkg-config \
+    pypy \
+    wget \
+    zlib1g-dev
+
+# Setup variables. Even though changing these may cause unnecessary invalidation of
+# unrelated elements, grouping them together makes the Dockerfile read better.
+ENV PROVISIONING /provisioning
+ENV CELERY_USER someone
```
georgepsarakis: Let's just change this to `developer` to reduce complexity and options. Having a consistent username will also help with possible bug reports regarding permission errors. Perhaps `docker/.env` is not required as well, if you proceed with this change.

docker/Dockerfile
```diff
@@ -0,0 +1,72 @@
...
+WORKDIR $HOME
+COPY --chown=1000:1000 requirements $HOME/requirements
+COPY --chown=1000:1000 docker/entrypoint /entrypoint
+RUN chmod gu+x /entrypoint
+
+# Define the local pyenvs
+RUN pyenv local python2.7 python3.4 python3.5 python3.6
+
+# Setup one celery environment for basic development use
+RUN pyenv exec pip install \
```
georgepsarakis: The current structure seems to also benefit the Docker image build cache, nice!

docker/docker-compose.yml
```diff
@@ -0,0 +1,34 @@
...
+    build:
+      context: ..
+      dockerfile: docker/Dockerfile
+    environment:
+      TEST_BROKER: pyamqp://rabbit:5672
+      TEST_BACKEND: redis://redis
+      PYTHONUNBUFFERED: 1
+      PYTHONDONTWRITEBYTECODE: 1
+      REDIS_HOST: redis
+      WORKER_LOGLEVEL: INFO
```
georgepsarakis: Perhaps change this to `DEBUG` since you probably want to view all logs during development?

georgepsarakis:
@Chris7 great work so far! If you can address the comments & rebase to latest master, I think this is ready for merge. As we are preparing to release 4.2.0 these days, perhaps it would be good if this could be included.

docker/Dockerfile
Chris7: Developer does sound much better.

The reason `.env` is useful is so `docker-compose` can be properly provisioned with volume overrides (`- ../celery:/home/$CELERY_USER/celery`) instead of copying a chosen user everywhere. However, I forgot that build-time args can be useful for reducing duplication of definitions. I changed this to be:
```
ARGS CELERY_USER
```
which will provision the user on build time and added the build-time arg to the docker-compose file. Thus, the only place `CELERY_USER=developer` appears is in the `.env` file (and this is also the single place to update).

To prevent the inevitable bug from someone running `docker build -t xxx .` for some reason, I added in this snippet as well: `RUN : "${CELERY_USER:?CELERY_USER build argument needs to be set and non-empty.}"` that ensures this build variable is set.

This keeps the options limited and working for the traditional case

georgepsarakis:
@Chris7 just tested your branch, works as promised. @auvipy @thedrow do you mind if I merge this and get it in the 4.2.0 release? Seems quite valuable to have it included mostly for visibility in the documentation, since it is intended for development usage.

thedrow:
@georgepsarakis Not at all. Go right ahead :)

georgepsarakis:
@thedrow when I am trying to merge, the Github dropdown only allows two options:
- `Rebase and merge`
- `Squash and merge`

Did you change any Github settings? My main concern with this approach, is that I found it much harder to work with rebased commits when constructing the changelog, since the PR number is not added, as happens with the merge commit. In any case, rebase is fine by me, even better than asking contributors to rebase their branches, just let me know please.

thedrow:
We changed that setting in 2016, I think. We usually squash the commits.

PR #4484, opened 2018-01-12
Adding the CMDS for Celery and Celery Beat to Run on Azure WebJob

PauloPeres:
This is what worked for us!

## Description
Created the .cmd files to be used on Azure.
Just going into Web Jobs on Azure servers and creating a Continuous Web Job, and uploading the zip files of the celery should work.
Each folder should have a separeted Web Job.
Also whoever use should take a look where their celery package is

This pull request don't fix any bugs, it's just a new "Helper" for the ones who want to use Celery into Azure servers.


thedrow:
Thanks!

Issue #4309, opened 2017-10-02
Results backend has minimal data


johnarnold:
This came up in:
https://github.com/mher/flower/pull/724
and
https://github.com/mher/flower/issues/542

Flower subscribes to celery events and caches them.  However, if flower misses an event (or many, i.e. if it's rebooted), it's cache is inconsistent.  Games played with persisting the cache don't solve the missed event problem.

The simple suggestion is "get state from the results backend;" however, the results backend has minimal metadata about the task.

Is there a design decision specifically to NOT update the results backend with "TaskState" data?  I get that events are intended for realtime consumption, and make sense for in-flight state changes (task sent, etc) but it seems like the worker would write the task state to the backend when it's "finished" with a task?

I started going down the road of "implement a snapshot camera, archive to database," but that seems really redundant with setting up a backend.

jobec:
We subclassed some Celery classes to have more info in our redis result backend. More particularly the name of the task and when it finished.

Celery settings:
```python
CELERY_RESULT_BACKEND = 'path.to.your.module:ExtendedRedisBackend'
CELERY_REDIS_HOST = 'localhost'
CELERY_REDIS_PORT = '6379'
```
Class extensions
```python
from datetime import datetime

from celery import Celery
from celery.backends.redis import RedisBackend
from celery.result import AsyncResult
from kombu.utils.encoding import bytes_to_str
from kombu.utils.objects import cached_property
from tzlocal import get_localzone


class ExtendedCelery(Celery):
    @cached_property
    def AsyncResult(self):  # noqa: N802
        """
        Override the default result class of Celery
        """
        return self.subclass_with_self('path.to.your.module:ExtendedAsyncResult')


class ExtendedRedisBackend(RedisBackend):
    def _store_result(self, task_id, result, state,
                      traceback=None, request=None, **kwargs):

        if state in self.READY_STATES:
            date_done = datetime.now(get_localzone()).isoformat()
        else:
            date_done = None

        meta = {
            'status': state,
            'result': result,
            'traceback': traceback,
            'children': self.current_task_children(request),
            'task_id': bytes_to_str(task_id),
            'task': request.task,
            'date_done': date_done,
        }
        self.set(self.get_key_for_task(task_id), self.encode(meta))
        return result


class ExtendedAsyncResult(AsyncResult):
    @property
    def task(self):
        return self._get_task_meta().get('task')

    @property
    def date_done(self):
        return self._get_task_meta().get('date_done')

```
Celery app
```python
import os

[...]

PR #4490, opened 2018-01-17
Add task properties to AsyncResult, store in backend

johnarnold:
This pr extends AsyncResult and Backend in order to store additional properties of the task execution.

### DIFF
--- a/celery/app/defaults.py
+++ b/celery/app/defaults.py
@@ -180,7 +180,7 @@ NAMESPACES = Namespace(
             type='float', old={'celery_task_result_expires'},
         ),
         persistent=Option(None, type='bool'),
-        extended=Option(True, type='bool'),
+        extended=Option(False, type='bool'),
         serializer=Option('json'),
         backend_transport_options=Option({}, type='dict'),
     ),
--- a/celery/app/defaults.py
+++ b/celery/app/defaults.py
@@ -180,6 +180,7 @@ NAMESPACES = Namespace(
             type='float', old={'celery_task_result_expires'},
         ),
         persistent=Option(None, type='bool'),
+        extended=Option(True, type='bool'),
         serializer=Option('json'),
         backend_transport_options=Option({}, type='dict'),
     ),
--- a/celery/app/task.py
+++ b/celery/app/task.py
@@ -92,6 +92,7 @@ class Context(object):
     errbacks = None
     timelimit = None
     origin = None
+    task = None
     _children = None   # see property
     _protected = 0
 
@@ -128,6 +129,7 @@ class Context(object):
             'retries': self.retries,
             'reply_to': self.reply_to,
             'origin': self.origin,
+            'task': self.task
         }
 
     @property
--- a/celery/app/task.py
+++ b/celery/app/task.py
@@ -92,7 +92,7 @@ class Context(object):
     errbacks = None
     timelimit = None
     origin = None
-    task = None
+    task_name = None
     _children = None   # see property
     _protected = 0
 
@@ -129,7 +129,7 @@ class Context(object):
             'retries': self.retries,
             'reply_to': self.reply_to,
             'origin': self.origin,
-            'task': self.task
+            'task_name': self.task_name
         }
 
     @property
--- a/celery/backends/base.py
+++ b/celery/backends/base.py
@@ -671,23 +671,25 @@ class BaseKeyValueStoreBackend(Backend):
             'date_done': date_done,
         }
 
-        if request:
-            request_meta = {
-                # Some unit tests mock the properties, casting to string to
-                # avoid serialization errors
-                'name': str(getattr(request, 'task', None)),
-                'args': getattr(request, 'args', None),
-                'kwargs': getattr(request, 'kwargs', None),
-                'worker': str(getattr(request, 'hostname', None)),
-                'retries': getattr(request, 'retries', None),
-                'queue': str(request.delivery_info.get('routing_key'))
-                if hasattr(request, 'delivery_info') and
-                request.delivery_info else None
-            }
-            if getattr(request, 'group', None):
-                request_meta['group_id'] = request.group
+        if request and getattr(request, 'group', None):
+            meta['group_id'] = request.group
 
-            meta.update(request_meta)
+        if self.app.conf.find_value_for_key('extended', 'result'):
+            if request:
+                request_meta = {
+                    # Some unit tests mock the properties, casting to string to
+                    # avoid serialization errors
+                    'name': getattr(request, 'task', None),
+                    'args': getattr(request, 'args', None),
+                    'kwargs': getattr(request, 'kwargs', None),
+                    'worker': getattr(request, 'hostname', None),
+                    'retries': getattr(request, 'retries', None),
+                    'queue': request.delivery_info.get('routing_key')
+                    if hasattr(request, 'delivery_info') and
+                    request.delivery_info else None
+                }
+
+                meta.update(request_meta)
 
         self.set(self.get_key_for_task(task_id), self.encode(meta))
         return result
--- a/celery/backends/base.py
+++ b/celery/backends/base.py
@@ -666,14 +666,13 @@ class BaseKeyValueStoreBackend(Backend):
             'status': state,
             'result': result,
             'traceback': traceback,
+            'children': self.current_task_children(request),
             'task_id': bytes_to_str(task_id),
             'date_done': date_done,
         }
 
         if request:
             request_meta = {
-
-                'children': self.current_task_children(request),
                 # Some unit tests mock the properties, casting to string to
                 # avoid serialization errors
                 'name': str(getattr(request, 'task', None)),
--- a/celery/backends/base.py
+++ b/celery/backends/base.py
@@ -672,9 +672,10 @@ class BaseKeyValueStoreBackend(Backend):
 
         if request:
             request_meta = {
+
+                'children': self.current_task_children(request),
                 # Some unit tests mock the properties, casting to string to
                 # avoid serialization errors
-                'children': self.current_task_children(request),
                 'name': str(getattr(request, 'task', None)),
                 'args': getattr(request, 'args', None),
                 'kwargs': getattr(request, 'kwargs', None),
--- a/celery/backends/base.py
+++ b/celery/backends/base.py
@@ -8,6 +8,7 @@
 """
 from __future__ import absolute_import, unicode_literals
 
+import datetime
 import sys
 import time
 from collections import namedtuple
@@ -16,8 +17,8 @@ from functools import partial
 from weakref import WeakValueDictionary
 
 from billiard.einfo import ExceptionInfo
-from kombu.serialization import dumps, loads, prepare_accept_content
 from kombu.serialization import registry as serializer_registry
+from kombu.serialization import dumps, loads, prepare_accept_content
 from kombu.utils.encoding import bytes_to_str, ensure_bytes, from_utf8
 from kombu.utils.url import maybe_sanitize_url
 
@@ -70,14 +71,13 @@ def unpickle_backend(cls, args, kwargs):
 
 
 class _nulldict(dict):
-
     def ignore(self, *a, **kw):
         pass
+
     __setitem__ = update = setdefault = ignore
 
 
 class Backend(object):
-
     READY_STATES = states.READY_STATES
     UNREADY_STATES = states.UNREADY_STATES
     EXCEPTION_STATES = states.EXCEPTION_STATES
@@ -332,6 +332,7 @@ class Backend(object):
     def get_state(self, task_id):
         """Get the state of a task."""
         return self.get_task_meta(task_id)['status']
+
     get_status = get_state  # XXX compat
 
     def get_traceback(self, task_id):
@@ -448,7 +449,6 @@ class Backend(object):
 
 
 class SyncBackendMixin(object):
-
     def iter_native(self, result, timeout=None, interval=0.5, no_ack=True,
                     on_message=None, on_interval=None):
         self._ensure_not_eager()
@@ -656,13 +656,39 @@ class BaseKeyValueStoreBackend(Backend):
 
     def _store_result(self, task_id, result, state,
                       traceback=None, request=None, **kwargs):
+
+        if state in self.READY_STATES:
+            date_done = datetime.datetime.utcnow()
+        else:
+            date_done = None
+
         meta = {
-            'status': state, 'result': result, 'traceback': traceback,
-            'children': self.current_task_children(request),
+            'status': state,
+            'result': result,
+            'traceback': traceback,
             'task_id': bytes_to_str(task_id),
+            'date_done': date_done,
         }
-        if request and getattr(request, 'group', None):
-            meta['group_id'] = request.group
+
+        if request:
+            request_meta = {
+                # Some unit tests mock the properties, casting to string to
+                # avoid serialization errors
+                'children': self.current_task_children(request),
+                'name': str(getattr(request, 'task', None)),
+                'args': getattr(request, 'args', None),
+                'kwargs': getattr(request, 'kwargs', None),
+                'worker': str(getattr(request, 'hostname', None)),
+                'retries': getattr(request, 'retries', None),
+                'queue': str(request.delivery_info.get('routing_key'))
+                if hasattr(request, 'delivery_info') and
+                request.delivery_info else None
+            }
+            if getattr(request, 'group', None):
+                request_meta['group_id'] = request.group
+
+            meta.update(request_meta)
+
         self.set(self.get_key_for_task(task_id), self.encode(meta))
         return result
 
@@ -769,7 +795,7 @@ class KeyValueStoreBackend(BaseKeyValueStoreBackend, SyncBackendMixin):
 class DisabledBackend(BaseBackend):
     """Dummy result backend."""
 
-    _cache = {}   # need this attribute to reset cache in tests.
+    _cache = {}  # need this attribute to reset cache in tests.
 
     def store_result(self, *args, **kwargs):
         pass
--- a/celery/backends/base.py
+++ b/celery/backends/base.py
@@ -677,7 +677,7 @@ class BaseKeyValueStoreBackend(Backend):
         if self.app.conf.find_value_for_key('extended', 'result'):
             if request:
                 request_meta = {
-                    'name': getattr(request, 'task', None),
+                    'name': getattr(request, 'task_name', None),
                     'args': getattr(request, 'args', None),
                     'kwargs': getattr(request, 'kwargs', None),
                     'worker': getattr(request, 'hostname', None),
--- a/celery/backends/base.py
+++ b/celery/backends/base.py
@@ -17,8 +17,8 @@ from functools import partial
 from weakref import WeakValueDictionary
 
 from billiard.einfo import ExceptionInfo
-from kombu.serialization import registry as serializer_registry
 from kombu.serialization import dumps, loads, prepare_accept_content
+from kombu.serialization import registry as serializer_registry
 from kombu.utils.encoding import bytes_to_str, ensure_bytes, from_utf8
 from kombu.utils.url import maybe_sanitize_url
 
--- a/celery/backends/base.py
+++ b/celery/backends/base.py
@@ -677,8 +677,6 @@ class BaseKeyValueStoreBackend(Backend):
         if self.app.conf.find_value_for_key('extended', 'result'):
             if request:
                 request_meta = {
-                    # Some unit tests mock the properties, casting to string to
-                    # avoid serialization errors
                     'name': getattr(request, 'task', None),
                     'args': getattr(request, 'args', None),
                     'kwargs': getattr(request, 'kwargs', None),
--- a/celery/contrib/testing/mocks.py
+++ b/celery/contrib/testing/mocks.py
@@ -13,6 +13,11 @@ except ImportError:
         from mock import Mock
 
 
+class PicklableMock(Mock):
+    def __reduce__(self):
+        return (Mock, ())
+
+
 def TaskMessage(name, id=None, args=(), kwargs={}, callbacks=None,
                 errbacks=None, chain=None, shadow=None, utc=None, **options):
     # type: (str, str, Sequence, Mapping, Sequence[Signature],
@@ -22,7 +27,7 @@ def TaskMessage(name, id=None, args=(), kwargs={}, callbacks=None,
     from celery import uuid
     from kombu.serialization import dumps
     id = id or uuid()
-    message = Mock(name='TaskMessage-{0}'.format(id))
+    message = PicklableMock(name='TaskMessage-{0}'.format(id))
     message.headers = {
         'id': id,
         'task': name,
--- a/celery/contrib/testing/mocks.py
+++ b/celery/contrib/testing/mocks.py
@@ -14,6 +14,7 @@ except ImportError:
 
 
 class PicklableMock(Mock):
+    """A subclass of Mock that is pickleable."""
     def __reduce__(self):
         return (Mock, ())
 
--- a/celery/contrib/testing/mocks.py
+++ b/celery/contrib/testing/mocks.py
@@ -15,6 +15,7 @@ except ImportError:
 
 class PicklableMock(Mock):
     """A subclass of Mock that is pickleable."""
+
     def __reduce__(self):
         return (Mock, ())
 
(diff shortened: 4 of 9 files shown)


celery/backends/base.py
```diff
@@ -647,10 +648,25 @@ def _forget(self, task_id):
...
+            'status': state,
+            'result': result,
+            'traceback': traceback,
             'children': self.current_task_children(request),
             'task_id': bytes_to_str(task_id),
+            'name': request.task,
+            'args': request.args,
+            'kwargs': request.kwargs,
+            'worker': request.hostname,
+            'date_done': date_done,
```
johnarnold: This 'date_done' is kind of hack. Is there a way to access the 'received', 'started', 'finished' values from the task cache from within this backend function?

johnarnold:
Fixes https://github.com/celery/celery/issues/4309

johnarnold:
@auvipy  Test case test_result.test_AsyncResult.test_get_request_meta added.

I also ran test_RedisBackend#test_get_set_forget  and other existing test cases successfully.

What is the isort error you mentioned?

auvipy:
check the travis build. i meant this https://pypi.python.org/pypi/isort  you can fix isort by running a command

celery/backends/base.py
```diff
@@ -647,11 +648,32 @@ def _forget(self, task_id):
 
     def _store_result(self, task_id, result, state,
                       traceback=None, request=None, **kwargs):
```
auvipy: is the changes in this file tested? report shows coverage decrease

johnarnold: @auvipy I'm working through them now.  There's some tests failing that I thought were unrelated to this change, but i'll double check.

johnarnold: @auvipy Ok, it looks pretty clean now. The only tests failing are for LoggingProxy issues which looks unrelated.

As an aside, this is my first change to Celery, and it's a big project to dig into. If you or anyone sees a better or more graceful way to accomplish this, please suggest.

johnarnold:
Should I flatten commits, or will you just squash-merge?

johnarnold:
I'm not really sure what to do about isort... maybe I'm using the wrong parameters, but it's saying the whole project needs to be fixed up.  Suggestions?

auvipy:
run $isort filename.py and you should be automatically fix isort issues in the files related to this pr

johnarnold:
@auvipy I tried.. it doesn't work.

travis-ci error:
```
isort-check runtests: commands[0] | isort --project celery --diff --order-by-type -rc -c /home/travis/build/celery/celery/celery /home/travis/build/celery/celery/t
  /home/travis/build/celery/celery$ /home/travis/build/celery/celery/.tox/isort-check/bin/isort --project celery --diff --order-by-type -rc -c /home/travis/build/celery/celery/celery /home/travis/build/celery/celery/t 
ERROR: /home/travis/build/celery/celery/celery/backends/base.py Imports are incorrectly sorted.
--- /home/travis/build/celery/celery/celery/backends/base.py:before	2018-01-22 03:07:05.686089
+++ /home/travis/build/celery/celery/celery/backends/base.py:after	2018-01-22 03:08:02.460470
@@ -16,8 +16,8 @@
 from weakref import WeakValueDictionary
 
 from billiard.einfo import ExceptionInfo
+from kombu.serialization import dumps, loads, prepare_accept_content
 from kombu.serialization import registry as serializer_registry
-from kombu.serialization import dumps, loads, prepare_accept_content
 from kombu.utils.encoding import bytes_to_str, ensure_bytes, from_utf8
 from kombu.utils.url import maybe_sanitize_url
 
ERROR: InvocationError: '/home/travis/build/celery/celery/.tox/isort-check/bin/isort --project celery --diff --order-by-type -rc -c /home/travis/build/celery/celery/celery /home/travis/build/celery/celery/t'
```
my output using --order-by-type:
```
net-task) johnar@netdev1-westus2:~/github/celery$ !664
isort --project celery --diff --order-by-type -rc -c celery/backends/base.py 
(net-task) johnar@netdev1-westus2:~/github/celery$
net-task) johnar@netdev1-westus2:~/github/celery$ isort --project celery --order-by-type celery/backends/base.py 
(net-task) johnar@netdev1-westus2:~/github/celery$ git status
On branch johnar/extendedresult
nothing to commit, working tree clean
```

my output not using --order-by-type:
```
(net-task) johnar@netdev1-westus2:~/github/celery$ isort celery/backends/base.py 
(net-task) johnar@netdev1-westus2:~/github/celery$ git status
On branch johnar/extendedresult
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

        modified:   celery/backends/base.py

no changes added to commit (use "git add" and/or "git commit -a")
(net-task) johnar@netdev1-westus2:~/github/celery$ isort --project celery --diff --order-by-type -rc -c celery/backends/base.py 
ERROR: /home/johnar/github/celery/celery/backends/base.py Imports are incorrectly sorted.
--- /home/johnar/github/celery/celery/backends/base.py:before   2018-01-22 17:19:31.524941
+++ /home/johnar/github/celery/celery/backends/base.py:after    2018-01-22 17:23:59.012049
```

If I use the default sort, it breaks the test. If I use --order-by-type (which is the way it's checked in), it still breaks the test.  What am I missing?

johnarnold:
@auvipy or anyone else... help?

auvipy:
@thedrow @AlexHill @georgepsarakis could any of you review

celery/backends/base.py
```diff
@@ -647,11 +647,36 @@ def _forget(self, task_id):
...
+
+        if state in self.READY_STATES:
+            date_done = datetime.datetime.utcnow()
```
georgepsarakis: Probably rename to `completed_at` ?

celery/backends/base.py
```diff
@@ -647,11 +647,36 @@ def _forget(self, task_id):
...
+        else:
+            date_done = None
+
         meta = {
-            'status': state, 'result': result, 'traceback': traceback,
-            'children': self.current_task_children(request),
```
georgepsarakis: I would suggest to leave this here. By adding the `children` key conditionally in the metadata dictionary, you may be breaking some other functionality.

celery/backends/base.py
```diff
@@ -647,11 +647,36 @@ def _forget(self, task_id):
...
+            'status': state,
+            'result': result,
+            'traceback': traceback,
             'task_id': bytes_to_str(task_id),
+            'date_done': date_done,
         }
+
+        if request:
+            request_meta = {
+                # Some unit tests mock the properties, casting to string to
```
georgepsarakis: I think this comment refers to the next line (`name` key).

celery/backends/base.py
```diff
@@ -647,11 +647,36 @@ def _forget(self, task_id):
...
+                # avoid serialization errors
+                'children': self.current_task_children(request),
+                'name': str(getattr(request, 'task', None)),
```
georgepsarakis: Perhaps all this logic here can be transferred as a method of `Context` and return this dictionary?

georgepsarakis: Consider using Python2/3 compatibility methods.

t/unit/tasks/test_result.py
```diff
@@ -367,6 +368,28 @@ def test_ready(self):
...
+            children=None,
+            args=['one', 'two'],
+            kwargs={'kwarg1': 'three'},
+            hostname="foo",
+            retries=1,
+            delivery_info={'routing_key': 'celery'}
+        )
+        x.backend.store_result(task_id="1", result='foo', state=states.SUCCESS,
+                               traceback=None, request=request)
+        assert x.name
```
georgepsarakis: I would suggest to be more explicit on these assertions:

```
assert x.args == ['one', 'two']
```

What do you think?

celery/backends/base.py
johnarnold: Done.

celery/backends/base.py
johnarnold: children requires the request object anyway, so putting it behind the conditional should not harm anything, and may guard against issues (principally, test case mocks)

celery/backends/base.py
johnarnold: moved comment... is this what you intended?

celery/backends/base.py
johnarnold: Good idea, but some of the tests don't send Context class as request :(

johnarnold: Can you be more specific?  I turned on py2/py3 inspections and also loaded py2 and py3 virtual envs and didn't get any warnings on these lines.. ?

t/unit/tasks/test_result.py
johnarnold: done!

johnarnold:
@georgepsarakis  I fixed several issues and replied to others, can you take another look? Thanks!

celery/backends/base.py
georgepsarakis: I understand, however my point was that now the `children` key will be missing from the dictionary, unless `request` is not `None`. This could have unwanted side-effects, getting `KeyError` exceptions at other places or certain flows.

celery/backends/base.py
georgepsarakis: This [library](https://github.com/celery/vine/blob/master/vine/five.py), also accessible through [kombu](https://github.com/celery/kombu/blob/master/kombu/five.py) provides the proper Python2/3 compatibility functions. [kombu.utils.encoding](https://github.com/celery/kombu/blob/master/kombu/utils/encoding.py) provides certain functions as well.

georgepsarakis: Great, thanks for checking it! :smile:

celery/backends/base.py
georgepsarakis: It referred to the `name` key, isn't that correct?

celery/backends/base.py
johnarnold: I see your point, the function returns a None if request is none.  Reverted to original behavior.

celery/backends/base.py
johnarnold: name, worker, queue.  I think they all break due to funky mocks without a cast.  Do you think I should change it?  If so what do you suggest?  thanks!

celery/backends/base.py
georgepsarakis: Not important, I just noted this because it encodes the type (`date`) when it actually is `datetime` and I find `done` too generic, even `ready` could be better (as in `READY_STATES`) :smile:

celery/backends/base.py
georgepsarakis: I see that tests are passing in Travis CI. Could you please clarify here, I am not sure I understand your comment?

johnarnold: Sorry, I meant that several tests break on name, worker, queue attributes, because the tests do mocks with weird values (e.g. Mock() instances)  .   For regular usage (not tests) the cast to string is not really needed, as it's a str->str op.  For the mentioned tests, the value ends up as a string repr of the Mock, (or in I think one case, a different class).

Just forcing to a string is draconian but effective in this case.

celery/backends/base.py
johnarnold: @georgepsarakis I'd like to make all the backends consistent if possible -- should we leave this as date_done, or change it (date_ready?) in the database backend as well (maybe make a backwards-compatible reference in database backend?)

georgepsarakis: You are very right. I found references of `date_done` in MongoDB, Cassandra and RDBMS backends. I think you can leave it as is then, sorry!

celery/backends/base.py
georgepsarakis: @johnarnold I am skeptical about this type casting, I don't think it is advisable to modify code logic for testing purposes. Could you revert and let Travis fail so I can examine the broken tests? I can then suggest changes on the tests.

celery/backends/base.py
```diff
@@ -655,11 +655,36 @@ def _forget(self, task_id):
...
             'task_id': bytes_to_str(task_id),
+            'date_done': date_done,
         }
+
+        if request:
+            request_meta = {
+                # Some unit tests mock the properties, casting to string to
+                # avoid serialization errors
+                'name': str(getattr(request, 'task', None)),
+                'args': getattr(request, 'args', None),
```
georgepsarakis: I wonder if there is any security concern with storing arguments. I am afraid we may need to allow users to opt-out from this. What do you think @johnarnold ?

johnarnold: @georgepsarakis Hmm... interesting question. I think if you have access the the results backend, you're already within the security boundary.

We also store the args in the broker when we serialize the task, and probably >80% of users are using the same redis server for both broker/results :)

We also include the args in the (send-task?) event / are  visible in flower.

I can certainly/easily add an option to disable, but would like to give others a chance to comment?

georgepsarakis: @johnarnold sure, let's wait to get other opinions on this. There are also some storage considerations along with the change (think of a large JSON payload. e.g from an API call, as argument being passed to the task). Broker messages are ephemeral, but results may be desired to be retained for longer time in some occasions.

I am inclined to have the setting and a warning in the logs that this behavior will be the default from the next minor version onward. If that sounds like too much work, I can help with the implementation.

@thedrow @auvipy would like to share your opinion perhaps?

celery/backends/base.py
johnarnold: @georgepsarakis I reverted the type casting.

celery/backends/base.py
johnarnold: @georgepsarakis I added a result_extended option, defaulted to True.  Does that look acceptable?

celery/backends/base.py
johnarnold: @georgepsarakis I did a couple rounds of test fixups:
- changing from Mock to Context() where possible to avoid json serialization errors
- changing Mock to PicklableMock to avoid pickle serialization errors
- calling app.set_current() where needed to avoid hitting the celery app trap

Unfortunately there are still some failing tests, e.g. in t/unit/worker/test_strategy.py  because they're using Mock() in properties of a Context object, and it's not super easy to fix, afaik.

Open to suggestions on how to fix.  Or, can go back to draconian type casting ?  Thanks!

celery/backends/base.py
```diff
@@ -656,13 +656,41 @@ def _forget(self, task_id):
...
+            'date_done': date_done,
         }
+
         if request and getattr(request, 'group', None):
             meta['group_id'] = request.group
+
+        if self.app.conf.find_value_for_key('extended', 'result'):
+            if request:
+                request_meta = {
+                    # Some unit tests mock the properties, casting to string to
```
georgepsarakis: Seems that this comment is no longer valid?

docs/userguide/configuration.rst
```diff
@@ -642,6 +642,16 @@ Default: No compression.
 Optional compression method used for task results.
 Supports the same options as the :setting:`task_serializer` setting.
 
+.. setting:: result_extended
+
+``result_extended``
+~~~~~~~~~~~~~~~~~~~~~~
+
+Default: ``True``
```
georgepsarakis: Perhaps this option should use `False` as default, for backwards compatibility.

celery/app/defaults.py
```diff
@@ -180,6 +180,7 @@ def __repr__(self):
             type='float', old={'celery_task_result_expires'},
         ),
         persistent=Option(None, type='bool'),
+        extended=Option(True, type='bool'),
```
georgepsarakis: Nice! :smile: 

Just a note though; as also noted below, I believe this should be set to `False` by default in order to maintain current behavior.

docs/userguide/configuration.rst
johnarnold: Done. thanks!

celery/app/defaults.py
johnarnold: Done. Thanks!

celery/backends/base.py
johnarnold: Removed. Thanks!

johnarnold:
@georgepsarakis @auvipy   Fixed the linting issues, please take a look at the unit tests results. thanks!

johnarnold:
All the tests pass now.  Travis hit one lame redis Timeout on one series, unrelated to the change.

auvipy:
is the patch coverage showing relevant info??

celery/app/task.py
```diff
@@ -92,6 +92,7 @@ class Context(object):
     errbacks = None
     timelimit = None
     origin = None
+    task = None
```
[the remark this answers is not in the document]
johnarnold: @thedrow Thanks for looking!   I did update the user guide, see:
https://github.com/celery/celery/pull/4490/files#diff-cc90971f3ec1a23279b1aa3e1426d247

Since it's internal, I'll change it to task_name.

thedrow:
I think that this new option should be mentioned in http://docs.celeryproject.org/en/latest/userguide/calling.html#results-options with a description of the use case for it.

celery/contrib/testing/mocks.py
```diff
@@ -13,6 +13,13 @@
         from mock import Mock
 
 
+class PicklableMock(Mock):
```
georgepsarakis: @johnarnold would you mind reminding me why this change is required? I also noticed that the `Mock` instance is losing its arguments, so is not fully restored. I can see that `Mock` instances are not pickleable in general, but I was wondering why this is now necessary.

I searched through our previous discussions and I don't think I found any reference. If there is any, please point me to it. Thanks!

t/unit/worker/test_request.py
```diff
@@ -629,6 +632,7 @@ def test_on_hard_timeout(self, patching):
         job.acknowledge = Mock(name='ack')
         job.task.acks_late = True
         job.on_timeout(soft=False, timeout=1337)
+        self.app.set_current()
```
georgepsarakis: Any idea why this is now required also? I am asking just to verify that test cases are still valid before/after the changes.

celery/contrib/testing/mocks.py
johnarnold: Sure, without it you get serialization errors.  It's a shame that TravisCI truncated those builds already, I would just link you.

There are many of them, but here is an example:

```
self = <t.unit.worker.test_request.test_Request object at 0x7fbc7a1edc18>

    def test_on_failure_Terminated(self):
        einfo = None
        try:
            raise Terminated('9')
        except Terminated:
            einfo = ExceptionInfo()
        assert einfo is not None
        req = self.get_request(self.add.s(2, 2))
>       req.on_failure(einfo)

worker/test_request.py:264: 
../../celery/worker/request.py:363: in on_failure
    'terminated', True, string(exc), False)
../../celery/worker/request.py:263: in _announce_revoked
    self.id, reason, request=self, store_result=self.store_errors,
../../celery/backends/base.py:191: in mark_as_revoked
    traceback=None, request=request)
../../celery/backends/base.py:322: in store_result
    request=request, **kwargs)
../../celery/backends/base.py:694: in _store_result
    self.set(self.get_key_for_task(task_id), self.encode(meta))
../../celery/backends/base.py:272: in encode
    _, _, payload = self._encode(data)
../../celery/backends/base.py:276: in _encode
    return dumps(data, serializer=self.serializer)
../../../../venv/celery/lib/python3.6/site-packages/kombu/serialization.py:221: in dumps
    payload = encoder(data)
/usr/lib/python3.6/contextlib.py:99: in __exit__
    self.gen.throw(type, value, traceback)
../../../../venv/celery/lib/python3.6/site-packages/kombu/serialization.py:54: in _reraise_errors
    reraise(wrapper, wrapper(exc), sys.exc_info()[2])
../../../../venv/celery/lib/python3.6/site-packages/vine/five.py:178: in reraise
    raise value.with_traceback(tb)
../../../../venv/celery/lib/python3.6/site-packages/kombu/serialization.py:50: in _reraise_errors
    yield
../../../../venv/celery/lib/python3.6/site-packages/kombu/serialization.py:221: in dumps
    payload = encoder(data)

obj = {'args': None, 'children': [], 'date_done': datetime.datetime(2018, 8, 15, 16, 37, 26, 979482), 'kwargs': None, ...}
dumper = <built-in function dumps>

    def pickle_dumps(obj, dumper=pickle.dumps):
>       return dumper(obj, protocol=pickle_protocol)
E       kombu.exceptions.EncodeError: Can't pickle <class 'unittest.mock.Mock'>: it's not the same object as unittest.mock.Mock

../../../../venv/celery/lib/python3.6/site-packages/kombu/serialization.py:350: EncodeError

Comment (johnarnold) [reply celery/contrib/testing/mocks.py]: Sure, there is a serialization error on Mock in Kombu when running t/unit/worker/test_request.py.
Here's a chunk of a bunch of tracebacks:
```

self = <t.unit.worker.test_request.test_Request object at 0x7fbc79ff8630>

    def test_on_failure_WorkerLostError_redelivered_None(self):
        einfo = None
        try:
            raise WorkerLostError()
        except:
            einfo = ExceptionInfo(internal=True)
        req = self.get_request(self.add.s(2, 2))
        req.task.acks_late = True
        req.task.reject_on_worker_lost = True
        req.delivery_info['redelivered'] = None
>       req.on_failure(einfo)

worker/test_request.py:342: 
../../celery/worker/request.py:367: in on_failure
    self.id, exc, request=self, store_result=self.store_errors,
../../celery/backends/base.py:157: in mark_as_failure
    traceback=traceback, request=request)
../../celery/backends/base.py:322: in store_result
    request=request, **kwargs)
../../celery/backends/base.py:694: in _store_result
    self.set(self.get_key_for_task(task_id), self.encode(meta))
../../celery/backends/base.py:272: in encode
    _, _, payload = self._encode(data)
../../celery/backends/base.py:276: in _encode
    return dumps(data, serializer=self.serializer)
../../../../venv/celery/lib/python3.6/site-packages/kombu/serialization.py:221: in dumps
    payload = encoder(data)
/usr/lib/python3.6/contextlib.py:99: in __exit__
    self.gen.throw(type, value, traceback)
../../../../venv/celery/lib/python3.6/site-packages/kombu/serialization.py:54: in _reraise_errors
    reraise(wrapper, wrapper(exc), sys.exc_info()[2])
../../../../venv/celery/lib/python3.6/site-packages/vine/five.py:178: in reraise
    raise value.with_traceback(tb)
../../../../venv/celery/lib/python3.6/site-packages/kombu/serialization.py:50: in _reraise_errors
    yield
../../../../venv/celery/lib/python3.6/site-packages/kombu/serialization.py:221: in dumps
    payload = encoder(data)

obj = {'args': None, 'children': [], 'date_done': datetime.datetime(2018, 8, 15, 16, 37, 27, 460001), 'kwargs': None, ...}
dumper = <built-in function dumps>

    def pickle_dumps(obj, dumper=pickle.dumps):
>       return dumper(obj, protocol=pickle_protocol)
E       kombu.exceptions.EncodeError: Can't pickle <class 'unittest.mock.Mock'>: it's not the same object as unittest.mock.Mock

../../../../venv/celery/lib/python3.6/site-packages/kombu/serialization.py:350: EncodeError
____________________ test_Request.test_terminate__pool_ref _____________________

wrapper = <class 'kombu.exceptions.EncodeError'>
include = (<class 'Exception'>,)
exclude = (<class 'kombu.exceptions.SerializerNotInstalled'>,)

    @contextmanager
    def _reraise_errors(wrapper,
                        include=(Exception,), exclude=(SerializerNotInstalled,)):
        try:
>           yield

../../../../venv/celery/lib/python3.6/site-packages/kombu/serialization.py:50: 

self = <kombu.serialization.SerializerRegistry object at 0x7fbc7b2ffcc0>
data = {'args': None, 'children': [], 'date_done': datetime.datetime(2018, 8, 15, 16, 37, 27, 706803), 'kwargs': None, ...}
serializer = 'pickle'
```

It's too bad TravisCI truncated the old builds.

I'll revert this, and the one below and let TravisCI re-run so you can see.

t/unit/worker/test_request.py
johnarnold: Sure.  There was an error with the app not being set, and the default app being used. It triggered the "celery app trap" when ran with pytest.  I'm not sure why this cropped up all of a sudden though.

I'll revert this one too and let TravisCI run, and you can take a look.   Thanks!

georgepsarakis:
@johnarnold fortunately tests on Travis CI seem to be passing after reverting the changes in tests, so they were not required. If you won't address this comment https://github.com/celery/celery/pull/4490#issuecomment-413101555 in this PR, would you mind opening a separate issue and reference me there?

@thedrow @auvipy I think this is ready for merge. We can handle the documentation in a separate PR probably as this has been quite extensive.

johnarnold:
@georgepsarakis I'll open a separate issue on the docs.  I just got real busy last week and didn't get to it :)

georgepsarakis:
@johnarnold no problem, thanks again for your efforts :smiley: !

celery/backends/base.py
caleb15: @auvipy is there any way to get this data? It would be useful to have.

PR #4501, opened 2018-01-24
Document chain, chord and group with task names

xavierhardy:
## Description

Document the use of chain, chord and group with task names (rather than task objects), through the use of the celery.Celery.signature method.


xavierhardy:
`ERROR: /home/travis/build/celery/celery/celery/bin/migrate.py Imports are incorrectly sorted.`

It seems to have nothing to do with my pull request, as I haven't edited the Python code at all. I could easily fix it, but I guess it would make more sense to fix this in another pull request.

PR #4527, opened 2018-02-06
Ensure base on_success is called when creating request

### LINKED ISSUE #4526: Custom requests should be allowed to provide an implementation of on_success
## Checklist

       version affected).

## Steps to reproduce
Write custom request with a ```on_success(...)``` method.
## Expected behavior
The method would be called when the request is complete.
## Actual behavior
The method is not called.

See #3977 for more details. The fix is probably to add a call to super [here](https://github.com/celery/celery/blob/master/celery/worker/request.py#L539).

cjh1:
This allows custom requests to provide an implementation.

## Description

Ensures that a custom requests on_success method is called.

Fixed #4526

### DIFF
--- a/celery/worker/request.py
+++ b/celery/worker/request.py
@@ -507,8 +507,6 @@ def create_request_cls(base, task, pool, hostname, eventer,
     default_time_limit = task.time_limit
     default_soft_time_limit = task.soft_time_limit
     apply_async = pool.apply_async
-    acks_late = task.acks_late
-    events = eventer and eventer.enabled
 
     class Request(base):
 
@@ -534,25 +532,5 @@ def create_request_cls(base, task, pool, hostname, eventer,
             # pylint: disable=attribute-defined-outside-init
             self._apply_result = maybe(ref, result)
             return result
-
-        def on_success(self, failed__retval__runtime, **kwargs):
-            # Ensure base class gets called for custom requests.
-            super(Request, self).on_success(failed__retval__runtime, **kwargs)
-
-            failed, retval, runtime = failed__retval__runtime
-            if failed:
-                if isinstance(retval.exception, (
-                        SystemExit, KeyboardInterrupt)):
-                    raise retval.exception
-                return self.on_failure(retval, return_ok=True)
-            task_ready(self)
-
-            if acks_late:
-                self.acknowledge()
-
-            if events:
-                self.send_event(
-                    'task-succeeded', result=retval, runtime=runtime,
-                )
 
     return Request
--- a/celery/worker/request.py
+++ b/celery/worker/request.py
@@ -536,6 +536,9 @@ def create_request_cls(base, task, pool, hostname, eventer,
             return result
 
         def on_success(self, failed__retval__runtime, **kwargs):
+            # Ensure base class gets called for custom requests.
+            super(Request, self).on_success(failed__retval__runtime, **kwargs)
+
             failed, retval, runtime = failed__retval__runtime
             if failed:
                 if isinstance(retval.exception, (
--- a/celery/worker/request.py
+++ b/celery/worker/request.py
@@ -507,30 +507,56 @@ def create_request_cls(base, task, pool, hostname, eventer,
     default_time_limit = task.time_limit
     default_soft_time_limit = task.soft_time_limit
     apply_async = pool.apply_async
+    acks_late = task.acks_late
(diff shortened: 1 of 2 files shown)


cjh1:
So ```test_on_success__no_events``` is failing. Now that we are calling ```super.on_success(...)``` we always get events [sent](https://github.com/celery/celery/blob/master/celery/worker/request.py#L332). Looking at the code its not clear to me why we have and on_success implementation in the ```Request``` class [here](https://github.com/celery/celery/blob/master/celery/worker/request.py#L320-L332) that then get overridden [here](https://github.com/celery/celery/blob/master/celery/worker/request.py#L538-L553). They appear to be the same appear from the condition event sending. I suggest adding a kwarg to the implementation in ```Requests``` to control when events get sent? I will push something but would appreciate some review from someone closer to the code.

celery/worker/request.py
```diff
@@ -535,21 +533,4 @@ def execute_using_pool(self, pool, **kwargs):
             self._apply_result = maybe(ref, result)
             return result
 
-        def on_success(self, failed__retval__runtime, **kwargs):
```
auvipy: doesn't removing this effect any public API?

cjh1: No I don't think so, my understanding is the super class will be the [Request](https://github.com/celery/celery/blob/master/celery/worker/request.py#L68) class that already has this method defined so its not being removed.

cjh1: I guess if the user provides a custom request that doesn't derive from [Request](https://github.com/celery/celery/blob/master/celery/worker/request.py#L68) then the API is changed, not sure the value of this use case.

thedrow:
Have you tested this with your own prod/staging environment? Just trying to make sure we won't cause any regressions just before a release.

cjh1:
@thedrow I have a toy test case, nothing in prod. I understand, I certainly don't want to cause a regression. We don't currently have a need for this, so may be safest not to include in the release.  I don't really understand the original implementation of [create_request_cls](https://github.com/celery/celery/blob/master/celery/worker/request.py#L504), so may be I am missing a use case, the original implementation was by @ask. When is the release planned?

cjh1:
I could also come up with a more conservative patch, that would only effect those that are using custom request classes, this might be the best approach, even if its not as clean, I will push a WIP commit to see what people think.

cjh1:
The latest commit limits any change to the situation where a custom request class has been provided, thoughts?

auvipy:
lets defer it for next release

cjh1:
Can someone with the power restart the Travis build, its failed downloading a package ...

georgepsarakis:
@cjh1 I restarted Travis CI, but it seems there is a flake8 error.

cjh1:
@georgepsarakis Thanks, I will fix up the flake8 issue.

georgepsarakis:
@cjh1 from my understanding and as far as I have tested there is no way for the user-defined custom class to have access to the variables passed to `create_request_cls`. For example `on_success` uses [acks_late](https://github.com/celery/celery/blob/master/celery/worker/request.py#L547) from the external scope.

I am afraid that this is a complicated case, with little gain for users and I would be reluctant to merge this. @thedrow could you please share your opinion too?

cjh1:
@georgepsarakis Correct but if you look at the implementation on [on_success in the Request class](https://github.com/celery/celery/blob/master/celery/worker/request.py#L320) it has access to the acks_late from the task, so a custom request implementation could access it from there. I am closing this for now as I don't have time to continue pushing this through.

georgepsarakis:
@cjh1 thanks a lot for your contribution so far and the suggestions! I will try and have another look to see if this can be implemented, and check the documentation as well.

cjh1:
@georgepsarakis No problem, thanks for looking at this!

mvaled:
Hi @georgepsarakis and @cjh1 

I do have a production/stage setting where I use the custom request.  Though I don't use the on_success customization.

I could deploy this PR in our stage server and see if all our tasks run.  That would give some degree of confidence about this PR.  I wouldn't have closed it, but blocked it.

auvipy:
this shouldn't be included in 4.2 release

thedrow:
@auvipy I agree.
@mvaled I think this enhancement requires a more throughout refactor unfortunately. Feel free to try and let us know though.

PR #4540, opened 2018-02-15
Prevent ResultSet from invoking backend.remove_pending_result (fixes #4539)

### LINKED ISSUE #4539: ResultSet incorrectly calls backend.remove_pending_result
The master branch currently defines `ResultSet` with:

    class ResultSet(...):
        ...
        def _on_ready(self):
            self.backend.remove_pending_result(self)
            if self.backend.is_async:
                self.on_ready()

However a typical implemenation of `backend.remove_pending_result` in `celery.backends.async` will assume that the value has an `id`

        def remove_pending_result(self, result):
            self._remove_pending_result(result.id)
            self.on_result_fulfilled(result)
            return result

Presumably, this call to `self.backend.remove_pending_result(self)` is really concerned with deregistering groups from the backend, in which case it should appear in the `GroupResult._on_ready`

This problem was exposed by #4131.  Prior to that, `_on_ready` was not called (which as a bug).

donkopotamus:
The master branch currently defines `ResultSet` with:

    class ResultSet(...):
        ...
        def _on_ready(self):
            self.backend.remove_pending_result(self)
            if self.backend.is_async:
                self.on_ready()

However a typical implementation of `backend.remove_pending_result` in `celery.backends.async` will assume that the value has an `id`

        def remove_pending_result(self, result):
            self._remove_pending_result(result.id)
            self.on_result_fulfilled(result)
            return result

Joining over a `ResultSet` will then break:

      File "...\site-packages\celery\backends\async.py", line 175, in remove_pending_result
        self._remove_pending_result(result.id)
    AttributeError: 'ResultSet' object has no attribute 'id'

Presumably, this call to `self.backend.remove_pending_result(self)` is really concerned with deregistering groups from the backend, in which case it should appear in the `GroupResult._on_ready`

This problem was exposed by #4131.  Prior to that, `_on_ready` was not called (which as a bug).  Fixes #4539


donkopotamus:
@auvipy Where should I include this test case?  To demonstrate that master is currently broken requires running an async backend and performing a `ResultSet(some_result).get()`

thedrow:
Preferably include an integration test demonstrating that this issue is resolved.

auvipy:
thanks for the integration test, but doesn't this need unit test too?

donkopotamus:
@auvipy @thedrow What sort of unit test would be appropriate, and what would it test over and above the added integration test?    This change is just patching a simple bug ... a `ResultSet` is **not** managed by a backend and so calling `remove_pending_result` is not appropriate.  (In contrast a `GroupResult` is managed by a backend and can be saved/deleted).

This problem was not previously exposed in 4.1 due to https://github.com/celery/celery/pull/4131 (internally I have been running a patched 4.1 with several fixes to `ResultSet` that has been running millions of tasks a day ... so its reasonably well tested)

Issue #4041, opened 2017-05-20
Celery Beat Periodic Task Duplicated Incessantly


Jarch09:
Celery 4.0.0
Python 3.5.2
(rest of report output below)

I recently pushed celery beat to production (I had been using celery without beat with no issues for several months). It had 1 task to run several times daily, but it duplicated the task many times per second, gradually overwhelming the capacity of my small elasticache instance and resulting in **OOM error**.

I was searching for similar errors online and found the following:
https://github.com/celery/celery/issues/943#issuecomment-8112993

While not exactly the same, changing timezone back to default settings seemed to solve the issue  for now. To be clear, previously I had:
enable_utc = False
timezone = 'America/New_York'

I changed these to:
enable_utc = True

and it seemed to solve the problem (for now).

** celery -A [proj] report:

software -> celery:4.0.0 (latentcall) kombu:4.0.0 py:3.5.2
            billiard:3.5.0.2 redis:2.10.5
platform -> system:Linux arch:64bit, ELF imp:CPython
loader   -> celery.loaders.app.AppLoader
settings -> transport:redis results:redis://[elasticache]

result_persistent: False
enable_utc: False
result_serializer: 'json'
include: [tasks -- redacted]
result_backend: 'redis://[elasticache]
task_create_missing_queues: True
task_acks_late: False
timezone: 'America/New_York'
broker_url: '[elasticache]'
task_serializer: 'json'
crontab: <class 'celery.schedules.crontab'>
broker_transport_options: {
 'visibility_timeout': 600}
task_time_limit: 200
task_always_eager: False
task_queues: 
    (<unbound Queue default -> <unbound Exchange default(direct)> -> default>,
 <unbound Queue render -> <unbound Exchange media(direct)> -> media.render>,
 <unbound Queue emails -> <unbound Exchange media(direct)> -> media.emails>,
 <unbound Queue other -> <unbound Exchange media(direct)> -> media.other>)
beat_schedule: {
    'lookup-emails-to-send': {   'args': [],
                                 'schedule': <crontab: 0 9,12,15,18,21 * * * (m/h/d/dM/MY)>,
                                 'task': 'send_emails'}}
accept_content: ['json']
task_default_exchange_type: 'direct'
task_routes: 
    ('worker.routes.AppRouter',)
worker_prefetch_multiplier: 1
BROKER_URL: [elasticache]

** Worker error message after OOM:
Traceback (most recent call last):
[...]

PR #4572, opened 2018-03-06
fixes shadow_name issue when using old Task

pachewise:
## Description
Attempts to fix the issue with #4041 on `master`, which seems to happen when using old-style Task/PeriodicTask classes (which are used as Task classes and not instances).

### DIFF
--- a/celery/app/task.py
+++ b/celery/app/task.py
@@ -530,7 +530,7 @@ class Task(object):
             args = args if isinstance(args, tuple) else tuple(args or ())
             args = (self.__self__,) + args
 
-        if self.__v2_compat__:  # TODO this or just not support shadow?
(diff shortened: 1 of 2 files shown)


pachewise:
Tried to write a unit test for this, but I'm not sure how to trigger the issue without using the default_app:

````python
import celery

@celery.task
def foo()
    return 5

f = foo.apply_async()
````

pachewise:
Renamed test to reflect that it's old, not deprecated - if that's the case though, we should change the docstring for `celery.task.base.Task` (which says it _is_ deprecated).

thedrow:
Thank you very much for your contribution! Feel free to add yourself to the AUTHORS file.

PR #4578, opened 2018-03-07
redis_run_task_one_time_only.py

tagashy:
a simple decorator to block the duplication of task with redis broker

## Description

pull request as asked on https://github.com/celery/celery/issues/3270 
the goal is to have a decorator that will check if a task run more than one time and block the second execution ( mostly a hack to patch the duplicate bug seen on the issue 3270)

### DIFF
--- a/celery/utils/redis_run_task_one_time_only.py
+++ b/celery/utils/redis_run_task_one_time_only.py
@@ -1,0 +1,23 @@
+from redis import Redis
+from functools import wraps
+from logging import error
+
+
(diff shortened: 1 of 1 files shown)


thedrow:
Thank you but we'd rather not include this kind of backend specific code. If we implement a feature we make it available to as many backends as possible.

As a side note, the code quality of this PR is not as good as it should be to be accepted.
The code opens a new Redis connection pool for each operation instead of reusing the backend's already existing connection pool.

PR #4583, opened 2018-03-09
Add unit test for celery.contrib.sphinx

lpsinger:
Note that this unit test reveals a bug: tasks are documented automatically (e.g. without explicitly adding a `.. autotask::` line) only if the `undoc-members` option is set. Try removing '`undoc-members'` from `autodoc_default_flags` in conf.py, and the test will fail!

I'll open a separate issue about the `undoc-members` bug.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -7,7 +7,7 @@ include TODO
 include setup.cfg
 include setup.py
 
-recursive-include t *.py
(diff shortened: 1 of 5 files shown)


lpsinger:
No longer needed because commit c8a326e was merged already as a part of #4584 (along with a second unintended commit).

PR #4584, opened 2018-03-09
WIP: Sphinx autodoc picks up tasks automatically only if `undoc-members` is set

lpsinger:
Sphinx autodoc picks up tasks automatically (e.g. without explicitly adding `.. autotask:: name.of.task`) only if the `undoc-members` option is set. This is demonstrated by removing the `undoc-members` flag from the default autodoc flags in the unit test from #4583.

The reason is that the task class and task instance have exactly the same docstring, which trips [this conditional in sphinx](https://github.com/sphinx-doc/sphinx/blob/v1.7.1/sphinx/ext/autodoc/__init__.py#L606-L613), which is intended to suppress repetition of the docstring of a class when documenting an instance of that class.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -7,7 +7,7 @@ include TODO
 include setup.cfg
 include setup.py
 
-recursive-include t *.py
+recursive-include t *.py *.rst
 recursive-include docs *
 recursive-include extra/bash-completion *
 recursive-include extra/centos *
--- a/t/unit/contrib/proj/conf.py
+++ b/t/unit/contrib/proj/conf.py
@@ -1,0 +1,10 @@
+from __future__ import absolute_import, unicode_literals
+
+import os
+import sys
+
+extensions = ['celery.contrib.sphinx', 'sphinx.ext.autodoc']
+autodoc_default_flags = ['members', 'undoc-members']
+autosummary_generate = True
+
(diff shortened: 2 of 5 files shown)


auvipy:
oops! it shows work in progress! @lpsinger can you verify this won't regress? if so can you add the fix?

lpsinger:
I'm sorry, I think I tried to do this in a confusing way. Commit c8a326e introduces a unit test for `celery.contrib.sphinx`, which passes. Commit 1029cfc removes the `undoc-members`, which was supposed to make the test fail, in order to illustrate the bug. But since it didn't fail, that must mean that the test was skipped because sphinx is not present. I'll open a second pull request to revert 1029cfc (though c8a326e should stay). Does that make sense?

lpsinger:
Yup, I see from travis that `test_sphinx` was skipped. That's why this passed.

lpsinger:
It would be good to install sphinx in one of the CI configurations so that the new test is actually run. I'm not familiar enough with tox or the layout of the requirements directory to be able to do this easily myself. Do you know how to do it?

auvipy:
@thedrow

lpsinger:
@auvipy, should I open a new issue for the bug?

auvipy:
sure

lpsinger:
@auvipy, I managed to fix the bug too: #4588.

PR #4611, opened 2018-03-21
Flush stderr before exiting with error code 1.

### LINKED ISSUE #4594: Please use `sys.exit()` or flush STDERR if failed on `platform.check_privileges`
https://github.com/celery/celery/blob/a3c377474ab1109a26de5169066a4fae0d30524b/celery/platforms.py#L794:24

The way it is now, the message could be lost because `os._exit()` does not flushes STDERR.

Why are we using `os._exit()` instead of the recommended `sys.exit()`? I can provide a PR if asked, but please clarify why the `os._exit()` before the patch is written.

wetneb:
Closes #4594.

## Description

As explained in #4594, in some environments (docker, kubernetes, ...), it is possible that Celery terminates with error code 1 without printing anything. This happens when it is run as root with default parameters.

I have searched for all calls to `os._exit(1)` in the code and manually flushed stderr just before them. (Only the one in `celery/platforms.py` is necessary to solve the issue, but I thought that could save sweat to others…)

### DIFF
--- a/celery/platforms.py
+++ b/celery/platforms.py
@@ -791,6 +791,7 @@ def check_privileges(accept_content):
                         uid=uid, euid=euid, gid=gid, egid=egid,
                     ), file=sys.stderr)
                 finally:
+                    sys.stderr.flush()
                     os._exit(1)
         warnings.warn(RuntimeWarning(ROOT_DISCOURAGED.format(
             uid=uid, euid=euid, gid=gid, egid=egid,
--- a/celery/utils/threads.py
+++ b/celery/utils/threads.py
@@ -74,6 +74,7 @@ class bgThread(threading.Thread):
                         self.on_crash('{0!r} crashed: {1!r}', self.name, exc)
                         self._set_stopped()
                     finally:
+                        sys.stderr.flush()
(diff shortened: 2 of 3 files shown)


wetneb:
I'm not too sure about the failing test, triggered in only one build - if someone knowledgeable could have a look (and maybe restart the build if it is a spurious failure) it would be great.

georgepsarakis:
I think this would not be necessary if you used [PYTHONUNBUFFERED](https://docs.python.org/2/using/cmdline.html#envvar-PYTHONUNBUFFERED), or am I wrong?

wetneb:
@georgepsarakis probably - but as a user I would prefer not to be presented with error messages no matter which buffering settings I use. It's not like forcing a flush is expensive in these three cases: it's just before termination anyway. What is the downside of adding these flushes? On paper, it might decrease your test coverage as I assume these hard failures are probably not covered by your tests… but I think that it is pretty clear that these additions are harmless, no? Anyway, if decreasing test coverage is the concern, we can keep only the flush that solves the issue, maybe? (Although that would be a bit sad too)

PR #4617, opened 2018-03-22
Allow join in eager canvas application

### LINKED ISSUE #4576: Eager Application Synchronous Subtask Guard Blocks Canvas
## Checklist

       version affected).

## Steps to reproduce

```
celery -A minimal_eager_chain report

software -> celery:4.2.0 (latentcall) kombu:4.1.0 py:2.7.13

```

Minimal script to reproduce: https://gist.github.com/npilon/9f3c8469a615081fc8454359945eebd7

## Expected behavior

- No error occurs

## Actual behavior

- We trip the synchronous subtask alarm:

```
>>> test.delay().get()
  File "<console>", line 1, in <module>
  File "/home/npilon/Documents/celery/celery/result.py", line 949, in get
    raise self.result
RuntimeError: Never call result.get() within a task!
```

npilon:
## Description

Permit async applied chains to join their results during eager application testing.

Fixes #4576

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -287,6 +287,10 @@ an overview of what's new in Celery 4.2.
 
     Contributed by **Vinod Chandru**.
 
+- **Development/Testing**: Allow eager application of canvas structures (#4576)
+
+    Contributed by **Nicholas Pilon**.
+
 Documentation, CI, Installation and Tests fixes:
 
 
--- a/celery/canvas.py
+++ b/celery/canvas.py
@@ -24,7 +24,7 @@ from vine import barrier
 from celery._state import current_app
 from celery.five import python_2_unicode_compatible
 from celery.local import try_import
-from celery.result import GroupResult
+from celery.result import GroupResult, allow_join_result
 from celery.utils import abstract
 from celery.utils.functional import _regen
 from celery.utils.functional import chunks as _chunks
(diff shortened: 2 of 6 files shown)


thedrow:
We'd prefer to have both unit and integration tests for this PR.
If you can provide them, we'll gladly include it in our next release.

npilon:
I'd love to provide unit tests, but the unit testing framework patches out the functionality that causes the bug I'm trying to fix: https://github.com/celery/celery/blob/master/t/unit/conftest.py#L139-L154

I'll see if I can provide integration tests.

npilon:
Ah, no, I see that fixture keeps an `_orig` for un-doing itself. I've added tests.

npilon:
What category should I use for the changelog entry?

auvipy:
you can check here to have some hint https://github.com/celery/celery/commit/8047ca9e818db8151dc6c80814600797ab94f5bc

npilon:
@auvipy I think I've adjusted the changelog files appropriately.

thedrow:
Thank you very much :)

courcelm:
Is this fix in 4.2rc2?

I have a test that run with CELERY_TASK_ALWAYS_EAGER = True
which call chord(header)(callback).

I get the following error:
RuntimeError: Never call result.get() within a task!

Without CELERY_TASK_ALWAYS_EAGER = True, this code runs fine.

thedrow:
@courcelm The fix is in 4.2rc2. If you believe something is wrong, please file an issue with a detailed analysis of what you think is wrong and possibly how to fix this.

jcushman:
As reported by @courcelm, this patch fixed the issue for `chain()` but not for `chord()`. I added a comment with a script to reproduce the problem on the original issue. I figured it might make sense to reopen the original issue instead of file a new one, since the issue is "Eager Application Synchronous Subtask Guard Blocks Canvas" and that covers the remaining problems. Let me know if you'd like a new issue filed instead.

PR #4685, opened 2018-04-27
Add Azure Block Blob Storage backend

c-w:
## Description

This pull request adds a new results backend. The backend is implemented on top of the [azure-storage library](https://github.com/Azure/azure-storage-python) which uses [Azure Blob Storage](https://azure.microsoft.com/en-us/services/storage/) for a scalable low-cost PaaS backend.

The backend was load tested via a simple nginx/gunicorn/sanic app hosted on a DS4 virtual machine (4 vCores, 16 GB RAM) and was able to handle 600+ concurrent users at ~170 RPS.

The commit also contains a live end-to-end test to facilitate verification of the backend functionality. The test is activated by setting the `AZUREBLOCKBLOB_URL` environment variable to `azureblockblob://{ConnectionString}` where the value for `ConnectionString` can be found in the `Access Keys` pane of a Storage Account resources in the Azure Portal:

Screenshot showing location of ConnectionString in Azure Portal

This pull request was created together with @ankurokok, @dkisselev, @evandropaula, @martinpeck and @michaelperel.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -69,11 +69,11 @@ before_install:
           fi
     - |
           if [[ "$TOXENV" == *dynamodb ]]; then
-              docker run -d -p 8000:8000 dwmkerr/dynamodb:38
+              docker run -d -p 8000:8000 dwmkerr/dynamodb:38 -inMemory
               while ! nc -zv 127.0.0.1 8000; do sleep 10; done
           fi
     - |
-          docker run -d -e executable=blob -t -p 10000:10000 arafato/azurite:2.6.5
+          docker run -d -e executable=blob -t -p 10000:10000 --tmpfs /opt/azurite/folder:rw arafato/azurite:2.6.5
           while ! nc -zv 127.0.0.1 10000; do sleep 10; done
           export AZUREBLOCKBLOB_URL="azureblockblob://DefaultEndpointsProtocol=http;AccountName=devstoreaccount1;AccountKey=Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==;BlobEndpoint=http://127.0.0.1:10000/devstoreaccount1;"
     - |
--- a/.travis.yml
+++ b/.travis.yml
@@ -68,18 +68,13 @@ before_install:
           fi
     - |
           if [[ "$TOXENV" == *dynamodb ]]; then
-              sudo apt-get update && sudo apt-get install -y default-jre supervisor
-              mkdir /opt/dynamodb-local
-              cd /opt/dynamodb-local && curl --retry 5 --retry-delay 1 -L http://dynamodb-local.s3-website-us-west-2.amazonaws.com/dynamodb_local_latest.tar.gz | tar zx
-              cd -
-              echo '[program:dynamodb-local]' | sudo tee /etc/supervisor/conf.d/dynamodb-local.conf
-              echo 'command=java -Djava.library.path=./DynamoDBLocal_lib -jar DynamoDBLocal.jar -inMemory' | sudo tee -a /etc/supervisor/conf.d/dynamodb-local.conf
-              echo 'directory=/opt/dynamodb-local' | sudo tee -a /etc/supervisor/conf.d/dynamodb-local.conf
-              sudo service supervisor stop
-              sudo service supervisor start
-              sleep 10
-              curl localhost:8000
+              docker run -d -p 8000:8000 dwmkerr/dynamodb:38
+              while ! nc -zv 127.0.0.1 8000; do sleep 10; done
           fi
+    - |
+          docker run -d -e executable=blob -t -p 10000:10000 arafato/azurite:2.6.5
+          while ! nc -zv 127.0.0.1 10000; do sleep 10; done
+          export AZUREBLOCKBLOB_URL="azureblockblob://DefaultEndpointsProtocol=http;AccountName=devstoreaccount1;AccountKey=Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==;BlobEndpoint=http://127.0.0.1:10000/devstoreaccount1;"
     - |
           wget -qO - https://packages.couchbase.com/ubuntu/couchbase.key | sudo apt-key add -
           sudo apt-add-repository -y 'deb http://packages.couchbase.com/ubuntu trusty trusty/main'
@@ -101,3 +96,10 @@ notifications:
 services:
     - rabbitmq
     - redis
+    - docker
+addons:
+  apt:
+    sources:
+      - debian-sid
+    packages:
+      - docker-ce
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,7 @@ env:
   - MATRIX_TOXENV=integration-rabbitmq
   - MATRIX_TOXENV=integration-redis
   - MATRIX_TOXENV=integration-dynamodb
+  - MATRIX_TOXENV=integration-azureblockblob
 matrix:
   include:
   - python: '3.5'
--- a/celery/backends/azureblockblob.py
+++ b/celery/backends/azureblockblob.py
@@ -1,0 +1,152 @@
+"""The Azure Storage Block Blob backend for Celery."""
+from __future__ import absolute_import, unicode_literals
+
+import re
+
+from kombu.utils import cached_property
+from kombu.utils.encoding import bytes_to_str
+
+from celery.exceptions import ImproperlyConfigured
+from celery.utils.log import get_logger
+
+from .base import KeyValueStoreBackend
+
+try:
+    import azure.storage as azurestorage
+    from azure.common import AzureMissingResourceHttpError
+    from azure.storage.blob import BlockBlobService
+    from azure.storage.common.retry import ExponentialRetry
+except ImportError:  # pragma: no cover
+    azurestorage = BlockBlobService = ExponentialRetry = \
+        AzureMissingResourceHttpError = None  # noqa
+
+__all__ = ("AzureBlockBlobBackend",)
+
+URL_RE = re.compile(r"azureblockblob://(?P<connection_string>.+)")
+
+LOGGER = get_logger(__name__)
+
+
+class AzureBlockBlobBackend(KeyValueStoreBackend):
+    """Azure Storage Block Blob backend for Celery."""
+
+    def __init__(self,
+                 url=None,
+                 container_name=None,
+                 retry_initial_backoff_sec=None,
+                 retry_increment_base=None,
+                 retry_max_attempts=None,
+                 *args,
+                 **kwargs):
+        super(AzureBlockBlobBackend, self).__init__(*args, **kwargs)
+
+        if azurestorage is None:
+            raise ImproperlyConfigured(
+                "You need to install the azure-storage library to use the "
+                "AzureBlockBlob backend")
+
+        conf = self.app.conf
+
+        self._connection_string = self._parse_url(url)
+
+        self._container_name = (
+            container_name or
+            conf["azureblockblob_container_name"])
+
+        self._retry_initial_backoff_sec = (
+            retry_initial_backoff_sec or
+            conf["azureblockblob_retry_initial_backoff_sec"])
+
+        self._retry_increment_base = (
+            retry_increment_base or
+            conf["azureblockblob_retry_increment_base"])
+
+        self._retry_max_attempts = (
+            retry_max_attempts or
+            conf["azureblockblob_retry_max_attempts"])
+
+    @classmethod
+    def _parse_url(cls, url):
+        match = URL_RE.match(url)
+        if not match:
+            raise ImproperlyConfigured("Invalid URL")
+
+        return match.group("connection_string")
+
+    @cached_property
+    def _client(self):
+        """Return the Azure Storage Block Blob service.
+
+        If this is the first call to the property, the client is created and
+        the container is created if it doesn't yet exist.
+
+        """
+        client = BlockBlobService(connection_string=self._connection_string)
+
+        created = client.create_container(
+            container_name=self._container_name, fail_on_exist=False)
+
+        if created:
+            LOGGER.info("Created Azure Blob Storage container %s",
+                        self._container_name)
+
+        client.retry = ExponentialRetry(
+            initial_backoff=self._retry_initial_backoff_sec,
+            increment_base=self._retry_increment_base,
+            max_attempts=self._retry_max_attempts).retry
+
+        return client
+
+    def get(self, key):
+        """Read the value stored at the given key.
+
+        Args:
+              key: The key for which to read the value.
+
+        """
+        key = bytes_to_str(key)
+        LOGGER.debug("Getting Azure Block Blob %s/%s",
+                     self._container_name, key)
+
+        try:
+            return self._client.get_blob_to_text(
+                self._container_name, key).content
+        except AzureMissingResourceHttpError:
+            return None
+
+    def set(self, key, value):
+        """Store a value for a given key.
+
+        Args:
+              key: The key at which to store the value.
+              value: The value to store.
+
+        """
+        key = bytes_to_str(key)
+        LOGGER.debug("Creating Azure Block Blob at %s/%s",
+                     self._container_name, key)
+
+        return self._client.create_blob_from_text(
+            self._container_name, key, value)
+
+    def mget(self, keys):
+        """Read all the values for the provided keys.
+
+        Args:
+              keys: The list of keys to read.
+
+        """
+        return [self.get(key) for key in keys]
+
+    def delete(self, key):
+        """Delete the value at a given key.
+
+        Args:
+              key: The key of the value to delete.
+
+        """
+        key = bytes_to_str(key)
+        LOGGER.debug("Deleting Azure Block Blob at %s/%s",
+                     self._container_name, key)
+
+        self._client.delete_blob(self._container_name, key)
--- a/celery/backends/azureblockblob.py
+++ b/celery/backends/azureblockblob.py
@@ -1,8 +1,6 @@
 """The Azure Storage Block Blob backend for Celery."""
 from __future__ import absolute_import, unicode_literals
 
-import re
-
 from kombu.utils import cached_property
 from kombu.utils.encoding import bytes_to_str
 
@@ -21,8 +19,6 @@ except ImportError:  # pragma: no cover
         AzureMissingResourceHttpError = None  # noqa
 
 __all__ = ("AzureBlockBlobBackend",)
-
-URL_RE = re.compile(r"azureblockblob://(?P<connection_string>.+)")
 
 LOGGER = get_logger(__name__)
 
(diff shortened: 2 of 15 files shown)


georgepsarakis:
@c-w thank you very much for your contribution! Is there any chance that we can run integration/functional test with this backend, without actually connecting to Azure?

c-w:
Hi @georgepsarakis. There is [Azurite](https://github.com/azure/azurite) for local emulation of Azure Storage which should be able to emulate all the APIs used in this pull request. If you wish, I can look into setting up Azurite for integration tests.

How often do you run the full integration test suite? Are you fundamentally opposed to making live calls against a backend or is it just an issue of setting up test credentials and the associated cost?

georgepsarakis:
@c-w Azurite seems as a great choice for running the tests.

@thedrow @auvipy what are your thoughts on this? I believe this could be merged (probably in a later version than 4.2), once we have a closer look. I am just asking if you are agreed, in order to continue looking into the integration testing option. Thanks.

auvipy:
as its a bit big feature we should skip 4.2

c-w:
That's great to hear. I'll get onto the integration tests via Azurite and push them onto this pull request.

We were also planning to make a few follow-up pull requests after this one:
- A [CosmosDB](https://azure.microsoft.com/en-us/services/cosmos-db/) storage backend.
- A [Storage Queue](https://azure.microsoft.com/en-us/services/storage/queues/) transport channel.
- A [ServiceBus](https://azure.microsoft.com/en-us/services/service-bus/) transport channel.

The overall idea is to use the Storage Queue transport channel with the Block Blob storage backend for simple workloads since the services are very cheap and combined only require setting up a single Azure resource. The combination of ServiceBus transport channel and CosmosDB storage backend on the other hand would enable more demanding workloads due to offering stronger latency guarantees, multi-zone failover, etc.

We already have some branches with work-in-progress code for each of these features, but it'll take a bit more time to clean them up and submit pull requests. If you have any thoughts about the above, just let me know.

c-w:
@georgepsarakis I've implemented the end-to-end test for the backend in 3ea28b6 and 1b9a263.

Travis CI now starts Azurite via Docker:

Screenshot showing Azurite starting via Docker on Travis

The required environment variables to light up the end-to-end test are then also set and the test executes:

Screenshot showing end-to-end test for Azure Block Blob backend running on Travis

Changing between Azurite and live Azure with this mechanism is also easy (just over-write the `AZUREBLOCKBLOB_URL` environment variable with an encrypted Travis environment variable).

auvipy:
great work ! @c-w

georgepsarakis:
Excellent work @c-w . I will have a closer look to the code and let you know if there are any suggestions.

celery/backends/azureblockblob.py
```diff
@@ -0,0 +1,160 @@
...
+                "You need to install the azure-storage library to use the "
+                "AzureBlockBlob backend")
+
+        conf = self.app.conf
+
+        self._connection_string = self._parse_url(url)
+
+        self._container_name = (
+            container_name or
+            conf.get(CONF_CONTAINER_NAME) or
```
georgepsarakis: It's slightly simpler to just use the default value of the `get` method here:

```
conf.get(CONF_CONTAINER_NAME, DEFAULT_CONTAINER_NAME)
```

celery/backends/azureblockblob.py
```diff
@@ -0,0 +1,160 @@
...
+            DEFAULT_CONTAINER_NAME)
+
+        self._retry_initial_backoff_sec = (
+            retry_initial_backoff_sec or
+            conf.get(CONF_RETRY_INITIAL_BACKOFF_SEC) or
```
georgepsarakis: Same comment as above for all configuration options.

celery/backends/azureblockblob.py
```diff
@@ -0,0 +1,160 @@
...
+        the container is created if it doesn't yet exist.
+
+        """
+        client = BlockBlobService(connection_string=self._connection_string)
+
+        created = client.create_container(
+            container_name=self._container_name, fail_on_exist=False)
+
+        if created:
+            LOGGER.info("Created container %s", self._container_name)
```
georgepsarakis: You can also specify that this action refers to Azure Block Blob storage in the log message.

celery/backends/azureblockblob.py
```diff
@@ -0,0 +1,160 @@
...
+
+        client.retry = ExponentialRetry(
```
georgepsarakis: I was expecting this class to be using constructor-based dependency injection, but [azure-storage sample](https://github.com/Azure/azure-storage-python/blob/19d845d58551bd36a04c46628dc42089ee269362/samples/advanced/client.py#L42) also uses the same pattern.

requirements/extras/azureblockblob.txt
```diff
@@ -0,0 +1,3 @@
+azure-storage>=0.36.0
```
georgepsarakis: It is not advisable to use non-exact tags for package versions, as they might lead to inconsistencies among installations and CI results. Could you please use the most recent usable version?

t/unit/backends/test_azureblockblob.py
```diff
@@ -0,0 +1,125 @@
...
+
+from celery.backends import azureblockblob
+from celery.backends.azureblockblob import AzureBlockBlobBackend
+from celery.exceptions import ImproperlyConfigured
+
+MODULE_TO_MOCK = "celery.backends.azureblockblob"
+
+
+@contextlib.contextmanager
+def patch_client():
```
georgepsarakis: I am not sure, but I don't think you need this context manager. `patch` will work as you require, see this [example](https://github.com/celery/celery/pull/4666/files#diff-5cb790f5c0c0aea16658640ccd102c19R149) of mocking an instance method.

t/unit/backends/test_azureblockblob.py
```diff
@@ -0,0 +1,125 @@
...
+
+        # ...but only once per backend instance
+        assert backend._client is not None
+        assert mock_blob_service_instance.create_container.call_count == 1
+
+    def test_get(self):
+        with patch_client() as mock_client:
+            self.backend.get(b"mykey")
+
+            mock_client.get_blob_to_text.assert_has_calls(
```
georgepsarakis: I think that perhaps `assert_called_once_with` is more suitable here.

t/unit/backends/test_azureblockblob.py
```diff
@@ -0,0 +1,125 @@
...
+                 call("celery", "mykey2")])
+
+    def test_delete(self):
+        with patch_client() as mock_client:
+            self.backend.delete(b"mykey")
+
+            mock_client.delete_blob.assert_called_once_with(
+                "celery", "mykey")
+
+    @skip.unless_environ("AZUREBLOCKBLOB_URL")
```
georgepsarakis: If I understand correctly, this test case probably belongs to the integration test suite.

tox.ini
```diff
@@ -47,8 +47,9 @@ setenv =
     dynamodb: TEST_BACKEND=dynamodb://@localhost:8000
     dynamodb: AWS_ACCESS_KEY_ID=test_aws_key_id
     dynamodb: AWS_SECRET_ACCESS_KEY=test_aws_secret_key
-PASSENV =
```
georgepsarakis: The conversion to lowercase is an improvement of course, but as a general note, I believe it is preferable to avoid making changes outside the scope of the PR (however small), especially in extensive feature requests, as they might confuse reviewers and anyone inspecting the commit history in the future. Again, just explaining my view, no need to revert this :smile: !

c-w: Totally agree on keeping diffs minimal, I'm a stickler for that in all of my code reviews, so we're definitely aligned there ;-)

However, I do believe that this is a required change. I was investigating for a while why the live test would still be skipped despite the `AZUREBLOCKBLOB_URL` variable being set, and it ended up being linked to the environment isolation that tox uses. Simply adding the `AZUREBLOCKBLOB_URL` to the `passenv` section though didn't fix it since the passenv property apparently needs to be lowercased (which is surprising given that configparser is supposed to be case insensitive).

.travis.yml
```diff
@@ -68,18 +68,13 @@ before_install:
...
-              sudo service supervisor start
-              sleep 10
-              curl localhost:8000
+              docker run -d -p 8000:8000 dwmkerr/dynamodb:38
+              while ! nc -zv 127.0.0.1 8000; do sleep 10; done
           fi
+    - |
+          docker run -d -e executable=blob -t -p 10000:10000 arafato/azurite:2.6.5
+          while ! nc -zv 127.0.0.1 10000; do sleep 10; done
+          export AZUREBLOCKBLOB_URL="azureblockblob://DefaultEndpointsProtocol=http;AccountName=devstoreaccount1;AccountKey=Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==;BlobEndpoint=http://127.0.0.1:10000/devstoreaccount1;"
```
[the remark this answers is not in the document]
georgepsarakis: I see, great, thanks for clarifying.

celery/backends/azureblockblob.py
c-w: That was my initial implementation but it failed the tests since the conf value were explicitly set to None in one of the mappings that the conf access checks. Moving the default values to `app/defaults.py` fixed this. Implemented in 44517e3.

celery/backends/azureblockblob.py
c-w: Implemented in 44517e3.

tox.ini
georgepsarakis: Wasn't aware of that, thanks for fixing this then :smile: !

celery/backends/azureblockblob.py
c-w: Done in f91b69b.

t/unit/backends/test_azureblockblob.py
c-w: Great catch! Done in 16ce838.

t/unit/backends/test_azureblockblob.py
c-w: Done in 16ce838.

requirements/extras/azureblockblob.txt
c-w: Could you provide some more guidance on what you require here? The azure-storage SDK has a bit of an odd versioning scheme where all versions below 0.36.0 are considered deprecated and the "new" SDK is versions 0.36.0 and above (see [compatibility section on the official repository](https://github.com/Azure/azure-storage-python#compatibility)). As such, this dependency statement basically means "use any of the new SDK versions". Would you like to pin this to a specific version instead?

t/unit/backends/test_azureblockblob.py
c-w: I'm having some issues running the integration tests locally via docker-compose so I've made a change and pushed it up to Travis in 9ba725d to see if the move of the test was successful. I'll tag you on the PR when everything is ready to be re-reviewed.

requirements/extras/azureblockblob.txt
georgepsarakis: Exactly, my suggestion was to pin the version for the new dependencies to the exact latest:

```
azure-storage==0.36.0
azure-common==...
azure-storage-common==...
```

t/unit/backends/test_azureblockblob.py
c-w: Alright, confirmed that integration tests ran and the new backend test executed and passed. Sample CI job log: https://travis-ci.org/celery/celery/jobs/372769687

requirements/extras/azureblockblob.txt
c-w: Alright, done in ce12efd.

docs/userguide/configuration.rst
```diff
@@ -598,6 +602,7 @@ Can be one of the following:
 .. _`CouchDB`: http://www.couchdb.com/
 .. _`Couchbase`: https://www.couchbase.com/
 .. _`Consul`: https://consul.io/
+.. _`AzureBlockBlob`: https://azure.microsoft.com/en-us/services/storage/
```
georgepsarakis: Correct me if I am wrong, but I think I found a more accurate link [here](https://azure.microsoft.com/en-us/services/storage/blobs/).

t/integration/test_backend.py
```diff
@@ -0,0 +1,31 @@
+from __future__ import absolute_import, unicode_literals
+
+import os
+
+from case import skip
+
+from celery.backends.azureblockblob import AzureBlockBlobBackend
+
+
```
georgepsarakis: Although this test has certainly value, it is better if you also add and activate the integration suite by changing the [tox](https://github.com/celery/celery/blob/master/tox.ini#L43) configuration. You simply need to specify `TEST_BROKER` and `TEST_BACKEND` variables accordingly.

docs/userguide/configuration.rst
c-w: Done in e6284b0.

t/integration/test_backend.py
c-w: Done in f17c3b6. Will verify on Travis and ping you once it's ready to be reviewed.

c-w:
This should now be ready for the next round of reviews, @georgepsarakis.

georgepsarakis:
@c-w thank you very much for the effort and changes so far. Integration tests seem to be running. 
However builds seem a bit slow. Would you like to give a try to Docker container `tmpfs` option?

```
docker run -d -e executable=blob --tmpfs /opt --tmpfs /tmp --tmpfs /var -t -p 10000:10000 arafato/azurite:2.6.5
```

DynamoDB is started without persistence so I think there is not much to be done, however the same may apply:
```
docker run -d --tmpfs /tmp --tmpfs /opt -p 8000:8000 dwmkerr/dynamodb:38
```

What do you think?

c-w:
That's very interesting. I didn't know about tmpfs in Docker; seems like a neat feature. I did a bit of research and used `docker diff {container}` to find the data directories of the dynamodb and azurite containers and mapped those to memory. Let's see if that speeds up the build time from the current ~8 minutes for azureblockblob and ~4 minutes for dynamodb.

Alternatively, as I mentioned in https://github.com/celery/celery/pull/4685#discussion_r185146182, we could also consider turning off the full integration tests for azureblockblob and dynamodb and instead only rely on the new backend integration test which only takes seconds to run. The idea here would be to use the core redis and rabbitmq integration tests to verify that the framework wiring is working as expected and then to verify additional backends like azureblockblob, dynamodb, cassandra, etc. only at the backend interface layer (under the assumption that if the backend contract is correctly implemented, the rest of the framework wiring will work as expected). This should massively speed up the build at a small loss of end-to-end verification. Thoughts?

c-w:
Looks like we save about 60% execution time on the azureblockblob integration tests by using tmpfs, so that was a great suggestion!

- [Pre-tmpfs CI](https://travis-ci.org/celery/celery/builds/373303656) has ~8 minutes per azureblockblob task.
- [Post-tmpfs CI](https://travis-ci.org/celery/celery/builds/373366249) has ~3 minutes per azureblockblob task.

Overall the CI now takes about 50% longer though due to the new integration tests:

- [Pre-azureblockblob CI](https://travis-ci.org/celery/celery/builds/373328674) has ~20 minutes for the full suite.
- [Post-azureblockblob CI](https://travis-ci.org/celery/celery/builds/373366249) has ~30 minutes for the full suite.

Let me know if this is acceptable, @georgepsarakis, or if you'd prefer to try some more optimizations.

thedrow:
@c-w The .tox directory is quite large and caching it might not provide us with the benefit we're after. In any case, if we choose to do so we'll need to also install and use [tox-pip-extensions](https://github.com/tox-dev/tox-pip-extensions/) to avoid recreating the virtual environment every time requirements change.
I'd rather discuss this after we merge this PR.

At this point I'd rather extend our CI time than to loose the guarantee that everything works as expected.
We need to include every result backend we have and broker we support in the integration test suite.
Celery had quite a few bugs that were uncovered very late due to infrequent use of some of the result backends.
In Celery 5, we will split those result backends to different packages. They'll have a separate CI process which will help us reduce the build time.

tox.ini
```diff
@@ -47,8 +47,12 @@ setenv =
...
+
+    azureblockblob: TEST_BROKER=redis://
+    azureblockblob: TEST_BACKEND=azureblockblob://DefaultEndpointsProtocol=http;AccountName=devstoreaccount1;AccountKey=Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==;BlobEndpoint=http://127.0.0.1:10000/devstoreaccount1;
```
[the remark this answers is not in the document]
c-w: This is a "well known" string to connect to the storage emulator, not a real secret. See earlier discussion https://github.com/celery/celery/pull/4685#discussion_r184890810 or the [official docs](https://docs.microsoft.com/en-us/azure/storage/common/storage-use-emulator#authenticate-with-shared-key-credentials).

t/integration/test_backend.py
```diff
@@ -0,0 +1,31 @@
...
+@skip.unless_module("azure")
+@skip.unless_environ("AZUREBLOCKBLOB_URL")
+class test_AzureBlockBlobBackend:
+    def test_live(self, manager):
+        backend = AzureBlockBlobBackend(
+            app=manager.app,
+            url=os.environ["AZUREBLOCKBLOB_URL"])
+
+        key_values = dict((("akey%d" % i).encode(), "avalue%d" % i)
```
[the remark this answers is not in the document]
c-w: Done in 0167476.

t/integration/test_backend.py
```diff
@@ -0,0 +1,31 @@
...
+                          for i in range(5))
+
+        for key, value in key_values.items():
+            backend.set(key, value)
+
+        actual_values = backend.mget(key_values.keys())
+        for value, actual_value in zip(key_values.values(), actual_values):
+            assert value == actual_value
```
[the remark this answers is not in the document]
c-w: Done in 0167476.

t/integration/test_backend.py
```diff
@@ -0,0 +1,31 @@
...
+
+from case import skip
+
+from celery.backends.azureblockblob import AzureBlockBlobBackend
+
+
+@skip.unless_module("azure")
+@skip.unless_environ("AZUREBLOCKBLOB_URL")
+class test_AzureBlockBlobBackend:
+    def test_live(self, manager):
```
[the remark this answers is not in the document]
c-w: Done in 0167476. NB: the test seems to take less than 1 second on Travis currently.

capture

celery/backends/azureblockblob.py
```diff
@@ -0,0 +1,152 @@
...
+            retry_increment_base or
+            conf["azureblockblob_retry_increment_base"])
+
+        self._retry_max_attempts = (
+            retry_max_attempts or
+            conf["azureblockblob_retry_max_attempts"])
+
+    @classmethod
+    def _parse_url(cls, url):
+        match = URL_RE.match(url)
```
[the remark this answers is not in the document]
c-w: Good catch. Refactored in 4dc8f31.

auvipy:
@c-w thanks! and please don't forget to chime in when azure related issues arise!

auvipy:
should we revert this before kombu 4.3 release?

c-w:
Sure thing, @auvipy; feel free to tag me on Azure-related issues.

auvipy:
Clemens, could you investigate the reasons of build failure after merging this PR?

thedrow:
Was this reverted because of test failures?

auvipy:
yes! I was confused, and reverted back!

auvipy:
yes!! what I did is again reverted the reverted pr! let's see what the build says!

c-w:
Looking into it now, @auvipy.

auvipy:
great thanks for investigating!

c-w:
Here's an example of a Travis build with the Azure Block Blob backend change and the Travis CI fix mentioned above: https://travis-ci.org/CatalystCode/celery/builds/424006344. For some reason, the tests for the Redis backend on Python 3 are now failing which I find surprising. I'll investigate why this is the case.

c-w:
Looks like the Redis test failures are transient and linked to timeouts/timing. Submitted the new pull request: https://github.com/celery/celery/pull/5025

PR #4689, opened 2018-04-28
Clean up redundent PY3 constants

thedrow:
## Description

This PR cleans up redundant PY3 constants. The PY3 constant we should use is located at `vine.five`.
They are just clogging up our code and possibly using a little bit of extra startup time and memory.


auvipy:
great job. thanks!

PR #4692, opened 2018-04-30
Enable docker mounting of test directory

c-w:
## Description

Currently there is a comment in the docker-compose file that indicates that the test directory can't be mounted due to compiled Python version mismatches. This pull request removes this limitation. This is achieved by cleaning up pyc files in the docker entrypoint script which avoids mismatches between compiled Python files on the docker host and/or from previous build steps by construction: before executing any command in the docker container, any remnant platform specific files are cleaned up and as such any subsequent py.test command will run in a clean environment.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -70,9 +70,7 @@ COPY --chown=1000:1000 celery $HOME/celery
 
 RUN pyenv exec pip install -e .
 
-# the compiled files from earlier steps will cause py.test to fail with
-# an ImportMismatchError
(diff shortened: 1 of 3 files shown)


c-w:
@auvipy I found it somewhat surprising that the build would fail for PyPy when I only changed files related to Docker (which has no interaction with the PyPy tests at all), so I re-triggered the build and it now passes. Looks like the PyPy test might have some flakiness or maybe it was just a temporary glitch in Travis.

PR #4698, opened 2018-05-01
Add test for mongodb+srv urls and fix url parsing to handle it correctly

### LINKED ISSUE #4668: MongoDB backend does not support mongodb+srv:// URL's
## Checklist

https://github.com/celery/celery/blob/master/celery/backends/mongodb.py#L143-L146

## Steps to reproduce

Set the MongoDB URL to one like:

```mongodb+srv://mongo.private.corp.example.com/celery?ssl=false```

## Expected behavior

This works.

## Actual behavior

This fails because the URL parsing does not match on `mongodb+srv://` instead trying to match `mongodb://` only.

### PR DESCRIPTION
**Pull request by josephtate**
## Description

This fixes bug #4668 (Fixes #4668) where passing a mongodb+srv url to celery causes a traceback in parsing before creating the MongoClient.

### DIFF
--- a/celery/app/backends.py
+++ b/celery/app/backends.py
@@ -63,7 +63,12 @@ def by_url(backend=None, loader=None):
         url = backend
         scheme, _, _ = url.partition('://')
         if '+' in scheme:
-            backend, url = url.split('+', 1)
+            if scheme.startswith('mongodb+'):
+                # e.g., mongodb+srv://
+                backend = url.split('+', 1)[0]
+            else:
+                # e.g., sqla+mongodb
+                backend, url = url.split('+', 1)
         else:
             backend = scheme
     return by_name(backend, loader), url
--- a/celery/app/backends.py
+++ b/celery/app/backends.py
@@ -57,18 +57,23 @@ def by_name(backend=None, loader=None,
 
 
 def by_url(backend=None, loader=None):
-    """Get backend class by URL."""
+    """Get backend class by URL. Return a class object and the url.
+
+    New in version 4.2.?: url is returned unmodified. Previously any leading
+    schema followed by a plus sign would be stripped off. Now backends that
+    require special formatting must handle that formatting in their own
+    constructors.
+    """
     url = None
     if backend and '://' in backend:
+        # There are two types of uris we'll see:
+        # * <backend>+<backend-defined-ext>:// e.g., mongodb+srv://
+        # * <backend>+<backend-subtype>:// e.g., cache+memory:// (cache is the
+        #   only one tested and documented thus far
         url = backend
         scheme, _, _ = url.partition('://')
         if '+' in scheme:
-            if scheme.startswith('mongodb+'):
-                # e.g., mongodb+srv://
-                backend = url.split('+', 1)[0]
-            else:
-                # e.g., sqla+mongodb
-                backend, url = url.split('+', 1)
+            backend, _, _ = scheme.partition('+')
         else:
             backend = scheme
     return by_name(backend, loader), url
--- a/celery/app/backends.py
+++ b/celery/app/backends.py
@@ -62,6 +62,7 @@ def by_url(backend=None, loader=None):
     Return a class object and the url to provide to the backend when it is
     created.
     """
+    orig_backend = backend
     url = None
     if backend and '://' in backend:
         # There are two types of urls we'll see:
@@ -71,12 +72,12 @@ def by_url(backend=None, loader=None):
         url = backend
         scheme, _, _ = url.partition('://')
         if '+' in scheme:
-            if scheme.startswith('mongodb+'):
-                # e.g., mongodb+srv://
-                backend, _, _ = scheme.partition('+')
-            else:
-                # e.g., sqla+mongodb
-                backend, url = url.split('+', 1)
+            backend, url = url.split('+', 1)
         else:
             backend = scheme
-    return by_name(backend, loader), url
+    cls = by_name(backend, loader)
+
+    if getattr(cls, 'expects_url_munging', True) or url is None:
+        return cls, url
+    else:
+        return cls, orig_backend
--- a/celery/app/backends.py
+++ b/celery/app/backends.py
@@ -57,23 +57,26 @@ def by_name(backend=None, loader=None,
 
 
 def by_url(backend=None, loader=None):
-    """Get backend class by URL. Return a class object and the url.
+    """Get backend class by URL.
 
-    New in version 4.2.?: url is returned unmodified. Previously any leading
-    schema followed by a plus sign would be stripped off. Now backends that
-    require special formatting must handle that formatting in their own
-    constructors.
+    Return a class object and the url to provide to the backend when it is
+    created.
     """
     url = None
     if backend and '://' in backend:
-        # There are two types of uris we'll see:
+        # There are two types of urls we'll see:
         # * <backend>+<backend-defined-ext>:// e.g., mongodb+srv://
-        # * <backend>+<backend-subtype>:// e.g., cache+memory:// (cache is the
-        #   only one tested and documented thus far
+        # * <backend>+<original_scheme>:// e.g., cache+memory://...,
+        #   db+postgresql+psycopg2://..., etc.
         url = backend
         scheme, _, _ = url.partition('://')
         if '+' in scheme:
-            backend, _, _ = scheme.partition('+')
+            if scheme.startswith('mongodb+'):
+                # e.g., mongodb+srv://
+                backend, _, _ = scheme.partition('+')
+            else:
+                # e.g., sqla+mongodb
+                backend, url = url.split('+', 1)
         else:
             backend = scheme
     return by_name(backend, loader), url
--- a/celery/backends/mongodb.py
+++ b/celery/backends/mongodb.py
@@ -141,9 +141,8 @@ class MongoBackend(BaseBackend):
                 # This enables the use of replica sets and sharding.
                 # See pymongo.Connection() for more info.
                 host = self.host
-                if isinstance(host, string_t) \
-                   and not host.startswith('mongodb://'):
-                    host = 'mongodb://{0}:{1}'.format(host, self.port)
+                if isinstance(url, string_t) and '://' not in url:
+                    url = 'mongodb://{0}:{1}'.format(url, self.port)
             # don't change self.options
             conf = dict(self.options)
             conf['host'] = host
--- a/celery/backends/mongodb.py
+++ b/celery/backends/mongodb.py
@@ -141,8 +141,8 @@ class MongoBackend(BaseBackend):
                 # This enables the use of replica sets and sharding.
                 # See pymongo.Connection() for more info.
                 host = self.host
-                if isinstance(url, string_t) and '://' not in url:
-                    url = 'mongodb://{0}:{1}'.format(url, self.port)
+                if isinstance(host, string_t) and '://' not in host:
+                    host = 'mongodb://{0}:{1}'.format(host, self.port)
             # don't change self.options
             conf = dict(self.options)
             conf['host'] = host
--- a/celery/backends/mongodb.py
+++ b/celery/backends/mongodb.py
@@ -54,6 +54,7 @@ class MongoBackend(BaseBackend):
     groupmeta_collection = 'celery_groupmeta'
     max_pool_size = 10
     options = None
+    expects_url_munging = False
 
     supports_autoexpire = False
 
--- a/t/unit/app/test_backends.py
+++ b/t/unit/app/test_backends.py
@@ -6,6 +6,8 @@ from case import patch
 from celery.app import backends
 from celery.backends.amqp import AMQPBackend
 from celery.backends.cache import CacheBackend
+from celery.backends.database import DatabaseBackend
+from celery.backends.mongodb import MongoBackend
 from celery.exceptions import ImproperlyConfigured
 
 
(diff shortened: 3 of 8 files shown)

### CODE REVIEW

**Comment by georgepsarakis** [comment]
@josephtate thanks for your contribution. Let me know once this is ready for review.

**Comment by josephtate** [comment]
Bad merge. I fixed this on the 3.1 branch and it didn't merge cleanly. I'll
fix and run tests before pushing.

**Comment by josephtate** [comment]
This ticket is ready for review, and if possible backport to 3.1, since we're stuck on that for a little while until we can digest 4.0.

File: celery/backends/mongodb.py [hunk 1]
```diff
@@ -141,8 +141,7 @@ def _get_connection(self):
                 # This enables the use of replica sets and sharding.
                 # See pymongo.Connection() for more info.
                 host = self.host
-                if isinstance(host, string_t) \
-                   and not host.startswith('mongodb://'):
+                if isinstance(host, string_t) and '://' not in host:
```
georgepsarakis: An alternative suggestion:

```
if isinstance(host, string_t) and not host.replace('+srv', '').startswith('mongodb://'):
```

The above will help identify directly the use case in the future. What do you think?

t/unit/backends/test_mongodb.py
```diff
@@ -163,6 +164,20 @@ def test_get_connection_no_connection_mongodb_uri(self):
...
+        with patch('pymongo.MongoClient') as mock_Connection:
+            mongodb_uri = self.srv_url
+            self.backend._connection = None
+            self.backend.host = mongodb_uri
+
+            mock_Connection.return_value = sentinel.connection
+
+            connection = self.backend._get_connection()
+            mock_Connection.assert_called_once_with(
+                host=mongodb_uri, **self.backend._prepare_client_options()
```
georgepsarakis: I assumed that `MongoClient` would be initialized with `mongodb://hostname.dom/database`. Isn't that the case?

celery/backends/mongodb.py
josephtate: No, that's wrong. The MongoClient does all the DNS lookups. We don't want to mess with a mongodb+srv url because we'll do it wrong. See the test for what I mean.

t/unit/backends/test_mongodb.py
josephtate: As I said above. MongoClient is passed the full uri unmodified and does all the DNS lookups (SRV, TXT, A).

josephtate: If you look at the test I added to Kombu, you can see what the MongoClient/parse_uri methods do:

https://github.com/celery/kombu/pull/859/files#diff-48b6aab62ddabad370445ef13ab32d41R90

bertjwregeer:
:+1:

This definitely is a better fix than just trying to look for `mongodb+srv` and passing that through unmodified which was my original proposal.

Celery should not attempt to be smart about what URL is provided, instead let the underlying Mongo library deal with it.

josephtate:
This patch (well the backport to 3.1) is working in our test environment and passes all our functional tests, so I think I've addressed all the url parsing errors with the new mongodb+srv:// scheme.

celery/backends/mongodb.py
amol-: I think that @josephtate point is correct, you want to consider `host` the host if it doesn't include any schema, and a schema is always specified through the usage of `://`. So it's fairly more reliable to check for the schema separator than checking for `mongodb://` because if mongo introduces new kind of schemas, celery we will continue to support them this was.

georgepsarakis: I believe this sounds as the more correct approach, thanks for clarifying.

celery/app/backends.py
```diff
@@ -63,7 +63,12 @@ def by_url(backend=None, loader=None):
         url = backend
         scheme, _, _ = url.partition('://')
         if '+' in scheme:
-            backend, url = url.split('+', 1)
+            if scheme.startswith('mongodb+'):
+                # e.g., mongodb+srv://
+                backend = url.split('+', 1)[0]
+            else:
+                # e.g., sqla+mongodb
```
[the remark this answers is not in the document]
josephtate: I think I want to make this method dumb. Just pass the unmodified url to the backend when its created and let each backend handle it however it needs to.

celery/app/backends.py
```diff
@@ -63,7 +63,12 @@ def by_url(backend=None, loader=None):
         url = backend
         scheme, _, _ = url.partition('://')
         if '+' in scheme:
-            backend, url = url.split('+', 1)
+            if scheme.startswith('mongodb+'):
```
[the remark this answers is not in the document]
josephtate: I want to dumb this down and do simple string matching for the backend type. Then let each backend parse its own url. MongodbBackend can handle mongodb+XXX and CacheBackend can deal with its own madness.

celery/app/backends.py
```diff
@@ -57,13 +57,23 @@ def by_name(backend=None, loader=None,
...
+        # There are two types of uris we'll see:
+        # * <backend>+<backend-defined-ext>:// e.g., mongodb+srv://
+        # * <backend>+<backend-subtype>:// e.g., cache+memory:// (cache is the
+        #   only one tested and documented thus far
         url = backend
         scheme, _, _ = url.partition('://')
         if '+' in scheme:
-            backend, url = url.split('+', 1)
+            backend, _, _ = scheme.partition('+')
         else:
```
josephtate: I made this dumb. Let CacheBackend (and other url munging backends) handle its own munging.

georgepsarakis: This change will break backwards compatibility, especially for the SQLAlchemy backend. Please keep the scope to MongoDB and let's make sure that every change here is backwards compatible.

georgepsarakis: @josephtate you may be right and we need to evaluate this more thoroughly. In order to maintain the scope in this PR, you can open a separate issue & PR to explain and implement your design. What do you think?

josephtate: You know what. Nevermind. I'll put this back. I don't care enough to re-architect this.

josephtate: I've got what I think is a good middle ground. I'll commit it and you can tell me what you think.

josephtate:
@georgepsarakis Any additional comments? I'm happy with the current state: it doesn't have backend specific checks in by_url, and still allows backends to decide behavior. It doesn't break existing backends, and 3rd party backends will still use the current url parsing system, but the mongodb backend isn't constrained by the bad parsing.

josephtate:
Any additional requests?

bertjwregeer:
> If I am not wrong, the same can be achieved if in this line in the MongoDB backend, you simply add a check to prepend/restore the mongodb+ prefix, so that it again becomes mongodb+srv://localhost. Isn't this the requirement here?

`srv://localhost` has no meaning, `mongodb+srv://localhost` does. Having to add back information that shouldn't have been stripped in the first place doesn't make much sense. There shouldn't be much, if any, URL munging happening in the first place, especially if the underlying libraries deal with it.

IMHO Celery is trying to be too smart, and even here you admitted that trying to be less smart by allowing potential future schemas: https://github.com/celery/celery/pull/4698/files#r185569254

If in the future MongoDB changes to `mongodb+txt://localhost` or any other URL format we are back to where we are now, and needing to add another check for "`txt://`" to re-add what was originally already there.

georgepsarakis:
@bertjwregeer the comment that you mentioned referred to a different change. I don't think that what happens here is wrong, it just happens that the `mongodb` alias that Celery chose for the backend, and the URL format that the MongoDB client requires, coincide.

If a different name had been chosen, perhaps it would be less confusing, consider this example:

```
mongo+mongodb+srv://localhost
```

Now it is more clear that the first component is the Celery backend alias and the remaining components are what MongoDB requires as a valid connection string.

My suggestion is to simply handle this MongoDB requirement internally. Also as for future changes in the MongoDB client, there is no need to handle each case separately:

```
if not self.url.startswith('mongodb') and '://' in self.url:
    self.url = 'mongodb+{}'.format(self.url)
```

Wouldn't the above work in all cases?

t/unit/app/test_backends.py
```diff
@@ -17,7 +17,8 @@ class test_backends:
     ])
     def test_get_backend_aliases(self, url, expect_cls, app):
         backend, url = backends.by_url(url, app.loader)
-        assert isinstance(backend(app=app, url=url), expect_cls)
```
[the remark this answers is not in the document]
josephtate: Instantiating the class is extra work. It's unnecessary. Test suites should avoid unnecessary work.

t/unit/backends/test_database.py
```diff
@@ -29,6 +30,23 @@ def __init__(self, data):
         self.data = data
 
 
+@skip.unless_module('sqlalchemy')
+class test_backend_by_url:
+    """We can't test these in t/unit/app/test_backends, because of importing in
+    various ci environments, so we repeat it here just for db lookups.
+    """
+
+    @pytest.mark.parametrize('url,expect_cls', [
```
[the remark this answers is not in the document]
josephtate: This test was implemented slightly differently from the other "test_backend_by_url" tests, but still belongs here.

celery/app/backends.py
```diff
@@ -57,13 +57,27 @@ def by_name(backend=None, loader=None,
 
 
 def by_url(backend=None, loader=None):
-    """Get backend class by URL."""
+    """Get backend class by URL.
+
+    Return a class object and the url to provide to the backend when it is
+    created.
+    """
+    orig_backend = backend
```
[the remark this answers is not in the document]
josephtate: No. You care so much about it, you fix it.

t/unit/backends/test_cache.py
```diff
@@ -20,6 +21,22 @@ def __init__(self, data):
         self.data = data
 
 
+class test_backend_by_url:
+    """We can't test these in t/unit/app/test_backends, because of importing in
```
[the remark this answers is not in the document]
josephtate:
```
$ ack -l def\ test_backend_by_url
t/unit/backends/test_mongodb.py
t/unit/backends/test_dynamodb.py
t/unit/backends/test_elasticsearch.py
t/unit/backends/test_couchdb.py
t/unit/backends/test_riak.py
t/unit/backends/test_couchbase.py
```

Plenty of precedent for this. I just added cache, database, and mongo tests to fill out the suite. DyanamoDB, Elasticsearch, Couchdb, and Riak were already there.

MongoDB doesn't get tested with the current Travis setup (probably an oversight?). Don't know about the rest. I just copied the existing paradigm for tests that require non-Travis.yaml libraries to be present. Personally, I think this is the right factoring. Adding all of these tests to test_backends.py will clutter it up with a bunch of skip.unless_module decorators, and since it doesn't actually import the module into the current scope, a local import would be needed for each as well.

celery/app/backends.py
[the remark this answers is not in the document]
georgepsarakis: @josephtate I care about code quality and readability, because it is a maintenance factor. This code is being (and will be) read by many developers, so this is also important for the project.

georgepsarakis:
@thedrow would you like perhaps to offer an additional opinion? I believe we should avoid changing how backend discovery and URL parsing works at a high level (at least for now) in order to accommodate MongoDB specifically. For a suggested alternative approach see https://github.com/celery/celery/pull/4698#issuecomment-389752545

josephtate:
On Thu, May 17, 2018 at 1:25 AM George Psarakis <notifications@github.com>
wrote:

> *@georgepsarakis* requested changes on this pull request.
>
> @josephtate <https://github.com/josephtate> to be honest I don't yet see
> the reason for making so extensive changes.
>
> The current backend URL discovery logic works:
>
> by_url('mongodb+srv://localhost')
> (celery.backends.mongodb.MongoBackend, u'srv://localhost')
>
>
This is not the problem. I wish Bert had added the original traceback to
the bug report, but I don't have it handy. The problem is that throughout
the MongoBackend, it treats that url as if it has been unmodified by the
by_url lookup method, but that method modifies ALL urls with + in the
scheme.

> If I am not wrong, the same can be achieved if in this line
> <https://github.com/celery/celery/pull/4698/files#diff-e00c561f36bc76fb8f2f5a57170e8e01R81>
> in the MongoDB backend, you simply add a check to prepend/restore the
> mongodb+ prefix, so that it again becomes mongodb+srv://localhost. Isn't
> this the requirement here?
>
> What do you think?
>

I tried that, and couldn't do it cleanly, and couldn't get tests to pass.

If I remember correctly, and it's been more than 2 weeks now, all I could
do was special case url.startswith('srv://'), so it definitely wouldn't be
future proof. I tried several ways to just modify MongoBackend, but nothing
was as clean as this without creating a special case for mongodb+srv
somewhere.

celery/app/backends.py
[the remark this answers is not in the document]
josephtate: According to the contributor's guide, I shouldn't have to worry about style and code cleanup, so I'm going to let you demonstrate how much you care about this by making the change when or after it's merged.

t/unit/backends/test_mongodb.py
josephtate: It seems like this is still a hangup, so I'll be explicit: MongoClient will fail if you pass it `mongodb://<host>/<db>...` because that hostname may not exist in DNS. One could ALSO add an A record that would get looked up, but MongoClient, when passed `mongodb+srv://<host>/<db>` will look up `SRV _mongodb._tcp.<host>`, and then `TXT <host>` to get additional parameters. The SRV record will point to one or more mongod servers, and will configure the client as `mongodb://<srv1>:<port1>[,<srv2>:<port2>..]/<db>?<TXT options>`, but unless celery does the lookup itself, just stripping off "+srv" before passing it to MongoClient will not work.

georgepsarakis:
I opened #4744 which includes the alternative implementation. Tests for the MongoDB backend are passing, but there is currently another issue, related with `eventlet` and `dnspython` imports that needs further research.

@josephtate @bertjwregeer thanks for your efforts and clarifications, I think we can close this now and continue the discussion in #4744.

Issue #4174, opened 2017-07-28
Potential race condition when detaching workers


arhoyling:
## Checklist
- [x] I have included the output of ``celery -A proj report`` in the issue.
      (if you are not able to do this, then at least specify the Celery
       version affected).
- [X] I have verified that the issue exists against the `master` branch of Celery.

## Versions
Verified on `celery` `3.1.25` and above (including tip).

## Report
Not applicable.

## Description

In `celery.platforms`, `after_chdir_do()` in `detached()` makes sure the pidfile is not stale by calling `_create_pidlock()` and releasing the pidlock soon after.

Although the comment says "*Doesn't actually create the pidfile, but makes sure it's not stale.*", `_create_pidlock()` does indeed call `Pidfile.acquire()` which **does** create the pidfile.
The fact that the release is performed on the same line (`_create_pidlock(pidfile).release()`) **does not** make it atomic.

Any other process monitoring for the existence of a pidfile (for example to make sure workers actually started) will (one day) see that *temporary* file and wrongly assume workers already loaded their `Blueprint`.

Surely their must be another way to check that the pidfile is not stale without actually creating it.
Why not extract the *stale* logic from `_create_pidlock()` and reusing it there?

```python
def can_create_pidlock(pidlock):
    if pidlock.is_locked() and not pidlock.remove_if_stale():
        print(PIDLOCKED.format(pidfile, pidlock.read_pid()), file=sys.stderr)
        return False
    return True

def _create_pidlock(pidfile):
    pidlock = Pidfile(pidfile)
    if not can_create_pidlock(pidlock):
        raise SystemExit(EX_CANTCREAT)
    return pidlock.acquire()

...
    def after_chdir_do():
        logfile and open(logfile, 'a').close()
        if pidfile and not can_create_pidlock(Pidfile(pidfile)):
            raise SystemExit(EX_CANTCREAT)
```

auvipy:
I am unsure it still exists on master? closing conditionally hoping to reopen if get report again on master

arhoyling:
We removed our work-around after upgrading to 4.1 and saw the race condition happen again (an external system noticed the temporary pidfile and wrongly assumed the celery node had started).
This still exists on master - this part of the code was unchanged. 

I am happy to write a patch for this if needed.

arhoyling:
Here are some simplified timelines to illustrate the problem:

[...]

PR #4704, opened 2018-05-03
Do not create temporary lockfiles before detaching processes (#4174)

arhoyling:
Although it is the responsibility of the child to create its pidlock,
the parent process was temporarily creating it when checking if a
similar process was already running. This could cause race conditions
with external processes - e.g. a process monitoring the existence of the
pidfile.

The fix still checks if a child process is running but does not attempt
to acquire the lock (even temporarily). All other pidfile
related behaviour is unchanged.

## Description

that version series.


celery/platforms.py
```diff
@@ -411,8 +417,8 @@ def after_chdir_do():
         # we need to know that we have access to the logfile.
         logfile and open(logfile, 'a').close()
         # Doesn't actually create the pidfile, but makes sure it's not stale.
-        if pidfile:
-            _create_pidlock(pidfile).release()
+        if pidfile and not _can_create_pidlock(Pidfile(pidfile)):
```
georgepsarakis: You could possibly extract this branch to a new function, something like:

```python
def _ensure_not_stale(pidfile):
    if pidfile:
      ...
```
This way you could even make the comment redundant :smile: !

celery/platforms.py
```diff
@@ -255,10 +255,16 @@ def create_pidlock(pidfile):
     return pidlock
 
 
+def _can_create_pidlock(pidlock):
```
georgepsarakis: How about transferring this to `Pidfile` class? I would suggest something like `is_not_acquired` or similar. What do you think?

celery/platforms.py
arhoyling: Good point

celery/platforms.py
arhoyling: I actually thought it would be better in the Pidfile class but I wanted to minimize changes.
I'll try to commit something soon.

arhoyling: Shouldn't we change `Pidfile.is_locked()` instead to remove the pidfile if stale, and only then report on whether the lock is acquired? Something like:

```python
def is_locked(self):
    return os.path.exists(self.path) and not self.remove_if_stale()
```
It seems semantically more accurate and it is hard to imagine only wanting to check if the file exists without wanting to know if it is legitimately so. It would also save us from growing the `Pidfile` interface.

With this we can implement the helper function:

```python
def _free_or_die(pidfile):
    pidlock = Pidfile(pidfile)
    if pidlock.is_locked():
        print(PIDLOCKED.format(pidlock, pidlock.read_pid()), file=sys.stderr)
        raise SystemExit(EX_CANTCREAT)
    return pidlock
```

And reimplement the other 2 functions:

```python
def _create_pidlock(pidfile):
    return _free_pidfile_or_die(pidfile).acquire()

def after_chdir_do():
    ...
    if pidfile:
        _free_pidfile_or_die(pidfile)
```

celery/platforms.py
```diff
@@ -255,11 +255,16 @@ def create_pidlock(pidfile):
     return pidlock
 
 
-def _create_pidlock(pidfile):
+def _free_pidfile_or_die(pidfile):
     pidlock = Pidfile(pidfile)
-    if pidlock.is_locked() and not pidlock.remove_if_stale():
-        print(PIDLOCKED.format(pidfile, pidlock.read_pid()), file=sys.stderr)
+    if pidlock.is_locked():
+        print(PIDLOCKED.format(pidlock, pidlock.read_pid()), file=sys.stderr)
```
georgepsarakis: Does the `pidlock` object representation contain more information?

I wouldn't advise changing something like this, as users may have already built some kind of alerting mechanism around this message pattern.

celery/platforms.py
```diff
@@ -153,7 +153,7 @@ def acquire(self):
 
     def is_locked(self):
         """Return true if the pid lock exists."""
-        return os.path.exists(self.path)
+        return os.path.exists(self.path) and not self.remove_if_stale()
```
georgepsarakis: I don't see this method used anywhere else, so you can redefine its purpose. However, `is_*` seem more informative and here there are side-effects. How about renaming it also to `maybe_unlock`?

georgepsarakis:
@arhoyling nice work so far! I added a few minor comments, let me know what you think. Also, if you can take some time to rebase, the code coverage difference will surely decrease.

celery/platforms.py
georgepsarakis: Although unlikely that users depend on this class, that is a very good point, thanks for taking the time to consider. I suggested the renaming for future readers and failed to take into account this possibility.

You could always revert and just mark `is_locked` as deprecated, by adding a deprecation warning (see `celery.utils.deprecated.warn` function) and simply create a new function (protected or public) with the name of your choice and the new functionality. What do you think?

georgepsarakis: Just noticed your latest commit, seems like a good solution. Let me have one final look please.

georgepsarakis (dismissed):
Will be included in the next release, thanks again :+1:

georgepsarakis:
@arhoyling do you have anything to add here? Otherwise we can merge.

georgepsarakis:
@auvipy I believe we can merge this. What do you think?

celery/platforms.py
```diff
@@ -229,6 +216,24 @@ def write_pid(self):
         finally:
             rfh.close()
 
+    def _is_stale(self):
+        """Check if the recorded pid in the pidfile is still alive."""
+        try:
+            pid = self.read_pid()
+        except ValueError as exc:
+            print('Broken pidfile found', file=sys.stderr)
```
thedrow: I'd prefer to use the logging module here.
What do you think?

celery/platforms.py
```diff
@@ -229,6 +216,24 @@ def write_pid(self):
...
+            return True
+        if not pid:
+            return True
+
+        try:
+            os.kill(pid, 0)
+        except os.error as exc:
+            if exc.errno == errno.ESRCH:
+                print('Stale pidfile exists', file=sys.stderr)
```
thedrow: I'd prefer to use the logging module here.
What do you think?

georgepsarakis: Not sure but it is highly probable that logging has not been properly configured yet at this point, plus it is better not to rely on logging level configuration etc, since this is a deployment-related message.

celery/platforms.py
auvipy: I think it's OK

auvipy:
please rebased

thedrow:
I can't edit this branch as it was opened from @arhoyling's master branch.
If we don't hear from @arhoyling soon, we need to create a new PR to rebase this.

PR #4709, opened 2018-05-05
Do not subscribe to Redis channels when results are ignored

### LINKED ISSUE #4707: redis backend does not unsubscribe from celery-task-meta-XXX channels when ignore_result=True
This is relevant to using the redis backend.

When I trigger a task, a "pubsub subscribe" command is issued, resulting in a 'celery-task-meta-XXX' entry in the pubsub channels list:
```
(py36) [host]:celery$ sudo redis-cli pubsub channels
1) "celery-task-meta-6e941dbb-006e-4e6e-8cec-2d0e024d6faf"
```
Because there is no return value, and because I don't want to block, I do not call AsyncResult.get(), which normally does the job of unsubscribing.

If the unsubscribe doesn't happen, the list of subscriptions grows without end, causing resource issues after several hundred thousand tasks.

Instead, I've tried two ways to prevent this problem:

1. include **ignore_result=True** in the @app.task annotator
2. to call AsyncResult.forget() on the return value from .delay()

I do not know if this is a bug or if I'm doing things incorrectly. 

This can be reproduced with this code, tasks.py (adapted from a contribution by @mleginus):

```
"""Celery tasks module."""
from celery import Celery
from kombu import Queue
from redis import StrictRedis
from time import sleep

class CeleryConfig(object):
    # celery configurations
    broker_url = 'redis://localhost:6379/1'
    task_default_queue = 'default'
    task_queues = (
        Queue('default', routing_key='parent.#,opbeat'),
    )
    imports = ('tasks',)
    result_backend = 'redis://localhost:6379/2'

def create_app(config):
    """Create a celery app instance."""
    celery_app = Celery("tasks")
    celery_app.config_from_object(config)
    return celery_app

app = create_app(CeleryConfig)

@app.task(ignore_result=True)
def enrich_single_call(data):
    """Single call to task that returns nothing. """
    pass

def run_test_task():
    """Run test task."""
    redis = StrictRedis()
    channels_before = len(redis.execute_command('PUBSUB CHANNELS'))
    print('Redis channels count before task execution {}'.format(channels_before))
    result = enrich_single_call.delay('hello')
    sleep(5)
    channels_after = len(redis.execute_command('PUBSUB CHANNELS'))
    print('Redis channels count after execution {}'.format(channels_after))

```
First, start a worker using tasks.py:
```
celery -A tasks worker -Q default -P solo
```

Then, import tasks.py in an interpreter and run
```
for _ in range(3):
    run_test_task()
```
The output is 
```
Redis channels count before task execution 0
Redis channels count after execution 1
Redis channels count before task execution 1
Redis channels count after execution 2
Redis channels count before task execution 2
Redis channels count after execution 3
```
The output indicates that the number of pubsub channels is growing.

I would expect the count to be 0 on all lines.

If we add 
```
result.get()
```
to the end of the test function, the count decrements.

If we add 
```
result.forget()
```
to the end of the test function, the count does not decrement.

-----
Issue #3812 and #4666 are related, but closed issues.

Based on the fix related to these issues, I tested the code above on the latest release candidate for celery:
```
pip3 freeze | grep celery
celery==4.2.0rc3
```
I'm using redis 4.0.7. I've tried this on Ubuntu 16 and mac os X

Thanks in advance for any help on this!

georgepsarakis:
## Description

The unnecessary channel subscriptions were occurring in two cases: 
- `backends.base.Backend.on_task_call` was being called unconditionally, which initiated the PubSub subscription.
- `AsyncBackendMixin.add_pending_result` also initiates a subscription.

The code excerpt given in #4707 was executed and produced no channel leak, neither on the worker nor on the consumer. Both `chord` and `AsyncResult` return values are retrieved correctly.

- [x] Avoid consumer Redis channel subscriptions if result is to be ignored.
- [x] Calling `AsyncResult.get()` now returns `None` directly, instead of subscribing to Redis channel.

Fixes #4707.

### DIFF
--- a/celery/app/base.py
+++ b/celery/app/base.py
@@ -711,6 +711,8 @@ class Celery(object):
             warnings.warn(AlwaysEagerIgnored(
                 'task_always_eager has no effect on send_task',
             ), stacklevel=2)
+
+        ignored_result = options.pop('ignore_result', False)
         options = router.route(
             options, route_name or name, args, kwargs, task_type)
 
@@ -735,11 +737,18 @@ class Celery(object):
 
         if connection:
             producer = amqp.Producer(connection, auto_declare=False)
+
         with self.producer_or_acquire(producer) as P:
             with P.connection._reraise_as_library_errors():
-                self.backend.on_task_call(P, task_id)
+                if not ignored_result:
+                    self.backend.on_task_call(P, task_id)
                 amqp.send_task_message(P, name, message, **options)
         result = (result_cls or self.AsyncResult)(task_id)
+        # We avoid using the constructor since a custom result class
+        # can be used, in which case the constructor may still use
+        # the old signature.
+        result.ignored = ignored_result
+
         if add_to_parent:
             if not have_parent:
                 parent, have_parent = self.current_worker_task, True
--- a/celery/app/task.py
+++ b/celery/app/task.py
@@ -525,6 +525,9 @@ class Task(object):
 
         preopts = self._get_exec_options()
         options = dict(preopts, **options) if options else preopts
+
+        options.setdefault('ignore_result', self.ignore_result)
+
         return app.send_task(
             self.name, args, kwargs, task_id=task_id, producer=producer,
             link=link, link_error=link_error, result_cls=self.AsyncResult,
--- a/celery/result.py
+++ b/celery/result.py
@@ -101,6 +101,19 @@ class AsyncResult(ResultBase):
         self.parent = parent
         self.on_ready = promise(self._on_fulfilled)
         self._cache = None
+        self._ignored = False
+
+    @property
+    def ignored(self):
+        """"If True, task result retrieval is disabled."""
+        if hasattr(self, '_ignored'):
+            return self._ignored
+        return False
+
+    @ignored.setter
+    def ignored(self, value):
+        """Enable/disable task result retrieval."""
+        self._ignored = value
 
     def then(self, callback, on_error=None, weak=False):
         self.backend.add_pending_result(self, weak=weak)
@@ -176,6 +189,9 @@ class AsyncResult(ResultBase):
             Exception: If the remote call raised an exception then that
                 exception will be re-raised in the caller process.
         """
+        if self.ignored:
+            return
+
         if disable_sync_subtasks:
             assert_will_not_block()
         _on_interval = promise()
--- a/celery/result.py
+++ b/celery/result.py
@@ -372,6 +372,11 @@ class AsyncResult(ResultBase):
     def __reduce_args__(self):
         return self.id, self.backend, None, None, self.parent
 
+    def __del__(self):
+        """Cancel pending operations when the instance is destroyed."""
+        if self.backend is not None:
+            self.backend.remove_pending_result(self)
+
     @cached_property
     def graph(self):
         return self.build_graph()
--- a/t/integration/test_tasks.py
+++ b/t/integration/test_tasks.py
@@ -1,9 +1,11 @@
 from __future__ import absolute_import, unicode_literals
 
+import pytest
+
 from celery import group
 
-from .conftest import flaky
-from .tasks import print_unicode, retry_once, sleeping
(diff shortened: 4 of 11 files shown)


celery/app/base.py
```diff
@@ -735,11 +737,16 @@ def send_task(self, name, args=None, kwargs=None, countdown=None,
...
+
         with self.producer_or_acquire(producer) as P:
             with P.connection._reraise_as_library_errors():
-                self.backend.on_task_call(P, task_id)
+                if not ignored_result:
+                    self.backend.on_task_call(P, task_id)
                 amqp.send_task_message(P, name, message, **options)
         result = (result_cls or self.AsyncResult)(task_id)
+        # Avoid constructor for backward compatibility
+        result.ignored = ignored_result
```
thedrow: I don't understand from this comment why we need to avoid the constructor for backwards compatibility.
Can you please elaborate?

celery/result.py
```diff
@@ -101,6 +101,17 @@ def __init__(self, id, backend=None,
         self.parent = parent
         self.on_ready = promise(self._on_fulfilled)
         self._cache = None
+        self._ignored = False
+
+    @property
+    def ignored(self):
+        if hasattr(self, '_ignored'):
```
thedrow: In which case we won't have `_ignored` if it's initialized in the constructor.

celery/app/base.py
georgepsarakis: I wanted a second opinion on this. I thought that since someone can use a custom result class, they may have overridden the constructor, so the signature will not contain the new argument. I thought that by using a setter, it would be 100% compatible with sub-classes.

celery/result.py
georgepsarakis: I introduced this check in order to account for sub-classes with overridden constructor, please see my comment above.

celery/app/base.py
thedrow: Now I understand. The comment is wrong though.
We avoid using the constructor since a custom request class can be used.

I think that requiring a Request class to implement a specific contract is 100% OK but we need to document it.

Either change the comment or adjust the documentation accordingly.

celery/result.py
```diff
@@ -101,6 +101,17 @@ def __init__(self, id, backend=None,
         self.parent = parent
         self.on_ready = promise(self._on_fulfilled)
         self._cache = None
+        self._ignored = False
+
+    @property
+    def ignored(self):
```
thedrow: Let's add a docstring for this property.

celery/result.py
```diff
@@ -101,6 +101,17 @@ def __init__(self, id, backend=None,
...
+        if hasattr(self, '_ignored'):
+            return self._ignored
+        return False
+
+    @ignored.setter
+    def ignored(self, value):
```
thedrow: Let's add a docstring for this property.

celery/app/base.py
georgepsarakis: You are right, I must adjust the comment, thanks for the suggestion.

celery/app/task.py
```diff
@@ -525,6 +525,9 @@ def apply_async(self, args=None, kwargs=None, task_id=None, producer=None,
 
         preopts = self._get_exec_options()
         options = dict(preopts, **options) if options else preopts
+
+        options.setdefault('ignore_result', self.ignore_result)
```
georgepsarakis: I think the only test cases missing now are the ones regarding passing correctly the `ignore_result` option.

t/integration/test_tasks.py
```diff
@@ -25,3 +25,25 @@ def test_unicode_task(self, manager):
             group(print_unicode.s() for _ in range(5))(),
             timeout=10, propagate=True,
         )
+
+    def test_ignoring_result_no_subscriptions(self):
```
thedrow: Shouldn't these tests run only when redis is available?

georgepsarakis:
@thedrow I don't really understand why the patch code coverage is below the threshold, if you have any ideas please let me know. Other than that, I think this is ready for review and potentially merging.

Issue #4710, opened 2018-05-05
Unable to use msgpack with groups


p-himik:
## Checklist
* [x] I have included the output of ``celery -A proj report`` in the issue.
<details><summary>Click to expand</summary><p>

```
software -> celery:4.2.0rc3 (windowlicker) kombu:4.1.0 py:3.6.5
            billiard:3.5.0.3 redis:2.10.6
platform -> system:Linux arch:64bit, ELF imp:CPython
loader   -> celery.loaders.app.AppLoader
settings -> transport:redis results:redis:///

ABSOLUTE_URL_OVERRIDES: {
 }
ADMINS: []
ALLOWED_HOSTS: []
APPEND_SLASH: True
AUTHENTICATION_BACKENDS: ['django.contrib.auth.backends.ModelBackend']
AUTH_PASSWORD_VALIDATORS: '********'
AUTH_USER_MODEL: 'auth.User'
BASE_DIR: '/home/p-himik/tmp/celery_test'
CACHES: {
 'default': {'BACKEND': 'django.core.cache.backends.locmem.LocMemCache'}}
CACHE_MIDDLEWARE_ALIAS: 'default'
CACHE_MIDDLEWARE_KEY_PREFIX: '********'
CACHE_MIDDLEWARE_SECONDS: 600
CELERY_ACCEPT_CONTENT: ['msgpack']
CELERY_BROKER_POOL_LIMIT: 0
CELERY_BROKER_TRANSPORT_OPTIONS: {
 'max_connections': 3}
CELERY_BROKER_URL: 'redis://localhost:6379//'
CELERY_REDIS_MAX_CONNECTIONS: 2
CELERY_RESULT_BACKEND: 'redis:///'
CELERY_RESULT_PERSISTENT: True
CELERY_RESULT_SERIALIZER: 'msgpack'
CELERY_TASK_IGNORE_RESULT: False
CELERY_TASK_SERIALIZER: 'msgpack'
CELERY_TASK_STORE_ERRORS_EVEN_IF_IGNORED: True
CSRF_COOKIE_AGE: 31449600
CSRF_COOKIE_DOMAIN: None
CSRF_COOKIE_HTTPONLY: False
CSRF_COOKIE_NAME: 'csrftoken'
CSRF_COOKIE_PATH: '/'
CSRF_COOKIE_SECURE: False
CSRF_FAILURE_VIEW: 'django.views.csrf.csrf_failure'
[...]

PR #4712, opened 2018-05-06
Testcase for #4710

thedrow:
An integration test for #4710.

### DIFF
--- a/t/integration/test_canvas.py
+++ b/t/integration/test_canvas.py
@@ -704,7 +704,8 @@ class test_group:
 
     @flaky
     @pytest.mark.xfail(raises=ExtraData, strict=True)
-    @pytest.mark.celery(task_serializer='msgpack', accept_content=['application/x-msgpack'])
+    @pytest.mark.celery(task_serializer='msgpack',
+                        accept_content=['application/x-msgpack'])
     def test_group_msgpack(self, manager):
         assert manager.inspect().ping()
 
--- a/t/integration/test_canvas.py
+++ b/t/integration/test_canvas.py
@@ -704,6 +704,7 @@ class test_group:
 
     @flaky
     @pytest.mark.xfail(raises=ExtraData, strict=True)
+    @pytest.mark.celery(task_serializer='msgpack', accept_content=['msgpack'])
     def test_group_msgpack(self, manager):
         assert manager.inspect().ping()
 
@@ -711,7 +712,7 @@ class test_group:
               group([
                     identity.si(b'arg1'),
                     identity.si(b'arg2')
-                    ])).delay(serializer='msgpack')
+                    ])).delay()
 
 
 def assert_ids(r, expected_value, expected_root_id, expected_parent_id):
--- a/t/integration/test_canvas.py
+++ b/t/integration/test_canvas.py
@@ -3,12 +3,12 @@ from datetime import datetime, timedelta
 from time import sleep
 
 import pytest
+from msgpack.exceptions import ExtraData
 
 from celery import chain, chord, group, signature
 from celery.backends.base import BaseKeyValueStoreBackend
 from celery.exceptions import ImproperlyConfigured, TimeoutError
 from celery.result import AsyncResult, GroupResult, ResultSet
-from msgpack.exceptions import ExtraData
 
 from . import tasks
 from .conftest import get_active_redis_channels, get_redis_connection
--- a/t/integration/test_canvas.py
+++ b/t/integration/test_canvas.py
@@ -2,6 +2,7 @@ import re
 from datetime import datetime, timedelta
 from time import sleep
 
+from msgpack.exceptions import ExtraData
 import pytest
 
 from celery import chain, chord, group, signature
@@ -700,6 +701,17 @@ class test_group:
         )                                       # [42, 42] due to unrolling
         res = sig.delay()
         assert res.get(timeout=TIMEOUT) == [42, 42]
+
+    @flaky
+    @pytest.mark.xfail(raises=ExtraData, strict=True)
+    def test_group_msgpack(self, manager):
+        assert manager.inspect().ping()
+
+        chain(identity.si(b'arg'),
+              group([
+                    identity.si(b'arg1'),
+                    identity.si(b'arg2')
+              ])).delay(serializer='msgpack')
 
 
 def assert_ids(r, expected_value, expected_root_id, expected_parent_id):
--- a/t/integration/test_canvas.py
+++ b/t/integration/test_canvas.py
@@ -704,7 +704,7 @@ class test_group:
 
     @flaky
     @pytest.mark.xfail(raises=ExtraData, strict=True)
-    @pytest.mark.celery(task_serializer='msgpack', accept_content=['msgpack'])
+    @pytest.mark.celery(task_serializer='msgpack', accept_content=['application/x-msgpack'])
     def test_group_msgpack(self, manager):
         assert manager.inspect().ping()
 
--- a/t/integration/test_canvas.py
+++ b/t/integration/test_canvas.py
@@ -9,6 +9,7 @@ from celery import chain, chord, group, signature
 from celery.backends.base import BaseKeyValueStoreBackend
 from celery.exceptions import ImproperlyConfigured, TimeoutError
 from celery.result import AsyncResult, GroupResult, ResultSet
+from kombu.exceptions import DecodeError
 
 from . import tasks
 from .conftest import get_active_redis_channels, get_redis_connection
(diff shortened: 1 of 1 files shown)


thedrow:
@georgepsarakis @auvipy Any idea why I can't switch the accepted content type?
I didn't do anything wrong but it still seems that we don't allow this.
http://docs.celeryproject.org/en/latest/userguide/configuration.html#accept-content

auvipy:
from error msg seems the serialization is tied to json hardcoded anywhere. need time to investigate. not sure if it's intended

georgepsarakis:
@thedrow @auvipy it seems that the testing app configuration is correct though, when reaching [here](https://github.com/celery/celery/blob/120770929f4a37c5373a378b75b5c41a99702af9/celery/contrib/testing/app.py#L16).

auvipy (commented):
restarted the unit tests

thedrow:
Seems like this is the same problem I had with the auth serializer.
The worker already started with some configuration and we don't provide a way to reset it and reload the configuration through the pytest plugin. If we do provide one, I don't know about it.

auvipy:
I am not fully sure either, right now.

maybe-sybr:
Freshened this on top of current master and can get an expected exception if I hack `application/x-msgpack` into the accept types in the kombu decoder. The task serializer marking appears to be working fine but the accept_content one isn't. Tracked it down to our instantiation of the kombu mailbox not actually passing the app's `accept_content` option to the mailbox so IIUC no non-JSON serializaed task bodies would current work. :o  Once I patched the mailbox instantiation, the test just needed to assert on a kombu exception rather than the msgpack one since it gets re-raised as a `DecodeError`.

I think we'll need another issue to track that mailbox instantiation thing to get some more task serialiser tests in before we merge this test case. Worth explicitly noting, the originally reported issue is still relevant and the exception described there is raised on task payload deserialisation in the integration test added here.

maybe-sybr:
I've pushed a standalone diff as fix/mailbox-accept-content (940478fc7) which makes the default control class instantiate its kombu mailbox with the configured mimetypes as well as the required json one. I'm not entirely sure if this is the right fix though since the naming implies that we shouldn't be passing non-control-plane messages via this mailbox... It has a simple sanity test so we could merge it if it's not an insane thing.

@celery/core-developers - is anyone able to provide some insight into the control class and if allowing it to decode application mime types is a sensible thing to do, or if I've misunderstood something?

thedrow:
I think that the patch is fine.

maybe-sybr:
I've revisited this and am pretty confident that the branch I pushed shouldn't be merged. That patch is to the Control class which I believe should only ever be handling JSON, not user data serialised with a user selected serializer like msgpack. The issue with this test case FWICT is that we don't have control over the consumer being spun up in a worker to handle the message we're attempting to send to it. The commits on this branch are quite happily setting the task serialiser upon submission but since the a consumer has never actually been instantiated with the correct accept types configuration, it's quite rightly refusing to decode the untrusted data type.

I can't find any prior art in the integration tests for spinning up consumers with specific configurations and I don't really understand the testing fixtures which are spinning up the app/worker/consumer(s) for the integration tests, so I'm spinning in the same place more than doing anything useful at the moment. I think I need someone with a bit more knowledge of the integration testing workers to pitch in on this one. Happy to elaborate more on what I know wrt message serialiser config if someone else knows how to manipulate or replace the worker/consumer(s) which are used to run the integration tests.

For now, I'm going to push this branch freshened against current master and re-assign back to @thedrow to get some input.

maybe-sybr:
@thedrow and I chatted about this last night and he suggested that an alteration to the app instance(s) which we use in testing to set the default queue name to a hash on something like the test name would help us isolate the instances and support test cases like this one. I've created #6527 to track this work.

maybe-sybr:
@matusvalo - I noticed that #6757 landed on master which makes a similar looking change to the commit I linked a few comments upward (ref: [0]). I ended up not thinking that the `Control` mailbox class should be using non-JSON serialisers but TBH I was never sure about that. Could you have a look over the diff for this PR and my commit at [0] and let me know if you have any thoughts? Could we re-tool the tests here for merge? Did your change potentially just fix #4710 for us?

* [0] https://github.com/celery/celery/commit/940478fc7a81a5bbfc2ac1d232c218410a418123

matusvalo:
>  I ended up not thinking that the Control mailbox class should be using non-JSON serialisers but TBH I was never sure about that. 

For now the main benefit of non-JSON serializers of Control mailbox is that it is easy fix for the issues with non-json parameters passed to task. Otherwise the protocol is fine with pure JSON and there is no need for non-JSON serializers. Moreover, There is one pending issue #6759 which is still not fixed.

> Could you have a look over the diff for this PR and my commit at [0] and let me know if you have any thoughts? Could we re-tool the tests here for merge? 
I need to have deeper look into the issue, PR and your commit. I will let you know.

> Did your change potentially just fix #4710 for us?
If I understand the issue, I suppose not.

matusvalo:
I have spent some time with this PR and it seems working for me (after some fixes - I have used `apply_async` instead of `delay` since in `delay` you cannot pass options + I have set msgpack as accepted serializer):

t/integration/test_canvas.py :
```python
 920     @flaky
 921     # @pytest.mark.xfail(raises=ExtraData, strict=True)
 922     def test_group_msgpack(self, manager):
 923         breakpoint()
 924         assert manager.inspect().ping()
 925         chain(identity.si(b'arg'),
 926               group([
 927                     identity.si(b'arg1'),
 928                     identity.si(b'arg2')
 929               ])).apply_async(serializer='msgpack')
```

t/integration/conftest.py:
```python
 32 @pytest.fixture(scope='session')
 33 def celery_config():
 34     return {
 35         'broker_url': TEST_BROKER,
 36         'result_backend': TEST_BACKEND,
 37         'accept_content': {'json', 'msgpack'},
 38         'cassandra_servers': ['localhost'],
 39         'cassandra_keyspace': 'tests',
 40         'cassandra_table': 'tests',
 41         'cassandra_read_consistency': 'ONE',
 42         'cassandra_write_consistency': 'ONE'
 43     }
 ```
 
 With this changes the test is passing:
 ```
 $ pytest --full-trace -vv t/integration/test_canvas.py::test_group::test_group_msgpack
=========================================================================================================================== test session starts ============================================================================================================================platform linux -- Python 3.8.5, pytest-6.2.4, py-1.10.0, pluggy-0.13.1 -- /home/matus/dev/celery3.8/bin/python3.8
cachedir: .pytest_cache
rootdir: /home/matus/dev/celery, configfile: setup.cfg
plugins: subtests-0.4.0, case-1.5.3, timeout-1.4.2, celery-0.0.0a1
collected 1 item

t/integration/test_canvas.py::test_group::test_group_msgpack
>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> PDB set_trace (IO-capturing turned off) >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> /home/matus/dev/celery/t/integration/test_canvas.py(924)test_group_msgpack()
-> assert manager.inspect().ping()
(Pdb) c

>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> PDB continue (IO-capturing resumed) >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>PASSED                                                                                                                                                                                                  [100%]

============================================================================================================================ 1 passed in 6.76s =============================================================================================================================
 ```

File: t/integration/test_canvas.py [hunk 1]
```diff
@@ -701,6 +703,18 @@ def test_nested_group_group(self, manager):
         res = sig.delay()
         assert res.get(timeout=TIMEOUT) == [42, 42]
 
+    @pytest.mark.xfail(raises=DecodeError, strict=True)
```
thedrow, line 706: Should this still fail or does it fail because our test suite doesn't switch the configuration because we use the session worker?

PR #4724, opened 2018-05-09
Changed documentation on eventlet

bluewhale8202:
"You need to ensure your tasks don't perform blocking calls"
This advice seems wrong. As far as I know, green threads are meant to be used for "blocking I/O-bound" operations. For example `requests.get`, which also is used in celery/examples/eventlet/tasks.py, blocks until it gets the response. I think what really matters for eventlet is whether an operation is CPU-bound or I/O-bound, not whether it's blocking or not.

### DIFF
--- a/celery/app/base.py
+++ b/celery/app/base.py
@@ -203,7 +203,7 @@ class Celery(object):
     log_cls = 'celery.app.log:Logging'
     control_cls = 'celery.app.control:Control'
     task_cls = 'celery.app.task:Task'
-    registry_cls = TaskRegistry
+    registry_cls = 'celery.app.registry:TaskRegistry'
 
     _fixups = None
     _pool = None
--- a/docs/userguide/concurrency/eventlet.rst
+++ b/docs/userguide/concurrency/eventlet.rst
@@ -21,10 +21,16 @@ change how you run your code, not how you write it.
     * The event dispatch is implicit: meaning you can easily use Eventlet
       from the Python interpreter, or as a small part of a larger application.
 
-Celery supports Eventlet as an alternative execution pool implementation.
-It's in some cases superior to prefork, but you need to ensure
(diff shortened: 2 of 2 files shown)


thedrow:
This is not true since not all calls can be monkeypatched.
You could use pylibmc which is I/O bound but still blocks the entire event loop.

bluewhale8202:
Okay, now I see your point. I think there is a distinction between programming in blocking style and blocking the event loop in effect. I was confused because I thought that the document advised me to do programming in asynchronous style. Particularly, you can find this line in the same file

> Coroutines ensure that the developer uses a *blocking style* of programming that's similar to threading, but provide the benefits of non-blocking I/O. 

Although I understand the point and the documentation is not wrong, I still think it's not totally gentle for a beginner who can't come up edge cases like pylibmc. How about this version?

Celery supports Eventlet as an alternative execution pool implementation and in some cases superior to prefork. However, you need to ensure one task doesn't block the event loop too long. Generally, CPU-bound operations don't go well with Evenetlet. Also note that some I/O-bound operations, such as ones in pylibmc, cannot be monkeypatched and therefore block the event loop without benefitting from using Eventlet.

Thank you for your reviewing🙂

thedrow:
Replace "pylibmc" with most C extensions and we're good.
If you don't mind, this also applies to gevent and we can fix that as well.

thedrow:
@bluewhale8202 ping?

bluewhale8202:
@thedrow Sorry for late reply. I don't know if I can modify this pull request by myself. Should a make a new pull request?

And for the gevent documentation, I though it would try that after I have better understanding of celery than now, maybe one or two week later.

thedrow:
@bluewhale8202 Yes you can change this PR. Just push to your branch.

PR #4736, opened 2018-05-15
Use Redis coercion mechanism for converting query parameters

### LINKED ISSUE #4735: Redis connection timeouts are not coerced correctly
Found using celery 4.1.0.

## Steps to reproduce

Configure with redis as a backend with timeouts set in the url:

```
result_backend = 'redis://127.0.0.1:6379?socket_timeout=30&socket_connect_timeout=5'
```

## Expected behavior

Timeouts are applied and celery works normally.

## Actual behavior

Errors are raised by the socket library that the timeout needs to be a float.

Pull request incoming.

reversefold:
Fixes #4735 

Redis does coercion of the query parameters when it parses the URL but celery pre-parses it and sends the values in as keyword arguments. When timeouts are specified in the result backend URL they are passed in as strings instead of floats, causing an error when the timeout is set on the socket.

Ideally, celery should pass the URL to the redis client and let it handle the parsing. This patch doesn't do that, though, it just uses the redis client's coercion map on the query params.

---
This change is [Reviewable](https://reviewable.io/reviews/celery/celery/4736)

### DIFF
--- a/celery/backends/redis.py
+++ b/celery/backends/redis.py
@@ -252,7 +252,9 @@ class RedisBackend(base.BaseKeyValueStoreBackend, async.AsyncBackendMixin):
 
         for key, value in query.items():
             if key in redis.connection.URL_QUERY_ARGUMENT_PARSERS:
-                query[key] = redis.connection.URL_QUERY_ARGUMENT_PARSERS[key](value)
+                query[key] = redis.connection.URL_QUERY_ARGUMENT_PARSERS[key](
+                    value
+                )
 
         # Query parameters override other parameters
         connparams.update(query)
--- a/celery/backends/redis.py
+++ b/celery/backends/redis.py
@@ -29,6 +29,7 @@ except ImportError:
 
 try:
     import redis
+    import redis.connection
     from kombu.transport.redis import get_redis_error_classes
 except ImportError:  # pragma: no cover
     redis = None  # noqa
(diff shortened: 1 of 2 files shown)


celery/backends/redis.py
```diff
@@ -249,6 +250,10 @@ def _params_from_url(self, url, defaults):
         db = db.strip('/') if isinstance(db, string_t) else db
         connparams['db'] = int(db)
 
+        for key, value in query.items():
+            if key in redis.connection.URL_QUERY_ARGUMENT_PARSERS:
+                query[key] = redis.connection.URL_QUERY_ARGUMENT_PARSERS[key](value)
```
georgepsarakis: This line is longer than 80 characters, thus Travis CI checks are failing.

reversefold:
I've pushed a test case for coercing the timeouts in the redis url.

georgepsarakis:
@auvipy I thought that we were meaning to finalize the 4.2 release and then move on with merging new fixes/features.

auvipy:
OK. but can can include critical fixes only then if any

reversefold:
For now I've been using my own fork of celery 4.1.1 with this patch applied as it was causing errors in my application that weren't there before. This is technically a regression from celery before v4.

Is there anything I need to do to fix the appveyor checks? Those don't look like errors I introduced.

georgepsarakis:
@reversefold you can ignore Appveyor.

reversefold:
Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[celery/backends/redis.py, line 255 at r1](https://reviewable.io/reviews/celery/celery/4736#-LC_RZtL7z3Uw36uKTyK-r1-255:-LDNrx9P9GLo8vXm7h4W:b-896fix) ([raw file](https://github.com/celery/celery/blob/8ea69b597ee4ba757556cf50f454c78e5dea08d0/celery/backends/redis.py#L255)):*
<details><summary><i>Previously, georgepsarakis (George Psarakis) wrote…</i></summary><blockquote>

This line is longer than 80 characters, thus Travis CI checks are failing.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io/reviews/celery/celery/4736)*

thedrow:
I think we can merge this as it is hurting @reversefold and isn't a huge change.
@auvipy @georgepsarakis Any objections?

auvipy:
I am willing to getting this in before 4.2 finale

PR #4744, opened 2018-05-19
MongoDB Backend: accept DNS seedlist connection format

### LINKED ISSUE #4668: MongoDB backend does not support mongodb+srv:// URL's

georgepsarakis:
## Description

- [x] Allow MongoDB backend to accept connection strings starting with the `srv` scheme.
- [x] Add `dnspython` in MongoDB extra requirements.
- [x] Add MongoDB backend connection settings initialization test.

Resolves #4668 .

### DIFF
--- a/t/unit/backends/test_mongodb.py
+++ b/t/unit/backends/test_mongodb.py
@@ -121,6 +121,62 @@ class test_MongoBackend:
 
         mb = MongoBackend(app=self.app, url='mongodb://')
 
+    @patch('dns.resolver.query')
+    def test_init_mongodb_dns_seedlist(self, dns_resolver_query):
+        from dns.rdtypes.IN.SRV import SRV
+        from dns.rdtypes.ANY.TXT import TXT
+        from dns.name import Name
+
+        self.app.conf.mongodb_backend_settings = None
+
+        def mock_resolver(_, record_type):
+            if record_type == 'SRV':
+                return [
+                    SRV(0, 0, 0, 0, 27017, Name(labels=hostname))
+                    for hostname in [
+                        b'mongo1.example.com'.split(b'.'),
+                        b'mongo2.example.com'.split(b'.'),
+                        b'mongo3.example.com'.split(b'.')
+                    ]
+                ]
+            elif record_type == 'TXT':
+                return [TXT(0, 0, [b'replicaSet=rs0'])]
+
+        dns_resolver_query.side_effect = mock_resolver
+
+        # uri with user, password, database name, replica set,
+        # DNS seedlist format
+        uri = ('srv://'
+               'celeryuser:celerypassword@'
+               'dns-seedlist-host.example.com/'
+               'celerydatabase')
+
+        mb = MongoBackend(app=self.app, url=uri)
+        assert mb.mongo_host == [
+            'mongo1.example.com:27017',
+            'mongo2.example.com:27017',
+            'mongo3.example.com:27017',
+        ]
+        assert mb.options == dict(
+            mb._prepare_client_options(),
+            replicaset='rs0',
+            ssl=True
+        )
+        assert mb.user == 'celeryuser'
+        assert mb.password == 'celerypassword'
+        assert mb.database_name == 'celerydatabase'
+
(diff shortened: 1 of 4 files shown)


georgepsarakis:
Related discussion and certain details can be found in #4698 .

georgepsarakis:
The [build error](https://travis-ci.org/celery/celery/jobs/381090478) is caused by an import issue with [eventlet](https://github.com/eventlet/eventlet/issues/479). I need to look further into this.

georgepsarakis:
I guess we need to wait for [eventlet 0.24](https://github.com/eventlet/eventlet/milestone/9) to be released.

t/unit/backends/test_mongodb.py
```diff
@@ -121,6 +121,62 @@ def test_init_with_settings(self):
...
 
+    @patch('dns.resolver.query')
+    def test_init_mongodb_dns_seedlist(self, dns_resolver_query):
+        from dns.rdtypes.IN.SRV import SRV
+        from dns.rdtypes.ANY.TXT import TXT
+        from dns.name import Name
+
+        self.app.conf.mongodb_backend_settings = None
+
+        def mock_resolver(hostname, record_type):
```
josephtate: I think you're mocking at way too low a level. The pymongo interface is the one we're interested in, not DNSpython. I think this is much simpler if you just mock pymongo's url parsing method as in https://github.com/celery/kombu/pull/859/files#diff-48b6aab62ddabad370445ef13ab32d41R90

josephtate:
I'm ok with this implementation, but, it's only half of the fix. Kombu needs fixing too: https://github.com/celery/kombu/pull/859

t/unit/backends/test_mongodb.py
josephtate: Doing this removes the need for dnspython in the list of deps, and so doesn't get affected by that eventlet bug.

auvipy: can you handle the kombu part? Joseph

thedrow:
@georgepsarakis Is this ready for merge?

georgepsarakis:
@thedrow is the latest release for eventlet is 0.24.1.

I can add the new dependency, rebase and see if the tests pass.

auvipy:
I restarted the failing builds

georgepsarakis:
@auvipy I think there is an issue with Python3, I will have a look.

bertjwregeer:
Where are we on this change?

georgepsarakis:
Rebased from current master. If tests pass, this should be ready for merge.

auvipy:
restarted the failing unit tests builds

georgepsarakis:
@auvipy thanks for restarting, a small fix was required for Python3 though.

thedrow:
Can we merge this?
Do we need the Kombu equivalent?

georgepsarakis:
@thedrow I think it could be merged. Let me know if I should add release notes & documentation.

As far as Kombu is considered, it is not a prerequisite, but I guess it would be nice if both broker and backend URLs supported this format.

thedrow:
Documentation is necessary.
You can add this to the changlog if you wish. If not, I'll add it before the release.

thedrow:
@georgepsarakis Why is dnspython required for the mongodb result backend now?
Isn't it related to eventlet?

georgepsarakis:
@thedrow thanks for making the change, indeed better to delegate dependencies to `pymongo` instead. Sorry for not noticing this before!

thedrow:
Now how about writing the documentation on this one?

georgepsarakis:
@thedrow sure I can add it. Do we have other documentation tasks pending? I can open a PR and include these as well, if any.

thedrow:
Check our blockers. I think there is one.

PR #4746, opened 2018-05-21
Add test for ResultSet that includes a deliberate error

donkopotamus:
The `master` branch currently defines `ResultSet._on_ready` as:
```python
def _on_ready(self):
    if self.backend.is_async:
        self.on_ready()
```
Where `_on_ready` is called via the `barrier` `self._on_full`.  However, this is invalid in any situation where a member of the `ResultSet` raised an exception, as in that case the execution of `r.get()` will cause an exception to unexpectedly cascade through to any `join_native` over the set.  

```Python Traceback
celery/result.py:671: in get
    on_interval=on_interval,
celery/result.py:789: in join_native
    on_message, on_interval):
celery/backends/async.py:142: in iter_native
    for _ in self._wait_for_pending(result, no_ack=no_ack, **kwargs):
celery/backends/async.py:255: in _wait_for_pending
    on_interval=on_interval):
celery/backends/async.py:56: in drain_events_until
    yield self.wait_for(p, wait, timeout=1)
celery/backends/async.py:65: in wait_for
    wait(timeout=timeout)
celery/backends/rpc.py:63: in drain_events
    return self._connection.drain_events(timeout=timeout)
    ...
celery/result.py:519: in _on_ready
celery/result.py:213: in get
celery/result.py:329: in maybe_throw
    self.throw(value, self._to_remote_traceback(tb))
celery/result.py:322: in throw
    self.on_ready.throw(*args, **kwargs)
.tox/2.7-integration-rabbitmq/lib/python2.7/site-packages/vine/promises.py:217: in throw
reraise(type(exc), exc, tb)
```
NOTE: The first commit is simply an attempt to replicate the error within the tests, a subsequent commit will then provide a fix.


donkopotamus:
There is a single test failure in `python 3.4: integration.redis: test_chord.test_empty_header_chord` which timed out.  However this same test has succeeded in 2.7, 3.5 and 3.6.

celery/result.py
```diff
@@ -497,7 +497,6 @@ class ResultSet(ResultBase):
 
     def __init__(self, results, app=None, ready_barrier=None, **kwargs):
         self._app = app
-        self._cache = None
```
auvipy: my concern is removing cache attrib won;t cause any regression?

donkopotamus: Grepping the source indicates that the only place a `._cache` appears to be used is in this type's own `.get`, and in `backends/async.py`.  The latter appears irrelevant however, as in that case it is referring to `AsyncResult._cache` ... and the current `ResultSet._cache` is in an incompatible format to `AsyncResult._cache` so can't be being accessed there.

(`AsyncResult._cache` stores the complete metadata of the result, whereas `ResultSet._cache` stores just the return value ... which is the problem, you can't do that unless every single result in the list actually succeeded).

Further, users of `ResultSet` are just as likely to be accessing results via `join` as `get` ... and only `get` attempts to access `_cache` ... all the other methods for extracting results ignore it completely.

auvipy: redis integration tests are failing

auvipy:
try to increase the patch coverage

donkopotamus:
The single test failure is a timeout error which I'm happily ignoring as it seems to happen sporadically in these continuous integration tests.

auvipy:
thanks @donkopotamus Derek!

PR #4751, opened 2018-05-22
Enable automatic result expiration with couchbase backend

dfresh613:
## Description
Couchbase has the ability to support automatic result expiration by specifying a ttl keyword argument when saving results.

This PR implements using the couchbase ttl argument and uses the backend expires property for its value.

By default, the ttl value is None, so if backend.expires is None, it won't have any impact.

[Couchbase Documentation including ttl value args](http://docs.couchbase.com/sdk-api/couchbase-python-client-1.2.3/api/couchbase.html#)

Please note that this is a similar change to 4752. Please let me know if you'd like me to combine them

### DIFF
--- a/celery/backends/couchbase.py
+++ b/celery/backends/couchbase.py
@@ -39,6 +39,8 @@ class CouchbaseBackend(KeyValueStoreBackend):
     username = None
     password = None
     quiet = False
+    supports_autoexpire = True
+
     timeout = 2.5
 
     # Use str as couchbase key not bytes
@@ -103,7 +105,7 @@ class CouchbaseBackend(KeyValueStoreBackend):
             return None
 
     def set(self, key, value):
-        self.connection.set(key, value)
+        self.connection.set(key, value, ttl=self.expires)
 
     def mget(self, keys):
         return [self.get(key) for key in keys]
--- a/celery/backends/couchbase.py
+++ b/celery/backends/couchbase.py
@@ -47,7 +47,8 @@ class CouchbaseBackend(KeyValueStoreBackend):
     key_t = str_t
 
     def __init__(self, url=None, *args, **kwargs):
-        super(CouchbaseBackend, self).__init__(*args, **kwargs)
+        super(CouchbaseBackend, self).__init__(expires_type=int, *args,
+                                               **kwargs)
         self.url = url
 
         if Couchbase is None:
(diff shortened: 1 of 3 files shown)


georgepsarakis:
@dfresh613 thanks for your contribution! Let me know when this is ready for review.

georgepsarakis:
As a side-note, Couchbase should be also included in the note regarding result backend expiration [here](http://docs.celeryproject.org/en/latest/userguide/configuration.html#result-expires).

celery/backends/couchbase.py
```diff
@@ -39,6 +39,8 @@ class CouchbaseBackend(KeyValueStoreBackend):
     username = None
     password = None
     quiet = False
+    supports_autoexpire = True
```
georgepsarakis: Since expiration time should be in seconds, I think you should also pass `expires_type=int` to the constructor see the [Redis backend](https://github.com/celery/celery/blob/0a78c6a092c46d013d30d1a2719e4cbd65f28c31/celery/backends/redis.py#L157).

dfresh613:
Hi, I've made the requested changes to set expires_type and also updated the documentation.

Are there any unit tests I can add for this? As the expiration is enforced from couchbase, i'm not sure if unit tests are needed.

celery/backends/couchbase.py
```diff
@@ -39,13 +39,16 @@ class CouchbaseBackend(KeyValueStoreBackend):
...
+
     timeout = 2.5
 
     # Use str as couchbase key not bytes
     key_t = str_t
 
     def __init__(self, url=None, *args, **kwargs):
-        super(CouchbaseBackend, self).__init__(*args, **kwargs)
+        super(CouchbaseBackend, self).__init__(expires_type=int, *args,
```
georgepsarakis: Perhaps a better form would be to add this option in the keyworded arguments if it does not exist:

```
kwargs.setdefault('expires_type', int)
```

celery/backends/couchbase.py
```diff
@@ -103,7 +106,7 @@ def get(self, key):
             return None
 
     def set(self, key, value):
-        self.connection.set(key, value)
+        self.connection.set(key, value, ttl=self.expires)
```
georgepsarakis: A unit test could check that the proper expiration TTL (or `None`) is passed correctly.

dfresh613: Thanks for the suggestions! I've updated my PR with unit tests for the expires property (much of it was based off of the redis backend unit tests). Please let me know if there's anything else should change

georgepsarakis: @dfresh613 well done. Just try and fix the Travis CI issues and this should be ready for merge.

PR #4754, opened 2018-05-22
Fix gevent import

lepture:
`GreenletExit` is not in `__all__` in greenlet.py which can not be imported by Python 3.6.


lepture:
@auvipy I think this is the only one.

Issue #4387, opened 2017-11-14
Release version 4.2.0


georgepsarakis:
I think that it would be a good idea to release a new version that includes all bug fixes so far. @thedrow @auvipy thoughts?

I can help with the changelog.

auvipy:
I am always in favor of releasing smaller bug fixe release as early as possible. I prefer django style release

thedrow:
There is at least one new feature in master https://github.com/celery/celery/commit/0d5b840af1890a9a499a339aa3256445b43837dc
We'll need to cherry-pick all the bug fixes.

thedrow:
I really want to release a new version with https://github.com/celery/celery/pull/4358 but it needs test coverage.

georgepsarakis:
@thedrow I would suggest that we avoid the complexity of a release branch. Let's simply go to the next minor version (4.2.0), in order to be compliant with semantic versioning. What do you think?

thedrow:
Yeh let's do that.

auvipy:
considering the low resource and time we have, that should be the best fit

Fokko:
Any estimate when 4.2 will be released?

johnarnold:
Any updates on 4.2?  The time / timezone / eta bugfixes seem critical...

thedrow:
There is one PR that we have to merge which causes memory leaks when using the Redis results backend.
Check [the milestone ](https://github.com/celery/celery/milestone/19) is a for progress. We're almost there.

thedrow:
FYI, I'm getting married at the 28th of December so I don't think we'll be able to release before that.
I'd like us to get a release out the door by the 7th of January.
@georgepsarakis @auvipy Please prepare a rolling PR with the release notes for pyamqp, kombu, and celery before that.

thedrow:
@Fokko @johnarnold If you'd like to help we still have [4 pull requests which require test coverage](
https://github.com/celery/celery/pulls?q=is%3Apr+is%3Aopen+label%3A%22Status%3A+Needs+Test+Coverage+%E2%9C%98%22+milestone%3Av4.2) and/or rebasing before we release 4.2.

AlexHill:
I've just added #4369 to the 4.2 milestone as I think I'll be able to get to it in the next week or so.

Skeen:
Milestone 4.2 is passed it's due date and several issues are still open (33/98 issues are still open (33%)).
I'm new to the project and therefore have a question regarding this.

[...]

PR #4770, opened 2018-05-27
Minor documentation tweaks for broken links

juannyG:
Some documentation hyperlinks are not generating correctly.

Here's what I was able to find and update:
* The `app` parameter in [celery.schedules documentation](http://docs.celeryproject.org/en/master/reference/celery.schedules.html#celery-schedules) links to ~@Celery, which in the browser attempts to open an email client.
* A `datetime` reference in [celery.schedules `nowfun` parameter description](http://docs.celeryproject.org/en/master/reference/celery.schedules.html#celery-schedules)
* The [tasks userguide](http://docs.celeryproject.org/en/master/userguide/tasks.html#automatic-retry-for-known-exceptions) was not generating the desired hyperlink to the `Celery.task` method.
* The [Celery.send_task method's `result_cls` parameter type](http://docs.celeryproject.org/en/master/reference/celery.html#celery.Celery.send_task) generated ~@AsyncResult, which also attempted to open an email client.

### DIFF
--- a/celery/app/base.py
+++ b/celery/app/base.py
@@ -699,7 +699,7 @@ class Celery(object):
 
         Arguments:
             name (str): Name of task to call (e.g., `"tasks.add"`).
-            result_cls (~@AsyncResult): Specify custom result class.
+            result_cls (AsyncResult): Specify custom result class.
         """
         parent = have_parent = None
         amqp = self.amqp
--- a/celery/app/task.py
+++ b/celery/app/task.py
@@ -471,10 +471,10 @@ class Task(object):
                 :func:`kombu.compression.register`.
                 Defaults to the :setting:`task_compression` setting.
 
-            link (~@Signature): A single, or a list of tasks signatures
+            link (Signature): A single, or a list of tasks signatures
                 to apply if the task returns successfully.
 
-            link_error (~@Signature): A single, or a list of task signatures
+            link_error (Signature): A single, or a list of task signatures
                 to apply if an error occurs while executing the task.
 
             producer (kombu.Producer): custom producer to use when publishing
--- a/celery/bin/base.py
+++ b/celery/bin/base.py
@@ -144,7 +144,7 @@ class Command(object):
     """Base class for command-line applications.
 
     Arguments:
-        app (~@Celery): The app to use.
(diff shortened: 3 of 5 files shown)


juannyG:
Any guidance regarding these automated checks would be greatly appreciated. Although, they're very surprising, given that my updates are all comment/doc related :thinking:

georgepsarakis:
@juannyG thanks for your contribution. The build errors are unrelated to the PR, we need to look more to find what is causing them.

juannyG:
@georgepsarakis - after doing some digging, I believe the 2.7 integration tests are failing due to the recent release of pyamqp==2.3.0 - more specifically the change of `unicode` to `string_t` @ https://github.com/celery/py-amqp/pull/182/files#diff-e78ab3d38dc527a2efca0e44ffe7c2f9L88 . Seems like this may not be isolated to the Celery test suite: https://github.com/celery/py-amqp/issues/191

I went into the tox `2.7-integration-rabbitmq` path, forced a pip install of pyamqp==2.2.2 in that environment and the tests started running and also passing. The `str_to_bytes` method used to transform the args of the frame being written in pyamqp originally did a `unicode` type check for python2. The change to do a type check of `string_t` seems to be the source of errors. Seems like the path of least resistance would be to pin the pyamqp dependency to 2.2.2 (for the time being at least...)?

Should I open an issue here with these details? Or pyamqp? Both? :-)

That's as far as I got today...I'll try to look more into this tomorrow.

georgepsarakis:
@juannyG thanks a lot for troubleshooting this. @thedrow I think the `pyamqp` 2.3.0 release was yesterday indeed?

hsiaoyi0504:
@georgepsarakis I can confirm that **pyamqp** was updated yesterday. https://pypi.org/project/amqp/#history

juannyG:
:+1: Looks like [pyamqp 2.3.1 is out](https://github.com/celery/celery/issues/4387#issuecomment-392528468)

Y'all beat me too it - I was about to put up a PR in pyamqp and saw @thedrow had already merged one! <3

thedrow:
The build passes. Thank you very much.

Issue #4768, opened 2018-05-26
In 4.2.0rc4, executing a group within a chain results in a TypeError


badloop:
### Celery Report Output
```sh
software -> celery:4.2.0rc4 (windowlicker) kombu:4.2.0 py:3.6.5
            billiard:3.5.0.3 py-amqp:2.2.2
platform -> system:Darwin arch:64bit imp:CPython
loader   -> celery.loaders.app.AppLoader
settings -> transport:amqp results:redis://127.0.0.1:6379/0

broker_url: 'amqp://guest:********@127.0.0.1:5672//'
result_backend: 'redis://127.0.0.1:6379/0'
```
## Redis Server version
`Redis server v=4.0.9 sha=00000000:0 malloc=jemalloc-4.0.3 bits=64 build=22c6ee099916a58b`

## RabbitMQ Version
`RABBITMQ_DEBIAN_VERSION=3.7.4-1`

## Steps to reproduce
Create two tasks, execute task 1 in a chain and then execute task 2 twice in a group in the chain, i.e. - 
```
( task1 | group( task2, task2) )
## Expected behavior
All tasks execute successfully
## Actual behavior
```

```sh
Exception raised outside body: EncodeError(TypeError("Object of type 'TypeError' is not JSON serializable",),):
Traceback (most recent call last):
  File "/Users/aaron/IdeaProjects/sdn-conductor-worker/venv/lib/python3.6/site-packages/kombu/serialization.py", line 50, in _reraise_errors
    yield
  File "/Users/aaron/IdeaProjects/sdn-conductor-worker/venv/lib/python3.6/site-packages/kombu/serialization.py", line 221, in dumps
    payload = encoder(data)
  File "/Users/aaron/IdeaProjects/sdn-conductor-worker/venv/lib/python3.6/site-packages/kombu/utils/json.py", line 69, in dumps
    **dict(default_kwargs, **kwargs))
  File "/usr/local/Cellar/python/3.6.5/Frameworks/Python.framework/Versions/3.6/lib/python3.6/json/__init__.py", line 238, in dumps
    **kw).encode(obj)
  File "/usr/local/Cellar/python/3.6.5/Frameworks/Python.framework/Versions/3.6/lib/python3.6/json/encoder.py", line 199, in encode
    chunks = self.iterencode(o, _one_shot=True)
[...]

PR #4790, opened 2018-06-01
Fixes issue 4768.

xirdneh:
PR #4106 added a GroupResult's parent when converting the object to a tuple for easier access.
The parent was being returned as is. The parent needed to be returned as a tuple.
This way the entire result will be serializable.
Fixes issue #4768 by constructing the first value in a `GroupResult`'s `as_tuple` return value the same way that `AsyncResult` does it.

### DIFF
--- a/celery/result.py
+++ b/celery/result.py
@@ -918,7 +918,10 @@ class GroupResult(ResultSet):
                                          ', '.join(r.id for r in self.results))
 
     def as_tuple(self):
-        return (self.id, self.parent), [r.as_tuple() for r in self.results]
+        return (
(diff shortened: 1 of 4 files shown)


xirdneh:
Ah, forgot to run the rest of the tests locally.
Fixing failing tests...

xirdneh:
While testing this I found out about another issue.
If one tries to build a graph with the result of the task that creates the group, it raises an exception because `GroupResult` is not hashable.
I'll open another issue and work on it on Monday.
Don't know if that would be another blocker for 4.2
Hopefully it won't.

thedrow:
Great job.
I'm going to merge this as this looks exactly like the right fix.

Issue #4081, opened 2017-06-08
Unable to run tasks under Windows


SPKorhonen:
Celery 4.x starts (with fixes #4078) but all tasks crash

## Steps to reproduce

Use **First Steps** tutorial (http://docs.celeryproject.org/en/latest/getting-started/first-steps-with-celery.html) 

celery -A tasks worker --loglevel=info
add.delay(2,2)

## Expected behavior

Task is executed and a result of 4 is produced

## Actual behavior

Celery crashes.

"C:\Program Files\Python36\Scripts\celery.exe" -A perse.celery worker -l info

 -------------- celery@PETRUS v4.0.2 (latentcall)
---- **** -----
--- * ***  * -- Windows-10-10.0.14393-SP0 2017-06-08 15:31:22
-- * - **** ---
- ** ---------- [config]
- ** ---------- .> app:         perse:0x24eecc088d0
- ** ---------- .> transport:   amqp://guest:**@localhost:5672//
- ** ---------- .> results:     rpc://
- *** --- * --- .> concurrency: 12 (prefork)
-- ******* ---- .> task events: OFF (enable -E to monitor tasks in this worker)
--- ***** -----
 -------------- [queues]
                .> celery           exchange=celery(direct) key=celery

[tasks]
  . perse.tasks.celery_add

[2017-06-08 15:31:22,685: INFO/MainProcess] Connected to amqp://guest:**@127.0.0.1:5672//
[2017-06-08 15:31:22,703: INFO/MainProcess] mingle: searching for neighbors
[2017-06-08 15:31:23,202: INFO/SpawnPoolWorker-5] child process 5124 calling self.run()
[2017-06-08 15:31:23,207: INFO/SpawnPoolWorker-4] child process 10848 calling self.run()
[2017-06-08 15:31:23,208: INFO/SpawnPoolWorker-10] child process 5296 calling self.run()
[2017-06-08 15:31:23,214: INFO/SpawnPoolWorker-1] child process 5752 calling self.run()
[2017-06-08 15:31:23,218: INFO/SpawnPoolWorker-3] child process 11868 calling self.run()
[2017-06-08 15:31:23,226: INFO/SpawnPoolWorker-11] child process 9544 calling self.run()
[2017-06-08 15:31:23,227: INFO/SpawnPoolWorker-6] child process 16332 calling self.run()
[2017-06-08 15:31:23,229: INFO/SpawnPoolWorker-8] child process 3384 calling self.run()
[...]

PR #4802, opened 2018-06-07
Clarification for Windows Instructions

W7PEA:
The guide as-is makes it appear as if you can just run setup and follow along with the demo. That is _not_ true on Windows. see #4081 for example. Hopefully this additional link will help the community get through the tutorial without hitting `Task handler raised error: ValueError('not enough values to unpack (expected 3, got 0)',)`

## Description

that version series.


W7PEA:
Hmmm... FORKED_BY_MULTIPROCESSING is not mentioned anywhere in the RabbitMQ docs....

Cancelling this till I sort it out.

PR #4803, opened 2018-06-07
Windows Not Supported

W7PEA:
Its very misleading to say you can run RabbitMQ on Windows and then also at the same time not support Celery on Windows. This doc change helps the community head down the correct path in the beginning as opposed to wasting time trying things that won't work (liek I did).

## Description

that version series.

### DIFF
--- a/docs/getting-started/first-steps-with-celery.rst
+++ b/docs/getting-started/first-steps-with-celery.rst
@@ -30,6 +30,8 @@ it's a good idea to browse the rest of the documentation.
 For example the :ref:`next-steps` tutorial will
 showcase Celery's capabilities.
 
+NOTE: To save you time, please be aware As of 4.0 `Celery does not support Windows <http://docs.celeryproject.org/en/latest/whatsnew-4.0.html#removed-features>`_.
+
 .. contents::
     :local:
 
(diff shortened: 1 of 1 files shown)


auvipy:
support was restored in 4.1

thedrow:
We do not guarantee that Celery will work under Windows. The unit tests run and we accept pull requests to provide support but we do not develop Celery with Windows in mind.
Celery 4.0 was broken on Windows as far as I can recall but we fixed that.
If you want to help maintain Celery on Windows, we'll welcome it of course.

You can run RabbitMQ on Windows. If you can't, that's not something we can handle. That's something for RabbitMQ's developers to handle.
If you need to do so, run RabbitMQ on Windows and Celery on Linux and everything should work.

thedrow:
@W7PEA Ping?

W7PEA:
I'm going to close this. Thanks for the notes. I decided to go another direction.

Along the journey, I re-rememembered the Windows Subsystem for Linux exists now. Not yet fully supported but it should be a path forward for devs on Windows with technologies that need a Linux command not available on Windows (in this case Fork). 
WSL: https://docs.microsoft.com/en-us/windows/wsl/about
System Commands: https://blogs.msdn.microsoft.com/wsl/2016/06/08/wsl-system-calls/

Issue #4758, opened 2018-05-22
Celery worker state changes to offline after a while


maxdebayser:
## Checklist

- [x] I have included the output of ``celery -A proj report`` in the issue.

```#>celery -A tasks report

software -> celery:4.1.0 (latentcall) kombu:4.1.0 py:3.5.2
            billiard:3.5.0.3 py-amqp:2.2.2
platform -> system:Linux arch:64bit imp:CPython
loader   -> celery.loaders.app.AppLoader
settings -> transport:amqp results:redis://redis-prod/0

result_backend: 'redis://redis-prod/0'
task_track_started: True
task_acks_late: True
task_routes: {
 '*': {'queue': 'queue1'}}
worker_send_task_events: True
result_serializer: 'json'
broker_connection_max_retries: 0
broker_heartbeat: 20
worker_pool_restarts: True
task_serializer: 'json'
worker_redirect_stdouts_level: 'INFO'
accept_content: ['json', 'pickle']
task_queue_max_priority: 10
broker_url: 'amqp://guest:********@rabbitmq-prod:5672//'
```

- [x] I have verified that the issue exists against the `master` branch of Celery.

No, this issue is happening in release 4.1.0

## Steps to reproduce

Not sure, it happens randomly.

## Expected behavior

Workers should stay online in flower unless there is a problem. If there really was a persistent problem I would expect commands like 
```
celery -A tasks control heartbeat
celery -A tasks status
```

to show the workers as offline as well.

## Actual behavior

After several hours running, some workers appear as offline in flower and `celery -A proj events --dump` records no heartbeats coming from these workers. At the same time 
```
celery -A tasks control heartbeat
celery -A tasks status
```
show all workers as online. 

When I restart RabbitMQ with 
```
rabbitmqctl stop_app && rabbitmqctl start_app
```
the workers return to the online state and at the same time, the other worker nodes log messages complaining about a "Substantial drift" of several hours in relation to the clocks of the nodes that were offline. So it seems that somewhere the missing heartbeats were bottled up.

Another way to make them online again is to delete the celeryev.* queues in RabbitMQ.

This problem affect both workers with the fork pool and the gevent pool.

## Discussion

[...]

PR #4804, opened 2018-06-08
Activate Kombu publish retry in Heartbeat

maxdebayser:
The lack of this retry option lead to a curious bug where the hearbeat connection would be lost if the connection was block for a smaller amount of time than the heartbeat interval

https://github.com/celery/celery/issues/4758

### DIFF
--- a/celery/worker/heartbeat.py
+++ b/celery/worker/heartbeat.py
@@ -46,6 +46,7 @@ class Heart(object):
                                  active=len(active_requests),
                                  processed=all_total_count[0],
                                  loadavg=load_average(),
+                                 retry=True,
                                  **SOFTWARE_INFO)
 
     def start(self):


maxdebayser:
Perhaps, if someone could give me some pointers. Basically what I've been doing is a black-box test discovered by @zwg-zero:

0) I set up rabbitmq server, a flower and a dummy worker with a celeryconfig.py with 10s heartbeat inerval.
1) I check the number of connections from the worker to rabbitmq, it should be 3:
```
#> ss -antp | grep celery
ESTAB      0      0      127.0.0.1:38754              127.0.0.1:5672                users:(("celery",pid=13378,fd=17))
ESTAB      0      0      127.0.0.1:38762              127.0.0.1:5672                users:(("celery",pid=13378,fd=19))
ESTAB      0      0      127.0.0.1:38758              127.0.0.1:5672                users:(("celery",pid=13378,fd=18))
```
2) I interrupt the packet flow for less than the heartbeat interval:
```
#> iptables -I INPUT -p tcp --dport 5672 -j DROP && sleep 5s && iptables -D INPUT 1
```
3) I check the number of connections from the worker to rabbitmq, it should still be 3, but with the bug on will be lost:
```
# ss -antp | grep celery
ESTAB      0      0      127.0.0.1:38754              127.0.0.1:5672                users:(("celery",pid=13378,fd=17))
ESTAB      0      0      127.0.0.1:38762              127.0.0.1:5672                users:(("celery",pid=13378,fd=19))
```


With the fix, the behavior is different:

```
#> ss -antp | grep celery
ESTAB      0      0      127.0.0.1:40502              127.0.0.1:5672                users:(("celery",pid=25098,fd=18))
ESTAB      0      0      127.0.0.1:40506              127.0.0.1:5672                users:(("celery",pid=25098,fd=19))
ESTAB      0      0      127.0.0.1:40498              127.0.0.1:5672                users:(("celery",pid=25098,fd=17))
#> iptables -I INPUT -p tcp --dport 5672 -j DROP && sleep 5s && iptables -D INPUT 1
#> ss -antp | grep celery
ESTAB      0      0      127.0.0.1:40506              127.0.0.1:5672                users:(("celery",pid=25098,fd=19))
ESTAB      0      0      127.0.0.1:40514              127.0.0.1:5672                users:(("celery",pid=25098,fd=22))
ESTAB      0      0      127.0.0.1:40498              127.0.0.1:5672                users:(("celery",pid=25098,fd=17))
```

The connection is lost, but a new one is opened, as can be seen by the change in file descriptor number.

I don't know how easy it would be to automate this test since it needs administrator priviledges to mess with iptables. Perhaps a white-box test that
mocks some of the lower level stuff would be better, but my knowledge of the code base is too limited to know exacly what needs to be done.

(For the record, the failed AppVeyor test has nothing to do with this path. The stack trace shows that it breaks somewhere elese)

auvipy:
ignore appveyor

auvipy:
@xirdneh do you have a suggestion for testing the change?

asfaltboy:
Anyone know why this was reverted? I have a similar issue (though using redis transport, celery version 4.2.1); in my case network issues on host cause workers to go offline, and they do not reconnect after network issue is resolved, which requires us to restart the worker.

If this is no longer an issue, how can we force celery to retry connecting to broker after such failures?

thedrow:
It causes a problem for other users. I can't remember which issue reported that.
Man I wish I referenced this in the commit message.

maxdebayser:
Maybe this should become a configuration option then.

Issue #4810, opened 2018-06-12
Autodiscovery fails for empty "related_name"


slint:
## Checklist

The issue is affecting `v4.2.0` and current `master` branch.

## Steps to reproduce

In previous versions doing something like:

```python
celery.autodiscover_tasks(['my_package.tasks'], related_name='')
```

was fine and loaded all of the tasks in `my_package/tasks.py` ([the `find_related_module` function was returning here](https://github.com/celery/celery/blob/71d024446bbe3e0c9cdffe63fda2c83db42c7fc3/celery/loaders/base.py#L262-L265)).

The same call with `v4.2.0` returns a `KeyError: my_package.tasks.'` (notice the trailing dot `.` at the end of the package), since it tries to build and import an invalid python package path [here](https://github.com/celery/celery/blob/9778f2bfe0bf0c4f3d554bdcb479ad769a349fc2/celery/loaders/base.py#L263).

I'm not sure if this is/was the intended behaviour, though it kind of makes sense to be able to specify any kind of import path without having to always conform to a specific suffix (i.e. `.tasks`), we have use-cases where packages organize tasks like this:

```
my_package.tasks.models
my_package.tasks.api
my_package.tasks.foobar
```

Maybe the intended behaviour was to also return [at this point](https://github.com/celery/celery/blob/9778f2bfe0bf0c4f3d554bdcb479ad769a349fc2/celery/loaders/base.py#L256).

lnielsen:
This seems to boil down to a Python 2.7 vs 3.x issue. 

[``importlib.import_module()``](https://github.com/celery/celery/blob/4.2/celery/loaders/base.py#L262-L265) throws a ``KeyError`` in Python 2.7 and an ``ImportError`` in Python 3. The Celery function ``find_related_module()`` only catches the ``ImportError`` so probably just need to also catch the ``KeyError`` to be Python 2.7 compatible.

auvipy:
could you send a pr?

slint:
I've already opened one here: https://github.com/celery/celery/pull/4813

PR #4813, opened 2018-06-13
Allow passing "related_name=None" for autodiscovery (#4810)

slint:
* Passing "None" as the "related_name" in "Celery.autodiscover_tasks"
  will just try to import the plain package.

* Removes usage of the "imp" module from tests.

* Adds extra tests for the "find_related_module" function.

### DIFF
--- a/celery/app/base.py
+++ b/celery/app/base.py
@@ -655,7 +655,8 @@ class Celery(object):
                 value returned is used (for lazy evaluation).
             related_name (str): The name of the module to find.  Defaults
                 to "tasks": meaning "look for 'module.tasks' for every
-                module in ``packages``."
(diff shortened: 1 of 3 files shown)


auvipy:
this qualify as 4.2.1?

slint:
Technically it's a "bugfix" for our project, since `4.2.0` broke calls like

```python
celery.autodiscover_tasks(['foo.tasks.api', 'bar.tasks', 'baz.some'], related_name='')
```

that were working for all previous versions.

So for us it would be awesome if it qualifies as `4.2.1`, in order to avoid pinning via `celery<4.2.0`

auvipy:
this shoud be in 4.2.1

drjova:
@thedrow @auvipy any chance to release a new version which include this?

PR #4825, opened 2018-06-19
Improve celery upgrade error handling

federicobond:


federicobond:
I wasn’t sure it merited a changelog entry, but I can add one. It’s just making sure that the program does not crash with missing arguments.

auvipy:
still it is an improvement

PR #4837, opened 2018-06-21
Add logic around when to requeue the requested message when using the RPC backend.

### LINKED ISSUE #4830: RPC backend leaks memory / messages in some situations
I'm still working on a reduced test-case I can share, but so far it seems that if you are using the RPC backend and call `state` (or other methods that attempt to get the current status of a result) then the RPC backend will leak memory.

The core of the issue seems to be that if you call `state` on the `AsyncResult` it queries out the RPC backend to get the status, and caches the result (if the status is `SUCCESS`), once the result is cached then calling `get()` on the `AsyncResult` does nothing since it just uses the cached result. BUT, while calling to get the current state the RPC backend requeues the message. This has two bad side-effects:
1. It ends up in the queue (until it expires?)
2. It ends up in the `_out_of_band` dictionary on the *next* message, but will never be removed from there.

This code won't fully run, but might illustrate what is happening:

```python
from time import sleep
from celery import states
from my_app import app

result = app.send_task('my_task')

# Loop twice.
for i in range(2):
    print("RUNNING", i)
    result = app.send_task('my_task')

    while result.state != states.SUCCESS:
        sleep(1)

    # This should be empty, but isn't.
    print(app.backend._out_of_band)
    print(app.backend.result_consumer._pending_messages)

    # Calling get should clean these up since you have to call get/forget.
    result.get()

    # Still not empty
    print(app.backend._out_of_band)
    print(app.backend.result_consumer._pending_messages)
```

Similar code just using `get()` seems to work OK:
```python
from time import sleep
from celery import states
from my_app import app

result = app.send_task('my_task')

# Loop twice.
for i in range(2):
    print("RUNNING", i)
    result = app.send_task('my_task')
    result.get()

    # This should be empty, but isn't.
    print(app.backend._out_of_band)
    print(app.backend.result_consumer._pending_messages)

    # Calling get should clean these up since you have to call get/forget.
    result.get()

    # Still not empty
    print(app.backend._out_of_band)
    print(app.backend.result_consumer._pending_messages)
```

I believe what is happening is that the `requeue` calling being made while trying to `get_meta_data` should only happen if the task is not in a "done" state, see https://github.com/celery/celery/blob/120770929f4a37c5373a378b75b5c41a99702af9/celery/backends/rpc.py#L257-L259

This is because (according [to the docs](http://docs.celeryproject.org/en/latest/userguide/tasks.html#rpc-result-backend-rabbitmq-qpid):

> The RPC result backend (rpc://) is special as it doesn’t actually store the states, but rather sends them as messages. This is an important difference as it means that a result *can only be retrieved once, and only by the client that initiated the task*. Two different processes can’t wait for the same result.

If you're curious why I'm not just using `get()` it is because I don't want to actually block on waiting for tasks, I'm keeping track of many sent tasks and periodically checking them for their status. I could potentially replace this with `get()` with a short timeout, but that doesn't seem like the right solution.

clokep:
## Description

This fixes the memory leak described in #4830. The leak happens because `RPCBackend.get_task_meta()` always requeues the received message (if it matches the task ID you're looking for). The message is always requeued, but the result is also sometimes cached onto the `AsyncResult` object, this means that calling `AsyncResult.get()` to "clean up" the object (as [required in the documentation](http://docs.celeryproject.org/en/latest/userguide/tasks.html#result-backends)) doesn't actually reach out to the backend to remove the message since it is cached already.

This avoids requeueing the message in two cases:
* If the task is "done" (i.e. one of the `READY_STATES` has been reached).
* If the current status of the task is already being cached (when using `result_cache_max`).

Note that calling `AsyncResult.get()` ends up following a different code-path that doesn't call into `get_task_meta()`, but manually polling for the `AsyncResult.state` shows this bug. See the sample code in the issue.

I tested this with all combinations of `result_cache_max` and `task_track_started` (which interacts with this code in a subtle way when the null-cache is used) and I got correct results in each of them.

Fixes #4830

### DIFF
--- a/celery/backends/rpc.py
+++ b/celery/backends/rpc.py
@@ -267,7 +267,8 @@ class RPCBackend(base.Backend, AsyncBackendMixin):
             # null-cache, see :setting:`result_cache_max`) there's also no
             # need to requeue it since the next query will just return what is
             # in the cache.
-            if result['status'] not in states.READY_STATES and task_id not in self._cache:
+            if (result['status'] not in states.READY_STATES and
+                    task_id not in self._cache):
                 latest.requeue()
 
             return result
--- a/celery/backends/rpc.py
+++ b/celery/backends/rpc.py
@@ -255,8 +255,22 @@ class RPCBackend(base.Backend, AsyncBackendMixin):
             self.on_out_of_band_result(tid, msg)
 
         if latest:
-            latest.requeue()
-            return self._set_cache_by_message(task_id, latest)
+            result = self._set_cache_by_message(task_id, latest)
+
+            # Choose whether to requeue the message.
+            #
+            # If the result has reached its end state there's no reason to
+            # requeue (it won't change again and the RPC backend assumes you
+            # can only receive each result once).
+            #
+            # Otherwise, if the result was cached (e.g. if not using the
+            # null-cache, see :setting:`result_cache_max`) there's also no
+            # need to requeue it since the next query will just return what is
+            # in the cache.
+            if result['status'] not in states.READY_STATES and task_id not in self._cache:
+                latest.requeue()
+
+            return result
         else:
             # no new state, use previous
             try:


clokep:
Please let me know how this looks or any other feedback. I'm motivated to fix this! 👍

clokep:
I fixed the `flake8` error, not sure what was going on with the Redis test. Locally that also fails on `master` for me.

clokep:
The failed travis builds seem like infrastructure issues. The appveyor one is still failing that redis test.

clokep:
@auvipy / @thedrow Yes, I can add a test for it. Any idea where an appropriate unit test for this would go? Is `t/unit/backends/test_rpc.py` the appropriate place? I didn't see any integration tests dealing with backends so wasn't sure how to create one, but maybe I was looking in the wrong place, could you point me in the right direction here? Thanks.

thedrow:
`t/unit/backends/test_rpc.py` is indeed the right place.
The right place for backend integration tests is `t/integration/test_tasks.py`.

clokep:
Thanks! 👍 I'll try to take a look at those and see if I can expand some test-cases to handle this case!

clokep:
Just a quick update for this -- @austintrose and @tmwbook are doing some further testing of this, we'll likely deploy this to our production environment to see if it fixes our memory leak.

clokep:
This is still on my mind. 😄 I was able to fix a couple of other memory leaks (unrelated to Celery) + this and our memory usage has been pretty flat now for a couple of weeks. I'll try to find some time soon to work on the unit/integration test for this.

clokep:
@thedrow / @auvipy Do either of you know the intent of the *design of the RPC backend*? I'm frankly struggling on why you would ever requeue these messages, but maybe I'm making assumptions about my use of Celery.

thedrow:
See https://github.com/celery/celery/blob/master/docs/history/whatsnew-3.1.rst#new-rpc-result-backend regarding the intent if the RPC backend.

celery/backends/rpc.py
```diff
@@ -255,8 +255,23 @@ def get_task_meta(self, task_id, backlog_limit=1000):
...
+            #
+            # If the result has reached its end state there's no reason to
+            # requeue (it won't change again and the RPC backend assumes you
+            # can only receive each result once).
+            #
+            # Otherwise, if the result was cached (e.g. if not using the
+            # null-cache, see :setting:`result_cache_max`) there's also no
+            # need to requeue it since the next query will just return what is
+            # in the cache.
+            if (result['status'] not in states.READY_STATES and
```
thedrow: Just a suggestion: We can extract this to a helper function to make the code more readable.
The comment can be a docstring if needed.

auvipy:
can we draw a final line of this pr before 4.3 release?

clokep:
@auvipy I probably won't get to finishing this up before the 4.3 release. I think this PR is still missing something as we *sometimes* seem to still be requeueing in our production environment.

clokep:
After running with this for a while, this doesn't seem to fully solve the issue. I'm unsure whether to continue pursuing this as a potential solution or if I'm misunderstanding the design of the RPC backend.

thedrow:
Can you provide more details?

clokep:
I'm no longer working with RabbitMQ. I still think that there's some sort of logic bug here, but not 100% sure if I'm understanding the design of the RPC backend properly.

PR #4839, opened 2018-06-22
Fix memory leak of AsyncResult 

monsterxx03:
#4131 fixed a memory leak issue by remove weakref to a bound method. But when I upgraded my prod celery to 4.2.0, I come to this memory leak issue again. The test script in the issue also report memory leak in 4.2.0 :   https://github.com/celery/celery/pull/4131#issuecomment-346949954

The root case is `AsyncResult` defined  `__del__` method, but the `self.on_ready` created a ref cycle as well. `__del__` will prevent gc to collect ref cycle objects.

A simple script to test:

```
import resource
import gc

from celery import Celery

app = Celery()
app.conf.update(BROKER_URL='redis://localhost:6379/1')

@app.task
def dummy():
    return '1'


def print_mem():
    print 'Memory usage: %s (kb)' % resource.getrusage(resource.RUSAGE_SELF).ru_maxrss

def run():
    for i in range(10000):
        dummy.delay()
        if i % 1000 == 0:
            print_mem()

run()
print gc.garbag
```

Before remove `__del__`:

```
Memory usage: 31736 (kb)
Memory usage: 32916 (kb)
Memory usage: 34628 (kb)
Memory usage: 36432 (kb)
Memory usage: 38292 (kb)
Memory usage: 39812 (kb)
Memory usage: 41668 (kb)
Memory usage: 42608 (kb)
Memory usage: 43576 (kb)
Memory usage: 44592 (kb)

[<AsyncResult: 9ad49b99-73e3-41de-a972-3ab196874f91>, <AsyncResult: 7638fc5a-6c14-4fe5-bbe7-1d19ab992582>, <AsyncResult: 95a158b6-9c1b-408e-9364-679b01e3d36f>, <AsyncResult: 6bd7cac9-4a65-47ae-97ee-8a0bb0b24cee>.......]
```
After remove `__del__`:

```
Memory usage: 31828 (kb)
Memory usage: 31960 (kb)
Memory usage: 31960 (kb)
Memory usage: 31960 (kb)
Memory usage: 31960 (kb)
Memory usage: 31960 (kb)
Memory usage: 31960 (kb)
Memory usage: 31960 (kb)
Memory usage: 32188 (kb)
Memory usage: 32188 (kb)

[]
```

### DIFF
--- a/celery/result.py
+++ b/celery/result.py
@@ -379,11 +379,6 @@ class AsyncResult(ResultBase):
     def __reduce_args__(self):
         return self.id, self.backend, None, None, self.parent
 
-    def __del__(self):
-        """Cancel pending operations when the instance is destroyed."""
-        if self.backend is not None:
-            self.backend.remove_pending_result(self)
-
     @cached_property
     def graph(self):
         return self.build_graph()


monsterxx03:
https://docs.python.org/2/library/gc.html#gc.garbage

georgepsarakis:
@monsterxx03 did you by any chance find how the reference cycle is created in the first place?

monsterxx03:
The problem is here: https://github.com/celery/celery/blob/4.2/celery/result.py#L102

AsyncResult -> self.on_ready -> promise -> self._on_fulfilled

since _on_fulfilfilled is an instance method of AsyncResult, or ref back to AsyncResult, the cycle complete.

At first, I try to set `weak=True` for  `promise(self._on_fulfilled)`, actually, it works, the memory leak issue disappear, but later, I found this pr #4131 , weak ref not work on bound method, I think it makes sense, what really need is [WeakMethod](https://docs.python.org/3/library/weakref.html#weakref.WeakMethod), but it's only available after python3.4. And I agree the point in https://github.com/celery/celery/pull/4131#issuecomment-346949954  ref cycle is not a problem, python gc can handle it. BTW, the problem in 4131 is the func passed to promise never called if `weak=True`

Then I'm confusing why memory leak still happen(I thought gc can handle ref cycle), then I checked history of celery/result.py,  found the `__del__` method is introduced in 2636251a1249634258c1910c6400bf2ccf28b8ed ,  okay, ref cycle is not a problem, unless we didn't define a `__del__` for it, it will cause object uncollectable  https://docs.python.org/2/library/gc.html#gc.garbage

To clean up:

AsyncResult has a ref cycle, and define `__del__` which caused the memory leak, and it's 100% reproducible under python2.7.

In my test, add weak=True for promise or remove the `__del__` can both fix the memory leak of `AsyncResult`, but  weak=True will make `_on_fulfilled` never called, so I remove `__del__`.

I think the best way is resolve the ref cycle and keep `__del__`, but I don't see a obvious way so far, my quick fix just solve the memory issue(`__del__` won't be called in current code as well, since it won't be gc)  @georgepsarakis

georgepsarakis:
@monsterxx03 thanks for the detailed explanation.

I think that the best way to handle this is to actually backport [WeakMethod](https://github.com/python/cpython/blob/master/Lib/weakref.py#L36) for Python versions that do not support it. This should probably be introduced in [vine](https://github.com/celery/vine). [promise](https://github.com/celery/vine/blob/master/vine/promises.py#L90) objects should detect whether the passed function is bound, and use `WeakMethod` instead of `ref`.

How does that sound?

monsterxx03:
@georgepsarakis that would be great, the promise's `weak` parameter is easy to be mis-used if developer don't be careful.

BTW I tested this patch in my prod env for one day, the memory leak issue is fixed, at least for me

screen shot 2018-06-23 at 2 39 57 pm

It's a service always adding tasks to celery, calling `apply_async`. The celery worker processes don't have leak issue.

Waiting for a better release with vine upgrade, or merged as a tmp fix is both okay for me, thank you 👍

fernandotakai:
I was having the same issue after upgrading to celery 4.2 (kombu 4.2.1) -- my task producers where exponentially leaking memory while the consumers are quite ok.

After applying this patch, memory usage went back to stable.

image

(those drops on the graph are restarts, while the constant line at the end is after the patch was applied)

JioHooney:
Has this problem been resolved?

thedrow:
I don't think so.
See https://github.com/celery/celery/pull/4839#issuecomment-399633253 and https://github.com/celery/vine/issues/21
None of us have found the time to properly resolve this issue.
@JioHooney PRs are always welcome.

codingEnzo:
@monsterxx03 i think you can try the test unit on py3
i found it's ok on my env (fedora27, python3.6.6, celery4.2.1), with your code

monsterxx03:
@codingEnzo this bug only effect python < 3.4.

When I posting this pr, my prod env is running on python2.7, will 100% come to this memory leak issue. And this PR is a workable workaround.

Recently, I migrate my prod env to python 3.6, celery won't come to memory leak any more, I'm using celery 4.2.1, without this patch.

I'm not sure merge this patch is good for python3.6, seems the behave is hard to test in unit test.

jsh2134:
Can confirm this same memory leak with Celery 4.2.1 and python2.7

georgepsarakis:
We also need to revert the changes in #4131 in order to take advantage of the fix in https://github.com/celery/vine/pull/24 , as mentioned above https://github.com/celery/celery/pull/4839#issuecomment-399624269 .

georgepsarakis:
@thedrow @auvipy @xirdneh any ideas on how to move forward? Please see also my comment above.

georgepsarakis:
Can someone please run some tests with:
- using the latest master of https://github.com/celery/vine
- reverted changes of https://github.com/celery/celery/pull/4131

georgepsarakis:
@austintrose @clokep any feedback on the previous suggestions would be appreciated!

ldav1s:
Experiencing a memory leak that seems to be from Celery 4.2.1.  I'm guessing it could be this, since it sounds very much like what we've been observing.  Here are some particulars of our Celery setup:

python 2.7.12
Ubuntu 16.04.1
celery 4.2.1
amqp 2.3.2
billiard 3.5.0.4
gevent 1.3.7
kombu 3.0.37
librabbitmq 2.0.0
vine 1.1.4
RabbitMQ 3.7.5

I'm going to try @georgepsarakis [request](https://github.com/celery/celery/pull/4839#issuecomment-447739820) on this thread with a slight variation on step 1 -- I'm going to use vine 1.2.0 instead of "the latest master" because it will be easier for us to deploy.  I will revert the changes suggested in the request.

I guess we aren't pumping as many messages through Celery as some in this thread are -- it takes a few days for us to notice problems from the leak instead of hours.  But we should be able to see if this makes a difference in a couple of days or so.

georgepsarakis:
@ldav1s that is great, any feedback is appreciated. At the time of writing, vine 1.2.0 had not been released, and that is why I suggested using the master branch. Tag 1.2.0 should contain all necessary changes: https://github.com/celery/vine/commit/f23ffa4fbee403bcccd44501fd4b362faf0c6f26

ldav1s:
Reverting #4131 didn't unapply cleanly on Celery 4.2.1.  Attached is a patch that should.

I installed vine 1.2.0, and applied the attached patch, which reverts #4131 on Celery 4.2.1.  Things seem to still work, but as I said in my previous comment, it'll take a day or two to see if the leak is really fixed.

[celery-pull-4131-as-applied.patch.txt](https://github.com/celery/celery/files/2799420/celery-pull-4131-as-applied.patch.txt)

ldav1s:
This [attempted fix](https://github.com/celery/celery/pull/4839#issuecomment-457859705) didn't appear to fix the leak.  Still seeing climbing memory at approximately the same slope, AFAICT.  I'll try the [original suggested patch](https://github.com/celery/celery/pull/4839/files) to see if this is indeed this leak.

georgepsarakis:
@ldav1s I assume you are using Redis as a backend, right? Could you perhaps share some code and the configuration that can reproduce the memory leak?

ldav1s:
No, RabbitMQ as a backend, which might be why it doesn't seem to affect things at runtime.

georgepsarakis:
@ldav1s RabbitMQ as a broker then and/or RPC backend? I think you might be observing issue https://github.com/celery/celery/issues/4843

ldav1s:
RabbitMQ as a broker definitely.  The app doesn't get back any results back so I haven't configured a backend to do that piece.  I noticed #4843 in my bug search.  I'm using a gevent pool, not a prefork pool.  I've also searched for broker connection loss/reconnection as described, and I haven't seen that.  I also noticed this [librabbitmq issue](https://github.com/celery/librabbitmq/issues/125), and I'm trying that one now.

monsterxx03:
@georgepsarakis  I did a test with my original script: https://gist.github.com/monsterxx03/3d2daef3d9add57d697d95c51a5266b2

python 2.7.15,  vine 1.2.0, latest celery master d9139a52e58b96b398bd49cca0611b86473fa962

Revert changes in #4131 : 
  
```
diff --git a/celery/result.py b/celery/result.py
index 2f737838c..94ea66d2d 100644
--- a/celery/result.py
+++ b/celery/result.py
@@ -99,7 +99,7 @@ class AsyncResult(ResultBase):
         self.id = id
         self.backend = backend or self.app.backend
         self.parent = parent
-        self.on_ready = promise(self._on_fulfilled)
+        self.on_ready = promise(self._on_fulfilled, weak=True)
         self._cache = None
         self._ignored = False
 
@@ -205,7 +205,7 @@ class AsyncResult(ResultBase):
             assert_will_not_block()
         _on_interval = promise()
         if follow_parents and propagate and self.parent:
-            on_interval = promise(self._maybe_reraise_parent_error)
+            on_interval = promise(self._maybe_reraise_parent_error, weak=True)
             self._maybe_reraise_parent_error()
         if on_interval:
             _on_interval.then(on_interval)
@@ -531,7 +531,7 @@ class ResultSet(ResultBase):
         self.on_ready = promise(args=(self,))
         self._on_full = ready_barrier or barrier(results)
         if self._on_full:
-            self._on_full.then(promise(self._on_ready))
+            self._on_full.then(promise(self._on_ready, weak=True))
 
     def add(self, result):
         """Add :class:`AsyncResult` as a new member of the set.
```

Before revert changes:

```
Memory usage: 29740 (kb)
Memory usage: 31040 (kb)
Memory usage: 32656 (kb)
Memory usage: 34520 (kb)
Memory usage: 36180 (kb)
Memory usage: 38012 (kb)
Memory usage: 39644 (kb)
Memory usage: 41468 (kb)
Memory usage: 43120 (kb)
Memory usage: 44200 (kb)
[<AsyncResult: 30b6832a-1041-49ff-9a50-2e41e40cf2b5>, <AsyncResult: 76c1365e-ba71-4120-b770-936723bd3987>, <AsyncResult: 5a2ed644-1850-4233-9683-aca022c6c107> ......]
```

After revert changes:

```
Memory usage: 29640 (kb)
Memory usage: 29812 (kb)
Memory usage: 29812 (kb)
Memory usage: 29812 (kb)
Memory usage: 29812 (kb)
Memory usage: 29812 (kb)
Memory usage: 29812 (kb)
Memory usage: 29812 (kb)
Memory usage: 29812 (kb)
Memory usage: 29812 (kb)
[]
```

I think the memory leak is fixed. Since my production is running on python3.6, so I can't verify it in production environment. 

May it helpful!

georgepsarakis:
@monsterxx03 great, thanks for the feedback! @thedrow @auvipy I think we should revert https://github.com/celery/celery/pull/4131 before the 4.3 release.

PR #4844, opened 2018-06-23
Drop workaround for older unsupported Pythons

jdufresne:
All supported modern Pythons support `memoryview`. It will never raise a `NameError`.

All supported modern Pythons define `struct.unpack_from`. It will never raise an `ImportError`.


jdufresne:
Nope. Just clean up. If it isn't a good time to accept, no worries.

I noticed it while working on fixing the bytes warning in #4846. It looks like dead code and its purpose could cause confusion to new contributors.

PR #4846, opened 2018-06-23
Fix BytesWarning in concurrency/asynpool.py

jdufresne:
When Python is executed with the `-b` CLI option, Celery issues the following warnings:

```
.../celery/concurrency/asynpool.py:1254: BytesWarning: Comparison between bytes and string
  header = pack(b'>I', size)
.../celery/concurrency/asynpool.py:814: BytesWarning: Comparison between bytes and string
  header = pack(b'>I', body_size)
.../celery/concurrency/asynpool.py:76: BytesWarning: Comparison between bytes and string
  return unpack(fmt, iobuf.getvalue())  # <-- BytesIO
```

This occurs due to passing a `bytes` object to the `fmt` argument of `struct` functions. The solution was borrowed from py-amqp: https://github.com/celery/py-amqp/pull/117#issuecomment-267181264

For a discussion on passing `str` instead of `bytes` to `struct` functions, see: https://github.com/python/typeshed/pull/669

Information on the `-b` CLI option: https://docs.python.org/3/using/cmdline.html#miscellaneous-options

> `-b`
>
> Issue a warning when comparing bytes or bytearray with `str` or `bytes` with int. Issue an error when the option is given twice (`-bb`).

### DIFF
--- a/celery/concurrency/asynpool.py
+++ b/celery/concurrency/asynpool.py
@@ -20,7 +20,6 @@ import gc
 import os
 import select
 import socket
-import struct
 import sys
 import time
 from collections import deque, namedtuple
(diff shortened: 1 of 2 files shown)


jdufresne:
Rebasd. Thanks!

PR #4847, opened 2018-06-23
Require the redis module for test test_timeouts_in_url_coerced

jdufresne:
The test requires the redis module to be installed. It is referenced in the function [`RedisBackend._params_from_url()`](https://github.com/celery/celery/blob/d20b8a5d469c80f48468e251cbe6451c798d1c29/celery/backends/redis.py#L254).

Fixes AppVeyor build failures.


auvipy:
good catch thanks!

PR #4850, opened 2018-06-26
Fix hybrid_to_proto2 with missing timelimit

23doors:
If `timelimit` is not defined in `body`, it will default to `None` value. Which will cause result in a crash here:
https://github.com/celery/celery/blob/master/celery/worker/request.py#L188 (`'NoneType' object is not iterable`).

Defaulting to `(None, None)` instead should fix it.


23doors:
@auvipy added a simple test case that used to fail before.

auvipy:
could you fix the flake8 and isort errors?

23doors:
should be fine now

PR #4858, opened 2018-06-27
Preserve order of return values from groups

### LINKED ISSUE #3781: Group tasks' results order getting messed up when using redis
I upgraded to celery v4.0.2 from an old v3.x version and now I'm facing a weird issue with groups.
So here is my tasks file to reproduce the issue:

```python
from celery import group, chord
from worker import app


@app.task(name='task')
def task(i):
    return i


@app.task(name='dummy')
def dummy(result):
    print result


@app.task(name='mainTask')
def mainTask():
    groupTask = group([task.s(i) for i in range(0,10)])
    job = chord(groupTask, dummy.s())()
    return job
```

The expected output of the variable `result` is:

> [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]

and this is indeed what I'm getting when using `amqp` as result backend. But when I use `redis`, the output gets messed up. Basically, the redis backend is serving the results in the order the group tasks get finished, and not in the order its assigned. A sample would be:

> [0, 1, 2, 3, 4, 8, 6, 9, 7, 5]

Is there something I'm missing? I've posted the same thing over in [stackoverflow](http://stackoverflow.com/questions/41800124/celery-v4-group-task-behaviour-with-redis-backend) as well.

lpsinger:
## Description

Preserve the order of return values from groups under the redis backend. Fixes #3781.

## Example

Here is `test.py`:

```python
#!/usr/bin/env python
import random
import time

from celery import Celery, group

app = Celery('hello', broker='redis://', backend='redis://')


@app.task(shared=False)
def f(x):
    time.sleep(random.uniform(0, 1))
    return x


@app.task(shared=False)
def test():
    return (group(f.s(i) for i in range(10)) | f.s()).delay()


if __name__ == '__main__':
    app.start()
```

Start the worker like this:

```
$ ./test.py worker
```

And test it like this:

```
$ ./test.py shell
>>> test().get()
[0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
```

In previous versions of Celery, the return values would generally be out of order.

### DIFF
--- a/celery/backends/redis.py
+++ b/celery/backends/redis.py
@@ -364,7 +364,7 @@ class RedisBackend(BaseKeyValueStoreBackend, AsyncBackendMixin):
                 decode, unpack = self.decode, self._unpack_chord_result
                 with client.pipeline() as pipe:
                     resl, = pipe \
-                        .zrangebyscore(jkey, '-inf', '+inf') \
+                        .zrange(jkey, 0, -1) \
                         .execute()
                 try:
                     callback.delay([unpack(tup, decode) for tup in resl])
--- a/celery/backends/redis.py
+++ b/celery/backends/redis.py
@@ -347,8 +347,8 @@ class RedisBackend(BaseKeyValueStoreBackend, AsyncBackendMixin):
         result = self.encode_result(result, state)
         with client.pipeline() as pipe:
             _, readycount, totaldiff, _, _ = pipe \
-                .zadd(jkey, group_index,
-                      self.encode([1, tid, state, result])) \
+                .zadd(jkey,
+                      {self.encode([1, tid, state, result]): group_index}) \
                 .zcount(jkey, '-inf', '+inf') \
                 .get(tkey) \
                 .expire(jkey, self.expires) \
--- a/celery/backends/redis.py
+++ b/celery/backends/redis.py
@@ -335,9 +335,11 @@ class RedisBackend(BaseKeyValueStoreBackend, AsyncBackendMixin):
     def on_chord_part_return(self, request, state, result,
                              propagate=None, **kwargs):
         app = self.app
-        tid, gid = request.id, request.group
+        tid, gid, group_index = request.id, request.group, request.group_index
         if not gid or not tid:
             return
+        if group_index is None:
+            group_index = '+inf'
 
         client = self.client
         jkey = self.get_key_for_group(gid, '.j')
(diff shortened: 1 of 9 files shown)


lpsinger:
The unit tests are passing now, but the integration tests are still failing.

georgepsarakis:
@lpsinger this is nice, but a rather extensive change. It also seems backwards incompatible, since the Redis structure is changed from List to Sorted Set, so it needs some more thought. I will have a look in the next few days, thanks.

celery/canvas.py
```diff
@@ -1142,6 +1149,7 @@ def _freeze_unroll(self, new_tasks, group_id, chord, root_id, parent_id):
         # pylint: disable=redefined-outer-name
         #   XXX chord is also a class in outer scope.
         stack = deque(self.tasks)
+        i = 0
```
georgepsarakis: Perhaps minor, but I think we could also name the variable `group_index` instead of `i`.

celery/worker/request.py
```diff
@@ -520,6 +520,11 @@ def _context(self):
         }, **embed or {})
         return Context(request)
 
+    @cached_property
+    def group_index(self):
+        # used by backend.on_chord_part_return to order return values in group
+        return self.request_dict['group_index']
```
georgepsarakis: Perhaps it is better to use `dict.get` here too.

celery/backends/redis.py
```diff
@@ -361,7 +364,7 @@ def on_chord_part_return(self, request, state, result,
                 decode, unpack = self.decode, self._unpack_chord_result
                 with client.pipeline() as pipe:
                     resl, = pipe \
-                        .lrange(jkey, 0, total) \
+                        .zrangebyscore(jkey, '-inf', '+inf') \
```
georgepsarakis: You could also use:

```python
zrange(jkey, 0, -1)
```

I think it would also make the mock client https://github.com/celery/celery/pull/4858/files#diff-5cb790f5c0c0aea16658640ccd102c19R97 a bit simpler to implement. Not really important, merely a suggestion though.

georgepsarakis:
It seems that latest flake8 has incompatibilities. Would you mind setting the [version](https://github.com/celery/celery/blob/70bb858889e846a86ac0d3e3a53f3fb475c04fbe/requirements/pkgutils.txt#L3) to [3.7.1](https://github.com/PyCQA/flake8/tree/3.7.1) so that the build proceeds?

celery/canvas.py
lpsinger: Done.

celery/worker/request.py
lpsinger: Done.

maybe-sybr:
@auvipy - I've rebased this changeset onto the latest master and it still seems to function happily, resolving a problem I was having with group result ordering on top of Redis today. Is there still appetite to get this merged? If @lpsinger (/ping) isn't able, I can push an alternate PR and prep for merge.

auvipy:
> @auvipy - I've rebased this changeset onto the latest master and it still seems to function happily, resolving a problem I was having with group result ordering on top of Redis today. Is there still appetite to get this merged? If @lpsinger (/ping) isn't able, I can push an alternate PR and prep for merge.

please take this

lpsinger:
@maybe-sybr, I'm very happy for you to carry this forward. Thank you!

maybe-sybr:
I've created #6218 to absorb the changes proposed here. The questions around changes to the Redis results backend typing are still relevant so I'll ping the commenter there to raise it again.

PR #4859, opened 2018-06-28
Py37 matrix

auvipy:


ldsink:
https://github.com/travis-ci/travis-ci/issues/9069

The ssl module in 3.7 no longer compiles with OpenSSL from Ubuntu 14.04.

`dist` maybe need change to `xenial`

georgepsarakis:
@ldsink you are right. I found [this approach](https://github.com/travis-ci/travis-ci/issues/9069#issuecomment-401924248) in the comments.

auvipy:
i will update accordingly

auvipy:
I have updated travis as per the comment

georgepsarakis:
@auvipy the failure has to do with Riak backend unit tests (`t/unit/backends/test_riak.py`). The tests will continue to fail, unless a new release on PyPI includes the fix. We could do the following:
- Ignore Riak tests for now on 3.7.
- Detect Python 3.7 and emit a [warning](https://docs.python.org/3/library/warnings.html) that this backend is unsupported on 3.7.

@thedrow what do you think?

thedrow:
I think that we should emit a warning and skip the tests for riak on Python 3.7.

auvipy:
I do agree with @thedrow

Issue #4500, opened 2018-01-24
Support `async` keyword in python 3.7


libre-man:
## Checklist

- [ ] I have included the output of ``celery -A proj report`` in the issue.
      (if you are not able to do this, then at least specify the Celery
       version affected).
- [x] I have verified that the issue exists against the `master` branch of Celery.

## Steps to reproduce
When importing celery in python 3.7 this fails with a syntax error. This is because `async` is used as a property on at least one object. The first error is in `worker.py` on line 243. This is still a issue on the master branch and I think it at least affects 4.1.0. 

## Expected behavior
I can import celery.

## Actual behavior
I cannot import celery when running python 3.7

auvipy:
3.7 is not officially supported yet. that will be supported on v5 of celery after 4.2 release

thedrow:
We could support 3.7 with 4.x as well. It will require some work though.

thedrow:
I don't know yet. We need to scan the code and see if we use async anywhere.

auvipy:
https://github.com/celery/celery/blob/master/celery/backends/async.py

thedrow:
Change all `async` names to `asynchronous` like we did in Kombu.

auvipy:
https://github.com/celery/celery/issues/4500

auvipy:
I will take care of those soon, for 4.2.1

m3talstorm:
@auvipy I hope I am not imposing, but I needed a fix for some Python3.7 testing, so I did a simple replace in the repo for 'async' -> 'asynchronous' (in the relevant places)., as seen above.

auvipy:
just check my open pr plz

m3talstorm:
@auvipy Ah, I didn't see it because the PR or the commits didn't include the title/number of this issue.

auvipy:
merged

george-silva:
I'm using Python 3.7.0 and just upgraded to version 4.2.1.

Error still persists:

```
  File "/home/george/projects/.virtualenvs/nswa/lib/python3.7/site-packages/celery/backends/redis.py", line 22
    from . import async, base
                      ^
SyntaxError: invalid syntax
```

huangsam:
@george-silva encountering the same issue as well with 4.2.1 on Python 3.7.

huangsam:
`4.2` branch still has the bad line: https://github.com/celery/celery/blob/4.2/celery/backends/redis.py#L22

[...]

PR #4863, opened 2018-06-28
#4500: Support `async` keyword in python 3.7

m3talstorm:
Renamed variables/files/import/etc from 'aysnc' to 'asynchronous' as to not conflict with Python3.7 changes.

Follows suggestions in #4500 and https://github.com/celery/kombu/pull/839


m3talstorm:
The failing landscape test is returning a:

> Server Error (500)

So not sure if something actually wrong with the merge or their service is down.

auvipy:
check the open pr's before starting duplicating efforts.

PR #4888, opened 2018-07-06
Allow Extraction of Chord Results On Error

npilon:
## Description

Make it possible to access the results of chord header tasks in the chord body's `on_error` handler. This is desirable in circumstances where some of the work from a parallel process can still be used even if other tasks involved in the process failed.

The major change are:

* to make it so the redis backend does not delete the aggregated results before determining if an error occurred
* to preserve the group ID from the header task's request in its result, so the error handler can find the group ID and use it to access the aggregated results

The integration test validates that the information available to the chord error handler is sufficient for accessing the chord header results even if the process is not precisely convenient, and without changing the external API or expected function signatures.

### DIFF
--- a/celery/backends/base.py
+++ b/celery/backends/base.py
@@ -660,7 +660,7 @@ class BaseKeyValueStoreBackend(Backend):
             'children': self.current_task_children(request),
             'task_id': bytes_to_str(task_id),
         }
-        if request and request.group:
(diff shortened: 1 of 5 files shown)


npilon:
Part of the travis matrix errored on installing couchbase; can someone re-run that?

Issue #4893, opened 2018-07-08
Worker Monitoring


Twista:
Hello everyone,

This is more like a question than a bug report.

I would like to have, let's say cron based script which will tell me how many workeres which what capacity are available at the moment.

I tried multiple approaches, but currently struggling with `--autoscale` option, as maximum capacity is not part of any report.

I have multiple workers (code I will be posting have just one worker for sake of simplicity). Currently using celery 4.1 w/ redis. While playing around with `celery.app.control.Inspect` I was able to get total number of workers be using `.ping()` method.

```python
inspect = celery.control.inspect()  # type: Inspect
pprint(inspect.ping())
```

workers run in --autoscale mode => `celery worker --task-events --app=.. --autoscale=8,3 --queues=.. -Ofair`

Now the trouble is to get maximum capacity (8 in this case, is the autoscaling is 3-8) through code.
I tried multiple methods on `Inspect` class, but so far I was able to get jsut 3 (which is not maximum capacity

I hope it all make sense and I would be happy for any advice how to monitor # of workers with theirs maximum capacity.
Sample outputs from my machine

**stats()**
```python
inspect = celery.control.inspect()  # type: Inspect
pprint(inspect.stats())
```

out:
```
{'celery@Michals-MacBook-Pro.local': {'broker': {'alternates': [],
                                                 'connect_timeout': 4,
                                                 'failover_strategy': 'round-robin',
                                                 'heartbeat': 120.0,
                                                 'hostname': 'localhost',
                                                 'insist': False,
                                                 'login_method': None,
                                                 'port': 6379,
                                                 'ssl': False,
                                                 'transport': 'redis',
                                                 'transport_options': {},
                                                 'uri_prefix': None,
                                                 'userid': None,
                                                 'virtual_host': '0'},
                                      'clock': '88',
                                      'pid': 69573,
                                      'pool': {'max-concurrency': 3,
                                               'max-tasks-per-child': 'N/A',
                                               'processes': [69581,
                                                             69582,
                                                             69583],
                                               'put-guarded-by-semaphore': False,
                                               'timeouts': [0, 0],
                                               'writes': {'all': '',
                                                          'avg': '0.00%',
                                                          'inqueues': {'active': 0,
                                                                       'total': 3},
                                                          'raw': '',
                                                          'strategy': 'fair',
                                                          'total': 0}},
                                      'prefetch_count': 3,
                                      'rusage': {'idrss': 0,
                                                 'inblock': 234,
[...]

PR #4897, opened 2018-07-10
Add autoscaler information on stats inspect command

xirdneh:
The `autoscaler` boot step class is missing an `info` method to return mix and max settings.
An info method is added returning `worker.autoscaler.info()` data.
Fixes issue raised by #4893


xirdneh:
Yes, I should've tagged this as "working on it".
I'm trying to figure out how the `control` tests work.
will do, thanks.

xirdneh:
I'm not sure about the code-quality one. When I try to look at it there's a 500 eror.
Any ideas? @auvipy ?

auvipy:
Ignore that! seems we have to remove it!

PR #4900, opened 2018-07-11
Added compatibility with python 3.7

kidoz:
In Python 3.7 removed re._pattern_type, see python/cpython#1646. Now in 3.7 need using re.Pattern instead it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ python:
   - '3.4'
   - '3.5'
   - '3.6'
+  - '3.7'
 os:
     - linux
 stages:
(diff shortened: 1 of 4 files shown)


kidoz:
On test env with regression tests i catch only this problem on celery 4.2.0.
On 3.6 also done testing with this changes. 

About unit testing, i trying to do it.

auvipy:
not talking about extra tests, ,just ,existing tests or code fixes for this new change if anything exists just investigate and change

kidoz:
Open new pull request with extend tests:
https://github.com/celery/celery/pull/4902

PR #4912, opened 2018-07-19
Fixed Pytest 3.4 logging error.

auvipy:
cleanups needed


auvipy:
ERROR at teardown of test_default_logger.test_setup_logging_subsystem_misc2 ――
request = <SubRequest 'sanity_logging_side_effects' for <Function 'test_setup_logging_subsystem_misc2'>>
    @pytest.fixture(autouse=True)
        root = logging.getLogger()
        rootlevel = root.level
        roothandlers = root.handlers
    
        yield
    
        this = request.node.name
            raise RuntimeError(CASE_LOG_LEVEL_EFFECT.format(this))
        if root_now.handlers != roothandlers:
>           raise RuntimeError(CASE_LOG_HANDLER_EFFECT.format(this))
E           RuntimeError: Test test_setup_logging_subsystem_misc2 modified handlers for the root logger
t/unit/conftest.py:244: RuntimeError

auvipy:
@nicoddemus hi Bruno, could you suggest to fix the error?

nicoddemus:
@auvipy, probably there's been some changes in the logging plugin in respect to the handlers applied during testing. That fixture is checking that nobody changes the handlers, I presume the fixture wants to check that *celery* is not messing with the handlers, not pytest.

Not sure what to suggest, it would be useful if we can get some background on what was the original purpose of `sanity_logging_side_effects` from celery developers.

nicoddemus:
@thedrow thanks for the explanation.

In that case I believe it makes sense to ignore pytest's own handlers. This patch works for me:

```diff
diff --git a/t/unit/conftest.py b/t/unit/conftest.py
index cd8e8e9b6..b25005387 100644
--- a/t/unit/conftest.py
+++ b/t/unit/conftest.py
@@ -230,9 +230,10 @@ def sanity_stdouts(request):

 @pytest.fixture(autouse=True)
 def sanity_logging_side_effects(request):
+    from _pytest.logging import LogCaptureHandler
     root = logging.getLogger()
     rootlevel = root.level
-    roothandlers = root.handlers
+    roothandlers = [x for x in root.handlers if not isinstance(x, LogCaptureHandler)]

     yield

@@ -240,7 +241,8 @@ def sanity_logging_side_effects(request):
     root_now = logging.getLogger()
     if root_now.level != rootlevel:
         raise RuntimeError(CASE_LOG_LEVEL_EFFECT.format(this))
-    if root_now.handlers != roothandlers:
+    newhandlers = [x for x in root_now.handlers if not isinstance(x, LogCaptureHandler)]
+    if newhandlers != roothandlers:
         raise RuntimeError(CASE_LOG_HANDLER_EFFECT.format(this))
```

It is not ideal because it uses the private API, but I don't expect this to change much between pytest versions (and if it does it should be simple to update the code as we only reference the class name).

auvipy:
thanks, Bruno and Omer! it does fix the error!

PR #4929, opened 2018-07-26
Update `CELERY_TASK_TRACK_STARTED` setting in the docs

nikmolnar:
## Description

This PR updates the docs to correctly reflect the change in Django setting from `CELERY_TRACK_STARTED`, which no longer works, to `CELERY_TASK_TRACK_STARTED`.

### DIFF
--- a/docs/history/whatsnew-4.0.rst
+++ b/docs/history/whatsnew-4.0.rst
@@ -472,7 +472,7 @@ a few special ones:
 ``CELERY_TASK_SERIALIZER``             :setting:`task_serializer`
 ``CELERYD_TASK_SOFT_TIME_LIMIT``       :setting:`task_soft_time_limit`
 ``CELERYD_TASK_TIME_LIMIT``            :setting:`task_time_limit`
-``CELERY_TRACK_STARTED``               :setting:`task_track_started`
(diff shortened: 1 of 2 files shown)


clokep:
I'm unsure if this should have been merged. I think that table is showing that the old name was `CELERY_TRACK_STARTED` and the new name is `task_track_started`, but you can use capitalized ones and a prefix of `celery` to make it `CELERY_TASK_TRACK_STARTED`?

nikmolnar:
> I think that table is showing that the old name was CELERY_TRACK_STARTED and the new name is task_track_started

@clokep: The base name has always been `task_track_started`. However, when using with Django, the name used to be `CELERY_TRACK_STARTED` and is now `CELERY_TASK_TRACK_STARTED`.

clokep:
@nikmolnar Right, and that table is showing how to translate those from 3.x to 4.x. The "new" Django names aren't shown in that table, only the base names.

nikmolnar:
@clokep Oh right, I see what you're saying. Sorry, I misunderstood that table at first. I was caught off guard by the change of the name as used with Django and was looking for somewhere to clarify in the docs. I agree that this wasn't the correct place.

PR #4936, opened 2018-07-30
Update dynamodb._get_or_create_table

cetanu:
Attempt to get the existing table, regardless of the experienced error.

For example (my example) a client might not have access to create the table.
This is missed when only catching ResourceInUseException (assumes that we have permission to create the table)

### DIFF
--- a/celery/backends/dynamodb.py
+++ b/celery/backends/dynamodb.py
@@ -199,11 +199,11 @@ class DynamoDBBackend(KeyValueStoreBackend):
             error_code = e.response['Error'].get('Code', 'Unknown')
 
             # If table exists, do not fail, just return the description.
-            if error_code == 'ResourceInUseException':
+            try:
                 return self._client.describe_table(
                     TableName=self.table_name
                 )
-            else:
+            except:
                 raise e
 
     def _wait_for_table_status(self, expected='ACTIVE'):


celery/backends/dynamodb.py
```diff
@@ -199,11 +199,11 @@ def _get_or_create_table(self):
             error_code = e.response['Error'].get('Code', 'Unknown')
```
thedrow: Do we need this variable if the condition is removed?
The build is failing because of that.

celery/backends/dynamodb.py
```diff
@@ -199,11 +199,11 @@ def _get_or_create_table(self):
...
 
             # If table exists, do not fail, just return the description.
-            if error_code == 'ResourceInUseException':
+            try:
                 return self._client.describe_table(
                     TableName=self.table_name
                 )
-            else:
+            except:
```
thedrow: I'm not a fan of bare excepts.
Is there a common base class for these type of exceptions?

cetanu:
Negatory; for my use-case, I simply patched the class instead:

```
from celery.app import backends


class DynamoBackend(DynamoDBBackend):
    def _get_or_create_table(self):
        """
        Hack to replace default behavior for DDB backends.

        The only overridden portion of this is the except block.

        The difference is, we return the table no matter what the
        exception is.
        The original only handles specific exceptions, and then attempts
        to create the table.
        """
        table_schema = self._get_table_schema()
        try:
            table_description = self._client.create_table(**table_schema)
            # In case we created the table, wait until it becomes available.
            self._wait_for_table_status('ACTIVE')
            return table_description
        except ClientError as e:
            # If table exists, do not fail, just return the description.
            try:
                return self._client.describe_table(
                    TableName=self.table_name
                )
            except:
                raise e

backends.BACKEND_ALIASES['dynamodb'] = 'myapp.worker:DynamoBackend'
```

I considered if this would be important to merge back into the main celery repo but I think it's not applicable because this was implemented to get around the restraints of an internal platform used for hosting our AWS infrastructure.

PR #4945, opened 2018-08-01
Fix service stop

marcusmchale:
The kill_workers function in this template init file was missing the $DAEMON_OPTS parameters. In my case this contained the work-dir option required to find the PID file.

### DIFF
--- a/extra/generic-init.d/celeryd
+++ b/extra/generic-init.d/celeryd
@@ -284,7 +284,8 @@ restart_workers () {
 
 
 kill_workers() {
-    _chuid kill $DAEMON_OPTS $CELERYD_NODES --pidfile="$CELERYD_PID_FILE"
+    _chuid kill $CELERYD_NODES $DAEMON_OPTS         \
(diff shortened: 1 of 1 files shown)


extra/generic-init.d/celeryd
```diff
@@ -284,7 +284,7 @@ restart_workers () {
 
 
 kill_workers() {
-    _chuid kill $CELERYD_NODES --pidfile="$CELERYD_PID_FILE"
+    _chuid kill $DAEMON_OPTS $CELERYD_NODES --pidfile="$CELERYD_PID_FILE"
```
georgepsarakis: Minor, but just for uniformity with the other commands (see `restart_workers` function above) can you place `$DAEMON_OPTS` after `$CELERYD_NODES`?

georgepsarakis:
@marcusmchale feel free to add yourself to the list of contributors.

extra/generic-init.d/celeryd
marcusmchale: done.

PR #4966, opened 2018-08-08
Travis CI and Tox: Upgrade from Python 3.6 --> 3.7

cclauss:

## Description

Travis CI and Tox: Upgrade from Python 3.6 --> 3.7.  This a different take on #4859

that version series.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,9 +26,8 @@ matrix:
   - python: '3.7'
     dist: xenial    # required for Python 3.7 (travis-ci/travis-ci#9069)
     env: TOXENV=pypy-unit PYPY_VERSION="pypy2.7-6.0.0"
-  - python: '3.7'
-    dist: xenial    # required for Python 3.7 (travis-ci/travis-ci#9069)
-    env: TOXENV=pypy-integration-rabbitmq PYPY_VERSION="pypy2.7-6.0.0"
+  - python: '3.6'
+     env: TOXENV=pypy-integration-rabbitmq PYPY_VERSION="pypy2.7-6.0.0"
   - python: '3.7'
     dist: xenial    # required for Python 3.7 (travis-ci/travis-ci#9069)
     env: TOXENV=pypy-integration-redis PYPY_VERSION="pypy2.7-6.0.0"
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,11 +2,6 @@ language: python
 sudo: required
 dist: trusty
 cache: pip
-python:
-  - '2.7'
-  - '3.4'
-  - '3.5'
-  - '3.6'
 os:
     - linux
 stages:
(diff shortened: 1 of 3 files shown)


cclauss:
# Tests that **pass** on Python 3.7:  TOXENV=...
1. flake8
2. flakeplus
3. bandit
4. pydocstyle
5. pypy-unit PYPY_VERSION="pypy2.7-6.0.0"
6. pypy-integration-redis PYPY_VERSION="pypy2.7-6.0.0"
7. pypy-integration-dynamodb PYPY_VERSION="pypy2.7-6.0.0"

# Tests that **fail** on Python 3.7:  TOXENV=...
1. apicheck
2. configcheck
3. isort-check
4. unit
5. pypy-integration-rabbitmq PYPY_VERSION="pypy2.7-6.0.0"

cclauss:
This PR documents which tests pass and which tests fail so we can know where we need to focus effort.  Such as https://github.com/timothycrosley/isort/pull/716

auvipy:
@jdufresne does this approach seems good to you for celery?

cclauss:
Either way, we have to get those five failing tests to pass on Python 3.7.

.travis.yml
```diff
@@ -22,18 +17,29 @@ env:
   - MATRIX_TOXENV=integration-dynamodb
 matrix:
   include:
+  - python: '2.7'
+  - python: '3.4'
+  - python: '3.5'
   - python: '3.6'
+  #- python: '3.7'
+  #  dist: xenial    # required for Python 3.7 (travis-ci/travis-ci#9069)
```
jdufresne: Should this be uncommented? This is the test for Python 3.7 environment, right?

.travis.yml
```diff
@@ -22,18 +17,29 @@ env:
...
+  - python: '3.7'
+    dist: xenial    # required for Python 3.7 (travis-ci/travis-ci#9069)
     env: TOXENV=pypy-unit PYPY_VERSION="pypy2.7-6.0.0"
```
jdufresne: Can you explain why the PYPY test need Python 3.7 as a default system Python? Shouldn't it only be concerned with the PYPY version?

tox.ini
```diff
@@ -17,7 +17,7 @@ deps=
     -r{toxinidir}/requirements/test.txt
 
     2.7: -r{toxinidir}/requirements/test-ci-default.txt
-    3.4,3.5,3.6: -r{toxinidir}/requirements/test-ci-default.txt
+    3.4,3.5,3.6,3.7: -r{toxinidir}/requirements/test-ci-default.txt
```
jdufresne: Could this line be merged with the one above it?

tox.ini
```diff
@@ -54,8 +54,10 @@ basepython =
     3.4: python3.4
     3.5: python3.5
     3.6: python3.6
+    3.7: python3.7
```
jdufresne: If this `tox.ini` were to use the [tox default factors](https://tox.readthedocs.io/en/latest/config.html#factors), we could avoid specifying the `basepython` all together. For example `py27`, `py36`, `py37`, ...

See: https://tox.readthedocs.io/en/latest/config.html#confval-basepython=NAME-OR-PATH

.travis.yml
cclauss: My understand was that this was an integration tests between systems running two different versions of Python (CPython 3.7 and Pypy 2.7).  If that is not the case then __python: pypy__ should be the setting because in that case, we do not need CPython.

tox.ini
cclauss: I believe it could be but I assumed that there was some method behind the was it was.

.travis.yml
cclauss: It is commented out because it is test 4 of the failing tests.  Python 3.7 unit fails.

tox.ini
cclauss: Explicit is better than implicit (at least until we get the failing tests to pass).

PR #4970, opened 2018-08-12
Add task.acks_on_failure_or_timeout option

mariokostelac:
As shown in https://github.com/celery/celery/issues/4797, acknowledging
SQS messages on failure or timing out make it hard to use dead letter
queues. This change is introducing a new option
acks_on_failure_or_timeout, making sure we can totally fallback on
native SQS message lifecycle, using redeliveries for retries (in case of
slow processing or failure) and transitions to dead letter queue after
the defined number of times.

### DIFF
--- a/celery/app/task.py
+++ b/celery/app/task.py
@@ -250,6 +250,13 @@ class Task(object):
     #: :setting:`task_acks_late` setting.
     acks_late = None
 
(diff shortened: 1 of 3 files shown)


auvipy:
thanks. will review it tomorrow.

auvipy:
thanks!

dejlek:
Awesome, thanks!

PR #4975, opened 2018-08-14
Fix a typo in /userguide/optimizing

XD-DENG:

## Description

Fix a typo in /userguide/optimizing
(http://docs.celeryproject.org/en/latest/userguide/optimizing.html)

screen shot 2018-08-14 at 1 43 23 pm

### DIFF
--- a/docs/userguide/optimizing.rst
+++ b/docs/userguide/optimizing.rst
@@ -225,11 +225,11 @@ have a buffer as small as 64KB but on recent Linux versions the buffer
 size is 1MB (can only be changed system wide).
 
 You can disable this prefetching behavior by enabling the
-:option:`-Ofair <celery worker -O>` worker option:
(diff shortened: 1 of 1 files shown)


georgepsarakis:
@thedrow the two forms should be equivalent. While this is not a typo, I don't think it is necessary to revert also.

XD-DENG:
Hi @georgepsarakis @thedrow @auvipy , I proposed this change because there is argument `-O` with possible values `default` and `fair`, while there is no argument `-Ofair` for `celery.bin.worker` ([link](http://docs.celeryproject.org/en/latest/reference/celery.bin.worker.html#cmdoption-celery-worker-o)).

Even the part I changed itself was redirecting to argument `-O`.

If this PR is going to be reverted, I would like to suggest adding `-Ofair` explicitly in the documentation of `celery.bin.worker`.

PR #4986, opened 2018-08-18
[WIP] converted list() to []

auvipy:
clean ups

### DIFF
--- a/t/integration/test_canvas.py
+++ b/t/integration/test_canvas.py
@@ -89,10 +89,10 @@ class test_chain:
 
         result = (before | connect | after).delay()
         result.get(timeout=TIMEOUT)
-        redis_messages = list(map(
+        redis_messages = [map(
             bytes_if_py2,
             redis_connection.lrange('redis-echo', 0, -1)
-        ))
+        )]
         before_items = \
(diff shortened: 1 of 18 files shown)


auvipy:
need help to fix the bandit issue :/

thedrow:
I fixed the bandit errors.

t/integration/test_canvas.py
```diff
@@ -89,10 +89,10 @@ def test_group_chord_group_chain(self, manager):
 
         result = (before | connect | after).delay()
         result.get(timeout=TIMEOUT)
-        redis_messages = list(map(
+        redis_messages = [map(
```
thedrow: This change is invalid.
The purpose of the code is to turn a `map()` object into a list.
The code here is completely different. It changes the value of `redis_messages` into `[map()]` from `[b'<string converted to bytes if python2>', b'<another string converted to bytes if python2>']`.

t/integration/test_canvas.py
```diff
@@ -114,10 +114,10 @@ def test_second_order_replace(self, manager):
 
         result = second_order_replace1.delay()
         result.get(timeout=TIMEOUT)
-        redis_messages = list(map(
+        redis_messages = [map(
```
thedrow: Ditto

PR #4997, opened 2018-08-21
Deadlock on inqueue close (#4185)

jengelman:
Issue: #4185 

Upon attempting to close a socket, AsynPool only removes the queue writer from the hub, not the reader.  This leads to a deadlock on the file descriptor, and eventually the worker stop accepting new tasks.  I moved the call to hub_remove from on_inqueue_close to destroy_queues, since the latter is called directly after the former and allows removing both fds in a single loop.

### DIFF
--- a/celery/concurrency/asynpool.py
+++ b/celery/concurrency/asynpool.py
@@ -737,7 +737,6 @@ class AsynPool(_pool.Pool):
                     fileno_to_inq.pop(fd, None)
                     active_writes.discard(fd)
                     all_inqueues.discard(fd)
-                    hub_remove(fd)
             except KeyError:
                 pass
         self.on_inqueue_close = on_inqueue_close
@@ -1247,7 +1246,7 @@ class AsynPool(_pool.Pool):
             if queue:
                 for sock in (queue._reader, queue._writer):
                     if not sock.closed:
-                    	self.hub_remove(sock)
+                        self.hub_remove(sock)
                         try:
                             sock.close()
                         except (IOError, OSError):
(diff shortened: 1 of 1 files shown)


thedrow:
This is awesome!
Is there a way to add a unit test for this to ensure that the reader and the writer are removed?

jengelman:
Hm, I honestly don't know how to write a unit test for this, since it's pretty deep in the internals.

Maybe creating a mock job, assigning it to a one of the worker queues, creating a process and setting job._to_write to that process, setting job._accepted to False, killing the processing, calling destroy_queues, and then checking the hub directly? AsynPool.destroy_queues is called in AsyncPool.on_partial_read, which is called in AsynPool.on_job_process_down.  Looking at the concurrency unit tests, there's nothing really like this, so any help would be appreciated!

MichaelrMentele:
Would love to see this get merged, can't use the `max-tasks-per-child` and `max-memory-per-child` settings--both seem to run into the linked issue.

auvipy:
Merging this for verification purpose, since some people reported using this removed their issue. will be looking for better tests for this.

jengelman:
Thanks for merging this in!  I've got a bit more time now so will take a crack at a followup PR with relevant tests.

auvipy:
looking forward to it

auvipy:
nope, you need to install from github master using pip

auvipy:
willing to release by November

mikolaje:
I'm using the celery==4.2.1 ; python==3.6.7  ubuntu 16.04. and Redis as backend
I used supervisor to start celery with command:
celery -A zapier2cloud worker -l info --autoscale=10,2

It still got stuck sometimes.

Issue #5000, opened 2018-08-22
Chaining Chords produces enormously big messages causing OOM on workers


brabiega:
## Checklist

- [x] I have included the output of ``celery -A proj report`` in the issue.
      (if you are not able to do this, then at least specify the Celery
       version affected).
- [ ] I have verified that the issue exists against the `master` branch of Celery.

```
software -> celery:4.2.1 (windowlicker) kombu:4.2.1 py:2.7.6
            billiard:3.5.0.4 redis:2.10.6
platform -> system:Linux arch:64bit, ELF imp:CPython
loader   -> celery.loaders.app.AppLoader
settings -> transport:redis results:db+mysql://celery:**@127.0.0.1/celery

task_serializer: 'json'
broker_heartbeat: 360
result_serializer: 'json'
event_queue_ttl: 60
prefetch_multiplier: 1
broker_url: u'redis://:********@localhost:6379/0'
task_acks_late: True
worker_redirect_stdouts_level: 'INFO'
database_result_engine_options: {
    'isolation_level': 'READ_COMMITTED'}
task_time_limit: 910
task_soft_time_limit: 900
accept_content: ['json']
task_track_started: True
result_backend: u'db+mysql://celery:********@127.0.0.1/celery'
max_tasks_per_child: 100
```

## Steps to reproduce
Tasks file
```python
@app.task()
def noop(*results):
    return
```

Example workflow
```python
from tasks import noop
                                                    
workflow = (                                        
    noop.s() |                                      
    chord([noop.s() for i in range(10)], noop.s()) |
    chord([noop.s() for i in range(10)], noop.s()) |
    chord([noop.s() for i in range(10)], noop.s()) |
    chord([noop.s() for i in range(10)], noop.s()) |
    chord([noop.s() for i in range(10)], noop.s()) |
    noop.s()                                        
)                                                   
                                                    
workflow.apply_async()                              
```

Patch kombu to see serialized message size
```diff
diff --git a/kombu/utils/json.py b/kombu/utils/json.py
index b2466a4..dee3476 100644
--- a/kombu/utils/json.py
+++ b/kombu/utils/json.py
@@ -65,8 +65,11 @@ _default_encoder = JSONEncoder
 def dumps(s, _dumps=json.dumps, cls=None,
           default_kwargs=_json_extra_kwargs, **kwargs):
     """Serialize object to json string."""
-    return _dumps(s, cls=cls or _default_encoder,
[...]

Issue #1921, opened 2014-03-14
Chord of chords & Running parallel chords


cool-hazcker:
When i run the following code

```
wf1 = chord(group(task1.s(), task2.s()), test_callback1.s())
wf2 = chord(group(task1.s(), task2.s()), test_callback2.s())
workflow = group(wf1, wf2)
workflow.delay()
```

test_callback1 and test_callback2 are not being executed. What am i doing wrong?
Is there a way to run chords in parallel? Or even run chords nested in other chords?

PR #5009, opened 2018-08-28
Added an integration test for #1921

thedrow:
Following #5000, it turns out we need an integration test for #1921 to avoid regressions while fixing #5000.
This PR adds it.


auvipy:
is the typeerror valid?

thedrow:
No. I'll fix it.

PR #5011, opened 2018-08-29
Add a concurrency model with ThreadPoolExecutor

whuji:
Hello,

I don't know if it would be useful to someone else, but I needed to implement a concurrency model based on thread (and not processes) in celery because I wanted to pass future objects between tasks and some coroutines and it is not pickable.

Of course it is not scale out and it has limitations (because of the GIL), but it can be useful for people who wants the share memory between tasks and another thread (for example the asyncio event loop) without blocking as the 'solo' concurrency model.

I am open to comment. Maybe this model won't be needed anymore in celery 5 as it could be replaced by an asyncio loop, which is not possible in celery 4. 
And I would be happy to help on a massive asyncio refactoring for Celery 5.

Regards

### DIFF
--- a/celery/concurrency/thread.py
+++ b/celery/concurrency/thread.py
@@ -1,0 +1,40 @@
+# -*- coding: utf-8 -*-
+"""Thread execution pool."""
+
+from concurrent.futures import wait, ThreadPoolExecutor
+from .base import BasePool, apply_target
+
+__all__ = ('TaskPool',)
+
+class ApplyResult(object):
+    def __init__(self, future):
+        self.f = future
+        self.get = self.f.result
+
+    def wait(self, timeout=None):
+        wait([self.f], timeout)
+
+class TaskPool(BasePool):
+    """Thread Task Pool."""
+
+    body_can_be_buffer = True
+    signal_safe = False
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.executor = ThreadPoolExecutor(max_workers=self.limit)
+
+    def on_stop(self):
+        self.executor.shutdown()
+        super().on_stop()
+
+    def on_apply(self, target, args=None, kwargs=None, callback=None,
+                 accept_callback=None, **_):
+        f = self.executor.submit(apply_target, target, args, kwargs, callback, accept_callback)
+        return ApplyResult(f)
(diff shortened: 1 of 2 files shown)


thedrow:
I've been meaning to do this myself. I'll review this as soon as possible.

celery/concurrency/thread.py
```diff
@@ -0,0 +1,40 @@
...
+        wait([self.f], timeout)
+
+class TaskPool(BasePool):
+    """Thread Task Pool."""
+
+    body_can_be_buffer = True
+    signal_safe = False
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
```
thedrow: `super()` calls are only valid with Python 3.
Celery 4.x still supports Python 2.7 so we'll need to adjust that.

celery/concurrency/thread.py
```diff
@@ -0,0 +1,40 @@
...
+        self.executor = ThreadPoolExecutor(max_workers=self.limit)
+
+    def on_stop(self):
+        self.executor.shutdown()
+        super().on_stop()
```
thedrow: `super()` calls are only valid with Python 3.
Celery 4.x still supports Python 2.7 so we'll need to adjust that.

celery/concurrency/thread.py
```diff
@@ -0,0 +1,40 @@
...
+
+    def on_apply(self, target, args=None, kwargs=None, callback=None,
+                 accept_callback=None, **_):
+        f = self.executor.submit(apply_target, target, args, kwargs, callback, accept_callback)
+        return ApplyResult(f)
+
+    def _get_info(self):
+        return {
+            'max-concurrency': self.limit,
+            'threads': len(self.executor._threads)
```
thedrow: I'm not a big fan of using private APIs. Can we change this to something more sensible?

thedrow: Seems like you are correct. There's no public API for that.
I think a TODO comment to change that once it lands is sufficient.

thedrow:
@whuji Did you get a chance to work on this?

whuji:
I have submitted a new RP : https://github.com/celery/celery/pull/5098

Duplicate of #5098

PR #5017, opened 2018-08-31
Update task.update_state to accept variadic kwargs

chdsbd:
## Description

This change enables passing extra fields through task.update_state to the database backend.

In my case, I have a database backend where I allow specifying the database connection for updating a task. Currently, I need to monkey patch this task method to enable passing this kwarg to the backend store_result method. With this change, extra fields can be automatically passed to the database backend.

The backend [store_result](https://github.com/celery/celery/blob/1afaa7b0498201b670f7cbf79643e178b03f83eb/celery/backends/base.py#L317)  method already passes kwargs through, so this change will match up with that existing functionality.


chdsbd:
@auvipy I added an arbitrary kwarg to an existing test case, which verifies this change works.

chdsbd:
I'm not sure why CI is failing.

auvipy:
i have restarted the build and reverted a merged pr which might causing the build error

chdsbd:
I merged in master and fixed some linting issues. I'm not sure why the build is failing, but I think it's unrelated to this change.

thedrow:
Build failures are indeed unrelated.
Thank you for your contribution!

PR #5044, opened 2018-09-10
Gracefully handles consumer decode error

sklarsa:
## Description

Gracefully handles a decode error on the consumer when deserializing a task using the v2 protocol.

If using the v2 protocol, and the consumer encounters an error when deserializing a message's body here https://github.com/celery/celery/blob/master/celery/worker/strategy.py#L136, kombu raises a DecodeError and the worker terminates.

This code block captures and handles the error (in the same way that DecodeErrors are handled in v1), allowing the worker to continue running if it encounters a message it can't serialize.

As it stands now, if the message is unable to be deserialized and the connection drops, it is actually placed back on the queue, at least in our configuration using RabbitMQ and ACKS_LATE.  This is extremely dangerous since one bad message can effectively take down an entire cluster of workers as they continue to fail to deserialize the message and shut down as RabbitMQ places the poisoned message back on the queue for another worker to retrieve.

### DIFF
--- a/celery/worker/consumer/consumer.py
+++ b/celery/worker/consumer/consumer.py
@@ -16,6 +16,7 @@ from time import sleep
 from billiard.common import restart_state
 from billiard.exceptions import RestartFreqExceeded
 from kombu.asynchronous.semaphore import DummyLock
+from kombu.exceptions import DecodeError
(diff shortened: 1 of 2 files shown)


auvipy:
I would request you to come up with another PR :) will discuss there further

PR #5054, opened 2018-09-17
Make task.acks_on_failure_or_timeout settable via settings

mariokostelac:
## Description
Turns out that task attributes are not really set from config unless
the attribute is set to `None` beforehand (or not set at all).

https://github.com/celery/celery/blob/a1fb557b6ecc494514e3480518be57b985d0927d/celery/app/task.py#L328-L330.

### DIFF
--- a/celery/app/defaults.py
+++ b/celery/app/defaults.py
@@ -222,6 +222,7 @@ NAMESPACES = Namespace(
     task=Namespace(
         __old__=OLD_NS,
         acks_late=Option(False, type='bool'),
+        acks_on_failure_or_timeout=Option(True, type='bool'),
         always_eager=Option(False, type='bool'),
         annotations=Option(type='any'),
         compression=Option(type='string', old={'celery_message_compression'}),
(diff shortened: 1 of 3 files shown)


mariokostelac:
@auvipy sure, no problem. Any chance I get a little bit of guidance on what kind of tests I should write here?

I've been looking into tests for default values, but haven't found similar stuff that I could look into.

mariokostelac:
As well, I'd appreciate some guidance on failing config test. I've added a section in docs, but not sure what else i have to do.

mariokostelac:
Found another interesting issue. Messages are not ack-ed, but worker stops draining other jobs after failed job.

auvipy:
thanks for digging it

mariokostelac:
I've spent a little bit more time and figured out I am a little bit out of depth here to just hack it in.

Since the flag I've introduced (`acks_on_failure_or_timeout`) is not performing as expected and possibly is very dangerous, I think we should revert the original change to avoid bugs.

I'd be interesting talking to somebody who implemented the same thing for other transport supporting dead letter queues. Developers docs could be good as well.

thedrow:
So this is already implemented in master.
I'm sorry I missed it.

@mariokostelac If you'd like to adjust the documentation to specifically mention SQS, please do so [here](https://github.com/celery/celery/blob/master/docs/getting-started/brokers/sqs.rst).

If not, let's open an issue about that so one of us won't forget to do so.

Thank you for your contribution anyway.

PR #5063, opened 2018-09-20
require django 1.11 LTS for celery 4.3

auvipy:
update minimal django version to django 1.11 for celery 4.3

### DIFF
--- a/celery/fixups/django.py
+++ b/celery/fixups/django.py
@@ -30,8 +30,8 @@ def _maybe_close_fd(fh):
 
 
 def _verify_django_version(django):
-    if django.VERSION < (1, 8):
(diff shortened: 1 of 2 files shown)


thedrow:
What's the motivation behind this change?

auvipy:
Only support the supported versions of django and this change will affect only 4.3.x release of celery, when django 1.11 wil be the last supported LTS. moreover django-celery-beats and django-celery-results will also drop django versions less than 1.11 in the next release.

gregflynn:
As a user not on Django 1.11 yet, this change made me sad today.

thedrow:
You have to understand, we can't keep supporting versions which have reached EOL.
There are too many dependencies & versions in Celery as it is so we strive to keep the maintenance overhead as small as possible.

PR #5065, opened 2018-09-21
Don't use mutable arguments for ScheduleEntry initializer

bjackson:

## Description
When setting the `kwargs` argument for a ScheduleEntry, the `kwargs` argument is declared as `kwargs={}`. This is a mutable function argument, and causes `kwargs` to accumulate and be used for each additional entry. This PR fixes that. Same thing happens with the `options` argument.


georgepsarakis:
@auvipy I don't think that specific tests are required here. Also, I restarted a failed build in Travis.

Issue #5047, opened 2018-09-12
Memory leak from writing in a loop to a broken rabbitmq heartbeat connection


rgeens:
## Checklist

Verified in both the released 4.2.1 and the master (which is reported as 4.2.0).

- [x] I have included the output of ``celery -A proj report`` in the issue.
      (if you are not able to do this, then at least specify the Celery
       version affected).
```
software -> celery:4.2.1 (windowlicker) kombu:4.2.1 py:3.6.5
            billiard:3.5.0.4 py-amqp:2.3.2
platform -> system:Linux arch:64bit, ELF imp:CPython
loader   -> celery.loaders.app.AppLoader
settings -> transport:pyamqp results:disabled

broker_url: 'amqp://guest:********@172.17.0.1:5672//'
```
- [x] I have verified that the issue exists against the `master` branch of Celery.

Version: https://github.com/celery/celery/tree/ed17e12b14051bac51b84593cf7c8cf21c77e010

```
software -> celery:4.2.0 (windowlicker) kombu:4.2.1 py:3.6.5
            billiard:3.5.0.4 py-amqp:2.3.2
platform -> system:Linux arch:64bit, ELF imp:CPython
loader   -> celery.loaders.app.AppLoader
settings -> transport:pyamqp results:disabled
broker_url: 'amqp://guest:********@172.17.0.1:5672//'
```

## Description

The celery workers of several different applications I work on would sporadically start leaking memory in a continuous fashion, until the machine ran out of memory. We initially tried to work around that by making use of the flag that automatically restart the workers after a certain number of tasks was completed. However, the leak turned out to be in the main worker process that manages workers, so this did not help.

I've found 3 issues that together cause this to happen. One of them clearly seems to be a celery/kombu issue, which is what this report is about. For the other two it's not clear yet but I'll include them for context and reproducibility.

The first issue is that a high CPU (or maybe IO) load can cause Celery workers to lose connections. There's been mention of this in https://github.com/celery/celery/issues/3377 and https://github.com/celery/celery/issues/3932 . I've been able to reproduce that by using the `stress` tool with cpu and io flags enabled. I see celery worker connection resets and missed heartbeats appearing in the logs when I do that. In particular I see the connection resets in the worker log for application A when restarting the worker for application B, while under that load. While I saw similar log messages right before memory leaking started to happen, the leaking didn't happen consistently. I believe this is because the connections that are killed are somewhat random, and some of them can recover from that while others don't.

[...]

PR #5066, opened 2018-09-22
Retry if the heartbeat connection dies

rgeens:
Otherwise it will keep trying to write to the broken connection and
memory will leak because the event dispatcher will keep appending
the message to _outbound_buffer in EventDispatcher._publish.

```python
    def _publish(self, event, producer, routing_key, retry=False,
                 retry_policy=None, utcoffset=utcoffset):
        exchange = self.exchange
        try:
            producer.publish(
                event,
                routing_key=routing_key,
                exchange=exchange.name,
                retry=retry,
                retry_policy=retry_policy,
                declare=[exchange],
                serializer=self.serializer,
                headers=self.headers,
                delivery_mode=self.delivery_mode,
            )
            if not self.buffer_while_offline:
                raise
            self._outbound_buffer.append((event, routing_key, exc))
```

What happens in #5047 is that after the AMQ connection dies, the try-except in the above code will first catch a ConnectionResetError and after that BrokenPipeError's. `buffer_while_offline` is False, so it will just keep appending to the buffer. Raising the exception doesn't help because it just gets reported and ignored further on. Inside `producer.publish` if `retry` is True, it'll check if the connection is still ok and if not re-open it. That's what this patch does.

If needed I can try to add unit tests. Tests seem broken in general in master at the moment, so that would require some further digging.

### DIFF
--- a/celery/worker/heartbeat.py
+++ b/celery/worker/heartbeat.py
@@ -46,6 +46,7 @@ class Heart(object):
                                  active=len(active_requests),
                                  processed=all_total_count[0],
                                  loadavg=load_average(),
+                                 retry=True,
                                  **SOFTWARE_INFO)
 
     def start(self):


rgeens:
I checked the other issue and it looks like the same problem but in a different spot. Because the default for `producer.publish` is to not retry, I suppose every call to it that can have its connection drop for some reason is vulnerable to it. Do you know why it doesn't retry by default and if there are any downsides to doing so?

rgeens:
Because if not then it might make more sense to replace both of our patches with just setting the default to True in `Producer.publish`. But I'm not familiar enough with the code to judge the impact of that at the moment.

auvipy:
I think the retry was accidentally missed

rgeens:
You mean the default value for publish's retry parameter or in the code that calls the publish function?

auvipy:
then I think this should get in.

PR #5068, opened 2018-09-24
Fix Request to pass Context to backend store_result functions.

kiyoya:
## Description

Fix `celery.worker.request.Request` class to pass `celery.app.task.Context` to backend's store_result functions. Since the class currently passes `self` to these functions, revoking a task results in corrupted task result data when you are using celery/django-celery-results for example.

### DIFF
--- a/celery/worker/request.py
+++ b/celery/worker/request.py
@@ -17,6 +17,7 @@ from kombu.utils.encoding import safe_repr, safe_str
 from kombu.utils.objects import cached_property
 
 from celery import signals
+from celery.app.task import Context
 from celery.app.trace import trace_task, trace_task_ret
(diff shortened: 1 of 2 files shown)


celery/worker/request.py
```diff
@@ -502,6 +506,19 @@ def group(self):
         # by parent process
         return self.request_dict.get('group')
 
+    @cached_property
+    def request(self):
```
thedrow: The property name is not very meaningful.
`some_request.request` does not explain what the property returns.
We'll need to rename it.
Furthermore, should this property be public? Is there a reason to expose this to Celery's users?
And one last thing, we need a docstring for this property.

Issue #4860, opened 2018-06-28
Exception Serialization error? Tasks raising exceptions stuck in "PENDING" state


johnarnold:
## Checklist

- [x] I have included the output of ``celery -A proj report`` in the issue.
      (if you are not able to do this, then at least specify the Celery
       version affected).
- [x] I have verified that the issue exists against the `master` branch of Celery.

## Steps to reproduce
Still working on a standalone repro, but what we see in production is this traceback from a task making requests module POSTs:
```
Client caught exception in _post(). Unable to retry.
Traceback (most recent call last):
  File "/var/lib/celery/venv/lib/python3.6/site-packages/urllib3/connection.py", line 171, in _new_conn
    (self._dns_host, self.port), sel
f.timeout, **extra_kw)
  File "/var/lib/celery/venv/lib/python3.6/site-packages/urllib3/util/connection.py", line 79, in create_connection
    raise err
  File "/var/lib/celery/venv/lib/python3.6/site-packages/urllib3/util/connection.py", line 69, in create_connection
    sock.connect(sa)
TimeoutError: [Errno 110] Connection timed out

During handling of the 
above exception, another exception occurred:

Traceback (most recent call last):
  File "/var/lib/celery/venv/lib/python3.6/site-packages/urllib3/connectionpool.py", line 600, in urlopen
    chunked=chunked)
  File "/var/lib/celery/venv/lib/python3.6/site-packages/urllib3/connectionpool.py", line 343, in _make_request
    self._validate_conn(conn)
  File "/va
r/lib/celery/venv/lib/python3.6/site-packages/urllib3/connectionpool.py", line 849, in _validate_conn
    conn.connect()
  File "/var/lib/celery/venv/lib/python3.6/site-packages/urllib3/connection.py", line 314, in connect
    conn = self._new_conn()
  File "/var/lib/celery/venv/lib/python3.6/site-packages/urllib3/connection.py", line 180, in _new_conn
    self, "F
ailed to establish a new connection: %s" % e)
urllib3.exceptions.NewConnectionError: <urllib3.connection.VerifiedHTTPSConnection object at 0x7f695118a4a8>: Failed to establish a new connection: [Errno 110] Connection timed out

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/var/lib/celery/venv/lib
/python3.6/site-packages/requests/adapters.py", line 445, in send
    timeout=timeout
[...]

PR #4864, opened 2018-06-29
JSON Serialization fails when exception arguments are not JSON-serializable Fixes #4860

tombo315:
## Description

Fixes #4860 When a using JSON as a serializer and an exception is raised within a task, simple exceptions are handled gracefully. E.g.

```
@app.task
def raise_exc():
   raise Exception('message')
```

However, when the arguments are not serializable, the task is acknowledged and processed but dies when trying to mark the task as failed. Subsequent events are not fired either E.g.

```
import requests

@app.task
def raise_exc():
   raise Exception(requests.ConnectionError)
```

```
Traceback (most recent call last):
  File "/home/thomasbo/v-net-task/lib/python3.6/site-packages/celery/app/trace.py", line 396, in trace_task
    I, R, state, retval = on_error(task_request, exc, uuid)
  File "/home/thomasbo/v-net-task/lib/python3.6/site-packages/celery/app/trace.py", line 338, in on_error
    task, request, eager=eager, call_errbacks=call_errbacks,
  File "/home/thomasbo/v-net-task/lib/python3.6/site-packages/celery/app/trace.py", line 172, in handle_error_state
    call_errbacks=call_errbacks)
  File "/home/thomasbo/v-net-task/lib/python3.6/site-packages/celery/app/trace.py", line 217, in handle_failure
    call_errbacks=call_errbacks,
  File "/home/thomasbo/v-net-task/lib/python3.6/site-packages/celery/backends/base.py", line 156, in mark_as_failure
    traceback=traceback, request=request)
  File "/home/thomasbo/v-net-task/lib/python3.6/site-packages/celery/backends/base.py", line 317, in store_result
    request=request, **kwargs)
  File "/home/thomasbo/Networking-scripts-tools/project/NetTask/net_task/extensions.py", line 52, in _store_result
    self.set(self.get_key_for_task(task_id), self.encode(meta))
  File "/home/thomasbo/v-net-task/lib/python3.6/site-packages/celery/backends/base.py", line 267, in encode
    _, _, payload = self._encode(data)
  File "/home/thomasbo/v-net-task/lib/python3.6/site-packages/celery/backends/base.py", line 271, in _encode
    return dumps(data, serializer=self.serializer)
  File "/home/thomasbo/v-net-task/lib/python3.6/site-packages/kombu/serialization.py", line 221, in dumps
    payload = encoder(data)
  File "/usr/lib/python3.6/contextlib.py", line 99, in __exit__
    self.gen.throw(type, value, traceback)
[...]

PR #5074, opened 2018-09-26
Fixed #3586 made celery respect exception types when using serializers

### LINKED ISSUE #3586: Celery does not respect exceptions types when using a serializer different than pickle.

auvipy:
rebased version of https://github.com/celery/celery/pull/3592

### DIFF
--- a/celery/backends/base.py
+++ b/celery/backends/base.py
@@ -9,6 +9,7 @@
 from __future__ import absolute_import, unicode_literals
 
 import datetime
+import inspect
 import sys
 import time
 from collections import namedtuple
@@ -34,7 +35,6 @@ from celery.utils.collections import BufferMap
 from celery.utils.functional import LRUCache, arity_greater
 from celery.utils.log import get_logger
 from celery.utils.serialization import (create_exception_cls,
-                                        ensure_serializable,
                                         get_pickleable_exception,
                                         get_pickled_exception)
 
@@ -236,9 +236,14 @@ class Backend(object):
         serializer = self.serializer if serializer is None else serializer
         if serializer in EXCEPTION_ABLE_CODECS:
             return get_pickleable_exception(exc)
+        # retrieve exception original module
+        exc_module = inspect.getmodule(type(exc))
+        if exc_module:
+            exc_module = exc_module.__name__
+
         return {'exc_type': type(exc).__name__,
-                'exc_message': ensure_serializable(exc.args, self.encode),
-                'exc_module': type(exc).__module__}
+                'exc_args': exc.args,
+                'exc_module': exc_module}
 
     def exception_to_python(self, exc):
         """Convert serialized exception to Python exception."""
--- a/celery/utils/serialization.py
+++ b/celery/utils/serialization.py
@@ -13,10 +13,8 @@ from inspect import getmro
 from itertools import takewhile
 
 from kombu.utils.encoding import bytes_to_str, str_to_bytes
-
 from celery.five import (bytes_if_py2, items, python_2_unicode_compatible,
                          reraise, string_t)
-
 
 from .encoding import safe_repr
 
--- a/celery/utils/serialization.py
+++ b/celery/utils/serialization.py
@@ -12,9 +12,11 @@ from importlib import import_module
 from inspect import getmro
 from itertools import takewhile
 
+from kombu.utils.encoding import bytes_to_str, str_to_bytes
+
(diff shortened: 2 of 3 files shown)


auvipy:
@johnarnold

auvipy:
this is weird! isort is giving different errors in local and CI!

auvipy:
seems an isort issue! should we move to black formatter?

auvipy:
decided to disable isort from travis https://github.com/celery/celery/pull/5082

thedrow:
I'm sorry I missed this.
Last important change in this area was done in #4864 which resolves #4860.
Are we sure all the tests pass and exceptions with arguments that are not JSON-serializable are still serialized correctly?
Do we still need `ensure_serializable()` since that was added in #4864?

auvipy:
I have some confusion in mind, I probably mixed up. let me investigate and get to you soon.

auvipy:
investigating properly it seems I totally overlooked the recent changes

cscetbon:
Hey @auvipy, any progress on that ?

auvipy:
no, actually I was busy with so many other issues of celery

PR #5082, opened 2018-09-28
disablled isort from travis

auvipy:
isort is buggy in travis though it works fine locally.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -8,7 +8,6 @@ envlist =
     apicheck
     configcheck
     pydocstyle
-    isort-check
     bandit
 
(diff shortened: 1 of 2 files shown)


auvipy:
@cscetbon could you review?

tox.ini
```diff
@@ -8,7 +8,6 @@ envlist =
     apicheck
     configcheck
     pydocstyle
```
auvipy: should we keep the tox part as is and disable the Travis part only?

auvipy:
will re add the tox part if needed

PR #5086, opened 2018-09-29
Update celeryd.conf

specialorange:
Move the directory of the program before the execution of the command/script

### DIFF
--- a/extra/supervisord/celeryd.conf
+++ b/extra/supervisord/celeryd.conf
@@ -3,15 +3,9 @@
 ; ==================================
 
 [program:celery]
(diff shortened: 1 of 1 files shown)


georgepsarakis:
@specialorange thanks for the PR. Could you please describe the changes and what they improve?

specialorange:
I thought I included them.  Sorry.

I moved the command to start celery after the declaration of the directory the command should be run in.   The conf settings are interpreted in serial, not allowing the correct directory to be run all the time.  This allows the conf to set the directory, then run the command.

PR #5091, opened 2018-10-06
Improved message signing (fixed #5056)

### LINKED ISSUE #5056: `auth` seems to be horribly broken
## Checklist

       version affected).

## Report

```
software -> celery:4.2.1 (windowlicker) kombu:4.2.1 py:3.6.6
            billiard:3.5.0.4 redis:2.10.6
settings -> transport:redis results:redis://redis:6379/

[... Django-related settings redacted ...]

CELERY_ACCEPT_CONTENT: ['auth']
CELERY_BROKER_URL: 'redis://redis:6379//'
CELERY_RESULT_BACKEND: 'redis://redis:6379/'
CELERY_RESULT_SERIALIZER: 'auth'
CELERY_SECURITY_CERTIFICATE: '/private/keys/celery/public.pem'
CELERY_SECURITY_CERT_STORE: '/private/keys/celery/*.pem'
CELERY_SECURITY_KEY: '********'
CELERY_TASK_SERIALIZER: 'auth'
```

### Security setup silently fails because of default settings

I want to be able to check whether incoming tasks are legitimate. Celery seems to offer message signing, which is perfect for this. I've first followed the docs and added this to the Celery setup script:

```
app.conf.update(
    security_key='/private/keys/celery/private.key',
    security_certificate='/private/keys/celery/public.pem',
    security_cert_store='/private/keys/celery/*.pem')

app.setup_security()
```

That wasn't sufficient enough, though. When using an invalid key, Celery wouldn't even report on that. I dug into the source and found an interesting bit in `celery/security/__init__.py` that checks whether the `task_serializer` setting equals `auth`. If it does not, the security setup is silently ignored. This  doesn't seem like a good thing to me, especially as this rather special behaviour doesn't seem to be noted in the docs.

After adding the configuration for `task_serializer='auth'`, certificates do get loaded and security seems to be set up. Celery also reports on invalid keys when this setting is correctly set.

### Enforcing signed messages

Now that signing of messages seemed to be set up, I encountered another error. I can still publish unsigned messages and the workers will happily pick them up. This just totally defeats the purpose of message signing. Why does the worker still process unsigned messages after explicitly setting up message signing?

I now proceeded to setting `accept_content=['auth']` in order to force Celery to only accept `auth`-serialized messages from the broker. After setting this up, unsigned messages finally get "dismissed", as the docs describe it.

### Unsigned messages crash workers

After sending an unsigned message (json serializer, for example) into the queue, the worker still picks them up. It will then check for the content type and notice that it doesn't match `auth`, which I had set up as the only accepted content type (again, otherwise it would totally defeat the purpose of setting up message signing). It will then raise a ContentDisallowed exception in `celery/worker/strategy.py`. This exception is never catched and will therefor *crash the entire worker*.

The worker will actually put the unsigned message *back into the queue after it's crashed on it*. Any other workers that pick the message up, will also crash and put the message back into the queue. This is a potential DOS flaw and this should never happen. Messages should be "dismissed", such as the docs state, but it should *not* crash the entire worker - and potentially all other workers.

### Messages with invalid signature crash workers

The same issue seems to occur for messages that have been signed by a certificate that is not currently in the worker's cert store. A `SecurityError` exception is raised in `celery/security/certificate.py` which is not properly handled. This is another potential DOS flaw, especially as this bug will also restore the unacknowledged message.

### New workers introduce themselves with an unsigned message

I plan on having multiple servers running Celery workers. This set up exposed another bug that may enable DOS attacks. New workers introduce themselves, but don't sign their "introduction message", so to say. I haven't looked into this procedure yet because of lack of time, but it's simple to reproduce:

* Configure workers to have `accept_content = ['auth']` let them use the `auth` serializer
* Spawn the first worker and leave it up
* Spawn a second worker
* Watch the first ánd second worker crash

The exception is as follows:

```
[2018-09-10 02:33:07,446: CRITICAL/MainProcess] Unrecoverable error: ContentDisallowed('Refusing to deserialize disabled content of type json (application/json)',)
  File "/usr/lib/python3.6/site-packages/celery/worker/worker.py", line 205, in start
    self.blueprint.start(self)
  File "/usr/lib/python3.6/site-packages/celery/bootsteps.py", line 119, in start
    step.start(parent)
  File "/usr/lib/python3.6/site-packages/celery/bootsteps.py", line 369, in start
    return self.obj.start()
  File "/usr/lib/python3.6/site-packages/celery/worker/consumer/consumer.py", line 317, in start
    blueprint.start(self)
  File "/usr/lib/python3.6/site-packages/celery/bootsteps.py", line 119, in start
    step.start(parent)
  File "/usr/lib/python3.6/site-packages/celery/worker/consumer/consumer.py", line 593, in start
    c.loop(*c.loop_args())
  File "/usr/lib/python3.6/site-packages/celery/worker/loops.py", line 91, in asynloop
    next(loop)
  File "/usr/lib/python3.6/site-packages/kombu/asynchronous/hub.py", line 354, in create_loop
    cb(*cbargs)
  File "/usr/lib/python3.6/site-packages/kombu/transport/redis.py", line 1040, in on_readable
    self.cycle.on_readable(fileno)
  File "/usr/lib/python3.6/site-packages/kombu/transport/redis.py", line 337, in on_readable
    chan.handlers[type]()
  File "/usr/lib/python3.6/site-packages/kombu/transport/redis.py", line 667, in _receive
    ret.append(self._receive_one(c))
  File "/usr/lib/python3.6/site-packages/kombu/transport/redis.py", line 697, in _receive_one
    message, self._fanout_to_queue[exchange])
  File "/usr/lib/python3.6/site-packages/kombu/transport/virtual/base.py", line 983, in _deliver
    callback(message)
  File "/usr/lib/python3.6/site-packages/kombu/transport/virtual/base.py", line 633, in _callback
    return callback(message)
  File "/usr/lib/python3.6/site-packages/kombu/messaging.py", line 624, in _receive_callback
    return on_m(message) if on_m else self.receive(decoded, message)
  File "/usr/lib/python3.6/site-packages/celery/worker/consumer/gossip.py", line 201, in on_message
    _, event = prepare(message.payload)
  File "/usr/lib/python3.6/site-packages/kombu/message.py", line 207, in payload
    return self._decoded_cache if self._decoded_cache else self.decode()
  File "/usr/lib/python3.6/site-packages/kombu/message.py", line 192, in decode
    self._decoded_cache = self._decode()
  File "/usr/lib/python3.6/site-packages/kombu/message.py", line 197, in _decode
    self.content_encoding, accept=self.accept)
  File "/usr/lib/python3.6/site-packages/kombu/serialization.py", line 256, in loads
    raise self._for_untrusted_content(content_type, 'disabled')
kombu.exceptions.ContentDisallowed: Refusing to deserialize disabled content of type json (application/json)
```

I'd expect the workers to introduce themselves with a signed message. I can add `json` to the `content_types` config but that, again, totally defeats the purpose of signing these messages.

### No test coverage of the security features

I had a look at the coverage of these security features after having found these bugs. I was surprised by the fact that Codecov reports that these security modules don't have any test coverage. No actual code is tested: https://codecov.io/gh/celery/celery/tree/master/celery/security

So. Please tell me I have horribly misconfigured all of this. I'm not confident in the phrasing that "Celery is written with security in mind" (from the [docs](http://docs.celeryproject.org/en/latest/userguide/security.html)) if these are actual security bugs. I'm also worried about the millions of production systems, in which people may not even know that the security features may not even be turned on (also see "Security setup silently fails because of default settings").

Let me know if you need any more information. I'll try to actively participate in this discussion.

bpereto:
## Description

this improves the message signing described in   
http://docs.celeryproject.org/en/latest/userguide/security.html#message-signing

This fixes #5056 . see the issue for ongoing discussion

---

* [X] better documentation how to configure (`accept_content = ['auth']` and `auth` serializer)
* [X] remove silent fail on `app.setup_security()` if not correctly configured
* [X] Unsigned Messages: `ContentDisallowed` should be dismissed (like a unkown task)
* [ ] Invalid Signatures: `SecurityError` should also be dismissed.
  *  a security error should take attention. some logging/alarming? I don't know what the standard mechanisms in celery are for this case.
* [ ] ~~Handle Pidbox/Event exceptions if serializers don't match (auth - json/json - auth)~~
  * [ ] ~~Gossip~~
  * [ ] ~~Mingle~~
* [X] New Worker should introduce themselves with a signed message
* [ ] ~~Proposal: A configuration setting for verifying the certificates against a CA.~~
* [X] Proposal: Make digest of OpenSSL Sign configurable
  * Change default digest from `sha1` to something modern (sha256) ? Performance impact should be considered. (https://pyopenssl.org/en/stable/api/crypto.html#digest-names)
* [X] Test coverage
* [X] Integration Test
* [X] Replace pyOpenSSL with pyca/crypthography


thedrow:
This is good work! Keep it up.

thedrow:
Could you switch the dependency we're using to cryptography? We can certainly merge this as it is but there is some upside in doing so.
It's much better supported on PyPy and even the authors of PyOpenSSL say "Note: The Python Cryptographic Authority strongly suggests the use of pyca/cryptography where possible. If you are using pyOpenSSL for anything other than making a TLS connection you should move to cryptography and drop your pyOpenSSL dependency."

bpereto:
Sure, I can replace that. Give me some time to do so (and a look into how crypthography works)
I will also add the CA Verify with the new library.

Also open at the moment is the handling of unexpected messages (auth - json, ether way) in the gossip and mingle protocol.

bpereto:
Some Tests are failing already in master?

bpereto:
cryptography seems to me a little bit complexer, an we need some things from the "hazardous" section.
https://cryptography.io/en/latest/hazmat/primitives/asymmetric/rsa/#signing
I can produce a little snippet which shows the lib in action.

thedrow:
The "hazardous" section is simply the low level API. I don't mind using it. It's still public.

thedrow:
And yes, we're currently experiencing some breakage.
We'll have to look at it before the release.

thedrow:
Please fix the integration test.

bpereto:
i don't get it why manager in `t/integration/test_security.p` is not available...

thedrow:
I'm glad I asked for an integration task.
It's still failing.

bpereto:
yeah but I found an issue on my testing and i don't know, why the integration task does not detect this:   
task.delay().get() raises a ContentNotAllowed Exception, because the only accepted serializer is auth (`accept_content`). The `result_serializer` remains in json, because we don't want encrypted content in our result backend.   
Now if I call `.get()` celery wants to decrypt the message from the result backend, which is obviously json.

My proposal is an new setting `result_accept_content` similiar to `accept_content`, to accept json content from the result backend, but enforce signed messages over the broker.

Your thoughts?

bpereto:
Forget certificate Validation as it's not provided yet from pyca/cryptography:
https://github.com/pyca/cryptography/issues/2381

thedrow:
Is not having certificate validation a problem?

bpereto:
> Is not having certificate validation a problem?

nope. the current status also has no certificate verification of the CA. Therefore this would be an extension. 
Also, the validity is not checked at the moment.

thedrow:
If it's not already done, I'd rather complete this PR as soon as possible and add new features later.

bpereto:
As my time is limited I would suggest to postpone `Handle Pidbox/Event exceptions if serializers don't match (auth - json/json - auth)` for the release as it would take some additional time so understand and modify the pidbox and even handlers.
The impact for now is, that a non-secured worker crashes, if a secured-worker joins or is already connected.

I can implement the dismiss of `Invalid Signatures: SecurityError should also be dismissed. ` in the next few days.
Since we have not discussed about alarming, we can also postpone this feature.

The last thing is stabilize the test suite for this PR. The current state (runtime or testsuite) is working on my machine, but fails in travis.

auvipy:
could you rebase and squash commits logically?

thedrow:
No need to squash as we squash before merging anyway.
Rebasing should resolve your build errors.

thedrow:
We fixed the problems we had with Python 3.7 support.
The build should still fail as it depends on a fix from Kombu (which was merged but not released) but at least now we know why it fails.

thedrow:
Upon further investigation it seems that none of the configuration changes related to security have been applied to the worker.

thedrow:
I opened #5269 regarding the test failure in our integration suite.
Since the test fails because of our testing infrastructure, I think we should mark it as xfailed and move on with merging this PR.
@auvipy @georgepsarakis @bpereto @xirdneh Any objections?

thedrow:
I think that there are two remaining issues.
We probably need to create a PR for them.

PR #5110, opened 2018-10-15
Fix migrate task to work with both v1 and v2 of the message protocol

hedleyroos:
`contrib.migrate` is only compatible with version 1 of the message protocol (see http://docs.celeryproject.org/en/latest/internals/protocol.html). The only issue is that task id has been removed from body in v2, but luckily the task id is present in both v1 and v2 in the message object under a different name - correlation_id. This PR fetches task id from the message object.

### DIFF
--- a/celery/contrib/migrate.py
+++ b/celery/contrib/migrate.py
@@ -380,7 +380,7 @@ def move_by_idmap(map, **kwargs):
         ...   queues=['hipri'])
     """
     def task_id_in_map(body, message):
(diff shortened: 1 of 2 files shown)


hedleyroos:
flake8 is now happy with the part I've modified.

jheld:
@auvipy what's the status of this?

auvipy:
surprisingly I missed this one. this needs to pull from master

PR #5114, opened 2018-10-15
fix: fix celery.Scheduler._when use utctimetuple replace timetuple

### LINKED ISSUE #5113: Why beat.Schedle _when return this???
## Steps to reproduce
**https://github.com/celery/celery/blob/master/celery/beat.py#L261**
```python
        return (mktime(as_now.timetuple()) +
                     as_now.microsecond / 1e6 +
                     (adjust(next_time_to_run) or 0))
```
What about the as_now timezone???
## Expected behavior
like this?
```python
return (as_now.timestamp() + (adjust(next_time_to_run) or 0))
```
But timestamp not support for python2.7?
```python
        return (mktime(as_now.utctimetuple()) +
                     as_now.microsecond / 1e6 +
                     (adjust(next_time_to_run) or 0))
```
Or, tell me why wirte like this.

This problem is found because of django-celery-beat.TzAwareCrontab not work as well.
Beacause this schedule.now() return a time with timezone.

yywing:
Fixes #5113

### DIFF
--- a/celery/beat.py
+++ b/celery/beat.py
@@ -10,6 +10,7 @@ import shelve
 import sys
 import time
 import traceback
+from calendar import timegm
 from collections import namedtuple
 from functools import total_ordering
 from threading import Event, Thread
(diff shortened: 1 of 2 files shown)


auvipy:
unit test is producing attribute error

yywing:
Is it return utc timestamp or local timestamp? 
I think return utc timestamp better. So I use `calendar.timegm` not `time.mktime`

https://docs.python.org/2/library/time.html


From | To | Use
-- | -- | --
seconds since the epoch | struct_time in UTC | gmtime()
seconds since the epoch | struct_time in local time | localtime()
struct_time in UTC | seconds since the epoch | calendar.timegm()
struct_time in local time | seconds since the epoch | mktime()

yywing:
And why ci fail? I found master is failed, too.

auvipy:
the failures are not related to this PR

Issue #5118, opened 2018-10-16
`date_done` object in metadata is not serializable.


xirdneh:
## Checklist

- [x] I have included the output of ``celery -A proj report`` in the issue.
      (if you are not able to do this, then at least specify the Celery
       version affected).
```
software -> celery:4.2.0 (windowlicker) kombu:4.2.1 py:3.6.6
            billiard:3.5.0.4 py-amqp:2.3.2
platform -> system:Linux arch:64bit imp:CPython
loader   -> celery.loaders.app.AppLoader
settings -> transport:amqp results:redis://redis:6379/0
```

- [x] I have verified that the issue exists against the `master` branch of Celery.
## Steps to reproduce

This is using: 

    - latest master
    - Django 2.1.2
    - rabbitmq as broker
    - redis as backend

Queue debug task:
```
@app.task(bind=True)
def debug_task(self):
    """Debug Task"""
    print('Request: {0!r}'.format(self.request))
    return "debug task"
```

## Expected behavior

Task gets queued and runs in the worker without any hiccups.

## Actual behavior
Task is not run and following error is displayed:
```
Traceback (most recent call last):
  File "/src/kombu/serialization.py", line 50, in _reraise_errors
    yield
  File "/src/kombu/serialization.py", line 221, in dumps
    payload = encoder(data)
  File "/src/kombu/utils/json.py", line 69, in dumps
    **dict(default_kwargs, **kwargs))
  File "/usr/local/lib/python3.6/json/__init__.py", line 238, in dumps
    **kw).encode(obj)
  File "/usr/local/lib/python3.6/json/encoder.py", line 199, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/usr/local/lib/python3.6/json/encoder.py", line 257, in iterencode
    return _iterencode(o, 0)
  File "/usr/local/lib/python3.6/site-packages/cyanide/data.py", line 18, in default
    return super(JSONEncoder, self).default(obj)
  File "/usr/local/lib/python3.6/json/encoder.py", line 180, in default
    o.__class__.__name__)
TypeError: Object of type 'datetime' is not JSON serializable
```

Also, running `make test` yields some of the result tests failed with the previous error.

## Solution

I believe this was introduced by https://github.com/celery/celery/pull/4490.
The `date_done` object in the metadata is added as a `datetime` object instead of its iso string representation.
**PR is on the way.**

One thing that concerns me is how come tests in travis-CI are running correctly and not catching this?
[...]

Issue #5096, opened 2018-10-07
Update our Contributing document


thedrow:
Our contribution guide is too old and contains obsolete information.
Furthermore, it does not mention the [Needs Test Coverage](https://github.com/celery/celery/labels/Status%3A%20Needs%20Test%20Coverage%20%E2%9C%98) label which is where most of our PRs are abandoned at.

xirdneh:
We probably should add a section on how to easily setup a local dev environment.
I usually run a standalone app in a docker container and point `$PYTHONPATH` to where I clone the celery project. This way when I do `import celery` in the code I'm importing from the cloned project and i can update it by hand.

Is there any prefer method?
If this is OK with everyone I can update the documentation since I would like it to be updated for next week's DjangoCon sprints.

xirdneh:
Good idea, thanks!
I'll look into Cirlce-CI

PR #5119, opened 2018-10-16
Fixes issue #5118

xirdneh:
`date_done` object in metadata is not serializable #5118 . Introduced in #4490.
Fixed by saving iso string when creating metadata dict for `date_done` and parsing it celery utils module when returning the value as attribute.

### DIFF
--- a/celery/backends/base.py
+++ b/celery/backends/base.py
@@ -658,9 +658,9 @@ class BaseKeyValueStoreBackend(Backend):
                       traceback=None, request=None, **kwargs):
 
         if state in self.READY_STATES:
-            date_done = datetime.datetime.utcnow()
+            date_done = datetime.datetime.utcnow().isoformat()
         else:
-            date_done = None
+            date_done = ''
 
         meta = {
             'status': state,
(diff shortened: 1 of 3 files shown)


celery/backends/base.py
```diff
@@ -658,9 +658,9 @@ def _store_result(self, task_id, result, state,
                       traceback=None, request=None, **kwargs):
 
         if state in self.READY_STATES:
-            date_done = datetime.datetime.utcnow()
+            date_done = datetime.datetime.utcnow().isoformat()
         else:
-            date_done = None
+            date_done = ''
```
thedrow: What's wrong with None?

xirdneh: Nothin wrong really. I thought it would be better to keep the value as a string instead of another object (null) when serializing. But if we want to standardize it to `None` I'm ok with it, too.

xirdneh:
This doesn't seem to be an issue running `tox`. I'm going to leave this PR here and try to figure out what's up but this is not a priority.

auvipy:
failures are not related

xirdneh:
This error does not show up when running `tox`.
I believe this is an issue with how the `Dockerfile` sets up the environment.
Working on it in tandem with contributing documentation.
Closing this PR in favor of #5096

PR #5134, opened 2018-10-23
Documentation - add example systemd celerybeat.service file

### LINKED ISSUE #4304: Celery Beat daemonization with systemd (clarify docs?)
I would like to daemonize launch of celery beat. I am using systemd.

[Periodic Tasks page](https://github.com/celery/celery/blob/fcec01f6e041a70e5ddd061beba5fccb32d74e24/docs/userguide/periodic-tasks.rst) in the docs says the following:

>To daemonize beat see [daemonizing](https://github.com/celery/celery/blob/fcec01f6e041a70e5ddd061beba5fccb32d74e24/docs/userguide/daemonizing.rst).

And I see that there are different initd scripts for celery and celery beat. However, the celery.service example for systemd works with celery multi only.

So what is the preferred way to run beat in production with systemd? Should I create a separate service for celery beat, or should I use single systemd service and supply celery multi with the `--beat` option, like suggested here: https://stackoverflow.com/a/23353596/5618728

The Periodic Tasks page says:

>You can also embed beat inside the worker by enabling the workers -B option, this is convenient if you’ll never run more than one worker node, but it’s not commonly used and for that reason isn’t recommended for production use:

The same probably applies to running `celery multi` with `--beat`, but I wanted to know for sure what are the best practices for systemd users.

nthall:

## Description

This commit adds an example systemd celerybeat.service file, and also adds a couple of celerybeat-specific vars to the example EnvironmentFile (Fixes #4304)

Despite the contributing guide specifying there should be no warnings when building docs, I saw several when building with my change, but none are related to this change.

## Scope

There's a deficiency in the documentation for Daemonization - the Periodic Tasks page suggests you can learn more about daemonizing celery beat there, but systemd users find nothing to help them, as noted in #4304. This is just to help folks daemonize celery beat under systemd, and avoid people using things like StackOverflow answers suggesting to start workers with --beat or other settings not recommended for production.

### DIFF
--- a/docs/userguide/daemonizing.rst
+++ b/docs/userguide/daemonizing.rst
@@ -450,6 +450,37 @@ This is an example configuration for a Python project:
     CELERYD_LOG_FILE="/var/log/celery/%n%I.log"
     CELERYD_LOG_LEVEL="INFO"
 
+    # you may wish to add these options for Celery Beat
+    CELERYBEAT_PID_FILE="/var/run/celery/beat.pid"
+    CELERYBEAT_LOG_FILE="/var/log/celery/beat.log"
+
+Service file: celerybeat.service
+----------------------------------------------------------------------
+
+This is an example systemd file for Celery Beat:
+
+:file:`/etc/systemd/system/celerybeat.service`:
+
+.. code-block:: bash
+
+  [Unit]
+  Description=Celery Beat Service
+  After=network.target
+
+  [Service]
+  Type=simple
+  User=celery
+  Group=celery
+  EnvironmentFile=/etc/conf.d/celery
+  WorkingDirectory=/opt/celery
+  ExecStart=/bin/sh -c '${CELERY_BIN} beat  \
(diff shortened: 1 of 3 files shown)


nthall:
I edited my initial message to include a comment about scope. As for the suggestion regarding the unit file, I can add the example as a file in extra/systemd/ but considering that the userguide page on Periodic Tasks specifically refers people to the Daemonization page for info on daemonizing beat, it seems like there should be something there as well. 

When adding that example unit file to extra/systemd/,  would it be welcome or unwelcome to also add CELERYBEAT_PID_FILE and CELERYBEAT_LOG_FILE vars to extra/systemd/celery.conf  as I've done in the doc page?

georgepsarakis:
@nthall thank you. I believe you can add them, but perhaps have them commented out, with an extra note above them that they concern the Celery Beat service and only that. What do you think?

nthall:
Sounds great, will do

PR #5135, opened 2018-10-23
Fix Sphinx support for shared_task decorated functions

jonafato:
Currently, functions decorated with `@app.task` are picked up by
Sphinx's `automodule` directive, but those decorated with `@shared_task`
are not. This is related to the use of `PromiseProxy` in `app.task` vs.
`Proxy` in `shared_task` and how each of those handle `__module__`. This
change allows `celery.contrib.sphinx` to detect and render `shared_task`
functions as well.

### DIFF
--- a/celery/contrib/sphinx.py
+++ b/celery/contrib/sphinx.py
@@ -35,7 +35,7 @@ from sphinx.domains.python import PyModulelevel
 from sphinx.ext.autodoc import FunctionDocumenter
 
 from celery.app.task import BaseTask
-from celery.local import PromiseProxy
+from celery.local import Proxy
 
 try:  # pragma: no cover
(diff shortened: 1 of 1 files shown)


jonafato:
Note: I don't know that this is where this change should be made (vs. an actual functional change to the `shared_task` decorator), but I attempted to address the issue without any changes to the core codebase. I initially thought my issue was solved by #4588, but due to how `Proxy` and `PromiseProxy` treat `__module__` differently, it needed to be more flexible. I don't know enough yet about Celery's internals to say whether or not this needs to handle other cases as well, but I'm happy to update the PR here if that's the case.

auvipy:
redis integration test failures are not related though.

auvipy:
@lpsinger could you share some insight?

PR #5137, opened 2018-10-23
Add s3 result backend

othalla:
## Description

This Pull request add a new s3 result backend.
It supports both :
- AWS S3
- custom object storage s3 compatible backend (like Scality)


auvipy:
did you check this package https://github.com/robgolding/celery-s3?

auvipy:
yes, I am positive about having this in, I just pointed the lib to have some insight from existing lib. thanks for your effort. try to fix the build failures first.

othalla:
@auvipy ok :)
I'm trying, just having issues with missing doc. I don't find where it is missing :/

thomas-illiet:
I like this feature, good luck to pass test :)

othalla:
sorry, closed this PR by mistake :)

othalla:
@auvipy looks like fine, still having integration tests failure but not linked to this PR

othalla:
@auvipy any news regarding this PR?

celery/backends/s3.py
```diff
@@ -0,0 +1,80 @@
...
+    def _get_s3_object(self, key):
+        key_bucket_path = self.base_path + key if self.base_path else key
+        return self._s3_resource.Object(self.bucket_name, key_bucket_path)
+
+    def get(self, key):
+        s3_object = self._get_s3_object(key)
+        try:
+            s3_object.load()
+            return s3_object.get()['Body'].read().decode('utf-8')
+        except botocore.exceptions.ClientError:
```
georgepsarakis: I think it is preferable to raise an exception here or better, check the error code in the exception:

```python
except botocore. exceptions.ClientError as e:
    if e.response['Error']['Code'] == "404":
```

requirements/extras/boto3.txt
```diff
@@ -0,0 +1 @@
+boto3>=1.4.6
```
georgepsarakis: I think we already have a requirement for boto3. Would you mind checking the DynamoDB requirements?

docs/userguide/configuration.rst
```diff
@@ -1143,6 +1154,103 @@ Example configuration
...
+.. _conf-s3-result-backend:
+
+S3 backend settings
+-------------------
+
+.. note::
+
+    This s3 backend driver requires :pypi:`boto3`.
+
+    To install, use :command:`boto3`:
```
georgepsarakis: Perhaps it is easier and more intuitive to just use the backend name instead of the library name directly. So that would be to install:

```
pip install celery[s3]
```

t/unit/backends/test_s3.py
```diff
@@ -0,0 +1,116 @@
+from __future__ import absolute_import, unicode_literals
+
+from case import patch, MagicMock
```
georgepsarakis: The unit tests seem fine, but apart from that, we have used [moto](https://github.com/spulec/moto) in the past, to add some functional tests. You could check the DynamoDB backend code.

celery/backends/s3.py
othalla: Hum i see but we don't have exceptions regarding result backends.
Considering that, best i can do is reraise boto error but this seems useless.
What do you think?

requirements/extras/boto3.txt
othalla: There is 3 requirements for boto3 which are SQS, DynamoDB and this PR.
By checking you mean merging 3 files?

docs/userguide/configuration.rst
othalla: Seems legit :)

celery/backends/s3.py
georgepsarakis: I am not aware of custom exceptions either. My suggestion was to just handle this specific error code and return `None` in this case, otherwise just reraise. What do you think?

requirements/extras/boto3.txt
georgepsarakis: SQS requirements are at kombu, right? Perhaps we could use a symlink for `extras/s3.txt` and `extras/dynamodb.txt` both pointing at this file `extras/boto3.txt`?

georgepsarakis: Thanks for pointing it out @othalla . Unfortunately I cannot come up with a better solution, that will also work with the current `setup.py`. Let's leave it as is then :smile: !

t/unit/backends/test_s3.py
othalla: @georgepsarakis thank you, i didn't knew this program and it looks nice :)
will do it and push again soon

georgepsarakis:
@auvipy yes, I believe so, see https://github.com/celery/celery/pull/5137#discussion_r238030224

othalla:
@georgepsarakis @auvipy just refactored tests using `moto`

georgepsarakis:
@auvipy I noticed this error https://github.com/celery/celery/issues/5175 being raised in Travis CI and this has probably been fixed. Do you have any suggestions?

othalla:
Let me know if i still have work to do!

othalla:
Any update regarding this PR?

auvipy:
please rebase again on top of master and try to pass all relevant tests.

othalla:
fixed

thomas-illiet:
Congratulations for passing the test :D

georgepsarakis:
@othalla thank you very much for your contribution!

PR #5145, opened 2018-10-24
ADD: local thread context pubsub client for RedisConsumer

### LINKED ISSUE #4363: AttributeError 'list' object has no attribute 'decode' with redis backend
## Checklist

```
software -> celery:4.1.0 (latentcall) kombu:4.1.0 py:3.5.2
            billiard:3.5.0.3 redis:2.10.5
settings -> transport:redis results:redis://:**@****************

task_ignore_result: True
accept_content: {'pickle'}
result_serializer: 'pickle'
result_backend: 'redis://:********@************************'
task_serializer: 'pickle'
task_send_sent_event: True
broker_url: 'redis://:********@************************'
```
redis-server version, both 2.x and 3.x

## Steps to reproduce

Hello, i'm not sure what can cause the problems and already tried to find a simillar solution, but no luck so far. Therefore opening issue here, hopefully it helps

The issue is described there as well (not by me): https://github.com/andymccurdy/redis-py/issues/612

So far the experience is, its happen in both cases, where backend is and isn't involved (means just when calling `apply_async(...)`)

Exception when calling `apply_async()`
attributeerror___list__object_has_no_attribute__decode_

Exception when calling `.get()` (also this one has int, instead of list)
attributeerror___list__object_has_no_attribute__decode_

Hope it helps

## Expected behavior

To not throw the error.

## Actual behavior

AttributeError: 'list' object has no attribute 'decode'

Thanks!

sihrc:

## Description
This PR introduces a local thread context for the ResultConsumer class for the Redis backend that works around the lack of thread safety in redis' Pubsub client.

Fixes #4363

### DIFF
--- a/celery/backends/redis.py
+++ b/celery/backends/redis.py
@@ -2,9 +2,9 @@
 """Redis result store backend."""
 from __future__ import absolute_import, unicode_literals
 
-import threading
 from functools import partial
 from ssl import CERT_NONE, CERT_OPTIONAL, CERT_REQUIRED
+import threading
 
 from kombu.utils.functional import retry_over_time
 from kombu.utils.objects import cached_property
@@ -103,11 +103,6 @@ class ResultConsumer(BaseResultConsumer):
         self._maybe_cancel_ready_task(meta)
 
     def start(self, initial_task_id, **kwargs):
-        if self._pubsub is None:
-            self._pubsub = self.backend._create_client(
-                **self.backend.connparams
-            ).pubsub(ignore_subscribe_messages=True)
-
         self._consume_from(initial_task_id)
 
     def on_wait_for_pending(self, result, **kwargs):
@@ -143,7 +138,12 @@ class ResultConsumer(BaseResultConsumer):
 
     @property
     def _pubsub(self):
-        return getattr(self._thread, "_pubsub", None)
+        if getattr(self._thread, "_pubsub", None) is None:
+            self._thread._pubsub = self.backend._create_client(
+                **self.backend.connparams
+            ).pubsub(ignore_subscribe_messages=True)
+
+        return self._thread._pubsub
 
     @_pubsub.setter
     def _pubsub(self, value):
--- a/celery/backends/redis.py
+++ b/celery/backends/redis.py
@@ -145,7 +145,7 @@ class ResultConsumer(BaseResultConsumer):
 
     @property
     def _pubsub(self):
-        return getattr(self._thread, "_pubsub", None)
+        return getattr(self._thread, "pubsub", None)
 
 class RedisBackend(BaseKeyValueStoreBackend, AsyncBackendMixin):
     """Redis task result store."""
--- a/celery/backends/redis.py
+++ b/celery/backends/redis.py
@@ -139,9 +139,9 @@ class ResultConsumer(BaseResultConsumer):
     @property
     def _pubsub(self):
         if getattr(self._thread, "_pubsub", None) is None:
-            self._thread._pubsub = self.backend._create_client(
-                **self.backend.connparams
-            ).pubsub(ignore_subscribe_messages=True)
+            self._thread._pubsub = self.backend.create_client().pubsub(
+                ignore_subscribe_messages=True
+            )
 
         return self._thread._pubsub
 
@@ -417,6 +417,9 @@ class RedisBackend(BaseKeyValueStoreBackend, AsyncBackendMixin):
 
     @cached_property
     def client(self):
+        return self._create_client(**self.connparams)
+
+    def create_client(self):
         return self._create_client(**self.connparams)
 
     def __reduce__(self, args=(), kwargs={}):
--- a/celery/backends/redis.py
+++ b/celery/backends/redis.py
@@ -2,6 +2,7 @@
 """Redis result store backend."""
 from __future__ import absolute_import, unicode_literals
 
+import threading
 from functools import partial
 from ssl import CERT_NONE, CERT_OPTIONAL, CERT_REQUIRED
 
@@ -76,7 +77,7 @@ logger = get_logger(__name__)
 
 
 class ResultConsumer(BaseResultConsumer):
-    _pubsub = None
+    _thread = threading.local()
 
     def __init__(self, *args, **kwargs):
         super(ResultConsumer, self).__init__(*args, **kwargs)
@@ -102,9 +103,11 @@ class ResultConsumer(BaseResultConsumer):
         self._maybe_cancel_ready_task(meta)
 
     def start(self, initial_task_id, **kwargs):
-        self._pubsub = self.backend.client.pubsub(
-            ignore_subscribe_messages=True,
-        )
+        if self._pubsub is None:
+            self._pubsub = self.backend._create_client(
+                **self.backend.connparams
+            ).pubsub(ignore_subscribe_messages=True)
+
         self._consume_from(initial_task_id)
 
     def on_wait_for_pending(self, result, **kwargs):
@@ -137,6 +140,14 @@ class ResultConsumer(BaseResultConsumer):
             key = self._get_key_for_task(task_id)
             self.subscribed_to.discard(key)
             self._pubsub.unsubscribe(key)
+
+    @property
+    def _pubsub(self):
+        return getattr(self._thread, "_pubsub", None)
+
+    @_pubsub.setter
+    def _pubsub(self, value):
+        self._thread._pubsub = value
 
 
 class RedisBackend(BaseKeyValueStoreBackend, AsyncBackendMixin):
--- a/celery/backends/redis.py
+++ b/celery/backends/redis.py
@@ -87,7 +87,8 @@ class ResultConsumer(BaseResultConsumer):
 
     def on_after_fork(self):
         try:
-            self.backend.client.connection_pool.reset()
+            if getattr(self._thread, "client", None) is not None:
+                self._thread.client.connection_pool.reset()
             if self._pubsub is not None:
                 self._pubsub.close()
         except KeyError as e:
@@ -139,11 +140,12 @@ class ResultConsumer(BaseResultConsumer):
     @property
     def _pubsub(self):
         if getattr(self._thread, "_pubsub", None) is None:
-            self._thread._pubsub = self.backend.create_client().pubsub(
+            self._thread.client = self.backend.create_client()
+            self._thread.pubsub = self._thread.client.pubsub(
                 ignore_subscribe_messages=True
             )
 
-        return self._thread._pubsub
+        return self._thread.pubsub
 
     @_pubsub.setter
     def _pubsub(self, value):
--- a/celery/backends/redis.py
+++ b/celery/backends/redis.py
@@ -147,10 +147,6 @@ class ResultConsumer(BaseResultConsumer):
 
         return self._thread.pubsub
 
-    @_pubsub.setter
-    def _pubsub(self, value):
-        self._thread._pubsub = value
-
 
 class RedisBackend(BaseKeyValueStoreBackend, AsyncBackendMixin):
     """Redis task result store."""
--- a/celery/backends/redis.py
+++ b/celery/backends/redis.py
@@ -104,6 +104,12 @@ class ResultConsumer(BaseResultConsumer):
         self._maybe_cancel_ready_task(meta)
 
     def start(self, initial_task_id, **kwargs):
+        if self._pubsub is None:
+            self._thread.client = self.backend.create_client()
+            self._thread.pubsub = self._thread.client.pubsub(
+                ignore_subscribe_messages=True
+            )
+
         self._consume_from(initial_task_id)
 
     def on_wait_for_pending(self, result, **kwargs):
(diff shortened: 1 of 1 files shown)


celery/backends/redis.py
```diff
@@ -102,9 +103,11 @@ def on_state_change(self, meta, message):
         self._maybe_cancel_ready_task(meta)
 
     def start(self, initial_task_id, **kwargs):
-        self._pubsub = self.backend.client.pubsub(
-            ignore_subscribe_messages=True,
-        )
+        if self._pubsub is None:
```
thedrow: Shouldn't we do this inside the property getter instead?

celery/backends/redis.py
```diff
@@ -2,6 +2,7 @@
 """Redis result store backend."""
 from __future__ import absolute_import, unicode_literals
 
+import threading
```
thedrow: Please ensure import order is correct according to isort.

celery/backends/redis.py
```diff
@@ -102,9 +103,11 @@ def on_state_change(self, meta, message):
         self._maybe_cancel_ready_task(meta)
 
     def start(self, initial_task_id, **kwargs):
-        self._pubsub = self.backend.client.pubsub(
-            ignore_subscribe_messages=True,
-        )
+        if self._pubsub is None:
+            self._pubsub = self.backend._create_client(
```
georgepsarakis: I think the usage of a (new) public method which is the non-cached equivalent of `RedisBackend.client` (see [here](https://github.com/celery/celery/pull/5145/files#diff-accb7090ee45fffb2262cbceeabef002R419)) would require much less testing and simplify this section.

celery/backends/redis.py
```diff
@@ -138,6 +141,14 @@ def cancel_for(self, task_id):
             self.subscribed_to.discard(key)
             self._pubsub.unsubscribe(key)
 
+    @property
```
georgepsarakis: If I understand the proposal of @thedrow :

```python
@property
def pubsub(self):
    if self._thread.pubsub is None:
        # Create pubsub client instance here
        self._thread.pubsub = pubsub_instance
    return self._thread.pubsub 
```
Thus the setter is not required.

thedrow: I don't think we should make `_pubsub` public but otherwise, @georgepsarakis is correct.

thedrow:
Is this ready for another round of reviews?

sihrc:
@thedrow  Yes, sorry for the inactivity. I attempted to make the changes indicated in the comments left by the reviewers. In regards to your previous inquiry about reproduction of this issue, I put together a small repo with an environment replicated via Docker that demonstrates this issue. https://github.com/sihrc/celery_repro I hope that can shed some light on how this might be added to the integration tests.

auvipy:
restarted the build

celery/backends/redis.py
```diff
@@ -138,6 +137,16 @@ def cancel_for(self, task_id):
...
+    def _pubsub(self):
```
georgepsarakis: With the current implementation, the `_pubsub` property will never be `None`. You can view several sections in the class making this check:
- https://github.com/celery/celery/pull/5145/files#diff-accb7090ee45fffb2262cbceeabef002R135
- https://github.com/celery/celery/pull/5145/files#diff-accb7090ee45fffb2262cbceeabef002R124
- https://github.com/celery/celery/pull/5145/files#diff-accb7090ee45fffb2262cbceeabef002R92

I would suggest the following:
- Create the Redis client and PubSub instance during initialization in the [start](https://github.com/celery/celery/pull/5145/files#diff-accb7090ee45fffb2262cbceeabef002L105) method.
- Have this property fetch the attribute from the thread local: 
```python
@property
def _pubsub(self):
    return getattr(self._thread, '_pubsub', None)
```

@thedrow does that sound logical? @sihrc I believe this approach will minimize the changes and effort as well.

celery/backends/redis.py
```diff
@@ -408,6 +417,9 @@ def ConnectionPool(self):
     def client(self):
         return self._create_client(**self.connparams)
```
georgepsarakis: I guess we could now call the new public method here (`create_client`), which could also be considered as a means of testing the new method, without actually adding any new test cases.

celery/backends/redis.py
thedrow: That's strange. If `_pubsub` is `None` nothing happens and nothing is logged.
Is it because it wasn't started yet?

sihrc:
Thank you for the guidance @georgepsarakis @thedrow. I've made changes to how _pubsub is handled - though I'd like to mention that the creation was at first in the `start` method, but I moved it out based on @thedrow  suggestion. Wanted to make sure we're tracking motivations for one over the other.

celery/backends/redis.py
jamespeacock: Setting this property does not work this way...I forked off this branch and this method was not the solution. The _pubsub property is delegated to the current self._thread property. See below because there is a typo in the property getter @sihrc

celery/backends/redis.py
jamespeacock: This needs to be 'pubsub' attribute as you are setting self._thread.pubsub in the start method.

jamespeacock: All in all, @sihrc I have tested your changes and they cleanly fix the multithreading problem we were facing when asynchronously getting task results back from redis backend. Great stuff Chris! I'm looking forward to this being approved and hopefully Celery team does a release of 4.2.2 because a lot of people are looking for these changes!

celery/backends/redis.py
thedrow: Would you mind contributing the fix?

jamespeacock: I won't be able to push to @sihrc's branch unless I have permission on his fork unfortunately. And I don't think opening a new PR from my fork is a great idea. Hoping Chris will see this in the next couple days.

celery/backends/redis.py
sihrc: Nice catch, thanks @jamespeacock !

sihrc:
I've been using this solution for some time now, and I'm wondering how this affects the number of redis connections open. It seems like connections are opened and then left to be idle while new connections are created.

celery/backends/redis.py
thedrow: The tests should have failed because of this error.
Please make sure they do fail.

georgepsarakis:
> I've been using this solution for some time now, and I'm wondering how this affects the number of redis connections open. It seems like connections are opened and then left to be idle while new connections are created.

@sihrc I think we have an open issue regarding a memory leak https://github.com/celery/celery/pull/4839, which could also extend to several resources such as open connections not being removed due to not being properly garbage-collected.

Now regarding the changes here, I would advise against defining the `threading.local` as a class attribute https://github.com/celery/celery/pull/5145/files#diff-accb7090ee45fffb2262cbceeabef002R80 . Although I cannot justify it exactly for this case, sharing the same mutable object reference among multiple instances may have unexpected side-effects.

celery/backends/redis.py
```diff
@@ -102,9 +104,12 @@ def on_state_change(self, meta, message):
         self._maybe_cancel_ready_task(meta)
 
     def start(self, initial_task_id, **kwargs):
-        self._pubsub = self.backend.client.pubsub(
-            ignore_subscribe_messages=True,
-        )
+        if self._pubsub is None:
```
jamespeacock: @sihrc Re: @georgepsarakis's concern about setting threading.local() as a class attribute, you could assign self._thread to threading.local() inside start() if it is not yet an attribute.

georgepsarakis: Thanks for the feedback @jamespeacock and also for testing this solution!

celery/backends/redis.py
tuxinhang1989: I think there should be one redis client and one connection_pool, we only need different pubsub objects in different threads.

tuxinhang1989:
I dont understand why `greenletDrainer` need a new greenlet to call `result_consumer.drain_events` infinitely, should't we call it in current greenlet, just like multi-thread env? In this case, this solution can
solve the case I mentioned above.

akhalyavka:
I'm slightly concerned about this PR. Use of TLS as storage for `PubSub` instances will readily exhaust a connection pool in a highly-concurrent environment. 
While very high concurrency is not usually the case with "pure" threads, it's usually the case with gevent/eventlet. 
Also, connections are never explicitly returned to connection pool, so this code implicitly relies on PubSub.__del__ being called, and that might not ever happen.

akhalyavka:
IMO, the proper approach here would be careful synchronization. 
Good news: subscribe/unsubscribe calls only write to socket and never read, so we actually don't need to synchronize `drain_events` with `consume_from` \ `cancel_for`. We only need to make sure that exactly one thread at a time handles `drain_events` and exactly one (not necessarily the same!) thread at a time handles `consume_from` \ `cancel_for`. 
The latter could be achieved with a simple lock on subscribe/unsubscribe calls and the former could be achieved if there was a separate thread that handles `drain_events` (which is already the case for gevent/eventlet per `greenletDrainer` implementation). I'm willing to submit a PR if we reach a consensus here.

Using one lock for all these three operations doesn't strike me as a good idea because there will be a lot of contention on that lock, resulting in possible performance issues.

jamespeacock:
Trying to summarize/recap where this stands:
As of early 2019, this PR needed a test to ensure that these changes don't break other things.
It is my understanding that that was the only outstanding request remaining.

Since then, @tuxinhang1989's concerns for greenlet environments were brought up. It is unclear to me if those concerns were solved by PR: https://github.com/getgoing/celery/pull/5 and if that PR was also merged into celery/celery. 

I am happy to make the test and PR it into @sihrc's branch since he is the owner here. However, I have noticed that the CI jobs on this PR have never succeeded - is that a separate issue @auvipy ?

Lastly, do we need the changes from the getgoing PR in order for this to be approved?

thedrow:
@matusvalo Would you say that this is also fixed by #6416?

matusvalo:
I need to allocate some time and get familiar with this.

thedrow:
@matusvalo This is a friendly reminder to check this PR and whether we should close it.

matusvalo:
@thedrow I will have a look ASAP. Thank you for reminder

matusvalo:
I have checked to code and I think that storing information to backend should be thread safe, since each thread have it's own backend- I have also checked to paths of creating Backend and Result consumer and each thread should have it's own. **BUT** I see there potential problem which can cause that bugs and even this patch is not fixing it:
https://github.com/celery/celery/blob/34c469758f39a4d2ba693d63952df95525666826/celery/backends/redis.py#L538-L540
The code which is fixing this is using client but property for accessing it is using `@cached_property` decorator. This decorator as long as I understand **is not thread safe** (Please double check it here: https://github.com/celery/kombu/blob/master/kombu/utils/objects.py) hence you will end up with shared instance of `client` between threads :-).

One possible and elegant solution is to change cached_property to cache values to threadstore and make everything threadsafe in celery but we need to be sure that there is nothing in celery using global storage....

matusvalo:
BTW sorry for late reply. :-)

thedrow:
@matusvalo 

There is an implementation of a thread safe cached property in [pydanny/cached-property](https://github.com/pydanny/cached-property).
We can introduce a dependency if we must.
Please file an issue on this.

The question here is whether we should close this PR or not.

matusvalo:
In my opinion we should close this issue. I will create an issue regarding cached property.

thedrow:
@sihrc Thank you for your contribution.
Unfortunately it wasn't in a state which we could consider merging.
In any case, this PR is superceeded by #6416.

tahamr83:
#6416 Doesn't fix the problem for us. Getting this stack trace with celery: 5.1.2 and gevent worker pool with redis backend. 

```
[2021-12-13 16:55:18,182: WARNING/MainProcess] Traceback (most recent call last):

[2021-12-13 16:55:18,182: WARNING/MainProcess]   File "src/gevent/greenlet.py", line 906, in gevent._gevent_cgreenlet.Greenlet.run

[2021-12-13 16:55:18,182: WARNING/MainProcess]   File "/Users/tahazaidi/.local/share/virtualenvs/agent_builder-vEaXNefQ/lib/python3.7/site-packages/celery/backends/asynchronous.py", line 80, in run
    self.result_consumer.drain_events(timeout=1)

[2021-12-13 16:55:18,182: WARNING/MainProcess]   File "/Users/tahazaidi/.local/share/virtualenvs/agent_builder-vEaXNefQ/lib/python3.7/site-packages/celery/backends/redis.py", line 152, in drain_events
    message = self._pubsub.get_message(timeout=timeout)

[2021-12-13 16:55:18,182: WARNING/MainProcess]   File "/Users/tahazaidi/.local/share/virtualenvs/agent_builder-vEaXNefQ/lib/python3.7/site-packages/redis/client.py", line 3299, in get_message
    return self.handle_message(response, ignore_subscribe_messages)

[2021-12-13 16:55:18,182: WARNING/MainProcess]   File "/Users/tahazaidi/.local/share/virtualenvs/agent_builder-vEaXNefQ/lib/python3.7/site-packages/redis/client.py", line 3315, in handle_message
    message_type = nativestr(response[0])

[2021-12-13 16:55:18,183: WARNING/MainProcess]   File "/Users/tahazaidi/.local/share/virtualenvs/agent_builder-vEaXNefQ/lib/python3.7/site-packages/redis/_compat.py", line 178, in nativestr
    return x if isinstance(x, str) else x.decode("utf-8", "replace")

[2021-12-13 16:55:18,183: WARNING/MainProcess] AttributeError: 'int' object has no attribute 'decode'

[2021-12-13 16:55:18,183: WARNING/MainProcess] 2021-12-13T11:55:18Z
[2021-12-13 16:55:18,183: WARNING/MainProcess]
[2021-12-13 16:55:18,183: WARNING/MainProcess] <Greenlet at 0x106fbeef0: <bound method greenletDrainer.run of <celery.backends.asynchronous.geventDrainer object at 0x10aeafa50>>> failed with AttributeError
```

PR #5170, opened 2018-11-12
Support getting all results of a chain as a body of a chord

adamrimon:
This pull request adds support for getting all results of a chain when used as a body of a chord.

For example:

Suppose we have some iterator of tasks as `header` and two tasks `step1` and `step2`.
We can now run: `result = chord(header)(step1.s() | step2.s())`
(or `result = (group(header) | step1.s() | step2.s())()`)

In a simple chain, we would be able to use `.parent` recursively to get all results in the chain.
But in this case, the parent of the last result in the body is overwritten with the result of the header group, thus blocks any access to the intermediate results.

This pull requests solves this by adding the header as a parent of the first result in the chain rather then the last.


thedrow (changes requested):
Thank you for the contribution.
Before we can proceed to review this, please make sure to write the appropriate unit and integration tests.

celery/canvas.py
```diff
@@ -1298,7 +1298,14 @@ def run(self, header, body, partial_args, app=None, interval=None,
         else:
             body.delay([])
 
-        bodyres.parent = header_result
+        bodyres_root = bodyres
+        seen = set()
+        while bodyres_root.parent is not None:
```
drsantos20, line 1303: Hi @Prilkop .parent can be None in this line?

auvipy: can you tackle this one with appropriate test?

PR #5192, opened 2018-11-22
Mention password encoding requirement in SQS docs

pawelad:
## Description
Simple SQS documentation clarification (you need to URL encode the password) based on comments in https://github.com/celery/kombu/issues/777

---

Fixes https://github.com/celery/kombu/issues/777

### DIFF
--- a/docs/getting-started/brokers/sqs.rst
+++ b/docs/getting-started/brokers/sqs.rst
@@ -32,7 +32,9 @@ where the URL format is:
 
     sqs://aws_access_key_id:aws_secret_access_key@
 
-you must *remember to include the "@" at the end*.
+You must remember to include the ``@`` sign at the end and URL encode the
+password so it can be parsed correctly. This can be accomplished with
+builtin ``urllib.parse.quote`` function.
 
 The login credentials can also be set using the environment variables
 :envvar:`AWS_ACCESS_KEY_ID` and :envvar:`AWS_SECRET_ACCESS_KEY`,
--- a/docs/getting-started/brokers/sqs.rst
+++ b/docs/getting-started/brokers/sqs.rst
@@ -35,6 +35,8 @@ where the URL format is:
 Please note that you must remember to include the ``@`` sign at the end and
 encode the password so it can be always parsed correctly. For example:
 
+.. code-block:: python
+
     from kombu.utils.url import quote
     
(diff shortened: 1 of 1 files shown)


docs/getting-started/brokers/sqs.rst
```diff
@@ -32,7 +32,9 @@ where the URL format is:
 
     sqs://aws_access_key_id:aws_secret_access_key@
 
-you must *remember to include the "@" at the end*.
+You must remember to include the ``@`` sign at the end and URL encode the
+password so it can be parsed correctly. This can be accomplished with
+builtin ``urllib.parse.quote`` function.
```
auvipy: would it be better to have an example of urllib.parse.quote?

georgepsarakis: @pawelad @auvipy how about adding this, that will work both on Python 2 and 3:

```python
from kombu.utils.url import quote

quote('the/secret/key', safe='') # 'the%2Fsecret%2Fkey'
```

pawelad: One thing I was hesitant about was the use of f strings, but removing them would make that snippet twice as long - let me know what you think.

docs/getting-started/brokers/sqs.rst
```diff
@@ -32,7 +32,17 @@ where the URL format is:
...
+encode the password so it can be always parsed correctly. For example:
+
+.. code-block:: python
+
+    from kombu.utils.url import quote
+    
+    aws_access_key = quote("ABCDEFGHIJKLMNOPQRST")
+    aws_secret_key = quote("ZYXK7NiynGlTogH8Nj+P9nlE73sq3")
+    
+    broker_url = f"sqs://{aws_access_key}:{aws_secret_key}@"
```
auvipy: sadly it's python3 only for now and the next release of celery 4.x version will also support python 2.7. so kindly make it work in all supported version of python and we can switch to  f string from next major versions.

georgepsarakis: I think this version (using `format`) is indeed better, as it covers all supported Python versions.

pawelad: OK, I updated the example - let me know if you'd like me to change anything else!

pawelad:
(Given the number of commits it's probably better to squash these ; )

auvipy:
we always squash merge commits :) thanks

PR #5201, opened 2018-11-27
Require redis version less than 3.0.0 for celery 3.1

reywood:
In versions of the redis library prior to 3.0.0, the `redis.Redis` and `redis.StrictRedis` classes had different parameter orders for the `setex` method. For `redis.Redis`, the definition is `setex(key, value, expires)`, whereas for `redis.StrictRedis`, the definition is `setex(key, expires, value)`. Celery 3.1 requires the parameter order used by `redis.Redis` (see [RedisBackend code here](https://github.com/celery/celery/blob/3.1/celery/backends/redis.py#L166)).

When redis 3.0.0 was released, `redis.Redis` and `redis.StrictRedis` essentially became the same class with `setex` defined with parameters in the order `key, expires, value` which breaks celery 3.1.

This is the simplest change to fix the problem. Alternatively, the `RedisBackend` class could be changed to use `redis.StrictRedis`, but given the legacy nature of celery 3.1, I'm not sure it's worth the work.


thedrow:
I'll try to release a new patch version of 3.1.x as soon as I can.

Issue #4286, opened 2017-09-25
Wrong configuration name CELERY_TASK_ACKS_LATE  in celery 4.1 doc


chenfengyuan:
celery doc url is [v4.1.0](http://docs.celeryproject.org/en/v4.1.0/userguide/configuration.html)
I suppose the correct ack_late configuration name is `CELERY_ACKS_LATE`(old version) or `task_acks_late`(new version), not `CELERY_TASK_ACKS_LATE`.
current(latest, 4.1) doc is misleading

thedrow:
PRs welcome :)

sposs:
I stumbled on this and I think it's relevant to add that for a django config, the name is `CELERY_TASK_ACKS_LATE`, as it's the `CELERY_`-prefixed of the upper-cased version of the new setting: `task_acks_late`. Slightly confusing...

PR #5205, opened 2018-11-29
fix: wrong configuration name CELERY_ACKS_LATE

aviadatsnyk:

## Description
I think https://github.com/celery/celery/pull/4291 was either a mistake, or the implementation should be fixed.
for now, using `CELERY_ACKS_LATE` does not work, while `CELERY_TASK_ACKS_LATE` does.
It also seems to me that the configuration in `task.py` adds `TASK` for all configs in `from_config` (https://github.com/celery/celery/blob/6651e145989d6e890c28273d470bf8fb3a2d5c2b/celery/app/task.py#L328)

### DIFF
--- a/docs/userguide/configuration.rst
+++ b/docs/userguide/configuration.rst
@@ -115,7 +115,7 @@ have been moved into a new  ``task_`` prefix.
 ``CELERY_SECURITY_CERTIFICATE``        :setting:`security_certificate`
 ``CELERY_SECURITY_CERT_STORE``         :setting:`security_cert_store`
 ``CELERY_SECURITY_KEY``                :setting:`security_key`
-``CELERY_ACKS_LATE``                   :setting:`task_acks_late`
+``CELERY_TASK_ACKS_LATE``                   :setting:`task_acks_late`
 ``CELERY_TASK_ALWAYS_EAGER``           :setting:`task_always_eager`
 ``CELERY_TASK_ANNOTATIONS``            :setting:`task_annotations`
 ``CELERY_TASK_COMPRESSION``            :setting:`task_compression`


thedrow:
@auvipy Have we verified this?

clokep:
I think this should be reverted -- that table refers to the *old* setting name and what to use now. So yes, `CELERY_ACKS_LATE` does not work anymore -- that's exactly what that table is saying!

I think there's been a few issues/PRs like this (a couple of which have gotten merged), it makes me wonder if that table is really clear enough or if something around it could be improved.

clokep:
@aviadatsnyk Ah-ha! You're right! That sounds more like a code bug than a documentation issue though! I did go back and check and the old name of that setting is [`CELERY_ACKS_LATE`](http://docs.celeryproject.org/en/3.1/configuration.html#celery-acks-late).

Dem0n3D:
@aviadatsnyk, what's going on? I've just spend all day to find this issue: CELERY_TASK_ACKS_LATE is NOT WORKING for me, but CELERY_ACKS_LATE does (celery 4.3.0). So, #4286 is correct for me.

Dem0n3D:
@aviadatsnyk, you are listed as the author of the commit 6bd31f19. I think that 6bd31f19 should be reverted. Let's ask the committer @auvipy.

Dem0n3D:
A proof https://github.com/celery/celery/blob/master/celery/app/task.py#L314

auvipy:
https://github.com/celery/celery/blob/master/docs/userguide/configuration.rst#task_acks_late

and code says the new lowercase config option is ask_acks_late which was previously different.

PR #5206, opened 2018-11-29
Make error reporting in filesystem.py easier to understand

bt2901:
Currently, the ` _find_path` silently returns `None` and then everything fails with "None has no attribute 'encode'", which isn't very clear.


auvipy:
replaced by #5289

PR #5211, opened 2018-12-03
raise exception if importerrornot caused by missing tasks module

PeterBeklemishev:
## Description
Django 2.0.9, Celery 4.2.0

If tasks module in django app (for example `myapp.tasks`) contains wrong import (for example `from django.db.models import Now`), no exception will be raised and importing from this module will be silently skipped.
If tasks from this module is not imported elsewhere, finding cause of this may be hard.

Cause of this is there all ImportError are treated as "this module does not have `tasks` submodule" (or other specidied related_name)

To solve this in last try-except block we need to check that ImportError `name` attribute equals name of the module we are trying to import, and if not equals - raise original exception, i.e. something like:

### DIFF
--- a/celery/loaders/base.py
+++ b/celery/loaders/base.py
@@ -266,6 +266,7 @@ def find_related_module(package, related_name):
     try:
         return importlib.import_module(module_name)
     except ImportError as e:
-        if getattr(e, 'name', module_name) != module_name:
+        import_exc_name = getattr(e, 'name', module_name)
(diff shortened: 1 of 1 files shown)


auvipy:
could you check the build failures?

PeterBeklemishev:
@auvipy rebase done

auvipy:
thanks, but this error:           
 raise effect
E               ImportError
/opt/python/3.6.7/lib/python3.6/unittest/mock.py:995: ImportError

PR #5213, opened 2018-12-03
Store task args inside MaxRetries exception

Anthony25:
## Description

Allow to easily get the parent task's parameters from an error callback. Before, args and kwargs were contained in the exception message, but it can be useful to store them as attributes inside the exception.


Anthony25:
Hi @auvipy,
I added a test for that, thanks!

auvipy:
any linking issue?

Anthony25:
Nope, I didn't open any issue as the change was quite small.

Thanks for the merge!

Issue #5220, opened 2018-12-04
Group of chains hangs forever


erindy:
I am trying to create a group of chains and when the group gets executed it just stops and hangs, looks very much like deadlock.

I checked rabbitmq and resultbackend and they are running just fine.

here are my dummy tasks to reproduce:

## Steps to reproduce
```
from pprint import pprint
from time import sleep

@app.task
def a(i):
    result = 'A %s' % i
    sleep((i%3)/ 10.0)
    pprint(result)
    return result


@app.task
def b(self,i):
    result = 'B %s' % i
    sleep((i%3)/  10.0)
    pprint(result)
    return result

@app.task
def c(self,i):
    result = 'C %s' % i
    sleep((i%3)/  10.0)
    pprint(result)
    return result
```

and I call them like this:

```
        chains = [chain(a.s(i), b.s(i), c.s(i)) for i in range(100)]
        g = group(*chains)


        result = g.apply_async()
```

also my settings are as follows:

```
# Sensible settings for celery
CELERY_ALWAYS_EAGER = False
CELERY_ACKS_LATE = True
CELERY_TASK_PUBLISH_RETRY = True
CELERY_DISABLE_RATE_LIMITS = False
CELERY_IGNORE_RESULT = False
CELERY_SEND_TASK_ERROR_EMAILS = False
CELERY_TASK_RESULT_EXPIRES = 600
CELERY_REDIS_MAX_CONNECTIONS = 2048
CELERY_TASK_SERIALIZER = "pickle"
CELERY_RESULT_SERIALIZER = "pickle"
CELERY_ACCEPT_CONTENT = ["pickle"]

CELERYD_HIJACK_ROOT_LOGGER = False
CELERYD_PREFETCH_MULTIPLIER = 1
CELERYD_MAX_TASKS_PER_CHILD = 1000

```

## Expected behavior
result should return an async key and the group should be triggered.

## Actual behavior
result = g.apply_async() will just block and stay like that indefinitely, None of the tasks starts in the worker and no error is thrown.

I think some deadlock is occurring somewhere because what happens has its symptoms and no error is thrown at all. The version of celery is 4.1.0 and python 3.6.4 and using rabbitmq as a message queue.

What is also strange is that when I enable CELERY_ALWAYS_EAGER to True then everything runs just fine but it runs locally of course which is not what I want.

danielmeppiel:
Hi, I have exactly the same issue on Celery 5.1.0
Trying to trigger a group of chains simply hangs

[...]

Issue #5138, opened 2018-10-24
task DONT run in complex canvas


vincenthcui:
hi, we are using canvas to design our asynchronous task system, group and chord parallel tasks, chain serial tasks together. But in some complex cases, tasks DONT run as expected. I write a demo and print output here, please help to check :)

## project report

    software -> celery:4.2.1 (windowlicker) kombu:4.2.1 py:2.7.10
                billiard:3.5.0.4 redis:2.10.6
    platform -> system:Darwin arch:64bit imp:CPython
    loader   -> celery.loaders.app.AppLoader
    settings -> transport:redis results:redis://127.0.0.1:6379/1
    
    broker_url: u'redis://127.0.0.1:6379/0'
    result_backend: u'redis://127.0.0.1:6379/1'

## Steps to reproduce

I write a demo here, you can easily run code below to reproduce the issue.

```python
from celery import Celery, group, chain, chord

app = Celery(__name__, broker='redis://127.0.0.1:6379/0', backend='redis://127.0.0.1:6379/1')


@app.task(name='t')
def t(x):
    print x
    return sum(x) + 1 if isinstance(x, list) else x + 1


if __name__ == '__main__':

    tasks = group(
        chain(t.s(), t.s(), t.s()),
        chain(t.s(), t.s(), t.s()),
    )

    tasks = group(
        chain(t.s(), tasks, t.s()),
    )

    # tasks DONT run after chord callbacks
    tasks = chord(tasks, t.s())
    tasks = chain(t.s(1), tasks, t.s(), t.s(), t.s())

    tasks.apply_async()

```

As expected, there should be 13 tasks to be run, but celery run 9 only, below is the output

	celery -A app worker -l info -E

	celery@VINCENTCUI-MC0 v4.2.1 (windowlicker)

	Darwin-17.7.0-x86_64-i386-64bit 2018-10-24 10:50:46

	[config]
	.> app:         app:0x10a0997d0
	.> transport:   redis://127.0.0.1:6379/4
	.> results:     redis://127.0.0.1:6379/5
	.> concurrency: 8 (prefork)
	.> task events: ON

	[queues]
	.> celery           exchange=celery(direct) key=celery

	[tasks]
	  . t

	[2018-10-24 10:50:46,587: INFO/MainProcess] Connected to redis://127.0.0.1:6379/4
	[2018-10-24 10:50:46,592: INFO/MainProcess] mingle: searching for neighbors
[...]

Issue #4681, opened 2018-04-25
Nested chords/groups fail to complete


aequitas:
## Checklist

- [x] I have included the output of ``celery -A proj report`` in the issue.
      (if you are not able to do this, then at least specify the Celery
       version affected).
- [x] I have verified that the issue exists against the `master` branch of Celery.

```
env/bin/celery -A nested_chord_group_test report

software -> celery:4.2.0rc2 (windowlicker) kombu:4.1.0 py:3.6.0
            billiard:3.5.0.3 redis:2.10.6
platform -> system:Darwin arch:64bit imp:CPython
loader   -> celery.loaders.app.AppLoader
settings -> transport:redis results:redis://localhost/

broker_url: 'redis://localhost:6379//'
result_backend: 'redis://localhost/'
```

I encounter an issue when nesting 2 chords with groups, the following construct:

```
    task = group([
        debug_task.si(1) |
        group([
            debug_task.si(2),
            debug_task.si(3)
        ]) | debug_task.si(4)
    ]) | debug_task.si(5, r=True)
```

The source below includes 3 variations with parts commented out that pass and as last example the failing construct.

Is this unsupported behaviour of the canvas or could this be a bug in canvas composition? 

Test source and output of an run is included below.

## Observations
- problem exists with current celery (4.1), upcoming release (4.2) and demonstrated (master)
- in the last (failing) example the `chord` value in task with number 4 is `None`.

## Steps to reproduce

Distilled test reproduction: 

```
# rm -r env; virtualenv env --python=python3
# env/bin/pip install https://github.com/celery/celery/archive/c26e30bad8e141e80f2f62900474121ac52476ac.zip redis
# clear; env/bin/celery worker -A nested_chord_group_test:app -l info & sleep 2; env/bin/python -m nested_chord_group_test; pkill -f nested_chord_group_test:app


from celery import Celery, group

app = Celery(broker='redis://localhost/', backend='redis://localhost/')


@app.task(name="debug_task", bind=True)
def debug_task(self, *args, **kwargs):
    print("{} {}".format(args, kwargs))
    print(self.request)
    return kwargs.get('r')


if __name__ == "__main__":
    # this example passes
    task = group([
        debug_task.si(1) |
        group([
            debug_task.si(2),
            # debug_task.si(3)
        ]) | debug_task.si(4)
    ]) | debug_task.si(5, r=True)

    print(task)
    assert task.apply_async().get(timeout=10) is True
    print("PASS")

    # this example passes
[...]

Issue #4579, opened 2018-03-08
Call .get() hangs indefinitely and a subtask is never called


dmirecki:
## Checklist

- [x] I have included the output of ``celery -A proj report`` in the issue.
      (if you are not able to do this, then at least specify the Celery
       version affected).

```
software -> celery:4.2.0 (latentcall) kombu:4.1.0 py:2.7.9
            billiard:3.5.0.3 py-amqp:2.2.2
platform -> system:Linux arch:64bit, ELF imp:CPython
loader   -> celery.loaders.app.AppLoader
settings -> transport:amqp results:redis://redis:6379/0
```
Problem occurs also on version 4.1.0

- [x] I have verified that the issue exists against the `master` branch of Celery.

## Steps to reproduce

```
@shared_task
def a(msg):
    print msg

job = (
    group(
        a.si(1),
        (a.si(2) | group(a.si(3), a.si(4)) | a.si(5)),
    ) | a.si(6)
)

job.delay().get()
```

## Expected behavior
The job should finish and `6` should be printed.

## Actual behavior
The job does not finish. The output is:
```
 [2018-03-08 09:23:48,047: INFO/MainProcess] Received task: tasks.a[16b045dc-0648-4cca-8034-cff9efe4cb04]  
 [2018-03-08 09:23:48,048: WARNING/ForkPoolWorker-6] 1
 [2018-03-08 09:23:48,050: INFO/ForkPoolWorker-6] Task tasks.a[16b045dc-0648-4cca-8034-cff9efe4cb04] succeeded in 0.00248648200068s: None
 [2018-03-08 09:23:48,056: INFO/MainProcess] Received task: tasks.a[7ade5126-1070-4a69-99db-d50d102f441a]  
 [2018-03-08 09:23:48,057: WARNING/ForkPoolWorker-2] 2
 [2018-03-08 09:23:48,070: INFO/MainProcess] Received task: tasks.a[30dcf45e-340c-4e18-a016-d5fccbcebbf9]  
 [2018-03-08 09:23:48,072: INFO/ForkPoolWorker-2] Task tasks.a[7ade5126-1070-4a69-99db-d50d102f441a] succeeded in 0.0142815270083s: None
 [2018-03-08 09:23:48,071: WARNING/ForkPoolWorker-3] 3
[...]

PR #5222, opened 2018-12-05
Fix chords with chains including sub chords in the group part

the-glu:
## Description

Currently, using chords in chords with a chain doesn't works.

```
    c = chord(
        group([
            chain(
                dummy.si(),
                chord(
                    group([dummy.si(), dummy.si()]),
                    dummy.si(),
                ),
            ),
            chain(
                dummy.si(),
                chord(
                    group([dummy.si(), dummy.si()]),
                    dummy.si(),
                ),
            ),
        ]),
        dummy.si()
    )

    c.delay().get()
```

The last dummy task is never run, when using the redis backend.

It seems that the freeze of the chord class doesn't pass the chord information to his body, leading to this issue.

This PR fix this case.

I wasn't able to fix a subcase: when then the chords are not explicitly set but automatically created when the chain see a group -> task. If you remove the chords in the lowers calls (keeping the chain dummy -> group [ dummy, dummy ] -> dummy ), it will still hang.

There is no test because I have no idea how to do this.

This fix: #5220 #4681 #4579
This partially fix: #5138 (because implicit chord)

I also saw another bugs during my search for this issue but they are closed and/or I didn't find them again.


celery/canvas.py
```diff
@@ -1203,7 +1203,10 @@ def freeze(self, _id=None, group_id=None, chord=None,
             self.tasks = group(self.tasks, app=self.app)
         header_result = self.tasks.freeze(
             parent_id=parent_id, root_id=root_id, chord=self.body)
-        bodyres = self.body.freeze(_id, root_id=root_id)
+
+        bodyres = self.body.freeze(
```
auvipy: better name body_result?

auvipy:
@erind could you check this patch?

celery/canvas.py
the-glu: I didn't changed any existing code, should I clean it up ?

the-glu:
Also for tests I have no idea how, and current tests are already failing, so I'm not sure what to do (nor if this break things...)

celery/canvas.py
auvipy: you could try

the-glu:
I added a basic test not working before but working now

auvipy:
thanks!

erindy:
@auvipy Hi, I am still having the same result, I added this line but it is still hanging:    

  bodyres = self.body.freeze(
            _id, root_id=root_id, chord=chord, group_id=group_id)

Putting a breakpoint in the mentioned line doesn't help because the flow of execution never gets there when I run group(*chains)()

auvipy:
@thedrow @georgepsarakis

thedrow:
@erindy Do you have another test case for us?

thedrow:
@erindy Ping?

auvipy:
should we merge this for 4.3? I think we can fix bugs in minor releases too.

thedrow:
I do want to merge this but I'm not sure which issues it fixes.
According to @erindy, it doesn't fix his issue. So maybe we should introduce his exact test case to this PR and see what happens.
This is going to make it to an RC of 4.3 one way or another.
I just want to make sure not to close unresolved issues prematurely.

the-glu:
Quick remark: I don't have time to test again, but it was working for me when I tested on #5220. However there is probably a detail I missed and/or I failed to reproduce the exact case.

It's probable that this MR doesn't fix all cases when this issue happens (or a similar one..)

It's working in my own case for my project however, so it will be a improvement :)

thedrow:
@the-glu I completely agree. It should be merged.
I'm postponing the merge until I'll get to inspect @erindy's test case more closely.

Issue #961, opened 2012-09-03
Broken requirements/default.txt after merge


daevaorn:
Looks like:

```
billiard>=2.7.3.12
<<<<<<< HEAD
python-dateutil>=2.1
kombu>=2.4.4,<3.0
=======
python-dateutil>=1.5,<2.0
kombu>=2.4.5,<3.0
>>>>>>> 3.0
```

I think somebody haven't done merge conflict resolution:-)

PR #5227, opened 2018-12-07
Address resizing of pool while a connection is in use. Allow option t…

hp685:
…o forcibly resize and quietly ignore the RuntimeError for Resource.resize() which it currently does.

## Description

This goes along with PR #961 of kombu (https://github.com/celery/kombu/pull/961/commits)
Setting force resizing of pool to be true to prevent silent ignore of run time errors and defaulting to pool of 10 connections even when broker_pool_limit is set to a lower value or 0 (disabled)

### DIFF
--- a/celery/app/base.py
+++ b/celery/app/base.py
@@ -299,6 +299,8 @@ class Celery(object):
 
         self.on_init()
         _register_app(self)
+        # Enable force resizing of connection and producer pools
+        pools.set_forced_resize(True)
 
     def _get_default_loader(self):
         # the --loader command-line argument sets the environment variable.


hp685:
@auvipy  Here is the corresponding celery PR for https://github.com/celery/kombu/pull/961/commits

hp685:
done

hp685:
Done. I've validated the test with updated kombu and all tests pass in
app.py. It will fail on master due to dependency on kombu update.

auvipy:
this feature needs kombu new release with your patch before getting accepted

Issue #964, opened 2012-09-06
Autoscale leaves orphan processes.


jorgeecardona:
Hi,

I don't know exactly how to reproduce this but I was getting a lot of problems using celeryd with autoscale together with supervisor, since supervisor restart celeryd if dies I was getting tons of orphans processes of celeryd after a couple of days.

I've change my configuration to just fix concurrency and everything is back to normal.

I imagine then that somehow the celeryd was dead and its child were passed to process 1, then supervisor started again, and this cycle may continue, but the problem is that the task weren't get executed and my rabbitmq was filled with messages.

I was using celery 3.0.1, don't know if this issue is present in the last version, and I don't have plans to move up to a newer version, seems almost always means to change some code.

Bye.

PR #5228, opened 2018-12-10
Force resizing of pool if a connection is in use. Not default to pool…

### LINKED ISSUE #5215: Can only execute a constant number of tasks. All other tasks wait. Issue: connection pool limit reset to 10 (default), intermittently at runtime
Using a broker pool limit of 0 (disabling connection pooling), my application, after running for several days or a few hours, will land into a state where connection pooling gets enabled during runtime. 
Subsequently, it can only process 10 tasks at a time while all other tasks wait until these tasks finish even though I have 40+ workers running. It is potentially an issue with the resize of `Resource` (ProducerPool) instance of kombu Resource class. 

Note this doesn't always happen and is intermittently reproducible but once it lands into that state, it stays in that state until a restart is done.

## Checklist [EDIT: I initially had the config from a wrong client. This is the config of the actual issue]
software -> celery:4.1.0 (latentcall) kombu:4.0.2 py:2.7.13
            billiard:3.5.0.3 librabbitmq:1.5.1
settings -> transport:librabbitmq results:disabled

Workers are started with autoscaling enabled and maxtasksperchild set to 1.

## Steps to reproduce
~~Intermittently send more number of tasks than the set concurrency of workers. This triggers use of autoscale feature for prefork pool.~~

## Expected behavior
Connection pool should always be at to the value specified by config
broker_pool_limit = 0  (I expect rabbitmq admin dashboard to show no pooled connections)
## Actual behavior
RabbitMQ dashboard shows 10 pooled connections even though pooling has been disabled by setting broker_pool_limit to 0

hp685:
… of 10.

## Description
This goes along with PR #964 of kombu (https://github.com/celery/kombu/pull/964/commits)
Uses force resizing of kombu pools to never default to pool of 10.

Fixes #5215

### DIFF
--- a/celery/app/base.py
+++ b/celery/app/base.py
@@ -1162,6 +1162,8 @@ class Celery(object):
         if self._pool is None:
             self._ensure_after_fork()
             limit = self.conf.broker_pool_limit
+            # Enable force resizing of pools
+            pools.set_forced_resize(True)
             pools.set_limit(limit)
             self._pool = pools.connections[self.connection_for_write()]
         return self._pool
--- a/t/unit/app/test_app.py
+++ b/t/unit/app/test_app.py
@@ -959,6 +959,12 @@ class test_App:
         app = CustomCelery(set_as_current=False)
         assert isinstance(app.tasks, TaskRegistry)
 
+    def test_force_resize_called(self):
+        from kombu import pools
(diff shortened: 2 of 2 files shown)


thedrow:
@hp685 Why did you close this PR?

PR #5232, opened 2018-12-13
Fix error callback processing for task based class (Fixes #4377)

### LINKED ISSUE #4377: Using a class based task as errback results in AttributeError '_header_'
## Checklist

       version affected).
[celery-report.txt](https://github.com/celery/celery/files/1456808/celery-report.txt)

## Steps to reproduce
Create a tasks file:
```
from celery import Celery,Task
celery = Celery('tasks', broker='redis://localhost:6379/6',backend='redis://localhost:6379/6')

@celery.task
def ErrTask1(req):
    print('failure')

class CustomTask(Task):
    def run(self,var):
        print('running %s' % var)
        if var == "err":
            raise Exception("err")
CustomTask = celery.register_task(CustomTask())

class ErrTask2(Task):
        def run(self):
                    print('failure')
ErrTask2 = celery.register_task(ErrTask2())

```
start celery with it:
`celery worker -A tasks:celery -l info`

Start a python shell and start the CustomTask with an error callback:

```
from tasks import CustomTask,ErrTask1,ErrTask2
a = CustomTask.apply_async(['err'],link_error=ErrTask1.s())
a.result
b = CustomTask.apply_async(['err'],link_error=ErrTask2.s())
b.result
```
## Expected behavior
When running with ErrTask1 as error callback ( the one with the decorator) we get the expected result:
```
>>> a = CustomTask.apply_async(['err'],link_error=ErrTask1.s())
>>> a.result
Exception(u'err',)
```
## Actual behavior
When running with ErrTask2 as error callback ( the one that is defined classbased) we get the Attributeerror:
```
>>> b = CustomTask.apply_async(['err'],link_error=ErrTask2.s())
>>> b.result
AttributeError(u"'ErrTask2' object has no attribute '__header__'",)
```
It seems the changed behaviour on the error callback in celery 4 has something to do with it. The class based tasks miss some attributes which doesn't seem to be an issue for normal execution but only for the error callback. Probably there is a difference in initialization between `celery.register_task()` and the decorator.

Workaround is to define the error callbacks with the task decorator.

AmbientLighter:
Celery tasks created with the @task decorator have the `__header__` property.
But Celery task created from Task class do not have this property.
That's why we have to check if this property exists before checking is it partial function.

This pull request fixes #4377

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -265,3 +265,4 @@ Tom Booth, 2018/07/06
 Axel haustant, 2018/08/14
 Bruno Alla, 2018/09/27
 Artem Vasilyev, 2018/11/24
+Victor Mireyev, 2018/12/13
--- a/celery/backends/base.py
+++ b/celery/backends/base.py
@@ -170,6 +170,13 @@ class Backend(object):
         for errback in request.errbacks:
             errback = self.app.signature(errback)
             if (
+                # Celery tasks type created with the @task decorator have the
+                # __header__ property, but Celery task created from Task
+                # class do not have this property.
+                # That's why we have to check if this property exists before
+                # checking is it partial function.
+                hasattr(errback.type, '__header__') and
+
                 # workaround to support tasks with bind=True executed as
                 # link errors. Otherwise retries can't be used
                 not isinstance(errback.type.__header__, partial) and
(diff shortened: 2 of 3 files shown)


AmbientLighter:
It looks like the Travis failure is not related to this pull request.

PR #5235, opened 2018-12-15
Code cleanup

othalla:

## Description

Small PR to clean some code.

### DIFF
--- a/celery/backends/amqp.py
+++ b/celery/backends/amqp.py
@@ -141,12 +141,11 @@ class AMQPBackend(BaseBackend):
         if cache and cached_meta and \
                 cached_meta['status'] in READY_STATES:
             return cached_meta
-        else:
-            try:
-                return self.consume(task_id, timeout=timeout, no_ack=no_ack,
-                                    on_interval=on_interval)
-            except socket.timeout:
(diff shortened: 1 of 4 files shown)


celery/backends/amqp.py
```diff
@@ -141,12 +141,11 @@ def wait_for(self, task_id, timeout=None, cache=True,
         if cache and cached_meta and \
                 cached_meta['status'] in READY_STATES:
             return cached_meta
-        else:
-            try:
-                return self.consume(task_id, timeout=timeout, no_ack=no_ack,
-                                    on_interval=on_interval)
-            except socket.timeout:
-                raise TimeoutError('The operation timed out.')
+        try:
```
thedrow: Why remove the else?

othalla:
Thank you.
FYI, plan to do lots of PR like this one, and refactor if i can :)

Just one question, is there any reason not to have pylint in CI?

thedrow:
We have flake8 which runs pylint.

othalla:
They're quite different. Basically, they both respect pep8. Pylint reports more things, sometimes too strict but it can be configured.

PR #5270, opened 2019-01-07
Report kernel version along with other platform details

thedrow:
Since it might be useful to know which kernel version we're running under, I added it to the `celery report` message.


auvipy:
could you check the flake8 error?

PR #5275, opened 2019-01-08
Use DynamoDB TTL to auto-expire old task results

crgwbr:
## Description

Implement a TTL attribute on the DynamoDB result backend and configure the DB to
auto expire documents using the TTL.

### DIFF
--- a/celery/backends/dynamodb.py
+++ b/celery/backends/dynamodb.py
@@ -51,9 +51,13 @@ class DynamoDBBackend(KeyValueStoreBackend):
     #: The endpoint URL that is passed to boto3 (local DynamoDB) (`default`)
     endpoint_url = None
 
+    #: Use DynamoDB TTL to auto-expire results
+    supports_autoexpire = True
+
     _key_field = DynamoDBAttribute(name='id', data_type='S')
     _value_field = DynamoDBAttribute(name='result', data_type='B')
     _timestamp_field = DynamoDBAttribute(name='timestamp', data_type='N')
+    _ttl_field = DynamoDBAttribute(name='ttl', data_type='N')
     _available_fields = None
 
     def __init__(self, url=None, table_name=None, *args, **kwargs):
@@ -123,7 +127,8 @@ class DynamoDBBackend(KeyValueStoreBackend):
         self._available_fields = (
             self._key_field,
             self._value_field,
-            self._timestamp_field
+            self._timestamp_field,
+            self._ttl_field
         )
 
         self._client = None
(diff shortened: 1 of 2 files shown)


crgwbr:
@auvipy @thedrow Added tests as requested.

auvipy:
check the flake8 error

t/unit/backends/test_dynamodb.py
```diff
@@ -158,6 +158,55 @@ def test_wait_for_table_status(self):
         self.backend._wait_for_table_status(expected='SOME_STATE')
         assert mock_describe_table.call_count == 2
 
+    def test_update_table_ttl_change_pending(self):
+        self.backend._client = MagicMock()
+        mock_describe_ttl = self.backend._client.describe_time_to_live = MagicMock()
```
georgepsarakis: [moto](https://github.com/spulec/moto) supports table TTL, so you could use it here instead of creating Mock objects, as a means of functional testing, especially since the local DynamoDB does not support TTL yet.

celery/backends/dynamodb.py
```diff
@@ -223,6 +229,43 @@ def _wait_for_table_status(self, expected='ACTIVE'):
             achieved_state = current_status == expected
             sleep(1)
 
+    def _update_table_ttl(self):
```
georgepsarakis: I think this operation should be optional, since some users may want to retain results.

celery/backends/dynamodb.py
```diff
@@ -246,8 +290,11 @@ def _prepare_put_request(self, key, value):
                     self._value_field.data_type: value
                 },
                 self._timestamp_field.name: {
-                    self._timestamp_field.data_type: str(time())
-                }
+                    self._timestamp_field.data_type: str(_now)
+                },
+                self._ttl_field.name: {
+                    self._ttl_field.data_type: str(int(_now) + int(self.expires))
```
georgepsarakis: Perhaps we should handle cases where `expires` is `None`/`0` which disables expiration, as seen in the [documentation](http://docs.celeryproject.org/en/latest/userguide/configuration.html#result-expires)? 

What happens if you set `NULL` as the TTL field value, if this is possible? Would that perhaps disable expiration for this document?

Additionally, DynamoDB should be added in the documentation list of result backends supporting expiration.

thedrow:
@crgwbr Any chance you'll get to finish this before 4.3 GA?

georgepsarakis:
@crgwbr would you like to resume work here perhaps?

PR #5279, opened 2019-01-10
requirements: use msgpack instead of msgpack-python

blueyed:
msgpack-python is deprecated, and msgpack should be used:
https://pypi.org/project/msgpack-python/

### DIFF
--- a/requirements/extras/msgpack.txt
+++ b/requirements/extras/msgpack.txt
@@ -1,1 +1,1 @@
-msgpack-python>=0.3.0
+msgpack


blueyed:
@auvipy 
Always.. ;)

At least it will bump the minimum supported/installable version, since `msgpack` starts at 0.5.0, and it will install msgpack==0.6.0 instead of msgpack-python==0.5.6 now.

There is a failing test (https://travis-ci.org/celery/celery/jobs/477754815#L3034), have not checked if this is a flaky / known failure.

PR #5287, opened 2019-01-14
Code improvements

othalla:

### DIFF
--- a/celery/canvas.py
+++ b/celery/canvas.py
@@ -1389,7 +1389,7 @@ class chord(Signature):
                 tasks = self.tasks.tasks  # is a group
             except AttributeError:
                 tasks = self.tasks
-            if tasks:
+            if len(tasks):
                 app = tasks[0]._app
             if app is None and body is not None:
                 app = body._app
--- a/celery/app/routes.py
+++ b/celery/app/routes.py
@@ -86,7 +86,7 @@ class Router(object):
                 return lpmerge(self.expand_destination(route), options)
         if 'queue' not in options:
             options = lpmerge(self.expand_destination(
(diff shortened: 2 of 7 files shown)


celery/canvas.py
```diff
@@ -1380,7 +1380,7 @@ def _get_app(self, body=None):
                 tasks = self.tasks.tasks  # is a group
             except AttributeError:
                 tasks = self.tasks
-            if tasks:
+            if len(tasks):
```
auvipy: the reason for this change?

othalla: using len is quite heavy to know if a list or something is empty as it requires parsing it totaly

clokep: In Python, calling `len()` on a `list` (or `dict` or `set`) is `O(1)` (see this [StackOverflow](https://stackoverflow.com/questions/12342457/what-is-the-big-o-notation-for-the-len-function-in-python) for more information). It *does not* require iterating the list.

Personally, I find it more readable without the `len()` though. I don't think there's a performance benefit or penalty either way.

othalla: okok i see

thedrow:
What's the status here?

othalla:
@thedrow i don't have any news about this PR, should i fix something?

thedrow:
Simply rebase. I need to see CI green before merging this.

othalla:
@thedrow  done rebase

clokep:
I think this had unhandled comments above about the use of `len()`?

thedrow:
Yeh we can remove that one...

othalla:
should i do something regarding your comments ?

thedrow:
Just remove len from that condition.

PR #5288, opened 2019-01-16
[WIP] Python 2 compat removal

auvipy:
clean-ups for celery 5. do not merge.


auvipy:
let's not duplicate the effort. there are many docs/bug issues which need triage or patch.

celery/__init__.py
```diff
@@ -1,12 +1,11 @@
-# -*- coding: utf-8 -*-
```
thedrow: why did you remove this? It's not related to Python 2

auvipy: # -*- coding: utf-8 -*- ? this? might be mistakenly.

thedrow: No it isn't.
https://www.python.org/dev/peps/pep-0263/

auvipy: https://github.com/pandas-dev/pandas/pull/26197 and django removed them as well

jdufresne: Starting with Python 3.0, utf-8 is the default encoding of source files:

https://www.python.org/dev/peps/pep-3120/

This makes the utf-8 encoding cookie redundant.

auvipy:
continued in https://github.com/celery/celery/pull/5684

PR #5289, opened 2019-01-16
Meaningful error messages for filesystem backend

larsrinn:
I took over #5206 because it was missing tests. There @bt2901 wrote

> Currently, the _find_path silently returns None and then everything fails with "None has no attribute 'encode'", which isn't very clear.

Apart from that, he also allowed urls to be `file://localhost/path/to/directory`. **Does this require a change in the documentation?** I'd say it's fine without.

For both changes I added tests but didn't really change anything of the actual code.

### DIFF
--- a/celery/backends/filesystem.py
+++ b/celery/backends/filesystem.py
@@ -20,6 +20,10 @@ except NameError:
 
 default_encoding = locale.getpreferredencoding(False)
 
+E_NO_PATH_SET = 'You need to configure a path for the file-system backend'
+E_PATH_NON_CONFORMING_SCHEME = (
+    'A path for the file-system backend should conform to the file URI scheme'
+)
 E_PATH_INVALID = """\
 The configured path for the file-system backend does not
 work correctly, please make sure that it exists and has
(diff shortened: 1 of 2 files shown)


auvipy:
could you check the failure if they are unrelated?

larsrinn:
The test failures are the same as in #5291 https://travis-ci.org/celery/celery/builds/480974665 

As both PRs touch entirely different areas of the code (and none of them areas that cause the test failures), I'd say the failing tests are not related to the changes.

The failing tests are:
* Python 2.7, pypy 2.7 
  *  `t/unit/utils/test_serialization.py::test_strtobool.test_default_table[off-False]` -> `AssertionError: Pattern 'Cannot coerce 'foo' to type bool' not found in 'Cannot coerce u'foo' to type bool'`
  * ~No idea why this started failing~
  * The test was introduced in #5276 and always failed on Python 2.
* Python 3.7
  * `t/unit/worker/test_loops.py::test_asynloop.test_poll_err_readable` -> `RuntimeError: generator raised StopIteration` 
  * this is because Python 3.7 converts `StopIteration` inside a generator to `RuntimeError`
* The redis related tests:
  * `t/integration/test_canvas.py::test_chord.test_single_task_header` -> `celery.exceptions.TimeoutError: The operation timed out.`
  * The first time this failure occured was in #5277

PR #5295, opened 2019-01-20
Fix redis integration tests

larsrinn:
#5277 introduced failing integration tests for redis on CI. I traced down the line actually causing the issue and reverted it: https://github.com/celery/celery/pull/5277/files#diff-e27381cb5c64031dd28aa6855c78cfddL1315

I don't know if the failing tests were caused by actual changes in the behaviour of the code or if the test code would have to be changed... If anyone knows what's going on here, feel free to take over.


georgepsarakis:
@larsrinn do you perhaps have a link to a failed Travis CI build that you can add here for context?

PR #5297, opened 2019-01-21
Fix AttributeError when `sig` is not a Signature.

### LINKED ISSUE #5265: Complex canvas when using Redis 3 raise AttributeError: 'dict' object has no attribute 'clone'
WARNING:  I'm still trying to collect all the necessary details and, if possible, provide a reproducible example.  For the time, the error is happening constantly in one server, but doesn't happen in another one with the same source code.

I'm getting the following:

```
  File "eggs/celery-4.2.1-py2.7.egg/celery/app/trace.py", line 439, in trace_task
    parent_id=uuid, root_id=root_id,
  File "eggs/celery-4.2.1-py2.7.egg/celery/canvas.py", line 1232, in apply_async
    return self.run(tasks, body, args, task_id=task_id, **options)
  File "eggs/celery-4.2.1-py2.7.egg/celery/canvas.py", line 1277, in run
    header_result = header(*partial_args, task_id=group_id, **options)
  File "eggs/celery-4.2.1-py2.7.egg/celery/canvas.py", line 953, in __call__
    return self.apply_async(partial_args, **options)
  File "eggs/celery-4.2.1-py2.7.egg/celery/canvas.py", line 978, in apply_async
    args=args, kwargs=kwargs, **options))
  File "eggs/celery-4.2.1-py2.7.egg/celery/canvas.py", line 1054, in _apply_tasks
    **options)
  File "eggs/celery-4.2.1-py2.7.egg/celery/canvas.py", line 557, in apply_async
    dict(self.options, **options) if options else self.options))
  File "eggs/celery-4.2.1-py2.7.egg/celery/canvas.py", line 573, in run
    task_id, group_id, chord,
  File "eggs/celery-4.2.1-py2.7.egg/celery/canvas.py", line 683, in prepare_steps
    task.link_error(errback)
  File "eggs/celery-4.2.1-py2.7.egg/celery/canvas.py", line 1016, in link_error
    sig = sig.clone().set(immutable=True)
AttributeError: 'dict' object has no attribute 'clone'
```

When trying to execute a complex canvas.  The canvas itself is dynamically created.  We have a single task that dispatches the actual calls.  So all, jobs are instances of this task with different arguments.  The task is: https://github.com/merchise-autrement/odoo/blob/merchise-10.0/odoo/jobs.py#L746

The canvas is constructed in: https://github.com/merchise/xopgi.base/blob/master/xopgi/xopgi_cdr/cdr_agent.py#L181

Basically, we have a layered graphs of events that depend on several evidences, which in turn depend on variables.  We compute each variable in a job.  Each variable job is linked to a group of jobs to update the dependent evidences.  Each evidence job is liked to a group of event jobs.  Each job has an on_error callback to signal the whole cycle had some errors.

I'm trying to detect if the issue comes from a particular workflow of jobs, or other cause.

I'm using celery 4.2.1, but I tested with master.

mvaled:
Fixes #5265.  

Test is pending since I've been unable to produce a test.

### DIFF
--- a/celery/canvas.py
+++ b/celery/canvas.py
@@ -1045,7 +1045,11 @@ class group(Signature):
         return self.tasks[0].link(sig)
 
     def link_error(self, sig):
-        sig = sig.clone().set(immutable=True)
+        if not isinstance(sig, Signature):
+            sig['immutable'] = True
+            sig = Signature.from_dict(sig)
+        else:
+            sig = sig.clone().set(immutable=True)
         return self.tasks[0].link_error(sig)
 
     def _prepared(self, tasks, partial_args, group_id, root_id, app,


mvaled:
@auvipy I tried to create a reproducible scenario, but failed to do so in a timely manner.   I as explained in the issue #5265, I had a hard time trying to create the test.  

I'm using this PR's branch in my servers for the time being.  At the moment, I cannot devote much time to this.  I will let you know if my schedules clears up a bit to try to pursue this.

mvaled:
BTW, I've just noticed the tests are failing.  I will make them pass; but, again, I don't have the time to create the required regression test now.

Issue #3631, opened 2016-11-29
WorkerLostError: Worker exited prematurely: exitcode 0 when shutting down worker


tabascoterrier:
I'm new to Celery, so do correct me if I'm missing the obvious!

When Ctrl+Cing (or sending SIGTERM) to the Celery process, Im getting WorkerLostErrors a lot of (but not all) the time:
```
[2016-11-29 16:18:07,837: ERROR/MainProcess] Task handler raised error: WorkerLostError('Worker exited prematurely: exitcode 0.',)
Traceback (most recent call last):
  File "/home/ubuntu/venv/lib/python3.5/site-packages/celery/worker/worker.py", line 203, in start
    self.blueprint.start(self)
  File "/home/ubuntu/venv/lib/python3.5/site-packages/celery/bootsteps.py", line 119, in start
    step.start(parent)
  File "/home/ubuntu/venv/lib/python3.5/site-packages/celery/bootsteps.py", line 370, in start
    return self.obj.start()
  File "/home/ubuntu/venv/lib/python3.5/site-packages/celery/worker/consumer/consumer.py", line 318, in start
    blueprint.start(self)
  File "/home/ubuntu/venv/lib/python3.5/site-packages/celery/bootsteps.py", line 119, in start
    step.start(parent)
  File "/home/ubuntu/venv/lib/python3.5/site-packages/celery/worker/consumer/consumer.py", line 593, in start
    c.loop(*c.loop_args())
  File "/home/ubuntu/venv/lib/python3.5/site-packages/celery/worker/loops.py", line 88, in asynloop
    next(loop)
  File "/home/ubuntu/venv/lib/python3.5/site-packages/kombu/async/hub.py", line 289, in create_loop
    events = poll(poll_timeout)
  File "/home/ubuntu/venv/lib/python3.5/site-packages/kombu/utils/eventio.py", line 84, in poll
    return self._epoll.poll(timeout if timeout is not None else -1)
  File "/home/ubuntu/venv/lib/python3.5/site-packages/celery/apps/worker.py", line 282, in _handle_request
    raise exc(exitcode)
celery.exceptions.WorkerShutdown: 1

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/ubuntu/venv/lib/python3.5/site-packages/billiard/pool.py", line 1224, in mark_as_worker_lost
    human_status(exitcode)),
billiard.exceptions.WorkerLostError: Worker exited prematurely: exitcode 0.
```

[...]

PR #5308, opened 2019-01-25
if exit with EX_OK, won't mark it as worker lost, it's just normal (w…

tinylambda:
https://github.com/celery/celery/issues/3631

### DIFF
--- a/celery/concurrency/asynpool.py
+++ b/celery/concurrency/asynpool.py
@@ -1088,7 +1088,7 @@ class AsynPool(_pool.Pool):
         """
         # if exit with EX_OK, won't mark it as worker lost,
         # it's just normal (warm) shutdown
-        if exitcode == EX_OK:
+        if exitcode == EX_OK and job.ready():
             return
 
         self.mark_as_worker_lost(job, exitcode)
--- a/celery/concurrency/asynpool.py
+++ b/celery/concurrency/asynpool.py
@@ -44,6 +44,7 @@ from celery.platforms import pack, unpack, unpack_from
 from celery.utils.functional import noop
 from celery.utils.log import get_logger
 from celery.worker import state as worker_state
+from celery.platforms import EX_OK
 
 # pylint: disable=redefined-outer-name
 # We cache globals and attribute lookups, so disable this warning.
@@ -1085,6 +1086,11 @@ class AsynPool(_pool.Pool):
         was assigned to exited by mysterious means (error exitcodes and
         signals).
         """
+        # if exit with EX_OK, won't mark it as worker lost,
+        # it's just normal (warm) shutdown
+        if exitcode == EX_OK:
+            return
+
         self.mark_as_worker_lost(job, exitcode)
 
     def human_write_stats(self):


tinylambda:
Why there are so many errors in the CI progress? anyone can help ?

tinylambda:
@auvipy can you review this PR ?

georgepsarakis:
@tinylambda the failures seem to be caused by https://github.com/jaraco/configparser/issues/30

I will restart the build jobs.

tinylambda:
> @tinylambda the failures seem to be caused by [jaraco/configparser#30](https://github.com/jaraco/configparser/issues/30)
> 
> I will restart the build jobs.

Thanks very

tinylambda:
> @auvipy can you review this PR ?

@ask @thedrow

celery/concurrency/asynpool.py
```diff
@@ -1085,6 +1086,11 @@ def on_job_process_lost(self, job, pid, exitcode):
         was assigned to exited by mysterious means (error exitcodes and
         signals).
         """
+        # if exit with EX_OK, won't mark it as worker lost,
```
georgepsarakis: If I am not mistaken, at this point it is not clear whether the process exited **after the job was completed**. Perhaps you also need to check if the `job.ready` [method](https://github.com/celery/billiard/blob/13a52dc1c3dd9ba2f01c255064b33ceb71a9cb85/billiard/pool.py#L1136) returns `True`.

tinylambda: Yes, it's really important to make sure the job is completes before the process exits (which is the warm shutdown case) !

I will improve it, thanks ！

tinylambda: I tested again and found that job.ready() is False, but exitcode is EX_OK, is it a expected state ?

I think the job process exit with EX_OK indicates that the job it executes is also OK, any suggestions ?

> If I am not mistaken, at this point it is not clear whether the process exited **after the job was completed**. Perhaps you also need to check if the `job.ready` [method](https://github.com/celery/billiard/blob/13a52dc1c3dd9ba2f01c255064b33ceb71a9cb85/billiard/pool.py#L1136) returns `True`.

georgepsarakis: @tinylambda I am quite certain that it is not necessary. A process can shutdown gracefully without necessarily completing the task.

thedrow: Please rephrase the comment.
"If the process exists with the EX_OK status code, don't mark the worker as lost"
I think it's clearer that way.
Also,for future reference please mention the relevant issue number.

celery/concurrency/asynpool.py
```diff
@@ -1085,6 +1086,11 @@ def on_job_process_lost(self, job, pid, exitcode):
...
+        # it's just normal (warm) shutdown
+        if exitcode == EX_OK and job.ready():
```
thedrow: Instead of the comment, maybe we should use a helper method?
It would be clearer.

thedrow:
@tinylambda Are you going to finish this PR?

tinylambda:
> @tinylambda Are you going to finish this PR?

Hi @thedrow ， Sorry for the later response.  I think there should be a much better solution to this issue, but I think, it needs a lot of knowledge on the design of Celery architecture and its implementations.

So now, I don't think my PR on this issue is a proper solution. 

I think it's OK to close this PR, and I will supply a better solution on this issue if I have enough time.

PR #5309, opened 2019-01-25
Default priority for all tasks (task_default_priority setting)

madprogrammer:
## Description

There is a `task_queue_max_priority` configuration setting to specify default max priority for all queues, but there is no `task_default_priority` setting to allow specifying an implicit default priority for all tasks.

I wanted all my tasks to implicitly have a "middle" default priority of 5 (while having x-max-priority of 10), and to be able to selectively tune some specific tasks to have a different priority. So I've added such functionality in this pull request. It has been tested in production and seems to work as designed.

### DIFF
--- a/docs/userguide/configuration.rst
+++ b/docs/userguide/configuration.rst
@@ -1923,6 +1923,16 @@ Default: :const:`None`.
 
 See :ref:`routing-options-rabbitmq-priorities`.
 
+.. setting:: task_default_priority
+
+``task_default_priority``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+:brokers: RabbitMQ
+
+Default: :const:`None`.
+
+See :ref:`routing-options-rabbitmq-priorities`.
+
 .. setting:: worker_direct
 
 ``worker_direct``
--- a/docs/userguide/configuration.rst
+++ b/docs/userguide/configuration.rst
@@ -1927,7 +1927,7 @@ See :ref:`routing-options-rabbitmq-priorities`.
 
 ``task_default_priority``
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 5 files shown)


madprogrammer:
@auvipy Test added

madprogrammer:
@auvipy issue fixed, coverage should be ok too now

t/unit/tasks/test_tasks.py
```diff
@@ -774,6 +778,97 @@ def yyy2():
...
+        @self.app.task(shared=False, priority=66)
+        def yyy4():
+            pass
+
+        @self.app.task(shared=False, bind=True, base=TaskWithPriority)
+        def yyy5(self):
+            pass
+
+        self.app.conf.task_default_priority = 42
+        old_send_task = self.app.send_task
```
georgepsarakis: I don't think you need to save and restore the method object, `self.app` is reinstantiated on each test setup.

docs/userguide/configuration.rst
```diff
@@ -1923,6 +1923,16 @@ Default: :const:`None`.
 
 See :ref:`routing-options-rabbitmq-priorities`.
 
+.. setting:: task_default_priority
+
+``task_default_priority``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+:brokers: RabbitMQ
```
georgepsarakis: Although this is a indeed a standard feature of RabbitMQ, the Redis transport also emulates priorities.

georgepsarakis:
@madprogrammer several Travis CI failures have already been resolved in current master branch, could you please rebase?

t/unit/tasks/test_tasks.py
madprogrammer: The restoration part is there because I saw same code in other existing test, and copied it into my test

PR #5313, opened 2019-01-27
task_inherit_parent_priority setting: allow child tasks to inherit priority from parent task

madprogrammer:
## Description

An addition to the other priority-related PR #5309. Allow subtasks to inherit priority set for parent task. When the setting is enabled, the following behavior will apply:

    c = celery.chain(
        add.s(2), # priority=None
        add.s(3).set(priority=5), # priority=5
        add.s(4), # priority=5
        add.s(5).set(priority=3), # priority=3
        add.s(6), # priority=3
    )

As well as this:

    @app.task(bind=True)
    def child_task(self):
        pass

    @app.task(bind=True)
    def parent_task(self):
        child_task.delay()

    # child_task will also have priority=5
    parent_task.apply_async(args=[], priority=5)

### DIFF
--- a/celery/app/base.py
+++ b/celery/app/base.py
@@ -729,6 +729,10 @@ class Celery(object):
                 if not parent_id:
                     parent_id = parent.request.id
 
+                if conf.task_inherit_parent_priority:
+                    options.setdefault('priority',
+                            parent.request.delivery_info.get('priority'))
+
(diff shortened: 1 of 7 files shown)


thedrow:
Can you resolve the conflicts and rebase?

PR #5318, opened 2019-01-31
dropped python 3.4 support from

auvipy:
python 3.4 is going to be EOL soon. I believe we should drop it from celery 4.3 release too. what do you people think? @thedrow @georgepsarakis


WisdomPill:
Please don't add any further delays to the release of 4.3, if celery 4.x is the latest to support python 2, why you should start suppressing a release of python 3 this early in the first place.

georgepsarakis:
@auvipy 3.4.10 seems to be scheduled for [release on March](https://www.python.org/dev/peps/pep-0429/), so there might still be users that need this version, plus it does not seem urgent.

auvipy:
ok deferring for v5

thedrow:
I agree we should push this to the next release.

Issue #5133, opened 2018-10-23
Eager tasks ignore CELERY_*_SERIALIZER settings


muzhig:
It just uses `json`.
The bug is on master, but not on last production version, probably easy to fix it now than later.
Relevant line is this: https://github.com/celery/celery/blob/master/celery/app/task.py#L529

auvipy:
what is the proposed fix?

clokep:
@muzhig Any steps to reproduce this issue? At first glance that code looks OK.

bjackson:
I have this issue as well. 

The producer returned from `app.producer_or_acquire(producer)` has its `serializer` attribute set to `None`. When `serializer.dumps` is called with `serializer=None` it will use the default encoder, which is `json`.

bjackson:
If you are using the pickle serializer, and you call the task with `serializer='pickle'` an `ContentDisallowed` exception will be raised.

bjackson:
I've submitted a PR to fix this issue: https://github.com/celery/celery/pull/5342

PR #5342, opened 2019-02-15
Use task_serializer for doing roundtrip serialization for eager tasks

bjackson:
## Description
Fixes https://github.com/celery/celery/issues/5133
When doing the roundtrip serialization for eager tasks, the task serializer will always be JSON unless the `serializer` argument is present in the call to `Task.apply_async`. If the serializer argument is present but is `'pickle'`, an exception will be raised as pickle-serialized objects cannot be deserialized without specifying to `serialization.loads` what content types should be accepted. The Producer's `serializer` seems to be set to `None`, causing the default to JSON serialization.

This PR will continue using (in order) the `serializer` argument to `Task.apply_async`, if present, or the `Producer`'s serializer if not `None`. If the `Producer`'s serializer is `None`, it will use the Celery app's `task_serializer` configuration entry as the serializer.

### DIFF
--- a/celery/app/task.py
+++ b/celery/app/task.py
@@ -531,14 +531,17 @@ class Task(object):
         if app.conf.task_always_eager:
             with app.producer_or_acquire(producer) as eager_producer:
                 serializer = options.get(
-                    'serializer', eager_producer.serializer
+                    'serializer',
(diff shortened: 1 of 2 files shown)


bjackson:
I'll add one.

PR #5355, opened 2019-02-22
Prepend to sys.path in the Django fixup instead of appending.

### LINKED ISSUE #5347: Django fixup appends to PYTHONPATH instead of prepending
Hi,

## Environment & Settings
**Celery version**: 4.2.1  (windowlicker)

# Steps to Reproduce

We are using Celery + Django in [dissemin](https://github.com/dissemin/dissemin/). We have a Django app named "statistics" which is actually a name conflicting with a Python 3 package from standard library. This should be fine in principle as long as the `PYTHONPATH` is set so that the local modules have precedence over the system ones.

When running Celery CLI, the system wide module has precedence over the local ones apparently.

I traced this issue back to [this `sys.path` tweak](https://github.com/celery/celery/blob/072dab85261599234341cc714b0d6f0caca20f00/celery/fixups/django.py#L60-L61), which is actually **appending** local path instead of prepending it.

I may have missed something, but is it important for some reason to append it and not prepend it in this context?

# Expected Behavior

Celery should load the local module as expected.

# Actual Behavior

```
# When going through celery CLI
sys.path == ['/Users/lverney/.local/share/virtualenvs/dissemin3/bin', '/Users/lverney/.local/share/virtualenvs/dissemin3/lib/python36.zip', '/Users/lverney/.local/share/virtualenvs/dissemin3/lib/python3.6', '/Users/lverney/.local/share/virtualenvs/dissemin3/lib/python3.6/lib-dynload', '/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6', '/Users/lverney/.local/share/virtualenvs/dissemin3/lib/python3.6/site-packages', '/Users/lverney/.local/share/virtualenvs/dissemin3/src/django-allauth', '/Users/lverney/tmp/dissemin']

# Without celery
sys.path == ['', '/Users/lverney/.local/share/virtualenvs/dissemin3/lib/python36.zip', '/Users/lverney/.local/share/virtualenvs/dissemin3/lib/python3.6', '/Users/lverney/.local/share/virtualenvs/dissemin3/lib/python3.6/lib-dynload', '/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6', '/Users/lverney/.local/share/virtualenvs/dissemin3/lib/python3.6/site-packages', '/Users/lverney/.local/share/virtualenvs/dissemin3/src/django-allauth']
```

We can see that current path is actually appended, not prepended. Therefore, "system" modules have precedence on the ones from the local project.

# Workaround

For people experiencing this issue, `PYTHONPATH=$(pwd) celery …` is a workaround.

Best,

/cc @wetneb who first noticed this issue.

wetneb:
This makes sure that project modules have precedence over system ones.
Closes #5347.

## Description

This follows @Phyks's suggestion of a fix for #5347, by prepending instead of appending to the system path, to ensure that the project modules are not hidden by system-wide ones.

### DIFF
--- a/celery/fixups/django.py
+++ b/celery/fixups/django.py
@@ -57,8 +57,10 @@ class DjangoFixup(object):
         self._worker_fixup = None
 
     def install(self):
-        # Need to add project directory to path
-        sys.path.append(os.getcwd())
+        # Need to add project directory to path.
+        # The project directory has precedence over system modules,
+        # so we prepend it to the path.
+        sys.path.prepend(os.getcwd())
 
         self._settings = symbol_by_name('django.conf:settings')
         self.app.loader.now = self.now
--- a/t/unit/fixups/test_django.py
+++ b/t/unit/fixups/test_django.py
@@ -91,7 +91,7 @@ class test_DjangoFixup(FixupCase):
             f.install()
             self.sigs.worker_init.connect.assert_called_with(f.on_worker_init)
             assert self.app.loader.now == f.now
-            self.p.append.assert_called_with('/opt/vandelay')
+            self.p.prepend.assert_called_with('/opt/vandelay')
 
     def test_now(self):
         with self.fixup_context(self.app) as (f, _, _):


wetneb:
Thank you a million @auvipy!

auvipy:
welcome 2 million :) @wetneb

clokep:
I don't think this is correct, Python lists don't have a `prepend` method. Was this tested?

Usually to add something as the first element you would use:
```python
sys.path.insert(0, os.getcwd())
```

This was reported by a user on IRC:
```
AttributeError: 'list' object has no attribute 'prepend'
```

I'd suggest we back this out.

wetneb:
Oops, sorry about that! That's a pretty epic fail on my part.

PR #5356, opened 2019-02-22
Fix call to list.prepend to use list.insert

wetneb:
This fixes #5355 which used the well known `list.prepend` method which does not actually exist. Oops.


auvipy:
strange!! didn't you run that locally?

wetneb:
Not in this exact version I confess! @Phyks had done the testing. Now I can confirm I have run this version.

Issue #4899, opened 2018-07-11
link_error tasks not run eagerly when using apply()


OskarPersson:
## Checklist

- [x] I have included the output of ``celery -A proj report`` in the issue.
      (if you are not able to do this, then at least specify the Celery
       version affected).
- [x] I have verified that the issue exists against the `master` branch of Celery.

Tested using the included docker setup

```
$ celery -A tasks report
No handlers could be found for logger "vagrant"

software -> celery:4.2.0 (windowlicker) kombu:4.2.1 py:2.7.14
            billiard:3.5.0.4 py-amqp:2.3.2
platform -> system:Linux arch:64bit imp:CPython
loader   -> celery.loaders.app.AppLoader
settings -> transport:pyamqp results:redis://redis/

broker_url: u'amqp://guest:********@rabbit:5672//'
result_backend: u'redis://redis/'
```

## Steps to reproduce
Add `tasks.py`:

```python
# tasks.py

@app.task
def on_failure():
    print 'from on_failure'

@app.task
def error():
    raise ValueError('oh no!')
```
Run the tasks:
```python
>>> from tasks import error, on_failure
>>> error.apply(link_error=on_failure.si()).get()
```

## Expected behavior
Both `error` and `on_failure` to be executed eagerly.

## Actual behavior
`error` is executed eagerly and `on_failure` is executed on a worker.

auvipy:
please send a pr if you have any intended fix in mind if there isn't any open pr to fix this issue.

PR #5373, opened 2019-03-08
Added integration coverage for link_error

thedrow:
It appears that retries in link_error are not working correctly.
They hang the worker.


thedrow:
> can you rebase this?

Sure.

thedrow:
It turns out we no longer hang when retrying error callbacks. :tada: 
I did have some problems with the tests since I expected the error callbacks to change the result of the task which is not the case and shouldn't be.

In addition I fixed #4899 because it allowed me to debug this issue more easily.

thedrow:
I'm going to fix the unit test failure now :)
It happened because of the extra condition I added.

Issue #5377, opened 2019-03-11
Add documentation to the task_acks_on_failure_or_timeout setting


thedrow:
# Description

#4970 did not include the proper documentation for the setting.

# Suggestions

We should ensure to document it's behaviour before 4.3 GA.

huangsam:
@thedrow I submitted a PR based on the missing gaps I inferred from source code that used the `task_acks_on_failure_or_timeout` setting. I assume that the docs are built from the `celery.app.task` module so I applied my changes there.

thedrow:
The configuration documentation is at https://github.com/celery/celery/blob/master/docs/userguide/configuration.rst.
I'll copy what you added, there.

thedrow:
Oh, seems like this was not introduced as a setting as well.
I'm going to do so now.

PR #5379, opened 2019-03-11
Add caveat for acks_on_failure_or_timeout

huangsam:

## Description

This pull request is intended to fix the docs mentioned in https://github.com/celery/celery/issues/5377. After looking at the source code, the docs do not imply that the setting only applies when `task_acks_late` is configured. Added some subtle messaging to aid future readers.

### DIFF
--- a/celery/app/task.py
+++ b/celery/app/task.py
@@ -255,6 +255,9 @@ class Task(object):
     #: When enabled messages for this task will be acknowledged even if it
     #: fails or times out.
     #:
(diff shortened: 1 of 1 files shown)


huangsam:
Looks like Travis CI is not operating correctly due to `The command "wget -qO - https://packages.couchbase.com/ubuntu/couchbase.key | sudo apt-key add -...` - and seems to have nothing to do with my changes.

Issue #5328, opened 2019-02-07
Update our documentation for the new compression methods introduced in Kombu 4.3


thedrow:
# Description

The new release of Kombu 4.3 introduced a few new optional compression methods.
They are currently not documented in Celery.

# Suggestions

We should document those new compression methods and instruct how to enable them.

huangsam:
It seems like [kombu.compression](https://github.com/celery/kombu/blob/v4.3.0/kombu/compression.py) reveals the compression options available in Kombu at the moment. Where would these settings be documented? It seems like it was referenced in the [Celery release notes](https://github.com/celery/celery/blob/v4.3.0rc1/docs/whatsnew-4.3.rst) at the moment.

huangsam:
It seems like it would need to be documented [in the README.rst](https://github.com/celery/celery/blob/v4.3.0rc1/README.rst) and [introduction.rst](https://github.com/celery/celery/blob/v4.3.0rc1/docs/getting-started/introduction.rst).

thedrow:
https://github.com/celery/celery/blob/master/docs/userguide/calling.rst#compression - this is the right place.

huangsam:
@thedrow I went ahead and created a PR to add compression scheme documentation at the location you pointed out.

PR #5384, opened 2019-03-13
Add documentation for new compression schemes

huangsam:

## Description

Attempting to fill in the missing gaps for Celery user guides regarding newly supported compression schemes. I added spacing between the compression schemes so as to allow more room for scheme sub-sections if necessary. This should resolve https://github.com/celery/celery/issues/5328.

### DIFF
--- a/docs/userguide/calling.rst
+++ b/docs/userguide/calling.rst
@@ -499,15 +499,40 @@ Celery can compress messages using the following builtin schemes:
 
 - `brotli`
 
+    brotli is optimized for the web, in particular small text
+    documents. It is most effective for serving static content
+    such as fonts and html pages.
+
 - `bzip2`
+
+    bzip2 creates smaller files than gzip, but compression and
+    decompression speeds are noticeably slower than those of gzip.
 
 - `gzip`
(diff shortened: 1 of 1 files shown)


huangsam:
Screen Shot 2019-03-12 at 10 16 50 PM

This is how the documentation looks with the changes.

huangsam:
Interestingly enough, running `make html` on the Sphinx docs with Python 3.7.2 fails while running `make html` with Python 3.6.6 works fine with a couple of warnings. Not sure if that's because of the particular Sphinx version that Celery is using under the hood.

huangsam:
Screen Shot 2019-03-12 at 11 26 37 PM

The PR has been modified to include descriptions for each compression scheme - to help the user in making an informed choice on what is best for their particular use case(s).

thedrow:
Looks good. The only thing that's missing is to say that Kombu has an extra dependency for most of those.
So `pip install kombu[brotli]` will install the appropriate brotli implementation.

thedrow:
Done. Let's ensure that the build passes and merge this.

PR #5386, opened 2019-03-13
Fix `None` serialization error in Redis backend

bosonogi:
## Description

When `result_expires = None` in Celery config the Redis backend fails with an error saying that it cannot use a `None` value for the `expire` command.

This change makes it so that the `expire` commands are executed only when keys need to be expired.


thedrow:
Can you please fix the lint errors?

thedrow:
Thanks!

Issue #5371, opened 2019-03-06
Unable to create SSL broker/backend connection to redis


jcohen02:
I'm having some issues making an SSL/TLS connection to redis for the Celery broker and backend.

I'm using: 
Celery: 4.3.0rc2 (master) 
Python: 3.6 
OS: Mac OS X. 
Related dependency versions: kombu (4.4.0),  redis-py (3.2.0)

Celery works fine using a non-SSL connection to redis. My redis/SSL setup uses redis behind stunnel and I can successfully connect to and use this deployment via py-redis directly and via other libraries.

It looks like this is related to #3830, however that issue seemed to have been resolved by adding the `broker_use_ssl` option and adding this option doesn't appear to be having any effect here.

To setup SSL, I'm adding the `broker_use_ssl` attribute to configuration parameters as described in the [docs](http://docs.celeryproject.org/en/master/userguide/configuration.html#broker-use-ssl). I'm also adding the same set of parameters for the `redis_backend_use_ssl` [option](http://docs.celeryproject.org/en/master/userguide/configuration.html#broker-use-ssl). I notice there is some information here about adding the parameters in the query string but I've opted to stick with providing the dict of parameters. 

```python
broker_url = 'rediss://localhost:6380'
key_file = '/path/to/client.key' 
cert_file = '/path/to/client.crt' 
ca_file = '/path/to/CAcert.pem'
    
app = Celery('app', broker=broker_url, backend=broker_url,
             broker_use_ssl = {
                 'keyfile': key_file, 'certfile': cert_file, 
                 'ca_certs': ca_file,  
                 'cert_reqs': ssl.CERT_REQUIRED 
            }, 
             redis_backend_use_ssl = { 
                 'keyfile': key_file, 'certfile': cert_file,
                 'ca_certs': ca_file,  
                 'cert_reqs': ssl.CERT_REQUIRED
            })
app.connection().connect()
app.send_task('a_task')
```

When I attempt to make and use the connection as shown above, I get an error:

```python
ValueError: A rediss:// URL must have parameter ssl_cert_reqs be CERT_REQUIRED, CERT_OPTIONAL, or CERT_NONE
```
It looks like the `rediss` URL scheme is being picked up but the other parameters are not. Apologies if I'm missing something in the documentation or elsewhere.

[...]

PR #5395, opened 2019-03-18
Fixes to SSL config handling

jcohen02:
Addresses issue with handling of SSL parameters as detailed in #5371.

 - Adds configuration provided via the `broker_use_ssl` and `redis_backend_use_ssl` parameters to the application config.

 - Adds a warning if a `redis://` URL scheme is used but SSL configuration is provided (parameters may not be correctly handled in this case)

 - Refactored checking of the `ssl_cert_reqs` parameter so that this is checked correctly whether the parameter is provided via the query string or in the SSL parameter dict.

 - Tests: Added app config test to check SSL configuration is correctly added to application configuration object. Fixed incorrect URL scheme in SSL backend test.

### DIFF
--- a/celery/app/base.py
+++ b/celery/app/base.py
@@ -271,12 +271,8 @@ class Celery(object):
         self.__autoset('broker_url', broker)
         self.__autoset('result_backend', backend)
         self.__autoset('include', include)
-        self.__autoset('broker_use_ssl',
-                       (kwargs['broker_use_ssl'] if 'broker_use_ssl'
-                        in kwargs else None))
-        self.__autoset('redis_backend_use_ssl',
-                       (kwargs['redis_backend_use_ssl'] if
-                        'redis_backend_use_ssl' in kwargs else None))
+        self.__autoset('broker_use_ssl', kwargs.get('broker_use_ssl'))
+        self.__autoset('redis_backend_use_ssl', kwargs.get('redis_backend_use_ssl'))
         self._conf = Settings(
             PendingConfiguration(
                 self._preconf, self._finalize_pending_conf),
--- a/celery/app/base.py
+++ b/celery/app/base.py
@@ -271,11 +271,11 @@ class Celery(object):
         self.__autoset('broker_url', broker)
         self.__autoset('result_backend', backend)
         self.__autoset('include', include)
-        self.__autoset('broker_use_ssl', 
-                       (kwargs['broker_use_ssl'] if 'broker_use_ssl' 
+        self.__autoset('broker_use_ssl',
+                       (kwargs['broker_use_ssl'] if 'broker_use_ssl'
                         in kwargs else None))
-        self.__autoset('redis_backend_use_ssl', 
-                       (kwargs['redis_backend_use_ssl'] if  
+        self.__autoset('redis_backend_use_ssl',
+                       (kwargs['redis_backend_use_ssl'] if
                         'redis_backend_use_ssl' in kwargs else None))
         self._conf = Settings(
             PendingConfiguration(
--- a/celery/app/base.py
+++ b/celery/app/base.py
@@ -271,6 +271,12 @@ class Celery(object):
         self.__autoset('broker_url', broker)
         self.__autoset('result_backend', backend)
         self.__autoset('include', include)
+        self.__autoset('broker_use_ssl', 
+                       (kwargs['broker_use_ssl'] if 'broker_use_ssl' 
+                        in kwargs else None))
+        self.__autoset('redis_backend_use_ssl', 
+                       (kwargs['redis_backend_use_ssl'] if  
+                        'redis_backend_use_ssl' in kwargs else None))
         self._conf = Settings(
             PendingConfiguration(
                 self._preconf, self._finalize_pending_conf),
--- a/celery/backends/redis.py
+++ b/celery/backends/redis.py
@@ -66,10 +66,9 @@ will not valdate the identity of the redis broker when connecting. This \
 leaves you vulnerable to man in the middle attacks.
 """
 
-W_REDIS_SSL_PARAMS_AND_SCHEME_MISMATCH = """
+E_REDIS_SSL_PARAMS_AND_SCHEME_MISMATCH = """
 SSL connection parameters have been provided but the specified URL scheme \
-is redis://. A Redis SSL connection URL should use the scheme rediss://. \
-An SSL connection will be attempted.
+is redis://. A Redis SSL connection URL should use the scheme rediss://.
 """
 
 E_REDIS_SSL_CERT_REQS_MISSING = """
@@ -258,10 +257,10 @@ class RedisBackend(BaseKeyValueStoreBackend, AsyncBackendMixin):
                           'ssl_cert_reqs']
 
         if scheme == 'redis':
-            # Check IF connparams or query string contain ssl params, if so show warning.]
+            # If connparams or query string contain ssl params, raise error
             if (any(key in connparams for key in ssl_param_keys) or
                     any(key in query for key in ssl_param_keys)):
-                logger.warning(W_REDIS_SSL_PARAMS_AND_SCHEME_MISMATCH)
+                raise ValueError(E_REDIS_SSL_PARAMS_AND_SCHEME_MISMATCH)
 
         if scheme == 'rediss':
             connparams['connection_class'] = redis.SSLConnection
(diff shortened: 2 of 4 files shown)


celery/backends/redis.py
```diff
@@ -230,25 +254,23 @@ def _params_from_url(self, url, defaults):
...
             connparams['db'] = path
 
+        ssl_param_keys = ['ssl_ca_certs', 'ssl_certfile', 'ssl_keyfile', 
+                          'ssl_cert_reqs']
+
+        if scheme == 'redis':
+            # Check IF connparams or query string contain ssl params, if so show warning.]
+            if (any(key in connparams for key in ssl_param_keys) or
+                any(key in query for key in ssl_param_keys)):
+                logger.warning(W_REDIS_SSL_PARAMS_AND_SCHEME_MISMATCH)
```
georgepsarakis: I think it may be preferable to raise an exception, in order to better inform the user that they are using a wrong configuration.

t/unit/app/test_app.py
```diff
@@ -385,6 +386,37 @@ def test_pending_configuration__raises_ImproperlyConfigured(self):
...
 
+    def test_pending_configuration__ssl_settings(self):
+        with self.Celery(broker='foo://bar',
+                         broker_use_ssl={
+                             'ssl_cert_reqs': ssl.CERT_REQUIRED,
+                             'ssl_ca_certs': '/path/to/ca.crt',
+                             'ssl_certfile': '/path/to/client.crt',
+                             'ssl_keyfile': '/path/to/client.key'}, 
+                         redis_backend_use_ssl = {
+                             'ssl_cert_reqs': ssl.CERT_REQUIRED,
```
georgepsarakis: Should we also check what happens if someone uses the string form?

celery/backends/redis.py
```diff
@@ -198,6 +204,24 @@ def __init__(self, host=None, port=None, db=None, password=None,
...
         if url:
             self.connparams = self._params_from_url(url, self.connparams)
+        
+        # If we've received SSL parameters via query string or the 
+        # redis_backend_use_ssl dict, check ssl_cert_reqs is valid. If set 
+        # via query string ssl_cert_reqs will be a string so convert it here
+        if ('connection_class' in self.connparams 
+            and self.connparams['connection_class'] == redis.SSLConnection):
+            ssl_cert_reqs = self.connparams.get('ssl_cert_reqs', 'MISSING')
+            if ssl_cert_reqs in [CERT_REQUIRED, 'CERT_REQUIRED']:
```
georgepsarakis: Perhaps an alternative not using multiple `if` branches, not sure if it is clearer, but one can see the transformation immediately:

```python
ssl_cert_reqs_missing = 'MISSING'  # could also be a constant
ssl_string_to_constant = { 'CERT_REQUIRED': CERT_REQUIRED, 
'CERT_OPTIONAL': CERT_OPTIONAL, 
'CERT_NONE': CERT_NONE }
ssl_cert_reqs = ssl_string_to_constant.get(ssl_cert_reqs, ssl_cert_reqs)
if ssl_cert_reqs == ssl_cert_reqs_missing:
    raise ...
# Other warnings here ...
```

What do you think?

celery/app/base.py
```diff
@@ -271,6 +271,12 @@ def __init__(self, main=None, loader=None, backend=None,
         self.__autoset('broker_url', broker)
         self.__autoset('result_backend', backend)
         self.__autoset('include', include)
+        self.__autoset('broker_use_ssl', 
+                       (kwargs['broker_use_ssl'] if 'broker_use_ssl'
```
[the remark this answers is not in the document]
jcohen02: Agreed, much more succinct. Apologies for my confusion, was thinking this would raise a `KeyError` if the key was missing! Fixed.

celery/backends/redis.py
jcohen02: Agreed. I was concerned if raising an exception would introduce a breaking change for anyone already using code with an incorrectly structured URL, however, I guess it's better to address this so if you're happy with raising an exception, that's great. Code updated.

t/unit/app/test_app.py
jcohen02: Yes, I guess an extra check for that would be sensible. in `unit/app/test_app.py` it's simply checking that values added via the `broker_use_ssl` and `redis_backend_use_ssl` parameters when creating an instance of the `Celery` object end up in the config so I think here the content of the parameters is not significant. However, if ssl_cert_reqs is provided in string form, it should end up in the config in constant form after processing so I've added a test `test_backend_ssl_certreq_str` to the redis backend tests in `t/unit/backends/test_redis.py`

celery/backends/redis.py
jcohen02: I agree the above looks much cleaner  - I'm assuming line 5 should have `ssl_cert_reqs_missing` as the second param to `get`? e.g.
```python
ssl_cert_reqs = ssl_string_to_constant.get(ssl_cert_reqs, ssl_cert_reqs_missing)
```
However, the current code works when the value for `ssl_cert_reqs` is either in string form or provided as a constant. With the above, wouldn't we need a check to see whether the provided value is a string first? 

Happy to switch to this approach of you think it's preferable.

jcohen02:
@georgepsarakis Sorry for missing the coding style errors - thanks for pointing this out, issues resolved.

thedrow:
The Windows build is failing.
Can you check why?

jcohen02:
> The Windows build is failing.
> Can you check why?

@thedrow Looks like this was caused by the new `connection_class` check which looks to see if the class is `redis.SSLConnection`. `redis` seems to evaluate to `None` in the AppVeyor tests causing the error. I've added `@skip.unless_module('redis')` to the test since this is not relevant if the `redis` module is not in use. Windows build now passing.

celery/backends/redis.py
georgepsarakis: You are right, sorry I missed that step:
```python
ssl_cert_reqs = self.connparams.get('ssl_cert_reqs', ssl_cert_reqs_missing)
ssl_cert_reqs = ssl_string_to_constant.get(ssl_cert_reqs, ssl_cert_reqs)
# instead of `if ssl_cert_reqs == ssl_cert_reqs_missing:`
if ssl_cert_reqs not in (CERT_NONE, CERT_REQUIRED, CERT_OPTIONAL):
    raise ...
```
An example taking the above step-by-step:
1. `ssl_cert_reqs` will be either `CERT_NONE`, `'CERT_NONE'` or `'MISSING'`.
2. `ssl_cert_reqs` will be converted to the constant if string: `'CERT_NONE'` becomes `CERT_NONE`. Otherwise the previous value is retained: variable value is already a constant or `'MISSING'` string.
3. We finally assert that the variable value is one of the valid constants.

How does that sound?

jcohen02: Sounds good. This gives us the following block to carry out all the necessary checks.

```python
if ('connection_class' in self.connparams and
            self.connparams['connection_class'] == redis.SSLConnection):
        ssl_cert_reqs_missing = 'MISSING'
        ssl_string_to_constant = { 'CERT_REQUIRED': CERT_REQUIRED,
                                   'CERT_OPTIONAL': CERT_OPTIONAL,
                                   'CERT_NONE': CERT_NONE }
        ssl_cert_reqs = self.connparams.get('ssl_cert_reqs', ssl_cert_reqs_missing)
        ssl_cert_reqs = ssl_string_to_constant.get(ssl_cert_reqs, ssl_cert_reqs)
        if ssl_cert_reqs not in ssl_string_to_constant.values():
            raise ValueError(E_REDIS_SSL_CERT_REQS_MISSING_INVALID)

        if ssl_cert_reqs == CERT_OPTIONAL:
            logger.warning(W_REDIS_SSL_CERT_OPTIONAL)
        elif ssl_cert_reqs == CERT_NONE:
            logger.warning(W_REDIS_SSL_CERT_NONE)
        self.connparams['ssl_cert_reqs'] = ssl_cert_reqs
```
This passes the tests. If all looks ok, I'll commit.

jcohen02: Update committed

huangsam:
@georgepsarakis it looks like the `flake8` checks passed in [the most recent build](https://travis-ci.org/celery/celery/builds/509367984) - was there anything else that you had in your feedback that still needs changing before you approve?

celery/backends/redis.py
```diff
@@ -198,6 +203,27 @@ def __init__(self, host=None, port=None, db=None, password=None,
 
         if url:
             self.connparams = self._params_from_url(url, self.connparams)
+
+        # If we've received SSL parameters via query string or the
+        # redis_backend_use_ssl dict, check ssl_cert_reqs is valid. If set
+        # via query string ssl_cert_reqs will be a string so convert it here
+        if ('connection_class' in self.connparams and
+                self.connparams['connection_class'] == redis.SSLConnection):
```
georgepsarakis: Equality check will work too but since this is a class, let's use the `is` operator.

jcohen02: Update committed. `flake8` re-run along with unit tests for all supported Python versions. Many thanks for reviewing.

PR #5396, opened 2019-03-19
Update urls.py because of new version of Django.

ammirsm:
FYI patterns have been removed in Django 1.10.
See release 1.10 notes: [https://docs.djangoproject.com/en/2.0/releases/1.10/](Django Document)


thedrow:
Thank you very much!

PR #5415, opened 2019-03-29
Add optional anchor_id parameter to tasks

madprogrammer:
## Description
In my project, I had a need to pass some extra ID to every subsequent child task, in a similar way how `root_id` is currently propagated to tasks. However, `root_id` is the actual ID of the root task, and I needed to be able to additionally pass an user-suppied value.

Here I present the `anchor_id` header, which, if set by user, will be propagated to every subsequent child task, similar with `root_id`. Example:

    foo.apply_async(args=[42], headers={"anchor_id": "bar"})

Every task called by `foo` will have `anchor_id` available in `self.request.anchor_id`. This would allow users to identify and group tasks together using their own information in `anchor_id`. For example, we have a pipeline of processing tasks related to some business resource, and `anchor_id` could contain such resource ID, so that every task could be able to fetch the resource from the database if needed, avoiding the need to pass an extra parameter to many tasks.

Hope this makes it easy to understand the nature of this change.

### DIFF
--- a/celery/app/amqp.py
+++ b/celery/app/amqp.py
@@ -304,7 +304,7 @@ class AMQP(object):
                    time_limit=None, soft_time_limit=None,
                    create_sent_event=False, root_id=None, parent_id=None,
                    shadow=None, chain=None, now=None, timezone=None,
-                   origin=None, argsrepr=None, kwargsrepr=None):
+                   origin=None, argsrepr=None, kwargsrepr=None, anchor_id=None):
         args = args or ()
         kwargs = kwargs or {}
         if not isinstance(args, (list, tuple)):
@@ -359,6 +359,7 @@ class AMQP(object):
                 'retries': retries,
                 'timelimit': [time_limit, soft_time_limit],
                 'root_id': root_id,
+                'anchor_id': anchor_id,
                 'parent_id': parent_id,
                 'argsrepr': argsrepr,
                 'kwargsrepr': kwargsrepr,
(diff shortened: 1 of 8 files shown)


auvipy:
what problem it will solve?

madprogrammer:
@auvipy I don't see any existing way to have a user-supplied value automatically propagated to child tasks.

thedrow:
I'm not sure I like this idea. What's wrong with using the root_id as the identifier of the common data you need to share between tasks?

madprogrammer:
The problem with root_id is that it's purpose is to store the actual ID of
the root task. Moreover, in vanilla celery the root_id can't be specified
by the user. `anchor_id` allows to have an independent ID, while retaining
the original `root_id` tracking functionality.

It's okay if you think this PR isn't suitable for mainline, it's just that
I've been using similar logic in my project for a while and found it
useful. For example, we have a web crawling system designed to download
metadata and files from different websites (sources). The `anchor_id` in
our case uiniquely indentifies the website being processed, while the
original `root_id` identifies the processing session. Each website will be
processed every few hours, having same `anchor_id` but different `root_id`s.

пн, 1 апр. 2019 г., 20:40 Omer Katz <notifications@github.com>:

thedrow:
I think we're not going to introduce this to mainline.

Celery 5 will introduce a context concept in one form or another which will help you do what you want in a more meaningful way.

madprogrammer:
Thanks for the review. No problem, provided there is going to be a more flexible replacement in Celery 5.

thedrow:
Right now we're in the architecture planning phase of Celery 5.
You're welcome to join us.

PR #5283, opened 2019-01-14
Implement version switches to remove deprecation warnings for collections.abc

halfdan:
## Description

As [suggested by @MRigal](https://github.com/celery/celery/issues/5180#issuecomment-451515057) this PR gets rid of a number of deprecation warnings with the collections.abc module by implementing import switches.

PR #5428, opened 2019-04-03
Add a version switch when importing Mapping

Smotko:

## Description

This fixes a deprecation warning with the collections.abc module.

It's the same solution as in PR #5283 applied to app/amqp.py.


Smotko:
@auvipy cool 👍 

Let me know if you need me to do anything else to get this merged to the maintenance branch. My change is trivial so I think changing the base branch should not cause any problems though.

Thanks!

PR #5435, opened 2019-04-03
Attribute Error should be caught in backends base exception 

cdknorow:
Celery looks for exceptions in sys.modules and catches a keyerror if it doesn't exist. However,  it is also possible that it finds the exec_module but finds the wrong one. In this case an Attribute Error gets raised. It would be nice if celery caught that exception and created a custom exception class that correctly propogates the error message instead of raising a new one.

### DIFF
--- a/celery/backends/base.py
+++ b/celery/backends/base.py
@@ -274,7 +274,7 @@ class Backend(object):
                     exc_type = from_utf8(exc['exc_type'])
                     try:
                         cls = getattr(sys.modules[exc_module], exc_type)
-                    except KeyError:
+                    except (KeyError, AttributeError):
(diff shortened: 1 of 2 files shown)


cdknorow:
This solves an issue where I'm seeing

cdknorow:
this solves an issue I'm seeing where django db models is raising a DoesNotExist  exception, but when celery tries to reraise this exception on the backend, it is finding a different  module which doesn't have the DoesNotExist attribute.

auvipy:
can you add a small test for this?

cdknorow:
@auvipy  done

auvipy:
thanks, will come back after making the master stable

PR #5451, opened 2019-04-09
Remove python 2.7 from appveyor

othalla:
## Description

Remove python 2.7 from appveyor CI

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -12,18 +12,9 @@ environment:
     # a later point release.
     # See: https://www.appveyor.com/docs/installed-software#python
 
-    - PYTHON: "C:\\Python27"
-      PYTHON_VERSION: "2.7.x"
-      PYTHON_ARCH: "32"
-
     - PYTHON: "C:\\Python34"
       PYTHON_VERSION: "3.4.x"
       PYTHON_ARCH: "32"
-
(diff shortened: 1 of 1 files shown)


auvipy:
thanks for this. Will merge this to celery5 version. no in master though. going to revert few things in master. wait till then

othalla:
@auvipy i think we should never stay with a RED CI.
Why not keep a branch for 4.3 versions with python 2.7 support ?
This will point that we will stop supporting it and start to work on v5.

auvipy:
master will support 4.3 with 2.7 support

matteius:
I agree that we should remove the support from the RED CI now because we can see that python 2.7 build started failing in appveryor when we merged in the requirement for vine>=5.0.0a1
whereas before it was fine on 1.3.0 -- was it perhaps upgrade to 5.0.0a1 because the prior one was >= 1.3.0 and so that would have started installing 5.0.0.a1 anyway?  I guess I don't really understand the true intent behind this commit:  https://github.com/celery/celery/commit/af86fe4dba0c2527a94dc36e00c5bfe48b9a7bba
https://ci.appveyor.com/project/ask/celery/builds/23512901

auvipy:
actually, the c5 development was mistakenly started in master branch!! so reverting some commits should be alright.

PR #5456, opened 2019-04-11
Use 'property' with 'abstractmethod' instead 'abstractproperty'

tothegump:
'@abc.abstractproperty' has deprecated since Python 3.3 
It is now possible to use property, property.getter(), property.setter() and property.deleter() with abstractmethod(), making this decorator redundant. 
[Python doc](https://docs.python.org/3/library/abc.html#abc.abstractproperty)


thedrow:
What about 2.7?

tothegump:
You are right, Python 2.7 can not use 'property' with 'abstractmethod'.
What should I do then? Which one is better choice:
1. close this PR
2. Waiting for Celery 5. (p.s. The branch 5.0-devel latest commit date is Feb 17, 2017, does it is in progress?)
3. write extra compatible code for Python 2.7

auvipy:
> You are right, Python 2.7 cannot use 'property' with 'abstractmethod'.
> What should I do then? Which one is a better choice:
> 
> 1. close this PR
> 2. Waiting for Celery 5. (p.s. The branch 5.0-devel latest commit date is Feb 17, 2017, does it is in progress?)
> 3. write an extra compatible code for Python 2.7

wait for c5

thedrow:
I've rebased this to be merged on Celery 5.
This should be merged to that branch.

auvipy:
great

thedrow:
Which tests do you have in mind?

PR #5462, opened 2019-04-15
Fixed bug 5411

### LINKED ISSUE #5411: Minor bug in celery utils time.py file
There is a function named remaining in the celery.utils.time.py file. In that function, I think the statement start = start.replace(tzinfo=now.tzinfo) (line 211) has to be replaced with start = start.astimezone(now.tzinfo).

I hope the expected working by using the statement (`start = start.replace(tzinfo=now.tzinfo)`) is to change the `start` variable's date and time according to the `now` variable's timezone. But the replace function in the above mentioned statement is just replacing the timezone of the `start` variable without changing the values of its date and time. I thought `astimezone` is the method required for our purpose (which converts the date and time in `start` variable to the given timezone argument).

So, I think `start = start.replace(tzinfo=now.tzinfo)` must be changed to `start = start.astimezone(now.tzinfo)`.

Please correct me incase I am wrong.

Module versions:
Python 3.5.2
celery==4.2.1
django-celery-beat==1.4.0

esh-b:

## This pull request attempts to fix the bug 5411
Fixes #5411 
The function `remaining` defined in `celery/utils/time.py` was not 
calculating the remaining time for a schedule correctly when the timezones of the `start` and the `now` variables (which are arguments to the function) are different.

The issue was because of the lines 
https://github.com/celery/celery/blob/f2cab7715cceafcae1343fdcdc65704e0a2c751f/celery/utils/time.py#L210-L212

The `start` variable's timezone must be replaced by the `now` variable's timezone ONLY when there the `start` and `now` variable have different `utcoffsets` due to DST. But the above `if` statement is run even if `start` and `now` variables are in different timezones.

Also, the unit-tests are added for the `remaining` function to check whether the next run times are calculated correctly if the `start` and `now` variables are in different timezones or they differ in utcoffset (because of DST).

### DIFF
--- a/celery/utils/time.py
+++ b/celery/utils/time.py
@@ -208,7 +208,7 @@ def remaining(start, ends_in, now=None, relative=False):
     """
     now = now or datetime.utcnow()
     if str(start.tzinfo) == str(now.tzinfo) and now.utcoffset() != start.utcoffset():
-        #DST started/ended
+        # DST started/ended
         start = start.replace(tzinfo=now.tzinfo)
     end_date = start + ends_in
     if relative:
--- a/celery/utils/time.py
+++ b/celery/utils/time.py
@@ -207,8 +207,8 @@ def remaining(start, ends_in, now=None, relative=False):
         ~datetime.timedelta: Remaining time.
     """
     now = now or datetime.utcnow()
-    if now.utcoffset() != start.utcoffset():
-        # Timezone has changed, or DST started/ended
+    if str(start.tzinfo) == str(now.tzinfo) and now.utcoffset() != start.utcoffset():
+        #DST started/ended
         start = start.replace(tzinfo=now.tzinfo)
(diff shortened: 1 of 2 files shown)


esh-b:
The appveyor build fails for Python 2.7.x environment with the following error 
```
class Thenable(Callable, metaclass=abc.ABCMeta):  # pragma: no cover
                                      ^
SyntaxError: invalid syntax
```
Is it fine? or the issue is because of the commits in my pull request?

thedrow:
> The appveyor build fails for Python 2.7.x environment with the following error
> 
> ```
> class Thenable(Callable, metaclass=abc.ABCMeta):  # pragma: no cover
>                                       ^
> SyntaxError: invalid syntax
> ```
> Is it fine? or the issue is because of the commits in my pull request?

Yes, we need to fix the Windows build. It's fine.

PR #5476, opened 2019-04-19
 Make AsynPool._proc_alive_timeout configurable

srafehi:
The timeout which determines how long to wait for a new worker process to startup is hard-coded to `4.0` seconds via the [PROC_ALIVE_TIMEOUT](https://github.com/celery/celery/blob/e6ea4331215c6ac1dbee38868f212c4d48872b36/celery/concurrency/asynpool.py#L85:1) constant. This change allows the value to be configured via a new configuration option `worker_proc_alive_timeout`.

This is necessary for worker processes with expensive setups that can exceed `4.0` seconds.

### DIFF
--- a/celery/app/defaults.py
+++ b/celery/app/defaults.py
@@ -318,6 +318,7 @@ NAMESPACES = Namespace(
         pool=Option(DEFAULT_POOL),
         pool_putlocks=Option(True, type='bool'),
         pool_restarts=Option(False, type='bool'),
+        proc_alive_timeout=Option(4.0, type='float'),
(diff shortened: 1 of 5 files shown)


thedrow:
This is not for 4.3.1 as it is a new feature.
Should we release 4.4.0?

auvipy:
that early? :D

auvipy:
but its good to have more bug fixed smaller versions like 4.4, 4.5 etc

auvipy:
Be sure vine is <5.0 for 4.4 of celery

PR #5478, opened 2019-04-22
Removed dangerous default mutable arguments from function definitions…

todd-cook:
In several places, Celery & Kombu use dangerous default arguments which may cause subtle, insidious bugs, e.g.:

# bug demo code, not in Celery:

In [1]: def append(number, number_list=[]):
   ...:     number_list.append(number)
   ...:     print(number_list)
   ...:     return number_list
   ...: 

In [2]: append(5)  # expecting: [5], actual: [5]
[5]
Out[2]: [5]

In [3]: append(7)  # expecting: [7], actual: [5, 7]
[5, 7]
Out[3]: [5, 7]

In [4]: append(2)  # expecting: [2], actual: [5, 7, 2]
[5, 7, 2]
Out[4]: [5, 7, 2]

--As you can see, mutable default arguments can cause unwanted caching and spurious persistent values 
that can make debugging difficult.

This PR changes the mutable default argument values with None and then checks and replaces once inside the function call; this is the standard way to avoid the unwanted, buggy behavior.

### DIFF
--- a/celery/app/base.py
+++ b/celery/app/base.py
@@ -995,7 +995,8 @@ class Celery(object):
         return key
 
     def _sig_to_periodic_task_entry(self, schedule, sig,
-                                    args=(), kwargs={}, name=None, **opts):
+                                    args=(), kwargs=None, name=None, **opts):
+        kwargs = {} if not kwargs else kwargs
         sig = (sig.clone(args, kwargs)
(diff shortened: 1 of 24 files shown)


auvipy:
is it possible to update the related tests as well?

auvipy:
I restarted the job of python 3.5, btw don't you think some test changes are needed?

Issue #5469, opened 2019-04-16
Error with default kwargs in __header__ attribute generation


dimavitvickiy:
# Checklist

- [ ] I have read the relevant section in the
  [contribution guide](http://docs.celeryproject.org/en/latest/contributing.html#other-bugs)
  on reporting bugs.
- [ ] I have checked the [issues list](https://github.com/celery/celery/issues?q=is%3Aissue+label%3A%22Issue+Type%3A+Bug+Report%22+-label%3A%22Category%3A+Documentation%22)
  for similar or identical bug reports.
- [ ] I have checked the [pull requests list](https://github.com/celery/celery/pulls?q=is%3Apr+label%3A%22PR+Type%3A+Bugfix%22+-label%3A%22Category%3A+Documentation%22)
  for existing proposed fixes.
- [x] I have checked the [commit log](https://github.com/celery/celery/commits/master)
  to find out if the bug was already fixed in the master branch.
- [x] I have included all related issues and possible duplicate issues
  in this issue (If there are none, check this box anyway).

## Mandatory Debugging Information

- [ ] I have included the output of ``celery -A proj report`` in the issue.
    (if you are not able to do this, then at least specify the Celery
     version affected).
- [ ] I have verified that the issue exists against the `master` branch of Celery.
- [ ] I have included the contents of ``pip freeze`` in the issue.
- [ ] I have included all the versions of all the external dependencies required
  to reproduce this bug.

## Optional Debugging Information

- [x] I have tried reproducing the issue on more than one Python version
  and/or implementation.
- [ ] I have tried reproducing the issue on more than one message broker and/or
  result backend.
- [ ] I have tried reproducing the issue on more than one version of the message
  broker and/or result backend.
- [ ] I have tried reproducing the issue on more than one operating system.
- [ ] I have tried reproducing the issue on more than one workers pool.
- [ ] I have tried reproducing the issue with autoscaling, retries,
  ETA/Countdown & rate limits disabled.
- [x] I have tried reproducing the issue after downgrading
  and/or upgrading Celery and its dependencies.

#### Related Issues

- None

#### Possible Duplicates

- None

## Environment & Settings

**Celery version**: 4.3.0

<details>
<summary><b><code>celery report</code> Output:</b></summary>
<p>

```
```

</p>
</details>

# Steps to Reproduce

## Required Dependencies

* **Minimal Python Version**: 3.7.1
* **Minimal Celery Version**: 4.3.0
[...]

PR #5485, opened 2019-04-24
Fix 5469

dimavitvickiy:
I5469 fix: change default kwargs behaviour

For previous solution if you have function for example:

def my_func(*, name='default', age, city='Kyiv'):
    ...
    
kwargsonly were `age` and `city`, because of split and fetching last 2 kwargs.

In this fix we use kwargs that have default value (fetch from spec.kwonlydefaults.keys()).
So it fixed bug defined in issue https://github.com/celery/celery/issues/5469

### DIFF
--- a/celery/utils/functional.py
+++ b/celery/utils/functional.py
@@ -232,11 +232,10 @@ def _argsfromspec(spec, replace_defaults=True):
     varargs = spec.varargs
     varkw = spec.varkw
     if spec.kwonlydefaults:
-        split = len(spec.kwonlydefaults)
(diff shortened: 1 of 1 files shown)


dimavitvickiy:
> it would be easier to review if you could add test for the oroposed changes :)

will do

auvipy:
ping

dimavitvickiy:
sorry, a bit busy atm

thedrow:
Unfortunately, this fix fails some of our other unit tests.
The bug, however, is still present.

PR #5499, opened 2019-05-03
Potential fix and an example test case for the OSError: [Errno 9] Bad file descriptor. #4457

### LINKED ISSUE #4457: Connection to broker lost. Trying to re-establish the connection: OSError: [Errno 9] Bad file descriptor
## Checklist

       version affected).

## Software
celery==4.1.0
kombu==4.1.0
amqp==2.2.2
Python 3.6.1
broker: rabbitmq 3.6.14
result backend: redis

## Steps to reproduce
1. celery -A proj worker -Q Q1 --autoscale=10,1 -Ofair --without-gossip --without-mingle --heartbeat-interval=60 -n Q1
2. celery lost connection to broker
3. after restarting affected worker the connection is successfully re-established and the worker starts processing tasks

## Expected behavior
celery should re-establish connection to broker

## Actual behavior
celery tries to re-establish connection to broker but fails with this error message (which is repeated every second) until manually restarted:
```
[user] celery.worker.consumer.consumer WARNING 2017-12-18 00:38:27,078 consumer: 
Connection to broker lost. Trying to re-establish the connection...
  File "/home/user/.envs/user/lib/python3.6/site-packages/celery/worker/consumer/consumer.py", line 320, in start
    blueprint.start(self)
  File "/home/user/.envs/user/lib/python3.6/site-packages/celery/bootsteps.py", line 119, in start
    step.start(parent)
  File "/home/user/.envs/user/lib/python3.6/site-packages/celery/worker/consumer/consumer.py", line 596, in start
    c.loop(*c.loop_args())
  File "/home/user/.envs/user/lib/python3.6/site-packages/celery/worker/loops.py", line 47, in asynloop
    obj.controller.register_with_event_loop(hub)
  File "/home/user/.envs/user/lib/python3.6/site-packages/celery/worker/worker.py", line 217, in register_with_event_loop
    description='hub.register',
  File "/home/user/.envs/user/lib/python3.6/site-packages/celery/bootsteps.py", line 151, in send_all
    fun(parent, *args)
  File "/home/user/.envs/user/lib/python3.6/site-packages/celery/worker/components.py", line 178, in register_with_event_loop
    w.pool.register_with_event_loop(hub)
  File "/home/user/.envs/user/lib/python3.6/site-packages/celery/concurrency/prefork.py", line 134, in register_with_event_loop
    return reg(loop)
  File "/home/user/.envs/user/lib/python3.6/site-packages/celery/concurrency/asynpool.py", line 476, in register_with_event_loop
    for fd in self._fileno_to_outq]
  File "/home/user/.envs/user/lib/python3.6/site-packages/celery/concurrency/asynpool.py", line 476, in <listcomp>
    for fd in self._fileno_to_outq]
  File "/home/user/.envs/user/lib/python3.6/site-packages/kombu/async/hub.py", line 207, in add_reader
    return self.add(fds, callback, READ | ERR, args)
  File "/home/user/.envs/user/lib/python3.6/site-packages/kombu/async/hub.py", line 158, in add
    self.poller.register(fd, flags)
  File "/home/user/.envs/user/lib/python3.6/site-packages/kombu/utils/eventio.py", line 67, in register
    self._epoll.register(fd, events)
OSError: [Errno 9] Bad file descriptor
```

matteius:
## Description

This adds a test case that demonstrates the issue of Connection to broker lost. (Fixes #4457) Trying to re-establish the connection: OSError: [Errno 9] Bad file descriptor.

Since after probing the code and closing processes the normal ways and seeing that the list generally does expand and contract as you would expect, and so I did the only thing that I could--what you would not expect, which is to have the OS close the file descriptor on you for *reasons*. 

This showed the same stack trace and code path as the failing production code, and so I've amended the code to do that loop more safely and to release the file descriptors that are known to be bad so we don't cycle them every time we hit that method.   

The only thing now that may be worth figuring out is that when the pool/workers shut down now that process has already been closed, and so we see in the test output at the end that is unsettling, but technically all the tests themselves do pass and make logical sense to me.  Here is the output I mention:

"Exception ignored in: <bound method _ConnectionBase.__del__ of <billiard.connection.Connection object at 0x7f933751d5f8>>
  File "/home/mdavis/.virtualenvs/celery/lib/python3.6/site-packages/billiard/connection.py", line 161, in __del__
  File "/home/mdavis/.virtualenvs/celery/lib/python3.6/site-packages/billiard/connection.py", line 400, in _close
OSError: [Errno 9] Bad file descriptor"


requirements/default.txt
```diff
@@ -1,4 +1,4 @@
 pytz>dev
 billiard>=3.6.0,<4.0
 kombu>=4.5.0,<5.0
-vine>=5.0.0a1
+vine==1.3.0
```
auvipy: thanks for this!

t/unit/worker/test_worker.py
```diff
@@ -29,7 +30,7 @@
 from celery.utils.nodenames import worker_direct
 from celery.utils.serialization import pickle
 from celery.utils.timer2 import Timer
-from celery.worker import components, consumer, state
+from celery.worker import autoscale, components, consumer, state
```
auvipy: lets see how it behaves

requirements/default.txt
matteius: Yay ... mostly green now, except apparently my test is not py2 compat yet as you can see from the veyor error:  https://ci.appveyor.com/project/ask/celery/builds/24275512/job/e4vhy7by4k9df0j8

At least its the only failking one for that build configuration.  I will see if I can recall that about py2 ...

auvipy: no need to worry about veyor, focus on travis only 4 now :dagger:

auvipy:
I would like to wait for few more peoples review. btw, do you think adding more test as viable option?

auvipy:
@sudhishvnair @ryanpetrello @tonal @cajbecu @wind2008hxy @rabbit8ge

auvipy:
Please test this by tomorrow to

matteius:
I will consider what additional test cases might make sense, but this is one of the few that seems to actually start the pool.  I learned some things from debugging the appveyor logs, which is that our appveyor configs run on win32 system and so they run into some of the existing billiard bugs around the non-blocking interprocess pipe/file descriptor things that differ from the linux OS.  I am in a unique position on a windows system to try and help improve that part of billiard but I did not get fully complete with that work in my current time window so will resume when I can, but I did realize from my debugging that the Async pool class depends on compat.setblocking/isblocking which currently return -        raise NotImplementedError('isblocking not implemented on win32')  -- I've begun to implement them but have run into some typical win32 api stumbling blocks.   

Anyway, I think other than adding more test cases to this, this should solve the bug report for celery4 for people trying to use autoscaling, though it will require someone actually running this code to see for sure if there aren't more spots in the code that need to be checked for file descriptors.   We already removed the autoscale setting from our workers in order to eliminate this error for the time being.

PR #5506, opened 2019-05-06
Fix install celery from source via setuptools > 20.1.1, < 20.10.0

kosmos342:

## Description

Use [documented environment markers](https://www.python.org/dev/peps/pep-0508/#environment-markers). Because `python_implementation` marker is ["undocumented setuptools legacy"](https://github.com/pypa/setuptools/blob/83c667e0b2a98193851c07115d1af65011ed0fb6/pkg_resources/_vendor/packaging/markers.py#L94) and is absent in some versions of setuptools, for example in version 20.7.0 which is default in ubuntu 16.04

```
> pip install setuptools==20.7.0
> pip install --upgrade --no-cache-dir --force-reinstall celery==4.3.0 --no-binary :all:
...
    error in celery setup command: 'extras_require' must be a dictionary whose values are strings or lists of strings containing valid project/version requirement specifiers.

...
```

### DIFF
--- a/requirements/extras/brotli.txt
+++ b/requirements/extras/brotli.txt
@@ -1,2 +1,2 @@
-brotlipy>=0.7.0;python_implementation=="PyPy"
-brotli>=1.0.0;python_implementation=="CPython"
+brotlipy>=0.7.0;platform_python_implementation=="PyPy"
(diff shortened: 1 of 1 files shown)


thedrow:
This should be done in Kombu's repository as well.

Issue #5513, opened 2019-05-11
Can not get complete message content if it is long 


rhuab:
Which version I use?
4.3.0

What problem I encounter?
Recently I find that if I send a large json string to RabbitMQ via celery.send_task, the worker can receive the string completely. If I execute celery.app.control.Inspect.active and some other monitor methods, what I get is only part of the string such as

{
"id": "6c8ce3e8-3b8f-4479-97ac-b00ed77c40ba",
"name": "accelerated_pack",
"args": "('{"components": {"CAP_BUILD": {"repo_name": "cap_build", "commit_git": "a617e6321605470e838211ef9034fb82e9384c91", "source_path": "/tmp/all_source/cap_build", "install_dir": "/tmp/all_install/cap_build", "build_dir": "/tmp/all_build/cap_build", "artifact_id": "2b215641c4249980337ae5a0e00ea890"}, "CENSE": {"repo_name": "licensing", "commit_git": "ce52f024829174e5f654211db478db237fd28b82", "source_path": "/tmp/all_source/cense", "install_dir": "/tmp/all_install/cense", "build_dir": "/tmp/all_build/cense", "artifact_id": "b7c41ef521922baf527c72fe5f135f33"}, "DAP_4_0": {"repo_name": "dap4.0", "commit_git": "0f6adcc6e54d86c3904c16a8fa1a39273e0c9ded", "source_path": "/tmp/all_source/dap4.0", "install_dir": "/tmp/all_install/dap4.0", "build_dir": "/tmp/all_build/dap4.0", "artifact_id": "c10e64c6ec6f708c8b1bdd47426fec40"}, "REDIS": {"repo_name": "redis", "commit_git": "8b838e2a4eb655c6aa1004c3ad3d50ff4128492a", "source_path": "/tmp/all_source/redis", "install_dir": "/tmp/all_install/redis",...',)",
"kwargs": "{}",
"type": "accelerated_pack",
"hostname": "test@85aa72eff43c",
"time_start": 1557480202.127367,
"acknowledged": true,
"delivery_info": {
"exchange": "",
"routing_key": "test",
"priority": 0,
"redelivered": false
},
"worker_pid": 11
}

You can find at the end of the key "args",  some unexpected characters "." appears, it just represents the rest part of the string. I debug the celery and its dependencies and can't get any clue.

jheld:
[...]

PR #5514, opened 2019-05-11
Fix an issue that celery inpect api can not get complete args via the…

rhuab:
#5513

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,20 +21,20 @@ env:
 matrix:
   include:
   - python: 3.7
-      env: MATRIX_TOXENV=integration-rabbitmq
-      stage: integration
+    env: MATRIX_TOXENV=integration-rabbitmq
+    stage: integration
 
   - python: 3.7
-      env: MATRIX_TOXENV=integration-redis
-      stage: integration
+    env: MATRIX_TOXENV=integration-redis
+    stage: integration
 
   - python: 3.7
-      env: MATRIX_TOXENV=integration-dynamodb
-      stage: integration
+    env: MATRIX_TOXENV=integration-dynamodb
+    stage: integration
 
   - python: 3.7
-      env: MATRIX_TOXENV=integration-azureblockblob
-      stage: integration
+    env: MATRIX_TOXENV=integration-azureblockblob
+    stage: integration
 
   - python: '3.7'
     env: TOXENV=flake8


rhuab:
In celery.worker.control, the function "active(state, **kwargs)" is used to generate a reply for command "active". It calls "requests.info()" where the bug exists

rhuab:
@auvipy Could you tell me why my PR get py2.7 error on CI job? I am not very familiar with py2.7 and do not know the test rule in details. Sorry for bothering, but I need some help.

rhuab:
> you will need to add failing test to get a proper review.

Sorry I don't understand what you exactly mean. It seems that py2.7 test failed at a place not related to my modification and a Travis job failed at downloading py3.5

auvipy:
that's a different case, but to get your PR accepted you will need to add a proper unit test. so focus on that first, please.

rhuab:
> that's a different case, but to get your PR accepted you will need to add a proper unit test. so focus on that first, please.

I will do that

celery/worker/request.py
```diff
@@ -77,100 +77,240 @@ class Request(object):
...
             'reply_to': properties.get('reply_to'),
             'correlation_id': properties.get('correlation_id'),
-            'delivery_info': {
-                'exchange': delivery_info.get('exchange'),
-                'routing_key': delivery_info.get('routing_key'),
-                'priority': properties.get('priority'),
-                'redelivered': delivery_info.get('redelivered'),
-            }
-
+            'hostname': self._hostname,
```
thedrow: Why have you removed this?
I don't think it's a good idea because Celery expects this elsewhere.

rhuab:
@auvipy 
I add the following features to celery.app.worker.request:
1. Set Request object members as protected (starts with '_'), but for compatibility sake I also set property and setter function for them
2. Update self._request_dict['args'] and self._request_dict['kwargs'] in constructor instead of on the request being executed. The dict is updated by reference rather than full copy to avoid memory usage burst in some situations.
3. Request.info() function returns complete information of task args and kwargs, not in incomplete format.
4. Add some tests to the project

celery/worker/request.py
rhuab: Actually I don't remove any member of this class, just add two new members "args" and "kwargs" to maintain complete information of args and kwargs sent by Producer.

rhuab:
Any suggestions?

celery/worker/request.py
thedrow: But you removed the `delivery_info` key.

rhuab: self._delivery_info is created at line 155, and is appened to request_dict at line 165. It is exposed as property "delivery_info" at line 173 as well.

rhuab:
Any other question?

thedrow:
None.
I'll review this again.

thedrow:
Huh? I didn't close this issue.

thedrow:
That was a git mishap.
Sorry... I'm fixing it.

thedrow:
Merged in 0f1883374b788302f5bfb8e1fdcf9c13b199b17c.

Issue #5511, opened 2019-05-10
TypeError when waiting for results using join_native in a group


ikalyvas:
# Checklist

- [x] I have read the relevant section in the
  [contribution guide](http://docs.celeryproject.org/en/latest/contributing.html#other-bugs)
  on reporting bugs.
- [x] I have checked the [issues list](https://github.com/celery/celery/issues?q=is%3Aissue+label%3A%22Issue+Type%3A+Bug+Report%22+-label%3A%22Category%3A+Documentation%22)
  for similar or identical bug reports.
- [x] I have checked the [pull requests list](https://github.com/celery/celery/pulls?q=is%3Apr+label%3A%22PR+Type%3A+Bugfix%22+-label%3A%22Category%3A+Documentation%22)
  for existing proposed fixes.
- [x] I have checked the [commit log](https://github.com/celery/celery/commits/master)
  to find out if the bug was already fixed in the master branch.
- [x] I have included all related issues and possible duplicate issues
  in this issue (If there are none, check this box anyway).

## Mandatory Debugging Information

- [x] I have included the output of ``celery -A proj report`` in the issue.
    (if you are not able to do this, then at least specify the Celery
     version affected).
- [x] I have verified that the issue exists against the `master` branch of Celery.
- [x] I have included the contents of ``pip freeze`` in the issue.
- [x] I have included all the versions of all the external dependencies required
  to reproduce this bug.

## Optional Debugging Information

- [ ] I have tried reproducing the issue on more than one Python version
  and/or implementation.
- [ ] I have tried reproducing the issue on more than one message broker and/or
  result backend.
- [ ] I have tried reproducing the issue on more than one version of the message
  broker and/or result backend.
- [ ] I have tried reproducing the issue on more than one operating system.
- [ ] I have tried reproducing the issue on more than one workers pool.
- [ ] I have tried reproducing the issue with autoscaling, retries,
  ETA/Countdown & rate limits disabled.
- [ ] I have tried reproducing the issue after downgrading
  and/or upgrading Celery and its dependencies.

#### Related Issues

- None

#### Possible Duplicates

- None

## Environment & Settings

**Celery version**:
        4.3.0 (rhubarb)                                                                                                                                                                              

<details>
<summary><b><code>celery report</code> Output:<br>software -> celery:4.3.0 (rhubarb)<br><> kombu:4.5.0 py:3.6.4<br>                                                                                                                              
[...]

PR #5524, opened 2019-05-15
test case for issue #5511

ikalyvas:
@thedrow  please check

### DIFF
--- a/t/integration/test_canvas.py
+++ b/t/integration/test_canvas.py
@@ -12,7 +12,7 @@ from .conftest import flaky, get_active_redis_channels, get_redis_connection
 from .tasks import (add, add_chord_to_chord, add_replaced, add_to_all,
                     add_to_all_to_chord, build_chain_inside_task, chord_error,
                     collect_ids, delayed_sum, delayed_sum_with_soft_guard,
-                    fail, identity, ids, print_unicode, raise_error,
+                    fail, identity, ids, print_unicode, raise_error, raise_custom_error,
                     redis_echo, second_order_replace1, tsum)
 
 TIMEOUT = 120
@@ -231,6 +231,14 @@ class test_group:
         result = g.apply_async()
         while not result.ready():
             pass
+
+
+    @flaky
+    def test_not_ready_with_http_exception(self):
+        g = group([add.s(1, 2), raise_custom_error.s()])
+        result = g.apply_async()
+        with pytest.raises(TypeError) as e:
+            result.join_native()
 
     @flaky
     def test_empty_group_result(self, manager):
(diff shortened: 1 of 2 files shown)


t/integration/test_canvas.py
```diff
@@ -232,6 +233,30 @@ def test_ready_with_exception(self):
         while not result.ready():
             pass
 
+
+    @flaky
+    def test_not_ready_with_http_exception(self):
+        g = group([add.s(1, 2), raise_custom_error.s()])
+        result = g.apply_async()
+        with pytest.raises(TypeError) as e:
```
thedrow: Just so I'll get this right, this **should** fail with a TypeError right?
Instead, it raises a HTTPError?

t/integration/test_canvas.py
```diff
@@ -232,6 +233,30 @@ def test_ready_with_exception(self):
...
+            result.join_native()
+
+    @flaky
+    def test_raises_wrong_exception(self):
+        g = group([add.s(1, 2), raise_custom_error.s()])
+        result = g.apply_async()
+        with pytest.raises(urllib.error.HTTPError) as e:
```
thedrow: Just so I'll get this right, this **should** fail with a HTTPError right?
Instead, it raises another exception?

Testcases should fail the build if there's a bug.

t/integration/test_canvas.py
```diff
@@ -232,6 +233,30 @@ def test_ready_with_exception(self):
...
+            result.join_native()
+
+
+    @flaky
+    def test_raise_correct_error(self):
+        g = group([add.s(1, 2), raise_error.s()])
+        result = g.apply_async()
+        with pytest.raises(ValueError) as e:
```
thedrow: Just so I'll get this right, this **should** fail with a ValueError right?
Instead, it raises another exception?

t/integration/test_canvas.py
ikalyvas: No, the opposite.
The expected behaviour, should be that raise_custom_error tasks since it raises explicitly the HTTPError to be propagated as it is to the join method.
Instead a TypeError is raised stating that the HTTPError is not called with all 5 mandatory arguments.
And the raise_custom_error as you can see it raises the HTTPError correctly.

t/integration/test_canvas.py
ikalyvas: No,this is correct.Just testing to see that indeed the method propagates correct the exception

t/integration/test_canvas.py
ikalyvas: Yes,you are correct.
It raises TypeError .Attached the report.html to get an idea.

failed_test

PR #5529, opened 2019-05-16
Fix eager function not returning result after retries.

clems4ever:
## Description

Using apply function does not return correct results after at least one
retry because the return value of successive call is not going back to
the original caller.

This fix was needed for testing purposes and does not impact asynchronous workflow. It has been used for weeks in our prod.


clems4ever:
Yep, I will add some. Thanks for the quick feedback.

celery/app/task.py
```diff
@@ -27,6 +27,14 @@
 from .registry import _unpickle_task_v2
 from .utils import appstr
 
+class RetryEagerException(Retry):
```
thedrow: Why do we need a different type?
Couldn't we include the eagerness information inside the Retry exception?

celery/app/task.py
```diff
@@ -769,11 +774,14 @@ def apply(self, args=None, kwargs=None,
...
             propagate=throw, app=self._get_app(),
         )
+
         ret = tracer(task_id, args, kwargs, request)
         retval = ret.retval
+        state = states.SUCCESS if ret.info is None else ret.info.state
         if isinstance(retval, ExceptionInfo):
             retval, tb = retval.exception, retval.traceback
-        state = states.SUCCESS if ret.info is None else ret.info.state
+        if isinstance(retval, RetryEagerException):
```
thedrow: Why moving the call here is the right fix?

celery/app/task.py
georgepsarakis: It should be easy to reuse [Retry](https://github.com/celery/celery/blob/92ced78168b1ad92268af039827a9df957fc90af/celery/exceptions.py#L140):

```python
def __init__(self, message=None, exc=None, when=None, **kwargs):
    self.signature = kwargs.get('signature')
    self.eager = kwargs.get('eager', False)
```

celery/app/task.py
```diff
@@ -769,11 +774,14 @@ def apply(self, args=None, kwargs=None,
             task.name, task, eager=True,
             propagate=throw, app=self._get_app(),
         )
+
         ret = tracer(task_id, args, kwargs, request)
         retval = ret.retval
+        state = states.SUCCESS if ret.info is None else ret.info.state
         if isinstance(retval, ExceptionInfo):
             retval, tb = retval.exception, retval.traceback
-        state = states.SUCCESS if ret.info is None else ret.info.state
```
georgepsarakis: Not sure this line change is necessary, I can't seem to find any effect. Would you mind reverting this, if not needed?

celery/app/task.py
```diff
@@ -769,11 +774,14 @@ def apply(self, args=None, kwargs=None,
...
+        if isinstance(retval, RetryEagerException):
+            return retval.signature.apply()
```
georgepsarakis: So essentially here we detect that a `Retry` exception type has been raised and call the task signature synchronously again?

thedrow:
I'm going to try a Github app that rebases PRs.
/rebase

clems4ever:
Sorry, did not get time to update the PR yet.
I will rebase and add the tests as suggested. I'm figuring out how to run my tests individually to be able to iterate efficiently.

auvipy:
looking forward to your work :)

mchataigner:
I've just added some tests.
However the CI seems broken due to some issue in the setup for python 2.7, 3,5, 3,6 and for python 3.7 the issue is with couchbase backend needing couchbase service to run…

auvipy:
> I've just added some tests.
> However the CI seems broken due to some issue in the setup for python 2.7, 3,5, 3,6 and for python 3.7 the issue is with couchbase backend needing couchbase service to run…

I'm merging this. If you have time to come with another PR to fix couchbase failure?

mchataigner:
Apart from ignoring the tests or adding couchbase in the CI, I don't know…

mchataigner:
Thanks for merging this

auvipy:
you can start by skipping that one......

mchataigner:
I don't really understand, in couchbase test there is already:
`@skip.unless_module('couchbase')`
and in Couchbase module :
```
try:
    from couchbase import Couchbase
    from couchbase.connection import Connection
    from couchbase.exceptions import NotFoundError
    from couchbase import FMT_AUTO
except ImportError:
    Couchbase = Connection = NotFoundError = None   # noqa
```
and the error is thrown during the init of Couchbase backend
```       if Couchbase is None:
            raise ImproperlyConfigured(
                'You need to install the couchbase library to use the '
                'Couchbase backend.',
            )
```
but if `Couchbase` is None, it means module couchbase is not available, hence it should skip couchbase tests…

mchataigner:
Ok I found the issue, Couchbase published a major update to their python library breaking the API.
I'm working on a patch.

auvipy:
awesome thanks!

PR #5536, opened 2019-05-22
Refactor Build stages

auvipy:


.travis.yml
```diff
@@ -10,9 +10,9 @@ python:
 os:
     - linux
 stages:
-  - lint
   - test
   - integration
+  - lint
```
thedrow: No.
Lint should be triggered first.
There's no point to run the entire build if the code is not up to our standards.

auvipy: it is un needed to check the code standard before the code passes the test. if you disagree lets open an issue and the community vote. its annoying to see the whole build stopped just for a simple white space or comma. I even saw django channels following this. Lets open an issue regarding test stage ordering than

thedrow: Ok let's open an issue...

auvipy: issue is open

PR #5539, opened 2019-05-24
django example - upgrade django and example tasks that use models

forrestp:
## Description
This change extends the django example project by adding a Widget model and manipulating instances of it within celery tasks.

While setting up celery with my existing django project, I ran into some issues interacting with models within a celery task. The docs/example didn't have a good example of this. I forked the celery repo and got a couple simple tasks that interact with Models via the Django ORM working in the example project - this helped me track down the mistake I made in my own project. I figured other people might want to see this in the docs/example.

To get this working, I upgraded django to >=2.0.0 (including updating the settings file) and celery to >=4.3.0 within the example project- hope this is ok, I had some trouble with older django versions with python 3.7 and didn't bother figuring it out. I didn't try running these examples with older celery versions. Happy to spend a bit more time getting things to work with older versions if people would like, though I feel like it is ok for the example to use more-recent django and celery releases.


georgepsarakis:
@forrestp could you please rebase? I think tests will pass now.

PR #5540, opened 2019-05-24
task.replace now works in .apply() as well as .apply_async()

aleyan:
## Description
With this change tasks that use `self.replace` will execute their replacements synchronously if they themselves executed synchronously. This is expected behavior because tasks with children like chains, groups, and chords are also execution synchronously if they are called with `apply()`. This PR does not change the behavior in `.apply_async()` case.

The motivation for this change to run complex tasks of many chains/chords/groups/replacements synchronously in tests.

Current code with `raise self.replace(...)` pattern will continue functioning in async contexts, but will not work in sync contexts. New `return self.replace(...)` pattern will work in both contexts.

Tests forthcoming, but I wanted to get temperature for this change beforehand.

### DIFF
--- a/celery/app/task.py
+++ b/celery/app/task.py
@@ -855,14 +855,17 @@ class Task(object):
     def replace(self, sig):
         """Replace this task, with a new task inheriting the task id.
 
+        Execution of the host task ends immediately and no subsequent statements
+        will be run.
+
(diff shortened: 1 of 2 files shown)


aleyan:
Added a few tests.

PR #5545, opened 2019-05-26
Handle http_auth in Elasticsearch backend results

Thib17:
## Description

Using elasticsearch results backend, it doesn't respect scheme and basic auth credentials from configured url.

This fix was needed for using authenticated elasticsearch with https. It has been used for weeks in our prod.

### DIFF
--- a/celery/backends/elasticsearch.py
+++ b/celery/backends/elasticsearch.py
@@ -38,6 +38,8 @@ class ElasticsearchBackend(KeyValueStoreBackend):
     scheme = 'http'
     host = 'localhost'
     port = 9200
+    username = None
+    password = None
     es_retry_on_timeout = False
     es_timeout = 10
     es_max_retries = 3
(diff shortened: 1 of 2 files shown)


celery/backends/elasticsearch.py
```diff
@@ -132,7 +137,9 @@ def _get_server(self):
             '%s:%s' % (self.host, self.port),
             retry_on_timeout=self.es_retry_on_timeout,
             max_retries=self.es_max_retries,
-            timeout=self.es_timeout
+            timeout=self.es_timeout,
+            scheme=self.scheme,
+            http_auth=(self.username, self.password),
```
georgepsarakis: The default seems to be `http_auth=None`, so I think it would be better to use this value instead of a tuple of empty strings.

celery/backends/elasticsearch.py
```diff
@@ -132,7 +137,9 @@ def _get_server(self):
             '%s:%s' % (self.host, self.port),
             retry_on_timeout=self.es_retry_on_timeout,
             max_retries=self.es_max_retries,
-            timeout=self.es_timeout
+            timeout=self.es_timeout,
+            scheme=self.scheme,
```
georgepsarakis: I guess this is `http` by default, correct?

celery/backends/elasticsearch.py
```diff
@@ -38,6 +38,8 @@ class ElasticsearchBackend(KeyValueStoreBackend):
     scheme = 'http'
     host = 'localhost'
     port = 9200
+    username = ''
```
georgepsarakis: Perhaps `None` would be easier to detect, since these parameters are optional.

Thib17:
Thank you for the really quick review.
I just added unit tests.

celery/backends/elasticsearch.py
Thib17: Done

celery/backends/elasticsearch.py
Thib17: Yes, I added a unit test to check that.

celery/backends/elasticsearch.py
Thib17: Done

PR #5581, opened 2019-06-11
Restart consumer if the celery connection drops

jagguli:
original commit by @bremac https://github.com/celery/celery/commit/385a60df09201a17ad646c71eb1c00255d0a4431?diff=unified

Previously if an ack or reject message failed because the connection
was unavailable then celery would stop accepting messages until it was
restarted. This problem is that the main celery loop is responsible for
detecting closed connections, but the connection errors would not always
reach the main loop.

There are three places in the celery that share a long-running AMQP connection:

  1. The core worker loop listens for new messages.
  2. The ack / reject code that runs after a task completes.
  3. The heartbeat loop that keeps the connection alive.

Neither of the first two are guaranteed to see an error if the connection drops.
The main listener may never see an error if the connection drops since it may be
swallowed by drain_events; the connection may drop while no work is being done,
so the ack / reject code will never be triggered.

Fortunately the heartbeat loop is guaranteed to see an error if the connection
dies: periodic writes to the socket will fail with a broken pipe error.
Unfortunately it runs in a separate thread, so previously connection errors were
swallowed silently.

This commit alters the heartbeat code so that heartbeat error are always
re-raised in the main loop. This triggers existing code in the worker that
restarts the worker, reestablishing the connection. With the fix in place I've
been able to trigger ten long-running (three minute) RCA queries without the
worker hanging; without the fix it became unavailable after one or two queries.


celery/worker/loops.py
```diff
@@ -28,11 +28,23 @@ def _quick_drain(connection, timeout=0.1):
 
 
 def _enable_amqheartbeats(timer, connection, rate=2.0):
-    if connection:
-        tick = connection.heartbeat_check
-        heartbeat = connection.get_heartbeat_interval()  # negotiated
-        if heartbeat and connection.supports_heartbeats:
-            timer.call_repeatedly(heartbeat / rate, tick, (rate,))
+    heartbeat_error = [None]
```
clokep: I'm not seeing why this is a `list` instead of just `None`. What benefit is there by making this a list and sometimes modifying the first parameter?

thedrow: It preallocates the list with one item.
It also avoids confusion and possible bugs with `list.append()`.

It's possible that this should be a list since it is reference counted and can be mutated from anywhere.

I'm OK with this, but I'd like the OP to verify.

clokep: Sure, although the pre-allocation seems like an unnecessary optimization.

My question is really -- why isn't this just a bare value?

thedrow: @jagguli Care to explain?

NickWoodhams:
Thanks for the work on this, would love to see it merged.

celery/worker/loops.py
```diff
@@ -80,6 +92,8 @@ def asynloop(obj, connection, consumer, blueprint, hub, qos,
                 raise WorkerShutdown(should_stop)
             elif should_terminate is not None and should_stop is not False:
                 raise WorkerTerminate(should_terminate)
+            elif heartbeat_error[0] is not None:
```
reutsharabani: I would change both `elif`s to just `if`s. Less mental burden.

celery/worker/loops.py
reutsharabani: don't you have to pre-allocate a mutable construct to modify it in `tick`? I'll wait for @jagguli 's explanation but I think that is it.

auvipy:
could you plz resolve conflicts?

auvipy:
one test is failing

drsantos20:
Do we already have test for this change?

thedrow:
> Do we already have test for this change?

One test has to be updated and an integration test would be nice.

celery/worker/loops.py
jagguli: pre-allocating because somewhere else `if not heartbeat:` will be true, there is more to it I tried a couple of different vairations and this is the simplest

auvipy: Jaggull, can you update tests for this PR?

madchap:
This would really nice to see merged.

auvipy:
> This would really nice to see merged.

need to add unit test. can you push that? did you try this patch in dev env?

jagguli:
I could use some help with writing unittests for this, I have no idea where to start.
The original author of this change is @bremac I'm just using this in my deployment and it seems to work

jagguli:
FYI I also have this change in kombu to raise BrokenPipeError https://github.com/celery/kombu/compare/master...StreetHawkInc:master

auvipy:
first, try to resolve the conflicts then check these tests https://github.com/celery/celery/tree/master/t/unit/worker

yshaban:
Would love to see this merged along with the following:
https://github.com/celery/kombu/pull/1231
https://github.com/celery/celery/pull/6528

thedrow:
I think I can work on the tests for this if no one else would.

PR #5583, opened 2019-06-12
Fix error in docs, use new setting name for task_routes and task_queues.

mohi7solanki:
## Description

I think the docs are still referring to the old name for task_routes and task_queues setting, these are now under the `task` namespace. `CELERY_ROUTES` and `CELERY_QUEUES` does not work,


clokep:
The left column of that table is supposed to refer to the old settings names. It is showing what to replace it with.

PR #5584, opened 2019-06-13
Support load result_backend setting in app.config_from_object.

245967906:

## Description

I define the following parameters in the celeryconfig.py.
celery_broker_url = 'redis://localhost/0'
celery_result_backend = 'redis://localhost/1'

When I read the configuration using the following method:
app. config_from_object('celeryconfig', namespace = 'celery')

Broker_url loads correctly, but result_backend is not set.

So I tried to fix it, If there is any questions, please let me know. thanks.

### DIFF
--- a/celery/app/utils.py
+++ b/celery/app/utils.py
@@ -114,8 +114,8 @@ class Settings(ConfigurationView):
     def result_backend(self):
         return (
             os.environ.get('CELERY_RESULT_BACKEND') or
-            self.get('CELERY_RESULT_BACKEND') or
-            self.get('result_backend')
+            self.get('result_backend') or
+            self.get('CELERY_RESULT_BACKEND')
         )
 
     @property
--- a/celery/app/utils.py
+++ b/celery/app/utils.py
@@ -114,7 +114,8 @@ class Settings(ConfigurationView):
     def result_backend(self):
         return (
             os.environ.get('CELERY_RESULT_BACKEND') or
-            self.get('CELERY_RESULT_BACKEND')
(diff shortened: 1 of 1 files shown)


245967906:
I define the following parameters in the  celeryconfig.py.
celery_broker_url = 'redis://localhost/0'
celery_result_backend = 'redis://localhost/1'

When I read the configuration using the following method:
app. config_from_object('celeryconfig', namespace = 'celery')

Broker_url loads correctly, but result_backend is not set.

So I tried to fix it, If there is any questions, please let me know.  thanks.

celery/app/utils.py
```diff
@@ -114,7 +114,8 @@ def broker_url(self):
     def result_backend(self):
         return (
             os.environ.get('CELERY_RESULT_BACKEND') or
-            self.get('CELERY_RESULT_BACKEND')
+            self.get('CELERY_RESULT_BACKEND') or
+            self.get('result_backend')
```
auvipy: on celery 4.x series, shouldn't result_backend be looked first?

245967906: > on celery 4.x series, shouldn't result_backend be looked first?

yes, you're right, i will change the order. but do you want to retain the support of CELERY_RESULT_BACKEND?

auvipy: that will be dropped in celery 5 :) so...

245967906: Emmm... Okay, now celery still 4.x, so let's keep it...

245967906:
@auvipy thank you!

Issue #3743, opened 2017-01-07
chain/chord not propagating exceptions as expected


zaro:
## Checklist

- [x] I have included the output of ``celery -A proj report`` in the issue.
      (if you are not able to do this, then at least specify the Celery
       version affected).
- [ ] I have verified that the issue exists against the `master` branch of Celery.

## Steps to reproduce

Tasks:
```python
@app.task
def a1(arg):
    print('Task A1, starring with arg:' + str(arg))
    if arg == 3:
        raise Exception('A1 Error')
    time.sleep(5)
    print('A1 done')
    return {'a1': True}

@app.task
def a2(arg):
    print('Task A2, starring with arg:' + str(arg))
    time.sleep(5)
    print('A2 done')
    return {'a2': True}

@app.task
def b1(arg):
    print('Task B1, starring with arg:' + str(arg))
    time.sleep(5)
    print('B1 done')
    return {'b1': True}

```

If we start a chrod() with 3 a1 task fails as expected:
```python
>>> chord([ a1.s(1)  , a1.s(2) , a1.s(3)  ])(b1.s()).get()
Traceback (most recent call last):
  File "<console>", line 1, in <module>
  File "/Users/zaro/dev/proj-dev/proj/.direnv/python-3.5.2/lib/python3.5/site-packages/celery/result.py", line 189, in get
    on_message=on_message,
  File "/Users/zaro/dev/proj-dev/proj/.direnv/python-3.5.2/lib/python3.5/site-packages/celery/backends/async.py", line 191, in wait_for_pending
    return result.maybe_throw(callback=callback, propagate=propagate)
  File "/Users/zaro/dev/proj-dev/proj/.direnv/python-3.5.2/lib/python3.5/site-packages/celery/result.py", line 294, in maybe_throw
    self.throw(value, self._to_remote_traceback(tb))
  File "/Users/zaro/dev/proj-dev/proj/.direnv/python-3.5.2/lib/python3.5/site-packages/celery/result.py", line 287, in throw
    self.on_ready.throw(*args, **kwargs)
  File "/Users/zaro/dev/proj-dev/proj/.direnv/python-3.5.2/lib/python3.5/site-packages/vine/promises.py", line 213, in throw
    reraise(type(exc), exc, tb)
  File "/Users/zaro/dev/proj-dev/proj/.direnv/python-3.5.2/lib/python3.5/site-packages/vine/five.py", line 176, in reraise
    raise value
[...]

PR #5587, opened 2019-06-13
Fix Chain Exceptions propagations

### LINKED ISSUE #3810: Group of Chains Exception/Error Propagation
## Checklist
-  I'm running Celery 4.0.2

## Steps to reproduce
tasks.py
```
@shared_task
def add(x,y):
    return x+y

@shared_task
def raising_exception():
    raise RuntimeError("BLAH")
```
test.py
```
from celery import chord, group
from my_app.tasks import xsum, raising_exception

tasks = group(add.si(1,2), raising_exception.si() | add.si(1,1) )
result = tasks.apply_async()

try:
    result.get()
except Exception as ex:
    raise ex
```
Run the following:
celery worker -A grizzly_project -c 8 &
python manage.py runserver

then in another terminal, run:
python test.py
## Expected behavior
When the "raising_exception" task fails, result.get() should propagate one task back in the chain from the last task (for the second part of the group) to receive the exception: " RuntimeError: BLAH" thus allowing me to catch the Exception similar to below:
```
  File "test.py", line 15, in <module>
    raise ex
celery.backends.base.RuntimeError: BLAH
```
This is the expected behaviour, according to the back propagation of errors in a chain that was implemented in the closing of Issue [1014](https://github.com/celery/celery/issues/1014).

## Actual behavior
Rather, when the result.get() statement executes, the program hangs. 

When I change the group statement from the above code to this such that the failing task is the last to execute:
```
tasks = group(add.si(1,2), add.si(1,1) | raising_exception.si() )
```
...the result.get() doesn't hang but rather the Exception is caught, as expected 

Also, if i simply make it just a chain (rather than a group of chains) with the failing task first:
```
tasks = raising_exception.si() | add.si(1,1)
```
...result.get() doesn't hang but accurately catches the Exception (granted a wait a second or two before executing result.get())

SharpEdgeMarshall:
## Description

Fixes #3810, #3743

This commit https://github.com/celery/celery/commit/2055cbd056f4d2822e0e88b22e36cfca363952a6 introduced the `on_interval` arg but there is a typo and the promise passed to `self.backend.wait_for_pending` is messed up.

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -269,3 +269,4 @@ Artem Vasilyev, 2018/11/24
 Victor Mireyev, 2018/12/13
 Florian Chardin, 2018/10/23
 Shady Rafehi, 2019/02/20
+Fabio Todaro, 2019/06/13
--- a/celery/result.py
+++ b/celery/result.py
@@ -205,7 +205,7 @@ class AsyncResult(ResultBase):
             assert_will_not_block()
         _on_interval = promise()
         if follow_parents and propagate and self.parent:
-            on_interval = promise(self._maybe_reraise_parent_error, weak=True)
+            _on_interval = promise(self._maybe_reraise_parent_error, weak=True)
             self._maybe_reraise_parent_error()
         if on_interval:
             _on_interval.then(on_interval)
--- a/t/integration/test_canvas.py
+++ b/t/integration/test_canvas.py
@@ -561,14 +561,14 @@ class test_chord:
         )
         res = c1()
         try:
-            res.get()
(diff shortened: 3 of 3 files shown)


auvipy:
could you just rebase from master?

SharpEdgeMarshall:
There are no tests about `on_interval` so what I could do is add an integration test with a chain that has a task failing in the middle.

auvipy:
good job on integration tests!!

auvipy:
the integration tests seem unrelated. but if you have time, could you try to fix the assertion error of existing failure?

PR #5614, opened 2019-06-22
Fix: Add a patch to support Python 2.7

shouc:
## Description

In Python 2.7 or below, there is no `name` instance for exceptions.   
Thus, whenever the user import a package that does not exist with Python 2.7 or below, following unexpected exception is raised:

```
  File "celery", line 11, in <module>
    load_entry_point('celery==4.4.0rc2', 'console_scripts', 'celery')()
  File "celery/__main__.py", line 16, in main
    _main()
  File "celery/bin/celery.py", line 322, in main
    cmd.execute_from_commandline(argv)
  File "celery/bin/celery.py", line 496, in execute_from_commandline
    super(CeleryCommand, self).execute_from_commandline(argv)))
  File "celery/bin/base.py", line 293, in execute_from_commandline
    self.on_error(UNABLE_TO_LOAD_APP_MODULE_NOT_FOUND.format(e.name))
AttributeError: 'exceptions.ImportError' object has no attribute 'name'
```
This pull request adds a small patch so that Python 2.7 or below user would not be bothered by this error and know what package does not exist.   
This patch could also be easily removed when Python 2.7 is deprecated.

### DIFF
--- a/celery/bin/base.py
+++ b/celery/bin/base.py
@@ -288,7 +288,13 @@ class Command(object):
         try:
             argv = self.setup_app_from_commandline(argv)
         except ModuleNotFoundError as e:
-            self.on_error(UNABLE_TO_LOAD_APP_MODULE_NOT_FOUND.format(e.name))
+            # In Python 2.7 and below, there is no name instance for exceptions
+            # Remove when Python 2.7 is not supported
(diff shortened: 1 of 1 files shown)


thedrow:
Thank you!

Issue #2724, opened 2015-07-24
Task result is truncated to 128 chars by workers


AEtrade:
Currently  when worker executes the task it returns truncated result  value.

Actual result length is > 128 characters but when accessing this task info through Flower API  its getting truncated .

Is there any parameter where we can increase the truncated length from 128 chars to more OR  suppress truncate feature.

Appreciate your help

Example:

Expected Result:
----------------------
'Hello How are you 1 Hello How are you 2 Hello How are you 3 Hello How are you 4 Hello How are you 5 Hello How are you 6 Hello How are you 7.
.
Observed Result:
----------------------
'Hello How are you 1 Hello How are you 2 Hello How are you 3 Hello How are you 4 Hello How are you 5 Hello How are you 6...

Thanks,
AEtrade.

thedrow:
Can you provide me with a test case that I can use to reproduce your problem?

jaylik:
Same problem here - with same versions on Windows result on database is not getting truncated but on Linux it does.

thedrow:
I think that the task's result is truncated only on event messages. If you get the result using ```result.get()``` you'll get the full result.
@jaylik Am I correct?

ask:
Yeah, events truncate the task result, this is because the events are used exclusively for monitoring.  We may be able to make this character limit configurable.

AEtrade:
Wondering when are you planning to push this feature. It will be really useful if we can have the task event result length configurable.

centralniak:
+1

Please do. 

I think it's a terrible idea to truncate using a hard limit. For me a truncated result is just as unhelpful as no result at all.

ask:
It's available in the master branch:

    app.Task.resultrepr_maxsize = 2000`

AEtrade:
Thank you!!

laixintao:
I was searching for this configuration for the whole night, seriously this should be documented.

laixintao:
Please take a look: https://github.com/celery/celery/pull/5635

@ask @auvipy

PR #5635, opened 2019-07-08
doc: update event's result max size configuration.

laixintao:
Just add a document for a configuration.

see:

https://github.com/celery/celery/issues/2724#issuecomment-16401350

### DIFF
--- a/docs/userguide/monitoring.rst
+++ b/docs/userguide/monitoring.rst
@@ -725,6 +725,11 @@ Run-time is the time it took to execute the task using the pool.
 (Starting from the task is sent to the worker pool, and ending when the
 pool result handler callback is called).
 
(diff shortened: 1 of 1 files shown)


thedrow:
I'm not sure the API was meant to be used that way.

laixintao:
How about moving it to the configurations?

https://docs.celeryproject.org/en/latest/userguide/configuration.html#id12

thedrow:
I guess it's possible to add a default value for result representation max size.

jheld:
Are there docs around the args and kwargs reprs?

thedrow:
I don't think so.

jheld:
I would have appreciated more docs. I've found the code, but it took awhile.

thedrow:
Well this is not the proper fix in my opinion.
Feel free to issue a new one.

PR #5644, opened 2019-07-16
Passthrough Celery Task Name into AsyncResult

srikiraju:
## Description

Simple changeset to allow `AsyncResult` to always be populated with a `task_name`. Though a comment in the initializer for `celery.Result` says that `task_name` in there is deprecated, I think it would be useful to keep.

### DIFF
--- a/celery/contrib/abortable.py
+++ b/celery/contrib/abortable.py
@@ -147,9 +147,9 @@ class AbortableTask(Task):
 
     abstract = True
 
-    def AsyncResult(self, task_id):
+    def AsyncResult(self, task_id, **kwargs):
         """Return the accompanying AbortableAsyncResult instance."""
-        return AbortableAsyncResult(task_id, backend=self.backend)
+        return AbortableAsyncResult(task_id, backend=self.backend, **kwargs)
 
     def is_aborted(self, **kwargs):
         """Return true if task is aborted.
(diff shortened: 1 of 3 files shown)


celery/contrib/abortable.py
```diff
@@ -147,9 +147,9 @@ class AbortableTask(Task):
 
     abstract = True
 
-    def AsyncResult(self, task_id):
+    def AsyncResult(self, task_id, **kwargs):
```
dejlek: I am always sceptical about having **kwargs. I know Python developers love to stick it everywhere but question here is do we really need it here when we know ahead it is used only to pass a single parameter...

srikiraju: I see, how do I get the name if we know the task id? I'm writing some internal wrappers around `AsyncResult` to measure timing, so its important to know the name of the task in question

srikiraju: So, I still haven't figured out how to solve my use-case. The problem I'm facing is the following:

We want to dump a metric for counting how long tasks take as perceived by the clientside when they wait for results. For this, I overrode `celery.AsyncResult` with an extended class with a timer.

When we call a task using `apply_async()`, everything is fine since the `Task:AsyncResult()` method in `celery/app/task.py` appropriately passes in `task_name` argument. This name is useful for dumping a metric/timer and for debugging later on. However, when we send a task using `send_task()`, this doesn't happen, which is what I'm adding in this PR. This avoids the confusion about when the argument is available is not.

srikiraju: @dejlek ^

srikiraju:
@auvipy I don't see a clear path to getting the functionality I need without this PR, can we re-open it?

PR #5671, opened 2019-08-08
Fix termination of asynloop

Jamim:
Currently, `asynloop` hangs on termination in case when `should_terminate` is not `None` and `should_stop` is `False`.

It seems reasonable to use `state.maybe_shutdown` in order to fix the issue and, on the other hand, to simplify the code.

Also, I believe that termination should have a higher priority than shutting down, so order of checking for termination and for shutting down should be changed.

P.S. A CI failure for Windows is expected and it's not related to these changes.

### DIFF
--- a/celery/worker/loops.py
+++ b/celery/worker/loops.py
@@ -5,7 +5,7 @@ import errno
 import socket
 
 from celery import bootsteps
-from celery.exceptions import WorkerLostError, WorkerShutdown, WorkerTerminate
+from celery.exceptions import WorkerLostError
 from celery.utils.log import get_logger
 
 from . import state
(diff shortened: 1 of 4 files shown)


celery/worker/loops.py
```diff
@@ -71,15 +71,7 @@ def asynloop(obj, connection, consumer, blueprint, hub, qos,
...
     try:
         while blueprint.state == RUN and obj.connection:
-            # shutdown if signal handlers told us to.
-            should_stop, should_terminate = (
-                state.should_stop, state.should_terminate,
-            )
-            # False == EX_OK, so must use is not False
-            if should_stop is not None and should_stop is not False:
-                raise WorkerShutdown(should_stop)
-            elif should_terminate is not None and should_stop is not False:
```
Jamim: `should_stop` in this line is the root cause of the issue.
It should be `should_terminate`, just like in the implementation of the `state.maybe_shutdown` method.

Jamim:
Hi @auvipy,
I've enhanced existing tests and also changed `state.maybe_shutdown`.
Please re-review the PR.
Thanks a lot!

PR #5672, opened 2019-08-08
Fix Travis CI job for Python 3.7

Jamim:
Hello,

The [typing](https://github.com/python/typing) package is a built-in for all currently supported Python versions, so it seems to be not required anymore.
Presence of this package leads to an error on CI for Python 3.7.

These changes also:
 - remove redundant options from `.travis.yml`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,4 @@
 language: python
-sudo: required
 dist: xenial
 cache: pip
 python:
@@ -57,11 +56,9 @@ matrix:
     stage: lint
   - python: pypy2.7-7.1.1
     env: TOXENV=pypy
-    dist: xenial
     before_install: sudo apt-get update && sudo apt-get install libgnutls-dev
   - python: pypy3.5-7.0
     env: TOXENV=pypy3
-    dist: xenial
     before_install: sudo apt-get update && sudo apt-get install libgnutls-dev
 
 before_install:
--- a/tox.ini
+++ b/tox.ini
@@ -61,7 +61,7 @@ basepython =
     flake8,apicheck,linkcheck,configcheck,pydocstyle,bandit: python3.7
     flakeplus: python2.7
 usedevelop = True
-install_command = python -m pip --disable-pip-version-check install {opts} {packages}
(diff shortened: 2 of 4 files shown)


.travis.yml
```diff
@@ -57,11 +56,9 @@ matrix:
     stage: lint
   - python: pypy2.7-7.1.1
     env: TOXENV=pypy
-    dist: xenial
```
georgepsarakis: I guess these are redundant due to the distribution global configuration setting: https://github.com/celery/celery/pull/5672/files#diff-354f30a63fb0907d4ad57269548329e3R2 ?

tox.ini
```diff
@@ -60,7 +60,6 @@ basepython =
     pypy3: pypy3
     flake8,apicheck,linkcheck,configcheck,pydocstyle,bandit: python3.7
     flakeplus: python2.7
-usedevelop = True
```
georgepsarakis: I think this would be necessary for local development and testing. What errors are you observing due to this flag?

.travis.yml
Jamim: Yes, you are right.

tox.ini
Jamim: Hello @georgepsarakis,

> I think this would be necessary for local development and testing.

It's not necessary for local development for sure since it's for `tox` only, and I believe that it's not necessary for local testing because all the tests pass.

> What errors are you observing due to this flag?

You can see a corresponding error on [the latest master build for Python 3.7 on Travis CI](https://travis-ci.org/celery/celery/jobs/569235785).

Jamim: python/typing#573 is probably related to this issue.

auvipy: can we make it False?

Jamim: Sorry, I missed that the `usedevelop` option [was introduced](https://github.com/celery/celery/commit/bd0ed23c81b20fd75c0e2188fcf78e4d74898953) in order to fix coverage measurement.
Let me try to find a better solution.

Jamim:
Hello @georgepsarakis and @auvipy,

I found a better solution that doesn't affect coverage.
Could you please re-review this PR?

Thanks!

Issue #5651, opened 2019-07-21
Unit test for issue https://github.com/celery/celery/issues/5496


auvipy:
https://github.com/celery/celery/issues/5496

Issue #5676, opened 2019-08-14
Test failing in master for 4.4.0rc3


auvipy:
- t/unit/backends/test_mongodb.py:124 test_MongoBackend.test_init_mongodb_dns_seedlist

        try:

            results = resolver.query('_mongodb._tcp.' + self.__fqdn, 'SRV',

                                     lifetime=self.__connect_timeout)

        except Exception as exc:

            if not encapsulate_errors:

                # Raise the original error.

                raise

            # Else, raise all errors as ConfigurationError.

>           raise ConfigurationError(str(exc))

E           pymongo.errors.ConfigurationError: mock_resolver() got an unexpected keyword argument 'lifetime'

auvipy:
probably the commit to checking https://github.com/celery/kombu/commit/0ad7d08c4cb4abce32e9d765d0bb16b92830451f

Issue #5496, opened 2019-05-02
Nested group(chain(group)) fails


p-eli:
# Checklist

- [x] I have read the relevant section in the
  [contribution guide](http://docs.celeryproject.org/en/latest/contributing.html#other-bugs)
  on reporting bugs.
- [x] I have checked the [issues list](https://github.com/celery/celery/issues?q=is%3Aissue+label%3A%22Issue+Type%3A+Bug+Report%22+-label%3A%22Category%3A+Documentation%22)
  for similar or identical bug reports.
- [ ] I have checked the [pull requests list](https://github.com/celery/celery/pulls?q=is%3Apr+label%3A%22PR+Type%3A+Bugfix%22+-label%3A%22Category%3A+Documentation%22)
  for existing proposed fixes.
- [ ] I have checked the [commit log](https://github.com/celery/celery/commits/master)
  to find out if the bug was already fixed in the master branch.
- [x] I have included all related issues and possible duplicate issues
  in this issue (If there are none, check this box anyway).

## Mandatory Debugging Information

- [x] I have included the output of ``celery -A proj report`` in the issue.
    (if you are not able to do this, then at least specify the Celery
     version affected).
- [ ] I have verified that the issue exists against the `master` branch of Celery.
- [ ] I have included the contents of ``pip freeze`` in the issue.
- [ ] I have included all the versions of all the external dependencies required
  to reproduce this bug.

## Optional Debugging Information

- [x] I have tried reproducing the issue on more than one Python version
  and/or implementation.
- [ ] I have tried reproducing the issue on more than one message broker and/or
  result backend.
- [ ] I have tried reproducing the issue on more than one version of the message
  broker and/or result backend.
- [ ] I have tried reproducing the issue on more than one operating system.
- [ ] I have tried reproducing the issue on more than one workers pool.
- [ ] I have tried reproducing the issue with autoscaling, retries,
  ETA/Countdown & rate limits disabled.
- [ ] I have tried reproducing the issue after downgrading
  and/or upgrading Celery and its dependencies.

#### Related Issues

- None

#### Possible Duplicates

- Subsequent groups within a chain fail #3585
- Consecutive groups in chain fails #4848 
- task/group chains fails in some scenarios #5467

## Environment & Settings

**Celery version**: 4.3.0

<details>
<summary><b><code>celery report</code> Output:

software -> celery:4.3.0 (rhubarb) kombu:4.5.0 py:3.5.3
billiard:3.6.0.0 redis:3.2.1
[...]

PR #5679, opened 2019-08-16
UT: add unit test for ResultSet.join_native

tothegump:
- #5651

### DIFF
--- a/t/unit/tasks/test_result.py
+++ b/t/unit/tasks/test_result.py
@@ -7,7 +7,7 @@ from contextlib import contextmanager
 
 import pytest
 
-from case import Mock, call, patch, skip
+from case import Mock, call, patch, skip, MagicMock
 from celery import states, uuid
 from celery.app.task import Context
 from celery.backends.base import SyncBackendMixin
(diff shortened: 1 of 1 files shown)


tothegump:
Why another one fails?

auvipy:
> Why another one fails?

the pymongo failing might be the result of new kombu release which switched from pymongo 2.x to 3.x. need to check more in-depth though. do you have the bandwidth to check that?

tothegump:
> > Why another one fails?
> 
> the pymongo failing might be the result of new kombu release which switched from pymongo 2.x to 3.x. need to check more in-depth though. do you have the bandwidth to check that?

Interesting. Seems relates to issue #5676 .
I'd like to have a look.

t/unit/tasks/test_result.py
```diff
@@ -469,6 +469,33 @@ def test_get(self):
         x.get()
         x.join_native.assert_called()
 
+    def test_join_native(self):
+        def _get_meta(tid, result=None, children=None):
+            return {
+                'status': states.SUCCESS,
+                'result': result,
+                # 'traceback': traceback,
```
thedrow: No code in comments please.

tothegump: Yes, you are right.

thedrow:
@tothegump Can you please explain how this unit test checks that #5496 is fixed?

tothegump:
> @tothegump Can you please explain how this unit test checks that #5496 is fixed?



> It's not clear to me how this test (while it is certainly welcome) is a good test to verify #5496.

In #5496 fails because, when results have GroupResult, which has no "_cache" attr.
Then, join_native can not process correctly this situation which iter_native returns.
So, in this unit test, just mock iter_native returns when GroupResult is an element of results.

tothegump:
I wish I can explain clearly.

thedrow:
Some comments would be nice.
Also, you should rename the test as it doesn't simply test `ResultSet.join_native`.
It does so with very specific conditions.
If you have other test cases in mind, please include them.

auvipy:
――――――――――――――――――――――――― test_GroupResult.test_eq_ne ――――――――――――――――――――――――――

self = <t.unit.tasks.test_result.test_GroupResult object at 0x7f27e6cd9f60>

>   ???

E   AttributeError: 'Celery' object has no attribute 'groupresult'

t/unit/tasks/test_result.py:682: AttributeError

 t/unit/tasks/test_result.py::test_GroupResult.test_eq_ne ⨯

tothegump:
> Some comments would be nice.
> Also, you should rename the test as it doesn't simply test `ResultSet.join_native`.
> It does so with very specific conditions.
> If you have other test cases in mind, please include them.

Make sense. I will do it later.

tothegump:
Looks like, I should do more work for this issue. I mark it as WIP and divide into it:
1. change ut name
2. add comment to this ut;
3. fix the other ut failings.

auvipy:
the UT is passing now, the failing Integration test is different & causing for a kombu issue

auvipy:
this is the root cause of IT failures https://github.com/celery/kombu/commit/052f7601d3a1e09263cc0ebda7ef8c16e6b90175 do you have time to investigate a better fix rather than reverting this?

thedrow:
I'll investigate.

PR #5694, opened 2019-08-25
Fix issue with `RuntimeError: OrderedDict mutated during iteration`

sshishov:
## Description
In Python > 3.5 OrderedDict is not allowed to be changed during iteration.

Event state contains tasks, workers and other information. Tasks are `kombu LRUCache` implementation which internally stores data as OrderedDict.
When we are trying to [deepcopy](https://docs.python.org/3/library/copy.html) state or [shelve](https://docs.python.org/3/library/shelve.html) it (tasks should be more than 2 otherwise mutation is not happened), then we got exception `RuntimeError: OrderedDict mutated during iteration`

To easy reproduce you can try to `deepcopy` state after it contains 2 tasks with appropriate `root_id`s inside.

Added test which will fail on Python > 3.5 without these changes.

### DIFF
--- a/celery/events/state.py
+++ b/celery/events/state.py
@@ -398,7 +398,7 @@ class Task(object):
     def parent(self):
         # issue github.com/mher/flower/issues/648
         try:
-            return self.parent_id and self.cluster_state.tasks[self.parent_id]
+            return self.parent_id and self.cluster_state.tasks.data[self.parent_id]
         except KeyError:
             return None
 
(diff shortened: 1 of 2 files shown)


sshishov:
This issue fixes https://github.com/mher/flower/issues/779 as currently persistent mode is not working on Python > 3.5 version

auvipy:
thanks!

sshishov:
Welcome!

PR #5700, opened 2019-08-28
Fixes #5106: Retry WorkerLostError fix for canvas

### LINKED ISSUE #5106: Retry tasks which failed because of SIGKILL
Hello,

We are currently using celery with Broker=RabbitMQ & Backend=Redis.  Our setup runs on Kubernetes with auto-scaling enabled.  (Our number of pods scale up / down based on resource utilization).  We use the Forked Process mode to run our celery worker.

When the size of our cluster increases, everything works great.  However, when the pods turn down, the tasks that were running on the pods get marked as FAILED.  The sequence of actions is as follows:
1. Resource consumption decreases
2. Kubernetes realizes that pods that were scaled up now need to be turned down
3. Kubernetes sends `SIGKILL` signal to the pods that should turn down
4. Celery intercepts the signals and turns down all the Forked Processes
5. The tasks that were running on the processes return their execution back to the Main Process
6. The main process marks all the running tasks as FAILED

When the main process is the one that marks the running operations as FAILED, the `retry` code is never executed.

We do want to recover these tasks as they are part of a long running process.  Celery should provide a way to recover these tasks.

Usiel:
## Description
If `acks_late` and `reject_on_worker_lost` are enabled we do not mark a task as failed, assuming that the task has been requeued. 
Previously a `WorkerLostError` in a single task would cause a potential parent canvas to fail, even though the task was actually requeued and completed successfully.

### DIFF
--- a/celery/worker/request.py
+++ b/celery/worker/request.py
@@ -496,18 +496,8 @@ class Request(object):
         if isinstance(exc, Retry):
             return self.on_retry(exc_info)
 
-        # These are special cases where the process wouldn't've had
-        # time to write the result.
-        if isinstance(exc, Terminated):
-            self._announce_revoked(
-                'terminated', True, string(exc), False)
-            send_failed_event = False  # already sent revoked event
-        elif isinstance(exc, WorkerLostError) or not return_ok:
-            self.task.backend.mark_as_failure(
-                self.id, exc, request=self._context,
-                store_result=self.store_errors,
-            )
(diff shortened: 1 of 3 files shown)


auvipy:
thanks for your contribution!

thedrow:
@auvipy I'm right about the milestone right?

auvipy:
you are!! how I missed this!!!

Issue #5702, opened 2019-08-28
Celery's ssl_cert_reqs names do not match Redis's


arel:
# Checklist

- [x] I have checked the [issues list](https://github.com/celery/celery/issues?q=is%3Aissue+label%3A%22Issue+Type%3A+Enhancement%22+-label%3A%22Category%3A+Documentation%22)
  for similar or identical enhancement to an existing feature.
- [x] I have checked the [pull requests list](https://github.com/celery/celery/pulls?q=is%3Apr+label%3A%22Issue+Type%3A+Enhancement%22+-label%3A%22Category%3A+Documentation%22)
  for existing proposed enhancements.
- [x] I have checked the [commit log](https://github.com/celery/celery/commits/master)
  to find out if the if the same enhancement was already implemented in the
  master branch.
- [x] I have included all related issues and possible duplicate issues in this issue
      (If there are none, check this box anyway).

#### Related Issues

- #5371 this issue has to do with confusion about specifying `ssl_cert_reqs` as a constant (`ssl.CERT_REQUIRED` versus a string `'CERT_REQUIRED'`.)

#### Possible Duplicates

- None

# Brief Summary

The official redis python library expects `ssl_cert_reqs` to be one of: `'none'`, `'optional'`, or `'required'`.

See: https://github.com/andymccurdy/redis-py/blob/3.3.7/redis/connection.py#L793

Celery expects this to be one of `'CERT_NONE'`, `'CERT_OPTIONAL'`, or `'CERT_REQUIRED'`.

This can lead to confusion because the same redis URL cannot be used for both celery, with a redis backend, and redis directly.

For example, the following redis urls are incompatible:

```bash
"rediss://:${REDIS_TOKEN}@${REDIS_HOST}?ssl_cert_reqs=none"  # works for redis, not celery
"rediss://:${REDIS_TOKEN}@${REDIS_HOST}?ssl_cert_reqs=CERT_NONE"  # works for celery, not redis
```

This confusion could be avoided by reconciling the two formats.

To avoid future confusion, and remaining backward compatible, I propose _at a minimum_ letting celery recognize the redis `ssl_cert_reqs` strings.

Ideally, the documentation would also be updated to prefer the standard redis strings (mentioning that the `CERT_*` strings are also valid but deprecated).

# Design

## Architectural Considerations

None

## Proposed Behavior

Change celery to map both formats to the same value, such as:

```python
            ssl_string_to_constant = {'CERT_REQUIRED': CERT_REQUIRED,
[...]

PR #5703, opened 2019-08-28
Accept standard redis ssl_cert_req names

arel:

## Description

Fixes: #5702

The official redis python library expects ssl_cert_reqs
to be one of 'none', 'optional', or 'required'.

Celery expects this to be one of 'CERT_NONE', 'CERT_OPTIONAL',
or 'CERT_REQUIRED'.

This can lead to confusion because the same redis URL cannot
be used for both celery, with a redis backend, and redis directly.

This change allows either term to be recognized by celery.

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -272,3 +272,4 @@ Florian Chardin, 2018/10/23
 Shady Rafehi, 2019/02/20
 Fabio Todaro, 2019/06/13
 Shashank Parekh, 2019/07/11
+Arel Cordero, 2019/08/29
--- a/celery/backends/redis.py
+++ b/celery/backends/redis.py
@@ -212,7 +212,10 @@ class RedisBackend(BaseKeyValueStoreBackend, AsyncBackendMixin):
             ssl_cert_reqs_missing = 'MISSING'
             ssl_string_to_constant = {'CERT_REQUIRED': CERT_REQUIRED,
                                       'CERT_OPTIONAL': CERT_OPTIONAL,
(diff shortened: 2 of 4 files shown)


arel:
Thanks, @auvipy. I added unit tests for the proposed change.

auvipy:
thanks, Arel!

jdavisp3:
Question about this fix -- doesn't `kombu` need a similar change given the code here: https://github.com/celery/kombu/blob/6ef0a2d074ca6c9fd3cb499a0097ae3b0e02211d/kombu/utils/url.py#L53

It seems to be still assuming the query parameter is going to be `CERT_REQUIRED` and not `required`.

auvipy:
seems yes! are you up for contrib?

jdavisp3:
@auvipy Happy to make a patch, I'm just wondering how @arel is getting this to work without a change to `kombu`, maybe I'm missing something?

jdavisp3:
> @auvipy Happy to make a patch, I'm just wondering how @arel is getting this to work without a change to `kombu`, maybe I'm missing something?

Well based on latest releases of `celery` and `kombu` this still doesn't work for me so I'll make a patch for `kombu`.

auvipy:
plz proceed

georgepsarakis:
> @auvipy Happy to make a patch, I'm just wondering how @arel is getting this to work without a change to `kombu`, maybe I'm missing something?

@jdavisp3 the changes here concern the Redis result backend. For symmetry, Kombu should use the same format for broker URLs.

jdavisp3:
Ah, thanks for explaining that @georgepsarakis .

PR #5717, opened 2019-09-08
Fix serialization and deserialization of nested exception classes

nijel:
## Description

Fixes serialization and deserialization of nested exception types. The typical use case are Django DoesNotExist per model exceptions (see eg. https://github.com/celery/celery/pull/5435#issuecomment-479623006).

### DIFF
--- a/celery/backends/base.py
+++ b/celery/backends/base.py
@@ -257,9 +257,10 @@ class Backend(object):
         serializer = self.serializer if serializer is None else serializer
         if serializer in EXCEPTION_ABLE_CODECS:
             return get_pickleable_exception(exc)
-        return {'exc_type': type(exc).__name__,
+        exctype = type(exc)
+        return {'exc_type': getattr(exctype, '__qualname__', exctype.__name__),
                 'exc_message': ensure_serializable(exc.args, self.encode),
-                'exc_module': type(exc).__module__}
+                'exc_module': exctype.__module__}
 
     def exception_to_python(self, exc):
         """Convert serialized exception to Python exception."""
@@ -273,7 +274,11 @@ class Backend(object):
                     exc_module = from_utf8(exc_module)
                     exc_type = from_utf8(exc['exc_type'])
                     try:
-                        cls = getattr(sys.modules[exc_module], exc_type)
+                        # Load module and find exception class in that
+                        cls = sys.modules[exc_module]
+                        # The type can contain qualified name with parent classes
+                        for name in exc_type.split('.'):
+                            cls = getattr(cls, name)
                     except (KeyError, AttributeError):
                         cls = create_exception_cls(exc_type,
                                                    celery.exceptions.__name__)
--- a/t/unit/backends/test_base.py
+++ b/t/unit/backends/test_base.py
@@ -36,6 +36,11 @@ class paramexception(Exception):
         self.param = param
 
 
+class objectexception(object):
+    class Nested(Exception):
+        pass
+
+
 if sys.version_info[0] == 3 or getattr(sys, 'pypy_version_info', None):
     Oldstyle = None
 else:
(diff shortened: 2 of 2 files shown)


t/unit/backends/test_base.py
```diff
@@ -198,6 +203,21 @@ def test_json_exception_arguments(self):
         y = self.b.exception_to_python(x)
         assert isinstance(y, Exception)
 
+    def test_json_exception_nested(self):
+        if sys.version_info < (3, 3):
```
auvipy: next version support py2.7 and 3.5+

t/unit/backends/test_base.py
```diff
@@ -198,6 +203,21 @@ def test_json_exception_arguments(self):
...
+        self.b.serializer = 'json'
+        print(repr(objectexception.Nested('msg')))
+        x = self.b.prepare_exception(objectexception.Nested('msg'))
+        assert x == {
+            'exc_message': ('msg',),
+            'exc_type': 'objectexception.Nested',
+            'exc_module': objectexception.Nested.__module__}
+        print(x)
+        y = self.b.exception_to_python(x)
+        print(repr(y))
```
auvipy: are the print() functions intended here?

t/unit/backends/test_base.py
nijel: this is needed for 2.7, so I think this is still needed

t/unit/backends/test_base.py
nijel: no, I will remove them

t/unit/backends/test_base.py
auvipy: what about if sys.version_info <(3.5) and (2.7): ?

nijel: The needed [__qualname__ attribute](https://docs.python.org/3/library/stdtypes.html#definition.__qualname__) was introduced in 3.3, so it's not present in 2.7. I'm not sure what are you trying to address. The condition has to be there, the change can not work in Python older than 3.3, including 2.7. I think it's better to clearly specify version than hiding that into only 2.7 case.

auvipy: so it is a breaking change? Should we defer this to celery 5.x which will be python 3.6+ only?

nijel: On 2.7 it will behave exactly as before. On 3.3+ it serializes full class name (taken from `__qualname__`) and is able to deserialize such classes.

t/unit/backends/test_base.py
```diff
@@ -198,6 +203,18 @@ def test_json_exception_arguments(self):
         y = self.b.exception_to_python(x)
         assert isinstance(y, Exception)
 
+    def test_json_exception_nested(self):
+        if sys.version_info < (3, 3):
```
thedrow: Please use [`@pytest.mark.skipif`](http://doc.pytest.org/en/latest/skipping.html#id1) instead.

celery/backends/base.py
```diff
@@ -257,9 +257,10 @@ def prepare_exception(self, exc, serializer=None):
         serializer = self.serializer if serializer is None else serializer
         if serializer in EXCEPTION_ABLE_CODECS:
             return get_pickleable_exception(exc)
-        return {'exc_type': type(exc).__name__,
+        exctype = type(exc)
+        return {'exc_type': getattr(exctype, '__qualname__', exctype.__name__),
```
thedrow: Please consider using our `qualname()` helper if possible:
https://github.com/celery/celery/blob/4c6c609273dedc75ef4af9f4fa402e4de6a9344b/celery/utils/imports.py#L32-L46

celery/backends/base.py
```diff
@@ -273,7 +274,9 @@ def exception_to_python(self, exc):
                     exc_module = from_utf8(exc_module)
                     exc_type = from_utf8(exc['exc_type'])
                     try:
-                        cls = getattr(sys.modules[exc_module], exc_type)
+                        cls = sys.modules[exc_module]
```
thedrow: Should we leave a comment here? Maybe we can extract this to a helper to clarify the code's intentions instead?
What do you think?

nijel: I've added comment, not sure if helper is useful here as it seems quite specific case...

thedrow: Yes, that should do it.

nijel:
Okay, `qualname` is not good fit here as it does include module name. That lead to test failures

PR #5604, opened 2019-06-15
Additional file descriptor safety checks with refactor of my prior work (issue #5299)

matteius:
## Description
This intends to address problems reported here: #5299
I think these code changes reduce some repeated blocks and leverage the same pattern I added here:  https://github.com/celery/celery/pull/5499/ 

I noticed that the two definitions of on_poll_start had a lot of duplication and so additionally I sought to represent them as a single method that accounts for their one logical difference.

PR #5722, opened 2019-09-10
Don't log a debug statement every time the celery event loop is polled

wlach:
## Description
When I run celery locally using the rc candidate, I get a huge amount of debugging messages like this in my console:

```
worker_1  | [2019-09-10 18:00:57,146: DEBUG/MainProcess] AsyncPool._create_write_handlers ALL: 4 ACTIVE: 0
worker_1  | [2019-09-10 18:00:57,759: DEBUG/MainProcess] AsyncPool._create_write_handlers ALL: 4 ACTIVE: 0
worker_1  | [2019-09-10 18:00:58,768: DEBUG/MainProcess] AsyncPool._create_write_handlers ALL: 4 ACTIVE: 0
worker_1  | [2019-09-10 18:00:59,147: DEBUG/MainProcess] AsyncPool._create_write_handlers ALL: 4 ACTIVE: 0
worker_1  | [2019-09-10 18:00:59,149: DEBUG/MainProcess] AsyncPool._create_write_handlers ALL: 4 ACTIVE: 0
worker_1  | [2019-09-10 18:00:59,776: DEBUG/MainProcess] AsyncPool._create_write_handlers ALL: 4 ACTIVE: 0
```

Checking the source, this looks like a debugging relic from PR #5604. I don't see any reason why it would be useful to log information at this level of detail in production. Even in a debugging context, it's difficult to find other log messages with this in the console.

/cc @matteius who added this

### DIFF
--- a/celery/concurrency/asynpool.py
+++ b/celery/concurrency/asynpool.py
@@ -775,9 +775,6 @@ class AsynPool(_pool.Pool):
         def on_poll_start():
             # Determine which io descriptors are not busy
             inactive = diff(active_writes)
-            logger.debug(
-                "AsyncPool._create_write_handlers ALL: %r ACTIVE: %r",
-                len(all_inqueues), len(active_writes))
 
             # Determine hub_add vs hub_remove strategy conditional
             if is_fair_strategy:


jezdez:
Thanks @wlach! @auvipy, is this something that should just not use the debug logger (or reconfigure the debug logger to not use stdout automatically)?

matteius:
I personally think that we should leave the debug statement in the code because it can be useful for those who are working on the library to see that level of detail.  In general, we never deploy our python applications to production to output debug level information -- we generally only capture INFO level and above and that is configured by the Python logger settings you utilize in your project.

wlach:
> I personally think that we should leave the debug statement in the code because it can be useful for those who are working on the library to see that level of detail. In general, we never deploy our python applications to production to output debug level information -- we generally only capture INFO level and above and that is configured by the Python logger settings you utilize in your project.

I appreciate that, but the volume generated by this statement is *really* high -- I'm seeing 4 of these lines every *second* with my celery configuration. Sometimes you need/want to see other debug log statements when developing locally (or trying to debug something), but this makes that almost impossible.

matteius:
@wlach  Ok so I did add that log statement, I thought it was I had just adjusted the messaging to be debug -- the log message is actually wrong.  It says AsyncPool._create_write_handlers -- but its actually the start of Worker.on_poll_start -- I suppose we just remove it then based on the high volume of calling that method.

Issue #5711, opened 2019-09-03
Appveyor builds fail because of windows and a test I added a while back


matteius:
When we accepted one of my prior works to fix 
I added two tests `test_with_autoscaler_file_descriptor_safety` and `test_with_file_descriptor_safety`  here:  https://github.com/celery/celery/blame/master/t/unit/worker/test_worker.py#L797

These two tests have always caused the appveyor build to crash at that point because it runs on Windows, and I recall having worked on the underlying windows socket issue when I was submitting that prior PR but I didn't quite get it figured out.   I am reporting this to track that it still needs to get resolved, and to try and recall what I remember the issues were about it.

Apparently I lost whatever branches I had on that progress, but it had something to do with how windows initializes blocking/non-blocking pipes differently than unix pipes.    I will update this issue when I figure out more about what was happening again, but you can see an example here of: t\unit\worker\test_worker.py ..................................s......Command exited with code 1

https://ci.appveyor.com/project/ask/celery/builds/26904785/job/628wbsjwqefqtsbu

Most appveyor builds get far enough to get that failure I helped introduced, and so I want to help fix it without destroying those tests because they pass in linux.

matteius:
Ok running locally I can get the individual stack traces on Windows:

fetytest t/unit/worker/test_worker.py -k test_with_autoscaler_file_descriptor_saf
============================= test session starts =============================
platform win32 -- Python 3.7.1, pytest-4.6.5, py-1.8.0, pluggy-0.12.0
rootdir: C:\Users\matteius\projects\celery_matt, inifile: setup.cfg
plugins: case-1.5.3
collected 67 items / 66 deselected / 1 selected

t\unit\worker\test_worker.py FE                                          [100%]

=================================== ERRORS ====================================
_ ERROR at teardown of test_WorkController.test_with_autoscaler_file_descriptor_safety _

tp = <class 'AssertionError'>, value = None, tb = None

    def reraise(tp, value, tb=None):
        try:
            if value is None:
                value = tp()
            if value.__traceback__ is not tb:
                raise value.with_traceback(tb)
>           raise value

..\..\.virtualenvs\celery\lib\site-packages\six.py:693:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
..\..\.virtualenvs\celery\lib\site-packages\six.py:693: in reraise
    raise value
[...]

PR #5729, opened 2019-09-13
this is not a real pytest mark, lets start by removing that

matteius:
This test still gives me problems on windows, I wrote it, it passes on linux but I am not sure what happened recently with the builds because I cannot find any information on why I would have added a pytest.mark.nothreads_not_lingering when it seems nothreads_not_lingering is not a real  pytest mark and I see some builds failing because of this in Travis CI -- so we just start by removing the mark and see what the next failure up is.

See:  https://github.com/celery/celery/issues/5711


auvipy:
integration test failing with kombu 4.6.5

PR #5747, opened 2019-09-23
Add links to configuration docs from Django first steps

kjpc-tech:
This adds two links from the Django first steps page in the docs to the celery configuration reference for easier access/visibility.


auvipy:
thanks!

PR #5749, opened 2019-09-24
Test chords with large headers

thedrow:
In the spirit of #5742, these test cases verify that we are able to execute chords with large headers.


auvipy:
thanks!

PR #5758, opened 2019-10-01
Allow start_worker to function without ping task

nezhar:
## Description

The current version causes an error when using `start_worker` and the task `celery.ping` is not defined. The PR adds an option to disable this if the check is not required.

This is useful when writing integration tests for celery tasks and a celery worker needs to be started. In Django this can be achieved by adding `start_worker` in the TransactionTestCase setUpClass (https://docs.djangoproject.com/en/2.2/topics/testing/tools/#django.test.TransactionTestCase).

### DIFF
--- a/celery/contrib/testing/worker.py
+++ b/celery/contrib/testing/worker.py
@@ -75,6 +75,7 @@ def start_worker(app,
                               pool=pool,
                               loglevel=loglevel,
                               logfile=logfile,
+                              perform_ping_check=perform_ping_check,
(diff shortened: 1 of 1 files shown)


nezhar:
I see that some tests are already broken by this change. I will adapt this, add a separate test case for this and update the PR.

auvipy:
the broken tests are not directly related to this PR. but to get this accepted it would be great if you could provide unit + integration tests

PR #5771, opened 2019-10-10
Increases max results size for MySQL backend

### LINKED ISSUE #461: MySQL result backend defaults to BLOB which may be too short for some results (LONGBLOB)
I was running into errors caused by loading truncated pickled results. Specifically, I would get an EOFError (see below). I eventually figured out that the `result` column of the celery_taskmeta MySQL table was not able to hold all of the pickled result data. By altering the table to use LONGBLOB instead of BLOB, I was able to deal with the large results (`alter table celery_taskmeta modify result LONGBLOB`). Is there a way to make this the default type in the future?

```
[2011-08-31 13:34:13,605: ERROR/PoolWorker-1] 
  File "/home/ubuntu/w/apis/trunk/apis/celerytasks/tasks/file_task_base.py", line 293, in process
    self.process_records()
  File "/home/ubuntu/w/apis/trunk/apis/celerytasks/tasks/proxmatch_file.py", line 381, in process_records
    self.wait_results(async_result_buffer)
  File "/home/ubuntu/w/apis/trunk/apis/celerytasks/tasks/proxmatch_file.py", line 435, in wait_results
    async_result.wait()
  File "/usr/local/lib/python2.6/dist-packages/celery/result.py", line 87, in wait
    return self.get(*args, **kwargs)
  File "/usr/local/lib/python2.6/dist-packages/celery/result.py", line 83, in get
    interval=interval)
  File "/usr/local/lib/python2.6/dist-packages/celery/backends/base.py", line 110, in wait_for
    status = self.get_status(task_id)
  File "/usr/local/lib/python2.6/dist-packages/celery/backends/base.py", line 206, in get_status
    return self.get_task_meta(task_id)["status"]
  File "/usr/local/lib/python2.6/dist-packages/celery/backends/base.py", line 227, in get_task_meta
    meta = self._get_task_meta_for(task_id)
  File "/usr/local/lib/python2.6/dist-packages/celery/backends/database.py", line 61, in _get_task_meta_for
    task = session.query(Task).filter(Task.task_id == task_id).first()
  File "/usr/local/lib/python2.6/dist-packages/sqlalchemy/orm/query.py", line 1770, in first
    ret = list(self[0:1])
  File "/usr/local/lib/python2.6/dist-packages/sqlalchemy/orm/query.py", line 1667, in __getitem__
    return list(res)
  File "/usr/local/lib/python2.6/dist-packages/sqlalchemy/orm/query.py", line 1960, in instances
    rows = [process[0](row, None) for row in fetch]
  File "/usr/local/lib/python2.6/dist-packages/sqlalchemy/orm/mapper.py", line 2600, in _instance
    populate_state(state, dict_, row, isnew, only_load_props)
  File "/usr/local/lib/python2.6/dist-packages/sqlalchemy/orm/mapper.py", line 2454, in populate_state
    populator(state, dict_, row)
  File "/usr/local/lib/python2.6/dist-packages/sqlalchemy/orm/strategies.py", line 150, in new_execute
    dict_[key] = row[col]
  File "/usr/local/lib/python2.6/dist-packages/celery/db/a805d4bd.py", line 55, in process
    return loads(value)
EOFError
```

remeika:
Fixes #461 

The database result backend stores data in a of type [`PickleType`](https://docs.sqlalchemy.org/en/13/core/type_basics.html#sqlalchemy.types.PickleType) column. When materialized on a MySQL database, this creates a [`BLOB`](https://dev.mysql.com/doc/refman/5.7/en/blob.html), which has a max row size of 2^16 bytes == 65 kB. 

This PR increases the max result size to 4GB when stored in MySQL by changing the backing column type from `BLOB` to `LONGBLOB`. It takes advantage of the fact that PickleType is implemented as a SqlAlchemy [`TypeDecorator`](https://docs.sqlalchemy.org/en/13/core/custom_types.html#augmenting-existing-types); so we can just override how the type determine its base implementation given a database dialect, whist still maintaining all the well tested un/marshaling code in `PickleType`.

### DIFF
--- a/celery/backends/database/__init__.py
+++ b/celery/backends/database/__init__.py
@@ -14,8 +14,8 @@ from celery.exceptions import ImproperlyConfigured
 from celery.five import range
 from celery.utils.time import maybe_timedelta
 
-from .models import Task, TaskExtended, TaskSet
 from .session import SessionManager
+from .models import task_storage_factory
 
 try:
     from sqlalchemy.exc import DatabaseError, InvalidRequestError
@@ -68,9 +68,6 @@ class DatabaseBackend(BaseBackend):
     # to not bombard the database with queries.
     subpolling_interval = 0.5
 
-    task_cls = Task
-    taskset_cls = TaskSet
-
     def __init__(self, dburi=None, engine_options=None, url=None, **kwargs):
         # The `url` argument was added later and is used by
         # the app to set backend by url (celery.app.backends.by_url)
@@ -78,8 +75,10 @@ class DatabaseBackend(BaseBackend):
                                               url=url, **kwargs)
         conf = self.app.conf
 
-        if self.extended_result:
-            self.task_cls = TaskExtended
+        self.task_cls, self.taskset_cls = task_storage_factory(
+            conf.get('database_large_result_storage'),
+            self.extended_result
+        )
 
         self.url = url or dburi or conf.database_url
         self.engine_options = dict(
--- a/celery/backends/database/models.py
+++ b/celery/backends/database/models.py
@@ -5,7 +5,9 @@ from __future__ import absolute_import, unicode_literals
 from datetime import datetime
 
 import sqlalchemy as sa
+from sqlalchemy.dialects.mysql import LONGBLOB
 from sqlalchemy.types import PickleType
+from sqlalchemy.ext.declarative import declared_attr
 
 from celery import states
 from celery.five import python_2_unicode_compatible
@@ -14,9 +16,20 @@ from .session import ResultModelBase
 
 __all__ = ('Task', 'TaskExtended', 'TaskSet')
 
+class PickleXlType(PickleType):
+    """PickleType with increased storage capacity on MySQL
+
+    Uses MySQL's LONGBLOB column type, which can store up to 4GB per row.
+    """
+
+    def load_dialect_impl(self, dialect):
+        if dialect.name == 'mysql':
+            return dialect.type_descriptor(LONGBLOB())
+        return super(PickleXlType, self).load_dialect_impl(dialect)
+
 
 @python_2_unicode_compatible
-class Task(ResultModelBase):
+class TaskBase(ResultModelBase):
     """Task result/status."""
 
     __tablename__ = 'celery_taskmeta'
@@ -26,7 +39,6 @@ class Task(ResultModelBase):
                    primary_key=True, autoincrement=True)
     task_id = sa.Column(sa.String(155), unique=True)
     status = sa.Column(sa.String(50), default=states.PENDING)
-    result = sa.Column(PickleType, nullable=True)
     date_done = sa.Column(sa.DateTime, default=datetime.utcnow,
                           onupdate=datetime.utcnow, nullable=True)
     traceback = sa.Column(sa.Text, nullable=True)
@@ -47,7 +59,7 @@ class Task(ResultModelBase):
         return '<Task {0.task_id} state: {0.status}>'.format(self)
 
 
-class TaskExtended(Task):
+class TaskExtendedBase(TaskBase):
     """For the extend result."""
 
     __tablename__ = 'celery_taskmeta'
@@ -61,7 +73,7 @@ class TaskExtended(Task):
     queue = sa.Column(sa.String(155), nullable=True)
 
     def to_dict(self):
-        task_dict = super(TaskExtended, self).to_dict()
+        task_dict = super(TaskExtendedBase, self).to_dict()
         task_dict.update({
             'name': self.name,
             'args': self.args,
@@ -74,7 +86,7 @@ class TaskExtended(Task):
 
 
 @python_2_unicode_compatible
-class TaskSet(ResultModelBase):
+class TaskSetBase(ResultModelBase):
     """TaskSet result."""
 
     __tablename__ = 'celery_tasksetmeta'
@@ -83,7 +95,6 @@ class TaskSet(ResultModelBase):
     id = sa.Column(sa.Integer, sa.Sequence('taskset_id_sequence'),
                    autoincrement=True, primary_key=True)
     taskset_id = sa.Column(sa.String(155), unique=True)
-    result = sa.Column(PickleType, nullable=True)
     date_done = sa.Column(sa.DateTime, default=datetime.utcnow,
                           nullable=True)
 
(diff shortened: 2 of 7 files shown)


thedrow:
This is a breaking change.
I'm not sure everyone will be OK with this.
Unfortunately we lack migrations for the database backend.

thedrow (changes requested):
The solution to the backwards incompatibility problem would be to make this behavior configurable.
Using a `LONGBLOB` field should be disabled by default so that existing users won't be affected by this schema change.

remeika:
Hmm I might need a bit of help looking into the above test failures. They seem completely unrelated to my branch. Does the master branch have any failing check on it?

thedrow:
Yes these test failures are unrelated.
We're doing our best to stabilize the build.

auvipy:
James, is it ready for review?

remeika:
Sorry, I want to write a few more integration tests before publishing it. 

James

ericraymond:
Any chance this will make it into 4.4? 👍 

> auvipy added this to the 4.5 milestone 10 days ago

celery/backends/database/models.py
```diff
@@ -14,9 +15,29 @@
...
+class PickleXlType(PickleType):
+    """PickleType with increased storage capacity on MySQL
+
+    Uses MySQL's LONGBLOB column type, which can store up to 4GB per row.
+    """
+
+    def load_dialect_impl(self, dialect):
+        if dialect.name == 'mysql':
+            return dialect.type_descriptor(LONGBLOB())
+        else:
```
reutsharabani: no need for `else` here

requirements/test-integration.txt
```diff
@@ -1,7 +1,8 @@
+pytest-rerunfailures>=6.0
 simplejson
--r extras/redis.txt
--r extras/dynamodb.txt
--r extras/azureblockblob.txt
 -r extras/auth.txt
+-r extras/azureblockblob.txt
+-r extras/database-integration.txt
```
reutsharabani: I think this like is the only change needed here.

docker/docker-compose.yml
```diff
@@ -37,3 +39,10 @@ services:
 
   azurite:
     image: arafato/azurite:2.6.5
+
+  mysql:
+    image: mysql
+    environment:
+        - MYSQL_DATABASES=resultsdb
```
reutsharabani: indent by two spaces

celery/backends/database/__init__.py
```diff
@@ -95,6 +95,10 @@ def __init__(self, dburi=None, engine_options=None, url=None, **kwargs):
         self.taskset_cls.__table__.name = tablenames.get('group',
                                                          'celery_tasksetmeta')
 
+        if conf.get('database_large_result_storage', False):
```
reutsharabani: no need for explicit `False` here.

remeika: Yup! Fixed.

docker/docker-compose.yml
remeika: Done!

requirements/test-integration.txt
remeika: The azure block blob integration tests are failing without this line, but you're right it has no place in this PR.

celery/backends/database/models.py
remeika: Done.

remeika:
Yikes! This was a gnarly issue. My one remaining issue is that I can't figure out how to clean up SQLAlchemy sessions between tests. In `t/integration/test_backend.py::test_DatabaseMySQLBackend`, both tests pass if they are the only one that runs. I'm going to keep working on it, but any helpful hints would be appreciated!

t/integration/test_backend.py
```diff
@@ -35,3 +41,44 @@ def test_get_missing(self, manager):
...
+    def mysql_url(self):
+        yield os.environ["MYSQL_URL"]
+    
+    def clear_db(self, url):
+        engine = create_engine(url)
+        engine.execute('DROP TABLE IF EXISTS celery_taskmeta;')
+        engine.execute('DROP TABLE IF EXISTS celery_tasksetmeta;')
+        engine.dispose()
+
+    # def test_store_result(self, mysql_url, app):
```
matteius: Why is this one commented out?

matteius (commented):
This looks great!  Testing middleware for databases (such as sqlite, etc. ) can always be a major pain since this fix is for a MySQL data type it should not fail integration on anything other than a MySQL DB backend, but then there is the reality of where tests get added and how the tests get run.

celery/backends/database/__init__.py
```diff
@@ -68,18 +68,17 @@ class DatabaseBackend(BaseBackend):
     # to not bombard the database with queries.
     subpolling_interval = 0.5
 
-    task_cls = Task
```
drsantos20: what this `cls` means? is reserved name?

t/integration/test_backend.py
```diff
@@ -35,3 +41,44 @@ def test_get_missing(self, manager):
...
+    #     self.clear_db(mysql_url)
+    #     tb = DatabaseBackend(mysql_url, app=app)
+    #     tb.store_result(uuid1(), {'fizz': 'buzz'}, states.SUCCESS)
+
+    def test_store_extended_result(self, mysql_url, extended_app):
+
+        self.clear_db(mysql_url)
```
drsantos20: should we `print` this vars?

drsantos20 (commented):
i left some comments, if im wrong please just ignore.

the reserved name i just left a message bc is more clear to have the verbose name instead the abreviation names from variables

ericraymond:
Is this PR still active?

auvipy:
you can retake this

auvipy:
just need to rebase to merge

thedrow:
> 
> 
> Yikes! This was a gnarly issue. My one remaining issue is that I can't figure out how to clean up SQLAlchemy sessions between tests. In `t/integration/test_backend.py::test_DatabaseMySQLBackend`, both tests pass if they are the only one that runs. I'm going to keep working on it, but any helpful hints would be appreciated!

@auvipy Was this ever resolved?

celery/backends/database/__init__.py
thedrow: It's just a class variable used to manufacture tasks.

celery/backends/database/__init__.py
```diff
@@ -68,18 +68,17 @@ class DatabaseBackend(BaseBackend):
...
         # The `url` argument was added later and is used by
         # the app to set backend by url (celery.app.backends.by_url)
         super(DatabaseBackend, self).__init__(expires_type=maybe_timedelta,
                                               url=url, **kwargs)
         conf = self.app.conf
 
-        if self.extended_result:
-            self.task_cls = TaskExtended
+        self.task_cls, self.taskset_cls = task_storage_factory(
+            conf.get('database_large_result_storage'),
```
thedrow, line 79: This setting requires documentation

celery/backends/database/models.py
```diff
@@ -100,3 +111,25 @@ def to_dict(self):
 
     def __repr__(self):
         return '<TaskSet: {0.taskset_id}>'.format(self)
+
+
+def task_storage_factory(large_results, extended_results):
```
thedrow, line 116: This function requires documentation.

t/integration/test_backend.py
thedrow: We can remove that line.

thedrow (changes requested):
I'm sorry it took so long for me to review this.
Overall this is a good PR.

It requires some finishing touches.

thedrow:
I'm scheduling this for 5.1 since it is almost done. We need to rebase and finish documenting this.

auvipy:
I think the approach followed in this https://github.com/celery/celery/pull/5907 is better then the approach of this PR. even if we increase results size of specific DB backend, we should make the field configurable first

PR #5801, opened 2019-10-29
update  docker config

auvipy:


auvipy:
some download links also needed to be updated before merge

PR #5805, opened 2019-10-30
Add auto expiry for DynamoDB backend

sveniu:
## Description

This adds auto-expire support for the DynamoDB backend, via the DynamoDB Time to Live feature.

Highlighting OG DynamoDB backend author @georgepsarakis in case he wants to take a look as well.


celery/backends/dynamodb.py
```diff
@@ -194,6 +222,17 @@ def _get_or_create_table(self):
                     self.table_name
                 )
             )
+            # Enable time-to-live on the table, ignoring whether or not TTL is
+            # currently specified as an option for the backend. This allows
+            # enabling it later. Until then, items are inserted without a value
+            # in the ttl field, meaning that DynamoDB will never expire them.
+            self._client.update_time_to_live(**self._get_ttl_specification())
```
georgepsarakis: TTL configuration should be applied conditionally, if `time_to_live_seconds` is not `None`.

georgepsarakis: According to boto3 [documentation](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/dynamodb.html?highlight=dynamodb#DynamoDB.Client.update_time_to_live):

> It can take up to one hour for the change to fully process. Any additional UpdateTimeToLive calls for the same table during this one hour duration result in a ValidationException .

So perhaps we need to catch the `ValidationException` and log it?

celery/backends/dynamodb.py
```diff
@@ -51,9 +51,16 @@ class DynamoDBBackend(KeyValueStoreBackend):
...
+    #: Item time-to-live in seconds (`default`)
+    time_to_live_seconds = None
+
+    # DynamoDB supports Time to Live as an auto-expiry mechanism.
+    supports_autoexpire = True
+
     _key_field = DynamoDBAttribute(name='id', data_type='S')
     _value_field = DynamoDBAttribute(name='result', data_type='B')
     _timestamp_field = DynamoDBAttribute(name='timestamp', data_type='N')
+    _ttl_field = DynamoDBAttribute(name='ttl', data_type='N')
```
georgepsarakis: Correct, this should be of type `Number`.

celery/backends/dynamodb.py
sveniu: This has been addressed, but perhaps not sufficiently:

The TTL handling is now improved so that trying to enable or disable it when it is already enabled/enabling or disabled/disabling, is successful.

Other cases are more challenging:
- TTL is already enabled, but on the wrong attribute: This requires disabling it and then re-enabling it, but the operation throttling prevents this from being done quickly.
- TTL is being enabled or disabled and we want to instead disable/enable it: The operation throttling prevents this from being done in quick succession.

In the challenging cases, the exception is simply raised. The exception itself has an informative `['Error']['Message']` field, but it will perhaps not be logged appropriately to the user?

Should I rather catch these and extract the error message, and log it explicitly with level 'error' or something?

sveniu: Addressed.

georgepsarakis: I think logging the error message and reraising the exception would be preferrable, what do you think?

georgepsarakis:
@sveniu will this work with existing installations? I am quite sure it will, but please let's consider backwards compatibility.

celery/backends/dynamodb.py
sveniu: Agreed and addressed. Let me know if it should be done differently.

sveniu:
> @sveniu will this work with existing installations? I am quite sure it will, but please let's consider backwards compatibility.

I've added some code to gracefully handle older boto3 versions not supporting the TTL methods. This is in addition to the earlier commit that bumps the requirements file to require a boto3 version that actually has TTL support.

As for server-side compatibility: There may be users who use older versions of the downloadable local DynamoDB without support for TTL. I think all AWS regions have TTL support at this moment. In any case, there would be a "normal" exception from boto3 and it would be logged in the exception handler.

celery/backends/dynamodb.py
```diff
@@ -118,6 +125,17 @@ def __init__(self, url=None, table_name=None, *args, **kwargs):
                     self.write_capacity_units
                 )
             )
+
+            try:
```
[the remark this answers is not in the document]
sveniu: The proposed alternative doesn't seem to handle casting to integer? If a user mistakenly tries do use `ttl_seconds=1d`, the alternative will succeed, but the code using `self.time_to_live_seconds` later will fail (when it relies on numeric addition to calculate the absolute expiry time).

Maybe I missed the point here. Let me know.

celery/backends/dynamodb.py
```diff
@@ -206,6 +225,86 @@ def _get_or_create_table(self):
...
+            cur_attr_name if status == 'ENABLED' else self._ttl_field.name
+        specification = self._client.update_time_to_live(
+            **self._get_ttl_specification(
+                ttl_attr_name=attr_name
+            )
+        )
+        logger.info(
+            'DynamoDB Table {} time-to-live enabled={} on field {}.'.format(
+                self.table_name,
+                False if self.time_to_live_seconds is None else True,
```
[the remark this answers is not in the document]
sveniu: Good idea. Addressed.

celery/backends/dynamodb.py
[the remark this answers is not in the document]
georgepsarakis: Of course you are right! My snippet was not complete after all, so the new version (and it does need exception handling):

```python
ttl = query.get('ttl_seconds', self.time_to_live_seconds)
if ttl:
    try:
        self.time_to_live_seconds = int(ttl)
    except ValueError as e:
        logger.error('TTL must be a number', exc_info=e)
        raise e
```

celery/backends/dynamodb.py
```diff
@@ -206,6 +225,156 @@ def _get_or_create_table(self):
...
+            error_message = e.response['Error'].get('Message', 'Unknown')
+            logger.error((
+                'Error describing Time to Live on DynamoDB table {table}: '
+                '{code}: {message}'
+            ).format(
+                table=self.table_name,
+                code=error_code,
+                message=error_message,
+            ))
+            raise e
```
georgepsarakis: This method has a considerable size, so perhaps you could extract these lines to a different method (no need to modify tests):

```python     
def _get_table_ttl_description(self):
    ...
    return description
```

And its use:

```python
description = self._get_table_ttl_description()
status = description['TimeToLiveDescription']['TimeToLiveStatus']
```

How does that sound?

auvipy:
this should be hold for 4.7

celery/backends/dynamodb.py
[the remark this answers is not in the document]
sveniu: I think it's a good idea. Addressed.

celery/backends/dynamodb.py
sveniu: Addressed.

sveniu:
@georgepsarakis With this nearing completion, how does merging work - should I squash (and possibly rewrite the commit message properly) and force-push a single commit, or does squashing happen from your side? I'd like to flesh out the commit message a bit, but no big deal if it's left as it was written in the first commit.

celery/backends/dynamodb.py
```diff
@@ -206,6 +226,161 @@ def _get_or_create_table(self):
...
+        """Enable or disable Time to Live on the table."""
+
+        # Verify client support for the DynamoDB TTL methods.
+        for method in ('update_time_to_live', 'describe_time_to_live'):
+            if not hasattr(self._client, method):
+                message = (
+                    "boto3 method '{method}' not found; ensure that "
+                    "boto3>=1.9.178 and botocore>=1.12.178 are installed"
+                ).format(method=method)
+                if not self._has_ttl():
```
reutsharabani: Can't this part stand on its own as a guard? Even at `debug` level, the log can be a distraction when its not really used.

celery/backends/dynamodb.py
```diff
@@ -206,6 +226,161 @@ def _get_or_create_table(self):
...
+                        'on table {table}'
+                    ).format(
+                        situation='already enabled' \
+                            if status == 'ENABLED' \
+                            else 'currently being enabled',
+                        table=self.table_name
+                    ))
+                    return description
+
+        elif status in ('DISABLED', 'DISABLING'):
```
reutsharabani: elif -> if ?

celery/backends/dynamodb.py
```diff
@@ -206,6 +226,161 @@ def _get_or_create_table(self):
...
+                description['TimeToLiveDescription']['AttributeName']
+            if self._has_ttl():
+                if cur_attr_name == self._ttl_field.name:
+                    # We want TTL enabled, and it is currently enabled or being
+                    # enabled, and on the correct attribute.
+                    logger.debug((
+                        'DynamoDB Time to Live is {situation} '
+                        'on table {table}'
+                    ).format(
+                        situation='already enabled' \
```
reutsharabani: usually mappings from enums to text is nicer in dictionaries than in syntax.

Maybe you can use something like:

    {(ttl, status): message}

To determine the text to be printed?

It may be more clumsy, only suggesting.

celery/backends/dynamodb.py
reutsharabani: +  I would move the boto validation code elsewhere

And then you could do:

    self.validate_boto_version()

It would make the code more direct. This method verification is a necessary nuisance which isn't at the heart of this operation and has enough code to justify its own method.

(it would also be nice to see all missing methods and not just the first, but that is also not a big deal).

thedrow:
> @georgepsarakis With this nearing completion, how does merging work - should I squash (and possibly rewrite the commit message properly) and force-push a single commit, or does squashing happen from your side? I'd like to flesh out the commit message a bit, but no big deal if it's left as it was written in the first commit.

We squash when we merge.

thedrow:
> this should be on hold for 4.5

We don't have many features in that release as far as I can see.
What's the risk of including this feature?

auvipy:
we can bring it in 4.4.1 then but after RC releases, its norm not to include new features as other big projects do.

celery/backends/dynamodb.py
sveniu: Addressed.

celery/backends/dynamodb.py
sveniu: I'm not sure about this one. I've left it as an if, elif, else now (added the else just now), which I think makes sense. Let me know if I should re-do it, though.

reutsharabani: I'm no authority but `elif` usually makes me look at the previous condition. It satisfies the negation of previous condition *and* the following condition. I don't think its needed here (or anywhere where you have short-circuiting `if ... return/raise)`)

IMO `if` is less interleaved than `elif` and easier to read.

But its fine either way, just wanted you to consider it.

thedrow:
We can't introduce new features in a patch version.
Patch versions are solely for bugs.

I'm asking again, what's the risk of including this feature in 4.4.0?

auvipy:
not much risk. just convention. but if all are OK I hve no problem to get this in or 4.4

georgepsarakis:
@thedrow thanks for merging!

PR #5807, opened 2019-10-31
events bootstep disabled if no events

anilaratna2:
## Description

When events are disabled, the Events bootstep should not be enabled and hence not start. Currently, because of this bug, an unnecessary connection to the broker is created.


anilaratna2:
Was able to test that with CELERY_SEND_EVENTS=False, Events boot step doesnt start.

Issue #5763, opened 2019-10-07
Dockerfile - Build failed with kombu<5.0,>=4.6.4 installation 


naggappan:
I am trying to build the celery docker container as mentioned in the docs to setup the dev environment. But docker build fails with below error,

**Command:**  docker-compose build celery
**docker_Version:** Docker version 19.03.2, build 6a30dfc
**docker_Compse_version:** docker-compose version 1.18.0, build 8dd22a9

**Error:**
```  
    Running setup.py install for unipath: started
    Running setup.py install for unipath: finished with status 'done'
[...]

PR #5809, opened 2019-10-31
Fixed Dockerfile

dipAch:

## Description

Changes to Dockerfile for enabling noninteractive image build.

1. Set the build argument to: `ARG DEBIAN_FRONTEND=noninteractive`
With this set, we are able to disable the prompt for `tzdata`, during the build. With this, the timezone will be by default set to `UTC` however, it can be changed/reconfigured later on to reflect appropriate timezone.

2. Add `sudo` to the `apt-get install ...` command, as `sudo` is not included by default. I believe we are using `sudo` in the script `install-couchase.sh`.
Ref - https://github.com/tianon/docker-brew-ubuntu-core/issues/48

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -2,6 +2,8 @@ FROM ubuntu:bionic
 
 ENV PYTHONIOENCODING UTF-8
 
+ARG DEBIAN_FRONTEND=noninteractive
+
 # Pypy is installed from a package manager because it takes so long to build.
 RUN apt-get update && apt-get install -y build-essential \
     libcurl4-openssl-dev \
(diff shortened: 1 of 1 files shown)


dipAch:
#5763

docker/Dockerfile
```diff
@@ -2,6 +2,8 @@ FROM ubuntu:bionic
 
 ENV PYTHONIOENCODING UTF-8
 
+ARG DEBIAN_FRONTEND=noninteractive
```
auvipy: Debian frontend?

dipAch: It refers to the interface that is presented during the package installation.
Please see this for more info: https://askubuntu.com/a/972528

dipAch: In our case, it was prompting for user input and waiting on it during the installation. So the frontend used was `readline`. Needed to change that to above, to make it silent.

dipAch: Also, using the `ARG` command so that the environment variable is not persisted post image build.

PR #5843, opened 2019-11-22
SQS - Reject on failure

galCohen88:
## Description
https://github.com/celery/kombu/issues/1127

Config:
broker acks_late=True 
acks_on_failure_or_timeout=False

Using SQS  worker stop consuming messages after task failure (exception), as the message will not be acked out of the local queue

```
from celery import Celery
app.conf.worker_prefetch_multiplier = 1

@app.task(base=BaseAsyncJobTask, acks_late=True, acks_on_failure_or_timeout=False)
def dead_letter_q_task():
    return 1/0

dead_letter_q_task.delay()
dead_letter_q_task.delay()
dead_letter_q_task.delay()
```

# Expected Behavior
I expect all the tasks to be executed, although the first task failed

# Actual Behavior

The first task was executed, but the rest weren't.

The problem originates from the prefetch multiplier feature, with acks_late=True and acks_on_failure_or_timeout=False flags.

The problem is that the prefetched task is remaining in the local queue although it was failed (because it will not be acked on failure). This causes the mechanism not fetching any more tasks, and workers are not executing anything.

### DIFF
--- a/celery/worker/request.py
+++ b/celery/worker/request.py
@@ -510,6 +510,9 @@ class Request(object):
                 send_failed_event = False
             elif ack:
                 self.acknowledge()
+            else:
+                logger.info('!!rejecting the task')
+                self.reject(requeue=False)
 
         # These are special cases where the process would not have had time
         # to write the result.
--- a/celery/worker/request.py
+++ b/celery/worker/request.py
@@ -511,7 +511,8 @@ class Request(object):
             elif ack:
                 self.acknowledge()
             else:
-                logger.info('!!rejecting the task')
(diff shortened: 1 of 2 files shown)


galCohen88:
Integration test with what component? I could not find any integration tests for Request class

auvipy:
thanks

auvipy:
not 100% sure the level worker crashes is right for this

galCohen88:
its not that the worker crashed, its the [can_consume](https://github.com/celery/kombu/blob/master/kombu/transport/virtual/base.py#L191) method would always return false, as the task is not acked locally

PR #5850, opened 2019-11-28
Fix a race condition when publishing a very large chord header

thedrow:

## Description

Sometimes when publishing a chord with a large header a race condition occurs
when the group result isn't saved before one of the tasks of the header completes.

This regression was introduced in #4443 when the saving of the group result was moved after
publishing the chord's header.

Even after rearranging the code so that `header_result.save()` would come first
there was a problem since the code that freezes the group didn't use the same task_id
as the one that the group was published with.
This is now also fixed.

### DIFF
--- a/celery/canvas.py
+++ b/celery/canvas.py
@@ -1190,7 +1190,7 @@ class group(Signature):
         try:
             gid = opts['task_id']
         except KeyError:
-            gid = opts['task_id'] = uuid()
+            gid = opts['task_id'] = group_id or uuid()
         if group_id:
             opts['group_id'] = group_id
         if chord:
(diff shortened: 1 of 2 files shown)


thedrow:
@AlexHill Any change that you'll take a look here since you're the one who introduced this regression?
@auvipy Waiting for your LGTM...

auvipy:
sadly all the integration tests are failing! can you check that?

thedrow:
> sadly all the integration tests are failing! can you check that?

Sure.

auvipy:
if you can fix it I can release another RC today or tomorrow and final version of 4.4 on November 7.

PR #5856, opened 2019-12-01
remove cache back end integrtion test.

auvipy:


thedrow:
Why did you do that?
What's the motivation?
We can xfail what we can't fix but without this, we won't catch new bugs on the cache backend.

PR #5482, opened 2019-04-24
Fixed a possible memory leak in worker

auvipy:

PR #5870, opened 2019-12-12
Fixed event capture from building infinite list

### LINKED ISSUE #4843: Continuous memory leak
There is a memory leak in the parent process of Celery's worker.
It is not a child process executing a task.
It happens suddenly every few days.
Unless you stop Celery, it consumes server memory in tens of hours.

This problem happens at least in Celery 4.1, and it also occurs in Celery 4.2.
Celery is running on Ubuntu 16 and brokers use RabbitMQ.

memory

Prodge:
`consume` is a generator returning an infinite list of `None`s.

Wrapping this in a `list` will slowly build this list, holding it in memory.

This is a very slow leak and is only visible over some time when there is a large volume of events.

Supersedes #5482. 

Fixes #4843.

### DIFF
--- a/celery/events/receiver.py
+++ b/celery/events/receiver.py
@@ -90,7 +90,8 @@ class EventReceiver(ConsumerMixin):
         unless :attr:`EventDispatcher.should_stop` is set to True, or
         forced via :exc:`KeyboardInterrupt` or :exc:`SystemExit`.
         """
-        return list(self.consume(limit=limit, timeout=timeout, wakeup=wakeup))
+        for _ in self.consume(limit=limit, timeout=timeout, wakeup=wakeup):
+            pass
 
     def wakeup_workers(self, channel=None):
         self.app.control.broadcast('heartbeat',


thedrow:
@auvipy What kind of test do you have in mind?

auvipy:
this can be addressed in point release.

drsantos20:
> @auvipy What kind of test do you have in mind?

Do you know if we already have a test for it? if yes you can implement covering what you added

celery/events/receiver.py
```diff
@@ -90,7 +90,8 @@ def capture(self, limit=None, timeout=None, wakeup=True):
         unless :attr:`EventDispatcher.should_stop` is set to True, or
         forced via :exc:`KeyboardInterrupt` or :exc:`SystemExit`.
         """
-        return list(self.consume(limit=limit, timeout=timeout, wakeup=wakeup))
+        for _ in self.consume(limit=limit, timeout=timeout, wakeup=wakeup):
```
georgepsarakis: Although, the fix seems correct, I could not find a connection with worker functionality. Event capturing is performed by specific commands only. Could you please explain how this method affects the worker? Thanks!

spidernello:
@Prodge a feedback on this one is highly appreciated!

auvipy:
> @Prodge a feedback on this one would be highly appreciated! we still experience memory leak issues and hoping a permanent fix is being released, thanks

can you try this patch?

spidernello:
I'm evaluating the possibility to patch staging as well as upgrading to celery 4.4.0 and perform some tests, I'll update you as soon as I get to that

spidernello:
I'm about to test the patch in our version of celery 4.2.1 in staging. I'll give some updates as soon as I have some feedback

spidernello:
I can confirm the proposed patch fixed the memory leak, we run some tests over the weekend and did not see any memory leak occurring

iloveminbong:
@auvipy Do you know when we can expect to see this in an official release?

spidernello:
I also wanted to share with you that once I applied this flag --without-gossip --without-mingle --without-heartbeat in the celery multi string I have seen a general improvement in the performances celery <--> rabbitmq - on the rabbitmq side we often were seeing disconnection and very high cpu utilisation. After applying this fix all these behaviours disappeared Hope this helps

auvipy:
> @auvipy Do you know when we can expect to see this in an official release?

ASAP! It may be the last day of Feb!

Issue #5872, opened 2019-12-13
Outdated optimization note about usage of librabbitmq


Alireza2n:
# Checklist

- [x] I have checked the [issues list](https://github.com/celery/celery/issues?utf8=%E2%9C%93&q=is%3Aissue+label%3A%22Category%3A+Documentation%22+)
  for similar or identical bug reports.
- [x] I have checked the [pull requests list](https://github.com/celery/celery/pulls?q=is%3Apr+label%3A%22Category%3A+Documentation%22)
  for existing proposed fixes.
- [x] I have checked the [commit log](https://github.com/celery/celery/commits/master)
  to find out if the bug was already fixed in the master branch.
- [x] I have included all related issues and possible duplicate issues in this issue
      (If there are none, check this box anyway).

#### Related Issues

#4693 

#### Possible Duplicates

- None

# Description
In current documentation, in the "Optimizing" section, [this paragraph](https://docs.celeryproject.org/en/latest/userguide/optimizing.html#librabbitmq) recommends using librabbitmq but in #4693 it's mentioned librabbitmq is not the best performing package anymore.

# Suggestions
I'd suggest to mark [this paragraph](https://docs.celeryproject.org/en/latest/userguide/optimizing.html#librabbitmq)  as "old" or remove it to prevent users from using it as a "best practice".

auvipy:
contribution is welcome

Alireza2n:
@auvipy Sure. I just wanted to make sure this still holds true.
Will make a PR as soon as possbile.

Issue #4693, opened 2018-05-01
Librabbitmq 2.0.0 with Python 3 gives TypeError: can't pickle memoryview objects


somghosh:
I am using the latest master branch of the git repo https://github.com/celery/librabbitmq and installing `librabbitmq==2.0.0` for Python 3.6 by following the instructions in the readme

## Using the development version

You can clone the repository by doing the following:

	$ git clone git://github.com/celery/librabbitmq.git

Then install it by doing the following:

	$ cd librabbitmq
	$ make install        # or make develop

This works fine (after installing certain binaries for c compliation in the OS), but when I then make a small `a+b` add task and call it with `add.delay(2,2)` it fails with the following error. I looked up and saw that Celery 4 uses json as serializer, so clearly it is not because if pickle serialization

1. Changing from librabbitmq to pyamqp broker works normally
2. Same exact situation in both MacOS and Ubuntu 16

> 	[2018-04-30 23:40:02,956: CRITICAL/MainProcess] Unrecoverable error:
> SystemError('<built-in method _basic_recv of Connection object at
> 0x1099ec5e8> returned a result with an error set',) 	Traceback (most
> recent call last): 	  File
> "/Users/somghosh/.virtualenvs/ctdb/lib/python3.6/site-packages/kombu/messaging.py",
> line 624, in _receive_callback
> 	    return on_m(message) if on_m else self.receive(decoded, message) 	  File
> "/Users/somghosh/.virtualenvs/ctdb/lib/python3.6/site-packages/celery/worker/consumer/consumer.py",
> line 570, in on_task_received
> 	    callbacks, 	  File "/Users/somghosh/.virtualenvs/ctdb/lib/python3.6/site-packages/celery/worker/strategy.py",
> line 145, in task_message_handler
> 	    handle(req) 	  File "/Users/somghosh/.virtualenvs/ctdb/lib/python3.6/site-packages/celery/worker/worker.py",
> line 221, in _process_task_sem
> 	    return self._quick_acquire(self._process_task, req) 	  File "/Users/somghosh/.virtualenvs/ctdb/lib/python3.6/site-packages/kombu/async/semaphore.py",
> line 62, in acquire
> 	    callback(*partial_args, **partial_kwargs) 	  File "/Users/somghosh/.virtualenvs/ctdb/lib/python3.6/site-packages/celery/worker/worker.py",
> line 226, in _process_task
[...]

PR #5873, opened 2019-12-13
Removed the paragraph about using librabbitmq as optimization practice. Refer to #5872

Alireza2n:
## Description

Removes an outdated optimization note (Issue #5872) about using `librabbitmq` since #4693 mentions this is not the case anymore.

### DIFF
--- a/docs/userguide/optimizing.rst
+++ b/docs/userguide/optimizing.rst
@@ -48,22 +48,6 @@ like adding new worker nodes, or revoking unnecessary tasks.
 General Settings
 ================
 
-.. _optimizing-librabbitmq:
-
-librabbitmq
------------
-
-If you're using RabbitMQ (AMQP) as the broker then you can install the
-:pypi:`librabbitmq` module to use an optimized client written in C:
-
-.. code-block:: console
-
-    $ pip install librabbitmq
-
-The 'amqp' transport will automatically use the librabbitmq module if it's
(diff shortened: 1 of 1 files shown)


docs/userguide/optimizing.rst
```diff
@@ -48,22 +48,6 @@ like adding new worker nodes, or revoking unnecessary tasks.
...
------------
-
-If you're using RabbitMQ (AMQP) as the broker then you can install the
-:pypi:`librabbitmq` module to use an optimized client written in C:
-
-.. code-block:: console
-
-    $ pip install librabbitmq
-
-The 'amqp' transport will automatically use the librabbitmq module if it's
```
clokep: This part still seems to be true, see kombu's code: https://github.com/celery/kombu/blob/81101877a09e8f4442555c6c74b4836f113e4f51/kombu/connection.py#L251-L252

#4693 says that librabbitmq is no longer the recommended package, but the code in kombu hasn't be updated to stop automatically converting the `amqp://` prefix to `librabbitmq://` if it is installed.

auvipy: we need to update kombu as well

Alireza2n: I've made a [PR on kombu](https://github.com/celery/kombu/pull/1225) to address this.

PR #5874, opened 2019-12-16
Update docs regarding Redis Message Priorities

shenek:
## Description

The original documentation about "Redis Message Priorities" is misleading.
I would expect that some kind of priority based scheduling is performed by default.

But that's not true because the `round_robin` queue cycle strategy is used by default. So no matter which priorities your tasks have, they are scheduled in the same way.

see https://github.com/celery/kombu/blob/master/kombu/transport/redis.py#L447

So this PR adds a text which mentions that when you want to use priority based scheduling,
you need to also to change the queue order strategy.

### DIFF
--- a/docs/userguide/routing.rst
+++ b/docs/userguide/routing.rst
@@ -262,7 +262,7 @@ While the Celery Redis transport does honor the priority field, Redis itself has
 no notion of priorities. Please read this note before attempting to implement
 priorities with Redis as you may experience some unexpected behavior.
 
-To start scheduling tasks based on priorities you need configure queue_order_strategy transport option.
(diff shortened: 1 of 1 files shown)


docs/userguide/routing.rst
```diff
@@ -262,6 +262,14 @@ While the Celery Redis transport does honor the priority field, Redis itself has
 no notion of priorities. Please read this note before attempting to implement
 priorities with Redis as you may experience some unexpected behavior.
 
+To start scheduling tasks based on priorities you need configure queue_order_strategy transport option.
```
auvipy: need to

shenek: Fixed. Thx.

PR #5875, opened 2019-12-16
Update 4.4.0 docs

AndrewPix:
## Description
Added missing note about "Task class definitions can have retry attributes (#5869)" commit to the changelog

Added info about this feature to tasks docs `docs/userguide/tasks.rst`

Removed note about Django 1.11 requirement since it was introduced in 4.3

PR #5879, opened 2019-12-17
Minor fix for recent tasks doc file update

AndrewPix:
I found a minor bug in my latest PR (https://github.com/celery/celery/pull/5875) that updates some docs. So this PR is just fixing an order of one doc paragraph.


AndrewPix:
@auvipy This is a pretty quick fix, could you review it

PR #5887, opened 2019-12-22
Have appveyor build relevant versions of Python.

matteius:
## Description
Recent PR's fail their AppVeyor builds.   Python 2 is failing with a warning that AppVeyor will remove support after January 1st, and python 3.4 build is failing on our own setup.py requirement that " Celery 4.0 requires CPython 3.5 or later".

> DEPRECATION: Python 2.7 will reach the end of its life on January 1st, 2020. Please upgrade your Python as Python 2.7 won't be maintained after that date. A future version of pip will drop support for Python 2.7. More details about Python 2 support in pip, can be found at https://pip.pypa.io/en/latest/development/release-process/#python-2-support

```
  File "setup.py", line 239, in <module>
    'celery = celery.contrib.pytest',
  File "C:\Python27\lib\site-packages\setuptools\__init__.py", line 145, in setup
    return distutils.core.setup(**attrs)
  File "C:\Python27\lib\distutils\core.py", line 151, in setup
    dist.run_commands()
  File "C:\Python27\lib\distutils\dist.py", line 953, in run_commands
    self.run_command(cmd)
  File "C:\Python27\lib\distutils\dist.py", line 972, in run_command
    cmd_obj.run()
  File "C:\Python27\lib\site-packages\setuptools\command\test.py", line 225, in run
    installed_dists = self.install_dists(self.distribution)
  File "C:\Python27\lib\site-packages\setuptools\command\test.py", line 209, in install_dists
    tr_d = dist.fetch_build_eggs(dist.tests_require or [])
  File "C:\Python27\lib\site-packages\setuptools\dist.py", line 721, in fetch_build_eggs
    replace_conflicting=True,
  File "C:\Python27\lib\site-packages\pkg_resources\__init__.py", line 791, in resolve
    raise VersionConflict(dist, req).with_context(dependent_req)
pkg_resources.ContextualVersionConflict: (python-dateutil 2.8.1 (c:\projects\celery\.eggs\python_dateutil-2.8.1-py2.7.egg), Requirement.parse('python-dateutil<2.8.1,>=2.1'), set(['botocore']))
```

See also for python 3.4:  https://ci.appveyor.com/project/ask/celery/builds/29484693/job/dsnjeiykm2u4262d
```
Executing: C:\Python34/python setup.py test
  File "setup.py", line 50, in <module>
    raise Exception(E_UNSUPPORTED_PYTHON % (PYIMP, '3.5'))
Exception: 
----------------------------------------
 Celery 4.0 requires CPython 3.5 or later
----------------------------------------
- For CPython 2.6, PyPy 1.x, Jython 2.6, CPython 3.2->3.3; use Celery 3.1:
    $ pip install 'celery<4'
- For CPython 2.5, Jython 2.5; use Celery 3.0:
    $ pip install 'celery<3.1'
- For CPython 2.4; use Celery 2.2:
    $ pip install 'celery<2.3'
Command exited with code 1
```

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -12,6 +12,11 @@ environment:
     # a later point release.
     # See: https://www.appveyor.com/docs/installed-software#python
 
+    - PYTHON: "C:\\Python27-x64"
+      PYTHON_VERSION: "2.7.x"
+      PYTHON_ARCH: "64"
+      WINDOWS_SDK_VERSION: "v7.0"
+
     - PYTHON: "C:\\Python35-x64"
       PYTHON_VERSION: "3.5.x"
       PYTHON_ARCH: "64"
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -41,6 +41,7 @@ install:
   - "%PYTHON%/python -m pip install -U pip setuptools"
   - "%PYTHON%/Scripts/pip.exe install -U eventlet"
   - "%PYTHON%/Scripts/pip.exe install -U -r requirements/extras/thread.txt"
+  - "%PYTHON%/Scripts/pip.exe install -U -r requirements/test-ci-default.txt"
 
 build: off
 
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -12,21 +12,23 @@ environment:
     # a later point release.
     # See: https://www.appveyor.com/docs/installed-software#python
 
-    - PYTHON: "C:\\Python27"
-      PYTHON_VERSION: "2.7.x"
-      PYTHON_ARCH: "32"
-
-    - PYTHON: "C:\\Python34"
-      PYTHON_VERSION: "3.4.x"
-      PYTHON_ARCH: "32"
-
-    - PYTHON: "C:\\Python27-x64"
-      PYTHON_VERSION: "2.7.x"
+    - PYTHON: "C:\\Python35-x64"
+      PYTHON_VERSION: "3.5.x"
       PYTHON_ARCH: "64"
-      WINDOWS_SDK_VERSION: "v7.0"
+      WINDOWS_SDK_VERSION: "v7.1"
 
-    - PYTHON: "C:\\Python34-x64"
-      PYTHON_VERSION: "3.4.x"
+    - PYTHON: "C:\\Python36-x64"
+      PYTHON_VERSION: "3.6.x"
+      PYTHON_ARCH: "64"
+      WINDOWS_SDK_VERSION: "v7.1"
+      
+    - PYTHON: "C:\\Python37-x64"
+      PYTHON_VERSION: "3.7.x"
+      PYTHON_ARCH: "64"
+      WINDOWS_SDK_VERSION: "v7.1"
+      
+    - PYTHON: "C:\\Python38-x64"
+      PYTHON_VERSION: "3.8.x"
       PYTHON_ARCH: "64"
       WINDOWS_SDK_VERSION: "v7.1"
 
--- a/requirements/extras/sqs.txt
+++ b/requirements/extras/sqs.txt
@@ -1,2 +1,2 @@
 boto3>=1.9.125
-pycurl
+pycurl==7.43.0.2  # Latest version with wheel built (for appveyor)
--- a/requirements/test-ci-default.txt
+++ b/requirements/test-ci-default.txt
@@ -21,4 +21,4 @@
 -r extras/azureblockblob.txt
 
 # SQS dependencies other than boto
-pycurl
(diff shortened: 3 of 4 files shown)


auvipy:
why 2.7 is being dropped?

matteius:
@auvipy I mentioned the answer in the description above, but I will reiterate.

> why 2.7 is being dropped?

Appveyor is a Windows CI so it builds everything in Windows and is already failing to fetch the python 2.7 environment, it prints out the deprecation warning:
```
DEPRECATION: Python 2.7 will reach the end of its life on January 1st, 2020. Please upgrade your Python as Python 2.7 won't be maintained after that date. A future version of pip will drop support for Python 2.7. More details about Python 2 support in pip, can be found at https://pip.pypa.io/en/latest/development/release-process/#python-2-support
```

Why should having appveyor (windows builds) always fail the build because it is configured to never work properly, be somehow a proof that the project still can run on python 2?

matteius:
Perhaps the real issue with 2.7 for Appveyor is:

> pkg_resources.ContextualVersionConflict: (python-dateutil 2.8.1 (c:\projects\celery\.eggs\python_dateutil-2.8.1-py2.7.egg), Requirement.parse('python-dateutil<2.8.1,>=2.1'), set(['botocore']))

but I am not sure there is a great fix to that either -- perhaps figure out what those things have to be pinned to and call out 2.7 with a different set of requirements.txt in the tox.ini ?  Just doesn't seem worth it for something that is going away -- have the Travis CI show that the 2.7 unit tests still pass should be sufficient evidence the code can be run on 2.7.

appveyor.yml
```diff
@@ -12,21 +12,23 @@ environment:
     # a later point release.
     # See: https://www.appveyor.com/docs/installed-software#python
 
-    - PYTHON: "C:\\Python27"
-      PYTHON_VERSION: "2.7.x"
-      PYTHON_ARCH: "32"
```
auvipy: just try to use 64bit and keep it, please. so that this can be merged.

matteius:
It can be merged as incremental progress -- it further reveals the next set of issues related to the windows celery build's.  After the latest run build failures are:
python2.7 --  couchbase will never support py2 on windows (requires py3)
python3.5 and 3.6 get the furthest along -- they eventually fail on the pylibmc install not finding the libmemcached headers:
` c:\users\appveyor\appdata\local\temp\1\pip-install-jw3v0vds\pylibmc\src\_pylibmcmodule.h(42): fatal error C1083: Cannot open include file: 'libmemcached/memcached.h': No such file or directory`
python3.7 and 3.8 fail earlier on the pycurl because the wheel files don't exist for those builds of python and so it fails to build Re:  libcurl  (we have only been able to work around this with appveyor by pinning to versions with prebuilt whl files that are compatible).

dejlek:
PyPy will continue supporting v2. Perhaps we should add PyPy 2 and 3 versions?

auvipy:
we will keep python 2 until the end of 2020

matteius:
We really need to differentiate project level support vs. appveyor build.   My chief argument here is that we have two build CIs right now:  Travis CI and Appveyor.  While Travis CI is designed to pass, we are locking ourselves into patterns that will always fail Appveyor builds, because Appveyor builds are Windows system builds.  

If the project dependency cannot build on Windows easily, then it will fail in Appveyor which is the problem we have today with all of the Appveyor builds.   We should design the happy path build case such that the master code repository passes as much as possible on both CIs -- why even have Appveyor if we aren't going to refine to have a happy path Green build of it?

I have been attempting through agile commit patterns tried to refine the exact blocker for each python version in the Appveyor output preventing it from building without introducing an additional impediment that would affect all builds to fail at an earlier step (in other words, reduce impediments to the builds succeeding).

For the builds that have gotten past the py2 issue, or the issue of finding a wheel for pycurl (some py3 builds have this), the latest issue is that none of the wheels line up for pylibmc.  You can tell because those builds run longer before failing in appveyor.
Its not that pylibmc has no wheels, it has some, but I think the weird naming pattern prevents them from being installed by default?  Still trying to figure that one out, but if we do, we might be able to get a few of the appveyor build configs to build green and that would be a victory in my book.

image

auvipy:
We got a plan to move from appveyor

matteius:
@auvipy What is the planned replacement for appveyor?  My rational argument would be to keep appveyor as a "Here is what a windows build looks like" CI and then add on any additional CI for the purpose appveyor was often perceived to of had.

auvipy:
check it out https://github.com/celery/ceps/issues/15#issuecomment-480147177

PR #5900, opened 2020-01-03
Very minor tweak to comment to improve docs

bernd-wechner:
As discussed here: 

https://stackoverflow.com/questions/58816271/celery-task-asyncresult-takes-task-id-but-is-documented-to-get-asyncresult-inst

this comment seems to flow to a very confusing and misleading piece of documentation here:

https://docs.celeryproject.org/en/latest/reference/celery.app.task.html#celery.app.task.Task.AsyncResult


bernd-wechner:
AppVeyor can fail on a change in a comment?

PR #5907, opened 2020-01-06
Database Backend: Use configured serializer, instead of always pickling.

martincapello:
Rebases the changes defined on #2881 (since they are almost 4 years old)  and refactors them a bit.
Original PR's comment:

> Use the BLOB as an sa.BLOB
> Serialise the result an add to the db as bytes

### DIFF
--- a/celery/backends/database/models.py
+++ b/celery/backends/database/models.py
@@ -5,7 +5,7 @@ from __future__ import absolute_import, unicode_literals
 from datetime import datetime
 
 import sqlalchemy as sa
-from sqlalchemy.types import PickleType
+#from sqlalchemy.types import PickleType
 
 from celery import states
 from celery.five import python_2_unicode_compatible
@@ -26,7 +26,7 @@ class Task(ResultModelBase):
                    primary_key=True, autoincrement=True)
     task_id = sa.Column(sa.String(155), unique=True)
     status = sa.Column(sa.String(50), default=states.PENDING)
-    result = sa.Column(PickleType, nullable=True)
+    result = sa.Column(sa.BLOB, nullable=True)
     date_done = sa.Column(sa.DateTime, default=datetime.utcnow,
                           onupdate=datetime.utcnow, nullable=True)
     traceback = sa.Column(sa.Text, nullable=True)
@@ -83,7 +83,7 @@ class TaskSet(ResultModelBase):
     id = sa.Column(sa.Integer, sa.Sequence('taskset_id_sequence'),
                    autoincrement=True, primary_key=True)
     taskset_id = sa.Column(sa.String(155), unique=True)
-    result = sa.Column(PickleType, nullable=True)
+    result = sa.Column(sa.BLOB, nullable=True)
     date_done = sa.Column(sa.DateTime, default=datetime.utcnow,
                           nullable=True)
 
--- a/celery/backends/database/models.py
+++ b/celery/backends/database/models.py
@@ -5,7 +5,6 @@ from __future__ import absolute_import, unicode_literals
 from datetime import datetime
 
 import sqlalchemy as sa
-#from sqlalchemy.types import PickleType
 
 from celery import states
 from celery.five import python_2_unicode_compatible
--- a/celery/backends/database/models.py
+++ b/celery/backends/database/models.py
@@ -25,7 +25,7 @@ class Task(ResultModelBase):
                    primary_key=True, autoincrement=True)
     task_id = sa.Column(sa.String(155), unique=True)
     status = sa.Column(sa.String(50), default=states.PENDING)
-    result = sa.Column(sa.BLOB, nullable=True)
+    result = sa.Column(sa.LargeBinary(length=(2**31)-1), nullable=True)
     date_done = sa.Column(sa.DateTime, default=datetime.utcnow,
                           onupdate=datetime.utcnow, nullable=True)
     traceback = sa.Column(sa.Text, nullable=True)
@@ -82,7 +82,7 @@ class TaskSet(ResultModelBase):
     id = sa.Column(sa.Integer, sa.Sequence('taskset_id_sequence'),
                    autoincrement=True, primary_key=True)
     taskset_id = sa.Column(sa.String(155), unique=True)
-    result = sa.Column(sa.BLOB, nullable=True)
+    result = sa.Column(sa.LargeBinary(length=(2**31)-1), nullable=True)
     date_done = sa.Column(sa.DateTime, default=datetime.utcnow,
                           nullable=True)
 
--- a/celery/backends/database/models.py
+++ b/celery/backends/database/models.py
@@ -25,7 +25,7 @@ class Task(ResultModelBase):
                    primary_key=True, autoincrement=True)
     task_id = sa.Column(sa.String(155), unique=True)
     status = sa.Column(sa.String(50), default=states.PENDING)
-    result = sa.Column(sa.LargeBinary(length=(2**31)-1), nullable=True)
+    result = sa.Column(sa.LargeBinary(length=(2 ** 31) - 1), nullable=True)
     date_done = sa.Column(sa.DateTime, default=datetime.utcnow,
                           onupdate=datetime.utcnow, nullable=True)
     traceback = sa.Column(sa.Text, nullable=True)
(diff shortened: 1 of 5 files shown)


celery/backends/database/models.py
```diff
@@ -26,7 +25,7 @@ class Task(ResultModelBase):
                    primary_key=True, autoincrement=True)
     task_id = sa.Column(sa.String(155), unique=True)
     status = sa.Column(sa.String(50), default=states.PENDING)
-    result = sa.Column(PickleType, nullable=True)
+    result = sa.Column(sa.BLOB, nullable=True)
```
georgepsarakis: Can we use `LargeBinary` with custom length instead so that the largest possible datatype is used (at least for MySQL):
- https://stackoverflow.com/a/43791881/920374
- https://dev.mysql.com/doc/refman/5.7/en/storage-requirements.html#data-types-storage-reqs-strings

I think PostgreSQL `BYTEA` columns have a limit of 1GB, so that RDBMS will be covered too.

georgepsarakis (commented):
Nice work! Some tests are breaking, strings seem to be converted to unicode now as far as I can tell. Could also please add a test for this backend, which uses the JSON serializer?

celery/backends/database/models.py
martincapello: Sure, will update it

martincapello:
Will try to add a test as soon as I can.

martincapello:
@georgepsarakis  I ended up creating a clone of the database backend tests with a different setup (the result serializer is set to json in this case). Does this makes sense? Or do you prefer a fixtures based approach modifying the current class?

martincapello:
I have an improvement for this...will push a commit on monday

georgepsarakis:
@martincapello it would be best if all files for this backend are contained within a single file. It would make it easier for future contributions to write tests that will cover all serializers.

Perhaps [parametrize](https://docs.pytest.org/en/latest/parametrize.html) helps here.

martincapello:
Oh yeah, I didn't mean to add a new file, just a new class into the same file. I tried to use parametrize, but ran into some issues that I will explain on monday.

martincapello:
This is the issue I ran into:
Let's say I choose to use this:

```python
    @pytest.mark.parametrize(
        'result_serializer', ['pickle', 'json'],
        ids=['using pickle', 'using json']
    )

```
For instance, for the following test (which I believe it should be renamed as "test_is_serialized", but we can discuss this later, I'll keep the name as is for now):
```python
    @pytest.mark.parametrize(
        'result_serializer', ['pickle', 'json'],
        ids=['using pickle', 'using json']
    )
    def test_is_pickled(self, result_serializer):
        self.app.conf.result_serializer = result_serializer
        tb = DatabaseBackend(self.uri, app=self.app)

        tid2 = uuid()
        result = {'foo': 'baz', 'bar': SomeClass(12345)}
        tb.mark_as_done(tid2, result)
        # is serialized properly.
        rindb = tb.get_result(tid2)
        assert rindb.get('foo') == 'baz'
        assert rindb.get('bar').data == 12345
```
When I run the above test, it fails for the json result serializer, because rindb.get('bar') doesn't have a member named "data", this is due to rindb.get('bar') returning a dict (which is what the json decoder returns as far as I understood).
So, a quick workaround makes the test look like this:
```python
    @pytest.mark.parametrize(
        'result_serializer', ['pickle', 'json'],
        ids=['using pickle', 'using json']
    )
    def test_is_pickled(self, result_serializer):
        self.app.conf.result_serializer = result_serializer
        tb = DatabaseBackend(self.uri, app=self.app)

        tid2 = uuid()
        result = {'foo': 'baz', 'bar': SomeClass(12345)}
        tb.mark_as_done(tid2, result)
        # is serialized properly.
        rindb = tb.get_result(tid2)
        assert rindb.get('foo') == 'baz'
        if result_serializer == 'json':      # <--- custom logic to workaround the issue
            assert rindb.get('bar').get('data') == 12345
        else:
            assert rindb.get('bar').data == 12345
```
As you can see, I needed to add some logic to determine the right assertion to do, which -IMO- goes against the spirit of using parametrize.
So, for this reason, as a first approach I decided to make a copy of the class that works for the json serializer. It worked, so I pushed a commit.
Then I realized that I could improve that.
So, what I propose now is to split the test_DatabaseBackend class in three classes:
- test_DatabaseBackend: would contain the tests that must pass for all the serializers. It would contain almost all the tests it has right now, except the test_is_pickled.
- test_DatabaseBackend_pickle_serializer: would contain only tests that pass when using the pickle serializer and we don't care what happens for other serializers. It would contain the test_is_pickled.
- test_DatabaseBackend_json_serializer: would contain only tests that pass when using the json serializer and we don't care what happens for other serializers. It would contain a test which is almost the same as test_is_pickled but with the fixed assertion for the json serializer.
If I wasn't clear, it might be easier to take a look at PR's code.

martincapello:
@georgepsarakis any thoughts on my latest updates?

celery/backends/database/models.py
```diff
@@ -26,7 +25,7 @@ class Task(ResultModelBase):
                    primary_key=True, autoincrement=True)
     task_id = sa.Column(sa.String(155), unique=True)
     status = sa.Column(sa.String(50), default=states.PENDING)
-    result = sa.Column(PickleType, nullable=True)
+    result = sa.Column(sa.LargeBinary(length=(2 ** 31) - 1), nullable=True)
```
auvipy: isn't this a breaking change?

celery/backends/database/models.py
```diff
@@ -83,7 +82,7 @@ class TaskSet(ResultModelBase):
     id = sa.Column(sa.Integer, sa.Sequence('taskset_id_sequence'),
                    autoincrement=True, primary_key=True)
     taskset_id = sa.Column(sa.String(155), unique=True)
-    result = sa.Column(PickleType, nullable=True)
+    result = sa.Column(sa.LargeBinary(length=(2 ** 31) - 1), nullable=True)
```
auvipy: a breaking change?

celery/backends/database/models.py
martincapello: Not sure...I believe this could break celery apps using a serializer other than pickle when the results backend is sqlalchemy based.
According to the sqlalchemy docs:

> PickleType builds upon the Binary type to apply Python’s pickle.dumps() to incoming objects, and pickle.loads() on the way out, allowing any pickleable Python object to be stored as a serialized binary field.

So, the database column should stay binary (I've been able to confirm this using postgres), the issue with PickleType is that Sqlalchemy pickles the objects automatically, despite having a json (or any but pickle) serializer configured for our app.

Hence, it seems that results from old task runs (tasks ran before this change) will have a pickled results instead of a json serialized one.

Can a celery client access to tasks results that were already ran in the past? I'm asking because am a celery noob.

georgepsarakis: > Can a celery client access to tasks results that were already ran in the past?

Yes, this is possible if someone has access to the result (task) ID.

This is indeed a breaking change, for configurations that used a different serializer than the pickle serializer. Perhaps this can be solved by allowing a transitional fallback to `pickle.loads` on a deserialization error, but only with a configuration option. So the new serializer support will be disabled by default, and we emit a warning for the deprecation in the next version.

How does that sound?

georgepsarakis:
> 
> When I run the above test, it fails for the json result serializer, because rindb.get('bar') doesn't have a member named "data", this is due to rindb.get('bar') returning a dict (which is what the json decoder returns as far as I understood).

I am not really sure how you end up getting this result when using the JSON serializer, since `SomeClass` instances cannot be encoded by the JSON encoder:

```python
assert rindb.get('bar').get('data') == 12345
```

I think the most easy way to test, would be to make the expected result parametric too, as seen in [other test cases](https://github.com/celery/celery/blob/5ab9ebeec4f5d34dac00545691e7daae62ed6448/t/unit/backends/test_database.py#L232).

martincapello:
@georgepsarakis 
> I am not really sure how you end up getting this result when using the JSON serializer, since `SomeClass` instances cannot be encoded by the JSON encoder:

`SomeClass` *was* not json serializable until I added this [json](https://github.com/celery/celery/pull/5907/files#diff-ec1ce96f3697749abc4c7c46f5393914R35) method.
 
> I think the most easy way to test, would be to make the expected result parametric too, as seen in [other test cases](https://github.com/celery/celery/blob/5ab9ebeec4f5d34dac00545691e7daae62ed6448/t/unit/backends/test_database.py#L232).  

Sorry, but I'm not following you here, because the expected result is the same (12345 in this case) for both serializers. Instead, the thing that changes is the structure of the `rindb` instance, which is the result object that have the actual values I want to test.

celery/backends/database/models.py
martincapello: > Yes, this is possible if someone has access to the result (task) ID.

Alright then, let's suppose that this issue never existed and the serialization was working fine for any serializer. Also suppose someone started using celery configured with the json serializer, and after some time he changes his mind and decides to use pickle serializer. In this case, if someone has access to a result (task) ID corresponding to a task ran before the serializer change, would in this case the same deserialization error be expected?

auvipy:
let's retarget this for celery 4.5

celery/backends/database/models.py
auvipy: sqla should have support jsonfield as well, need to check that

k-sparrow:
Any news on this?

PR #5910, opened 2020-01-08
Support configuring schema of a PostgreSQL database

### LINKED ISSUE #1013: Support storing results in a specific schema of a PostgreSQL database
I'm using celery in a PostgreSQL environment where I only have a single database.
So I use the "schema" idiom to keep non-strongly-related data apart.
That's why I'd like to be able to store the celery task results in a different schema than the default "public" one.

Source code modification would take place in `backends/database/models.py`, both the Task & TaskSet tables and their related Sequence.
Plus of course adding a configuration hook to specify the schema name.

Thanks!

soldag:
These changes make it possible to configure the schema in which the result tables should be created when using a PostgreSQL database as result backend. 

Fixes #1013

### DIFF
--- a/celery/app/defaults.py
+++ b/celery/app/defaults.py
@@ -244,6 +244,7 @@ NAMESPACES = Namespace(
         short_lived_sessions=Option(
             False, type='bool', old={'celery_result_db_short_lived_sessions'},
         ),
+        table_schemas=Option(type='dict'),
         table_names=Option(type='dict', old={'celery_result_db_tablenames'}),
     ),
     task=Namespace(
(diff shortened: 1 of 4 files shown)


soldag:
Is this simple test fine for you? It's hard to test if the tables are really created inside the configured schema, because sqlite is used for the unit tests and it doesn't support multiple schemas inside a database.

If not, could you provide some instructions on how to test this properly?

auvipy:
it's ok just flake8 /unit/backends/test_database.py:221:1: E303 too many blank lines (3)

soldag:
Alright, fixed it :+1:

soldag:
Can I count on this change to be released end of this month as stated on the 4.4.x milestone?

auvipy:
yes

soldag:
@auvipy  No release, yet? :thinking:

auvipy:
this will take at least 1 or 2 weeks more :)

PR #5915, opened 2020-01-11
Add progress for retry connections

### LINKED ISSUE #4556: Celery hangs indefinitely (instead of failing) if redis is not started
## Checklist

       version affected).

```
software -> celery:4.1.0 (latentcall) kombu:4.1.0 py:2.7.12
            billiard:3.5.0.3 redis:2.10.6
settings -> transport:redis results:disabled

broker_url: u'redis://localhost:6379/0'
```

## Steps to reproduce

Stop `redis-server.`
Initialize a celery instance configured with redis.

## Expected behavior

I expected an `OperationalError` or another exception communicating that celery cannot connect to the broker server.

## Actual behavior

It hangs indefinitely.

## Why this is important

In the current situation is not possible to write automated functional and integration tests that check whether the software handles failures when brokers are not reachable.

marfgold1:
This will show current retry progress so it will clear confusion about how many retries will be tried for connecting to broker.
I still leave the default value in 100, because If I change it, I'm afraid it will affect many celery apps that was okay or used to with the current default value 100 in `broker_connection_max_retries`.
May fixes #4556

### DIFF
--- a/celery/worker/consumer/consumer.py
+++ b/celery/worker/consumer/consumer.py
@@ -51,7 +51,7 @@ Trying to re-establish the connection...\
 """
 
 CONNECTION_RETRY_STEP = """\
-Trying again {when}...\
+Trying again {when}... ({retries}/{max_retries})\
 """
 
 CONNECTION_ERROR = """\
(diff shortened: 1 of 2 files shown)


marfgold1:
Yup I've tried it
image
You can see there is something like `(1/100)` etc after `Trying again in...`
Should I add a test for this small changes too? I'm not really sure how to write the test for this one....

marfgold1:
Does travis just not receive an output for 10mins? I've seen other commit also errored on the travis.
But I've already run the test and flakecheck in my commit and it's fine:
image

auvipy:
network issue on travis

nemesisdesign:
Great to see this merged! Thank you @marfgold1!

PR #5921, opened 2020-01-15
Handle Redis connection errors in result consumer

### LINKED ISSUE #5919: Handle Redis connection errors in result consumer
# Checklist

  master branch.

#### Related Issues

- None

#### Possible Duplicates

- None

# Brief Summary

When there is a connection error with Redis while executing a command, in most cases, the redis client will [discard the connection](https://github.com/andymccurdy/redis-py/blob/272d3139e1c82c2d89551f87d12df7c18d938ea2/redis/client.py#L885-L886), causing the next command sent to Redis to open a new connection. This allows applications to recover from connection errors by simply retrying, a property that is used in Celery, for example when setting keys in the Redis result backend: https://github.com/celery/celery/blob/d0563058f8f47f347ac1b56c44f833f569764482/celery/backends/redis.py#L324-L325

This is not the case however when the [connection to Redis is in a pubsub state](https://github.com/andymccurdy/redis-py/blob/272d3139e1c82c2d89551f87d12df7c18d938ea2/redis/client.py#L3397-L3414). The reason for that is that some state associated with the connection (namely the list of keys subscibed to). The Redis client doesn't keep track of this state, so it can't possibly restore it when creating a new connection and leaves the connection handling to the application code.

The Celery Redis result consumer uses pubsub in order to be notified when results are available, but doesn't handle connection errors at all, causing a result consumer to end up in a state where it can't connect to the result backend any more after a single connection error, as any further attempt will reuse the same faulty connection.

The solution would be to add error handling logic to the result consumer, so it will recreate the connection on connection errors and initialize it to the proper state.

# Design

None

## Proposed Behavior

Add error handling in all places the Redis result consumer sends a Redis command in a pubsub context:
* https://github.com/celery/celery/blob/d0563058f8f47f347ac1b56c44f833f569764482/celery/backends/redis.py#L127
* https://github.com/celery/celery/blob/d0563058f8f47f347ac1b56c44f833f569764482/celery/backends/redis.py#L142
* https://github.com/celery/celery/blob/d0563058f8f47f347ac1b56c44f833f569764482/celery/backends/redis.py#L148

We should catch all Redis connection errors, and call a new method that will reinitialize a pubsub connection in the proper state (discard the current connection from the pool, start the pubsub context, subscribe to all keys in `ResultConsumer.subscribed_to`) using the retry policy. If in `drain_events`, we should try to get new messages again. 

This will take care of most issues with connection errors. I see two remaining issues:
1.Some message might have been lost (sent between losing the connection and reconnecting). We could read all keys subscribed to right after reconnecting and before starting the pubsub context and call `on_state_change` for each existing key, but this might cause some messages to be delivered twice and I don't know how Celery will react to that.
2. If the connection can't be re-established despite the retries and reaches max-retries, the result consumer will end up with a faulty connection that can't be recovered from. This should be communicated somehow to the user (documentation, logging an explicit error message, custom exception).

## Proposed UI/UX

None

## Diagrams

N/A

## Alternatives

None

michamos:

## Description
This fixes #5919 by adding reconnection logic to the Redis result backend which restores the pubsub context and deals with tasks that changed state while the connection was down.

### DIFF
--- a/celery/backends/redis.py
+++ b/celery/backends/redis.py
@@ -88,6 +88,8 @@ class ResultConsumer(BaseResultConsumer):
         super(ResultConsumer, self).__init__(*args, **kwargs)
         self._get_key_for_task = self.backend.get_key_for_task
         self._decode_result = self.backend.decode_result
+        self._ensure = self.backend.ensure
+        self._connection_errors = self.backend.connection_errors
         self.subscribed_to = set()
 
     def on_after_fork(self):
@@ -98,6 +100,20 @@ class ResultConsumer(BaseResultConsumer):
         except KeyError as e:
             logger.warning(text_t(e))
         super(ResultConsumer, self).on_after_fork()
+
+    def _reconnect_pubsub(self):
+        self._pubsub = None
+        self.backend.client.connection_pool.reset()
+        # task state might have changed when the connection was down so we
+        # retrieve meta for all subscribed tasks before going into pubsub mode
+        metas = self.backend.client.mget(self.subscribed_to)
+        metas = [meta for meta in metas if meta]
+        for meta in metas:
+            self.on_state_change(self._decode_result(meta), None)
+        self._pubsub = self.backend.client.pubsub(
+            ignore_subscribe_messages=True,
+        )
+        self._pubsub.subscribe(*self.subscribed_to)
 
     def _maybe_cancel_ready_task(self, meta):
         if meta['status'] in states.READY_STATES:
@@ -124,9 +140,12 @@ class ResultConsumer(BaseResultConsumer):
 
     def drain_events(self, timeout=None):
         if self._pubsub:
-            message = self._pubsub.get_message(timeout=timeout)
-            if message and message['type'] == 'message':
-                self.on_state_change(self._decode_result(message['data']), message)
+            try:
+                message = self._pubsub.get_message(timeout=timeout)
+                if message and message['type'] == 'message':
+                    self.on_state_change(self._decode_result(message['data']), message)
+            except self._connection_errors:
+                self._ensure(self._reconnect_pubsub, ())
         elif timeout:
             time.sleep(timeout)
 
@@ -139,13 +158,19 @@ class ResultConsumer(BaseResultConsumer):
         key = self._get_key_for_task(task_id)
         if key not in self.subscribed_to:
             self.subscribed_to.add(key)
-            self._pubsub.subscribe(key)
+            try:
+                self._pubsub.subscribe(key)
+            except self._connection_errors:
+                self._ensure(self._reconnect_pubsub, ())
 
     def cancel_for(self, task_id):
+        key = self._get_key_for_task(task_id)
+        self.subscribed_to.discard(key)
         if self._pubsub:
-            key = self._get_key_for_task(task_id)
-            self.subscribed_to.discard(key)
-            self._pubsub.unsubscribe(key)
+            try:
+                self._pubsub.unsubscribe(key)
+            except self._connection_errors:
+                self._ensure(self._reconnect_pubsub, ())
 
 
 class RedisBackend(BaseKeyValueStoreBackend, AsyncBackendMixin):
--- a/celery/backends/redis.py
+++ b/celery/backends/redis.py
@@ -3,6 +3,7 @@
 from __future__ import absolute_import, unicode_literals
 
 import time
+from contextlib import contextmanager
 from functools import partial
 from ssl import CERT_NONE, CERT_OPTIONAL, CERT_REQUIRED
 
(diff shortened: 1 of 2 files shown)


celery/backends/redis.py
```diff
@@ -139,13 +158,19 @@ def _consume_from(self, task_id):
         key = self._get_key_for_task(task_id)
         if key not in self.subscribed_to:
             self.subscribed_to.add(key)
-            self._pubsub.subscribe(key)
+            try:
+                self._pubsub.subscribe(key)
```
georgepsarakis: Perhaps we could define a [context manager](https://docs.python.org/3.6/library/contextlib.html#contextlib.contextmanager) that handles the reconnection process.

celery/backends/redis.py
```diff
@@ -99,6 +101,20 @@ def on_after_fork(self):
             logger.warning(text_t(e))
         super(ResultConsumer, self).on_after_fork()
 
+    def _reconnect_pubsub(self):
+        self._pubsub = None
+        self.backend.client.connection_pool.reset()
+        # task state might have changed when the connection was down so we
+        # retrieve meta for all subscribed tasks before going into pubsub mode
+        metas = self.backend.client.mget(self.subscribed_to)
```
georgepsarakis: Can we use `mget` with channels?

celery/backends/redis.py
```diff
@@ -99,6 +101,20 @@ def on_after_fork(self):
...
+        metas = [meta for meta in metas if meta]
+        for meta in metas:
+            self.on_state_change(self._decode_result(meta), None)
+        self._pubsub = self.backend.client.pubsub(
```
georgepsarakis: I know this is a single statement, but the PubSub instance is also created in the `start` method. Do you think it would make sense to add a method just for that purpose?

celery/backends/redis.py
michamos: Done!

celery/backends/redis.py
michamos: No we can't. This uses the fact that `RedisBackend.set` not only publishes the task status update to the channel, but also assigns it to a key with the same name. See https://github.com/celery/celery/blob/90fe53f901cf2e71c570fa639aa9b39ad228fadd/celery/backends/redis.py#L329-L333
In normal operation the consumer listens to messages published to the channel, but when reconnecting it will fetch all messages from the keys instead. This is tested since https://github.com/celery/celery/pull/5921/files#diff-5cb790f5c0c0aea16658640ccd102c19R246 is `RedisBackend.set` and not `Redis.set`.

celery/backends/redis.py
michamos: I don't think it's worth it, as you said it's a single statement, but if you insist I can add it.

michamos:
@georgepsarakis @auvipy do you have any opinion on what should be done in case the max number of retries is reached without the connection being re-established? That's a state celery can't recover from, so I believe this should be communicated clearly to the user. For reference, this is what the worker does on unrecoverable errors: https://github.com/celery/celery/blob/90fe53f901cf2e71c570fa639aa9b39ad228fadd/celery/worker/worker.py#L208-L210
IMHO killing the process shouldn't be done celery doesn't "own" it unlike the worker, but logging a clear message would be good.

thedrow:
> @georgepsarakis @auvipy do you have any opinion on what should be done in case the max number of retries is reached without the connection being re-established? That's a state celery can't recover from, so I believe this should be communicated clearly to the user. For reference, this is what the worker does on unrecoverable errors:
> 
> https://github.com/celery/celery/blob/90fe53f901cf2e71c570fa639aa9b39ad228fadd/celery/worker/worker.py#L208-L210
> 
> IMHO killing the process shouldn't be done as celery doesn't "own" it unlike the worker, but logging a clear message would be good.

In my opinion you should notify the user that this error has occurred and do nothing else.
We can't retry forever.

The worker shouldn't crash, that's for sure.

frost-nzcr4:
Should the Redis backend settings be extended with something like **redis_max_reconnect_retries** which in result give us a reconnection timeout bounded to _redis_socket_connect_timeout_ option and equals to **redis_max_reconnect_retries** * _redis_socket_connect_timeout_. Or would one's add another option **redis_max_reconnect_timeout** to unbind of _redis_socket_connect_timeout_?

frost-nzcr4:
Some changes to Redis backend via #5952 that could help.

auvipy:
can you push on the branch?

michamos:
@frost-nzcr4 It's already configurable through the [retry policy](https://docs.celeryproject.org/en/stable/userguide/calling.html#retry-policy). Do you think there's any practical use for having a different timeout here?

frost-nzcr4:
Thanks, I should give a try with the policy-retry, that's probably the case in some of my troubles. Anyway I waiting for this PR in it's current form either.

michamos:
> In my opinion you should notify the user that this error has occurred and do nothing else.
> We can't retry forever.

Done. I believe all issues have been addressed and this PR can be merged.

PR #5934, opened 2020-01-24
Fixed problem with conflicting autoretry_for task parameter and Task.replace()

M1hacka:
## Description
When task has `autoretry_for=(Exception,)` attribute, it catches `celery.exceptions.Ignore` exception and makes retries of original task. This behavior is incorrect as it leads to multiple executions of task, even if it succeeds.

How to reproduce:
```python
@shared_task(queue='second')
def executed_task():
    print('hello world')

@shared_task(queue='first', autoretry_for=(Exception,), bind=True)
def replaced_task(self):
    sig = executed_task.s()
    return self.replace(sig)

if __name__ == '__main__':
    replaced_task.delay()
```
1. `replace_task.delay()` writes task to broker
2. `replaced_task` executes, plans `executed_task` over `apply_async` and raises `Ignore` exception
3. `autoretry_for` catches `Ignore` exception and plans retry of `replaced_task`
4. Repeat from step 2 until retry attempts end...
As a result, `executed_task` is called several times.

### DIFF
--- a/celery/app/base.py
+++ b/celery/app/base.py
@@ -23,7 +23,7 @@ from celery._state import (_announce_app_finalized, _deregister_app,
                            _register_app, _set_current_app, _task_stack,
                            connect_on_app_finalize, get_current_app,
                            get_current_worker_task, set_default_app)
-from celery.exceptions import AlwaysEagerIgnored, ImproperlyConfigured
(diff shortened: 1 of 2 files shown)


M1hacka:
Hmm, I'm not sure what do you expect me to do. I've already added test on 2 lines I've changed.

auvipy:
thanks! i was talking from codecov matrix

PR #5939, opened 2020-01-28
Added links to other issue trackers

thedrow:

## Description

When picking a template, you can also direct users to other places such as the community forum or other issue trackers.

This PR disables blank issues and points the users to the other issue trackers in case they have an issue that is not related to Celery specifically.


thedrow:
For some reason only one template is now displayed.
I'm emailing Github about it since it appears to be a bug but I think we should revert this one for now.

PR #5941, opened 2020-01-29
remove redundant raise from docstring

AlexRiina:
`throw` is True by default so the Retry exception will already get raised by calling `self.retry(countdown=60 * 5, exc=exc)`


Lexicality:
Hello, I was just wondering about this change because the note in the docstring immediately after this example recommends that you should include `raise` when calling `self.retry`:

> Although the task will never return above as retry raises an exception to notify the worker, we use raise in front of the retry to convey that the rest of the block won’t be executed.

Is this no longer best practice?

PR #5945, opened 2020-01-31
Added a default value for retries in worker.strategy.

abhishekakamai:
I was facing an issue when adding tasks directly to rabbitmq using pika instead of calling `task.apply_async`. The issue was the self.retry mechanism was failing. In `app/tasks.py` the line
`retries = request.retries + 1` was causing the issue. On further tracing I figured out that it was because the default `.get` value (`None`) was getting passed through this function and was raising 

> TypeError: unsupported operand type(s) for +: 'NoneType' and 'int'

### DIFF
--- a/celery/worker/strategy.py
+++ b/celery/worker/strategy.py
@@ -48,7 +48,7 @@ def hybrid_to_proto2(message, body):
         'shadow': body.get('shadow'),
         'eta': body.get('eta'),
         'expires': body.get('expires'),
-        'retries': body.get('retries'),
+        'retries': body.get('retries', 0),
         'timelimit': body.get('timelimit', (None, None)),
         'argsrepr': body.get('argsrepr'),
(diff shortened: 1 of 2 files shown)


bajubullet:
I actually wanted to create an issue for the same but the create issue button said that its for maintainers only. Also I have a question, should I change retries value here or update where the retry value is actually getting updated in task.py [here](https://github.com/celery/celery/blob/master/celery/app/task.py#L680)  or in the update function [here](https://github.com/celery/celery/blob/master/celery/app/task.py#L101).

If the current PR is fine, let me know. I'll add the necessary test cases.

auvipy:
generally, we don't merge PR without proper unit tests to avoid regression

bajubullet:
Added test cases for retries as you asked. Let me know if anything else is required.

bajubullet:
@auvipy any update on this?

auvipy:
you need to make the Travis green

PR #5951, opened 2020-02-03
Update __init__.py

Ixiodor:
> Fixed issue for object instance with result_backend=True (decode fails on multiple possibile None request)

### DIFF
--- a/celery/backends/database/__init__.py
+++ b/celery/backends/database/__init__.py
@@ -163,7 +163,7 @@ class DatabaseBackend(BaseBackend):
             data = task.to_dict()
             if data.get('args', None) is not None:
                 data['args'] = self.decode(data['args'])
-            if data.get('kwargs', None) is not None::
+            if data.get('kwargs', None) is not None:
                 data['kwargs'] = self.decode(data['kwargs'])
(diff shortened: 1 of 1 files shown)


celery/backends/database/__init__.py
```diff
@@ -161,9 +161,9 @@ def _get_task_meta_for(self, task_id):
                 task.status = states.PENDING
                 task.result = None
             data = task.to_dict()
-            if 'args' in data:
+            if 'args' in data and data['args'] is not None:
```
thedrow: This is better:
```suggestion
            if data.get('args', None) is not None:
```

celery/backends/database/__init__.py
```diff
@@ -161,9 +161,9 @@ def _get_task_meta_for(self, task_id):
...
                 data['args'] = self.decode(data['args'])
-            if 'kwargs' in data:
+            if 'kwargs' in data and data['kwargs'] is not None:
```
thedrow: This is better:
```suggestion
            if data.get('kwargs', None) is not None:
```

Ixiodor:
i'm not in touch with the source and structure knowledge.
I hope someone can help with this.

PR #5954, opened 2020-02-04
Remove PY3 conditionals

thedrow:

## Description

Remove `if PY3` conditions and replace them with their Python 3 supported code.


jheld:
What is the timeline for this? Would celery 5 (5.0) be more about py3 only and then working on other parts of the CEP that didn't get in yet? 5.1, 5.2, etc.

celery/local.py
```diff
@@ -395,14 +384,11 @@ def maybe_evaluate(obj):
...
+
 # im_func is no longer available in Py3.
 # instead the unbound method itself can be used.
-if sys.version_info[0] == 3:  # pragma: no cover
-    def fun_of_method(method):
-        return method
-else:
-    def fun_of_method(method):  # noqa
-        return method.im_func
+def fun_of_method(method):
```
georgepsarakis: I think this is only used [here](https://github.com/celery/celery/blob/6215f34d2675441ef2177bd850bf5f4b442e944c/celery/local.py#L498), so perhaps we can skip the function all together and just inline it?

celery/utils/serialization.py
```diff
@@ -267,27 +260,14 @@ def jsonify(obj,
...
-else:
-    def raise_with_context(exc):
-        exc_info = sys.exc_info()
-        if not exc_info:
-            raise exc
-        if exc_info[1] is exc:
-            raise
-        elif exc_info[2]:
-            reraise(type(exc), exc, exc_info[2])
+from vine.five import exec_
```
georgepsarakis: I think this import needs to be on top now since it is not in a conditional branch?

celery/utils/serialization.py
```diff
@@ -10,18 +10,14 @@
 
 from kombu.utils.encoding import bytes_to_str, str_to_bytes
 
-from celery.five import bytes_if_py2, items, reraise, string_t
-
+from celery.five import bytes_if_py2, items, string_t
```
georgepsarakis: Perhaps out of scope for this PR, but I wonder if we still need `bytes_if_py2` here.

celery/canvas.py
```diff
@@ -468,10 +463,9 @@ def __json__(self):
     def __repr__(self):
         return self.reprcall()
 
-    if JSON_NEEDS_UNICODE_KEYS:  # pragma: no cover
-        def items(self):
-            for k, v in dict.items(self):
-                yield k.decode() if isinstance(k, bytes) else k, v
+    def items(self):
+        for k, v in dict.items(self):
+            yield k.decode() if isinstance(k, bytes) else k, v
```
bertonha, line 468: looks like here is missing yield the `v` if is instance of bytes.

thedrow: That may be the case but we need to fix this on master if it is so.

thedrow:
The tests pass locally.

thedrow:
After investigation, it seems that the test passes locally but fails on tox.

PR #5955, opened 2020-02-04
 Fix labels on Github issue templates.

martey:
Fixes #5938, which broke most of the issue templates.

Currently, navigating to https://github.com/celery/celery/issues/new/choose now only displays the "Minor Version Release Checklist" template:

Screenshot from 2020-02-04 12-53-40

This pull request escapes the labels in the template files (so that their colons don't cause the template to break) ~~and changes the "Documentation Bug Report" template to only use one label (since using multiple escaped labels also breaks the template)~~.

### DIFF
--- a/.github/ISSUE_TEMPLATE/Bug-Report.md
+++ b/.github/ISSUE_TEMPLATE/Bug-Report.md
@@ -1,7 +1,7 @@
 ---
 name: Bug Report
 about: Is something wrong with Celery?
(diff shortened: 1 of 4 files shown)


martey:
I've pushed a new commit which fixes the issue with the Documentation Bug Report's template multiple labels.

thedrow:
So that was the problem!
Their documentation is very unclear about that point.

PR #5962, opened 2020-02-18
Attempt to reload imported task modules

chbndrhnns:

### DIFF
--- a/celery/loaders/base.py
+++ b/celery/loaders/base.py
@@ -90,7 +90,7 @@ class BaseLoader(object):
 
     def import_module(self, module, package=None):
         want_reload_module = sys.modules.get(module, None)
-        if want_reload_module:
+        if want_reload_module and getattr(want_reload_module, '__spec__'):
             importlib.reload(want_reload_module)
         return importlib.import_module(module, package=package)
 
--- a/celery/loaders/base.py
+++ b/celery/loaders/base.py
@@ -89,6 +89,9 @@ class BaseLoader(object):
         return self.import_from_cwd(module)
 
     def import_module(self, module, package=None):
+        want_reload_module = sys.modules.get(module, None)
+        if want_reload_module:
+            importlib.reload(want_reload_module)
         return importlib.import_module(module, package=package)
 
     def import_from_cwd(self, module, imp=None, package=None):
--- a/t/unit/contrib/proj/foo.py
+++ b/t/unit/contrib/proj/foo.py
@@ -1,7 +1,7 @@
 from __future__ import absolute_import, unicode_literals
 
 from celery import Celery, shared_task
-from xyzzy import plugh  # noqa
+from .xyzzy import plugh  # noqa
 
 app = Celery()
 
--- a/t/unit/contrib/proj/foo.py
+++ b/t/unit/contrib/proj/foo.py
@@ -1,7 +1,7 @@
 from __future__ import absolute_import, unicode_literals
 
 from celery import Celery, shared_task
-# from xyzzy import plugh  # noqa
+from xyzzy import plugh  # noqa
 
 app = Celery()
 
--- a/t/unit/contrib/proj/foo.py
+++ b/t/unit/contrib/proj/foo.py
@@ -1,7 +1,7 @@
 from __future__ import absolute_import, unicode_literals
 
 from celery import Celery, shared_task
-from xyzzy import plugh  # noqa
+# from xyzzy import plugh  # noqa
 
 app = Celery()
 
--- a/t/unit/contrib/test_pytest.py
+++ b/t/unit/contrib/test_pytest.py
@@ -59,9 +59,9 @@ def test_pytest_import_passes(celery_app, celery_worker):
     task.get()
 
 
-def test_pytest_second_import_fails(celery_app, celery_worker):
+def test_pytest_second_import_fails_without_patched_module(celery_app, celery_worker):
     task = celery_app.send_task("t.unit.contrib.proj.foo.bar")
-    with pytest.raises(Exception):
+    with pytest.raises(KeyError):
         task.get()
 
 
--- a/t/unit/contrib/test_pytest.py
+++ b/t/unit/contrib/test_pytest.py
@@ -1,3 +1,5 @@
+import importlib
+import sys
 import pytest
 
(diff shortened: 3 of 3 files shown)


t/unit/contrib/proj/foo.py
```diff
@@ -1,7 +1,7 @@
 from __future__ import absolute_import, unicode_literals
 
 from celery import Celery, shared_task
-from xyzzy import plugh  # noqa
+# from xyzzy import plugh  # noqa
```
thedrow: Why is this commented out?

celery/loaders/base.py
```diff
@@ -89,6 +89,9 @@ def import_task_module(self, module):
         return self.import_from_cwd(module)
 
     def import_module(self, module, package=None):
+        want_reload_module = sys.modules.get(module, None)
```
thedrow: Why would we always want to reload already imported modules?

t/unit/contrib/test_pytest.py
```diff
@@ -32,3 +34,37 @@ def test_noop():
         result.stdout.fnmatch_lines_random(
             "*PytestUnknownMarkWarning: Unknown pytest.mark.celery*"
         )
+
+
+@pytest.fixture
+def fixed_import(mocker):
+    def import_module(module, package=None):
```
thedrow: Why are we monkeypatching here if the code of `import_module()` is exactly the same as the monkey patched version?

t/unit/contrib/test_pytest.py
```diff
@@ -32,3 +34,37 @@ def test_noop():
...
+
+
+def test_pytest_import_passes(celery_app, celery_worker):
+    task = celery_app.send_task("t.unit.contrib.proj.foo.bar")
+    task.get()
+
+
+def test_pytest_second_import_fails(celery_app, celery_worker):
+    task = celery_app.send_task("t.unit.contrib.proj.foo.bar")
+    with pytest.raises(Exception):
```
thedrow: Can we get more specific with the exception type/message?

t/unit/contrib/proj/foo.py
chbndrhnns: When running `pytest t/unit/contrib/test_pytest.py`, I see this error:

```python

    from celery import Celery, shared_task
>   from xyzzy import plugh  # noqa
E   ModuleNotFoundError: No module named 'xyzzy'
```

I understand it might be working in the CI jobs, so I will comment it in, again.

celery/loaders/base.py
chbndrhnns: I do not have a good explanation *why* except that it works for me. You can see from the test cases that it does not work *without* reloading. I would appreciate any hint why this issue occurs is at all.

t/unit/contrib/test_pytest.py
chbndrhnns: Replaced with `KeyError`

t/unit/contrib/test_pytest.py
chbndrhnns: In the first place, I wanted to show that the issue occurs and this is why I added these test cases. I will extract the failing tests to the issue.

chbndrhnns:
> Please include a description of your solution to #5963.
> Describe the problem and ensure that it will close the relevant issue.
> How it fixes the problem? Have you considered other solutions?

I added this comment for now: 
https://github.com/celery/celery/issues/5963#issuecomment-588418429

chbndrhnns:
For me, it is hard to understand why the CI jobs here fail. Is someone able to help me?

auvipy:
that might be not related to this PR?

auvipy:
can you recheck the tests?

chbndrhnns:
I added another guard clause to make the tests pass.

However, teardown fails with the message below. I have no idea where I could have modified the logging level in my test.

```
t/unit/contrib/test_pytest.py:47 (test_pytest_second_with_new_app_fixtures_also_passes)
request = <SubRequest 'sanity_logging_side_effects' for <Function test_pytest_second_with_new_app_fixtures_also_passes>>

    @pytest.fixture(autouse=True)
        from _pytest.logging import LogCaptureHandler
        root = logging.getLogger()
        rootlevel = root.level
        roothandlers = [
            x for x in root.handlers if not isinstance(x, LogCaptureHandler)]
    
        yield
    
        this = request.node.name
>           raise RuntimeError(CASE_LOG_LEVEL_EFFECT.format(this))
E           RuntimeError: Test test_pytest_second_with_new_app_fixtures_also_passes modified the level of the root logger

../conftest.py:244: RuntimeError
```

auvipy:
―――――――――――――――― ERROR at teardown of test_pytest_import_passes ――――――――――――――――

request = <SubRequest 'sanity_logging_side_effects' for <Function test_pytest_import_passes>>

>   ???

E   RuntimeError: Test test_pytest_import_passes modified the level of the root logger

chbndrhnns:
> ―――――――――――――――― ERROR at teardown of test_pytest_import_passes ――――――――――――――――
> 
> request = <SubRequest 'sanity_logging_side_effects' for >
> 
> > ???
> 
> E RuntimeError: Test test_pytest_import_passes modified the level of the root logger

Could you please help me understand where in my test the logger gets modified? The tests are as simple as:

```python
def test_pytest_import_passes(celery_app, celery_worker):
    task = celery_app.send_task("t.unit.contrib.proj.foo.bar")
    task.get()


def test_pytest_second_with_new_app_fixtures_also_passes(celery_app, celery_worker):
    task = celery_app.send_task("t.unit.contrib.proj.foo.bar")
    task.get()
```

auvipy:
@georgepsarakis

t/unit/contrib/proj/foo.py
```diff
@@ -1,7 +1,7 @@
 from __future__ import absolute_import, unicode_literals
 
 from celery import Celery, shared_task
-from xyzzy import plugh  # noqa
+from .xyzzy import plugh  # noqa
```
georgepsarakis: Are we sure this is issue is not caused by something specific to the environment? Wouldn't we observe issues in Travis CI as well?

auvipy (changes requested):
lets check on new CI

PR #5972, opened 2020-02-27
Use configured db schema also for sequences

soldag:
The configured schema is currently not taken into account when defining sequence generators for id fields.

Followup of #5910


soldag:
@auvipy I already adapted the `test_table_schema_config` test case accordingly

auvipy:
sorry for overlooking :p

Issue #4999, opened 2018-08-22
AsyncResult.get()  stuck in high cpu usage


monsterxx03:
celery  4.2.1
gevent  1.3.5 
redis-py 2.10.6

With gevent.monkey.patch_all(), when try to get result of a delay executed task(shutdown worker or make task in long pending task queue), result.get() will cause very high cpu usage on the producer side.

Tested on python2.7 and python3.6, both have this issue.

Test script:

```
import gevent
import gevent.monkey
gevent.monkey.patch_all()

import time
from celery import Celery

app = Celery()
app.conf.update(BROKER_URL='redis://localhost:6379/1',
        CELERY_RESULT_BACKEND='redis://localhost:6379/2')

@app.task()
def dummy():
    return '1'

ret = dummy.delay()
ret.get()
```
strace -p {pid}, get:

```
getpid()                                = 10484
epoll_wait(3, [], 64, 1)                = 0
getpid()                                = 10484
epoll_wait(3, [], 64, 1)                = 0
getpid()                                = 10484
epoll_wait(3, [], 64, 1)                = 0
getpid()                                = 10484
epoll_wait(3, [], 64, 1)                = 0
getpid()                                = 10484
epoll_wait(3, [], 64, 1)                = 0
getpid()                                = 10484
epoll_wait(3, [], 64, 1)                = 0
getpid()                                = 10484
epoll_wait(3, [], 64, 1)                = 0
getpid()                                = 10484
epoll_wait(3, [], 64, 1)                = 0
getpid()                                = 10484
epoll_wait(3, [], 64, 1)                = 0
getpid()                                = 10484
epoll_wait(3, [], 64, 1)                = 0
getpid()                                = 10484
epoll_wait(3, [], 64, 1)                = 0
getpid()                                = 10484
epoll_wait(3, [], 64, 1)                = 0
getpid()                                = 10484
epoll_wait(3, [], 64, 1)                = 0
getpid()                                = 10484
```
It's flushing very quickly, seems gevent is checking something actively, after debuging, I find `ret.get()`  stucked in  `https://github.com/celery/celery/blob/master/celery/backends/asynchronous.py#L99`

If  change it to `sleep(0.1)`,  high cpu usage go way. Not sure what sleep(0) does here.

thedrow:
`sleep(0)` tells gevent that it is allowed to yield to other greenlets at this point.
What do you think the fix should be?

SharpEdgeMarshall:
[...]

PR #5974, opened 2020-02-27
Use gevent and eventlet wait() functions to remove busy-wait

tildedave:
Fixes issue #4999.

## Description

Calling AsyncResult.get() in a gevent context would cause the async Drainer to repeatedly call wait_for until the result was completed (in a CPU-consuming spin loop).  I've updated the code to have a specific implementation for gevent and eventlet that will cause wait_for to only return every "timeout" # of seconds, rather than returning immediately.

## Things I'd Like Feedback On

* The way I did this for the Eventlet interface was to rely on the private _exit_event member of the GreenThread instance; to do this without relying on a private member would require some additional changes to the backend Drainer interface so that we could wait for an eventlet-specific event in wait_for().  I can do this, just wanted to get some feedback before jumping in.
* Another option for this diff would be to use the threading.Event on the greenlet and call `self._shutdown.wait(timeout=timeout)` in the `wait_for` method of the Greenlet.  Pros are that would avoid needing to do different things between gevent/eventlet, but I'm not sure if that will behave correctly if the threading library isn't monkeypatched.  (EDIT: looks like it doesn't, pushed a test case)

### DIFF
--- a/celery/backends/asynchronous.py
+++ b/celery/backends/asynchronous.py
@@ -93,11 +93,6 @@ class greenletDrainer(Drainer):
         self._stopped.set()
         self._shutdown.wait(THREAD_TIMEOUT_MAX)
 
-    def wait_for(self, p, wait, timeout=None):
-        self.start()
-        if not p.ready:
-            sleep(0)
-
 
 @register_drainer('eventlet')
 class eventletDrainer(greenletDrainer):
@@ -107,6 +102,11 @@ class eventletDrainer(greenletDrainer):
         from eventlet import spawn
         return spawn
 
+    def wait_for(self, p, wait, timeout=None):
+        self.start()
+        if not p.ready:
+            self._g._exit_event.wait(timeout=timeout)
+
 
 @register_drainer('gevent')
 class geventDrainer(greenletDrainer):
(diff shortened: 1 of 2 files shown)


celery/backends/asynchronous.py
```diff
@@ -107,6 +102,11 @@ def spawn(self):
         from eventlet import spawn
         return spawn
 
+    def wait_for(self, p, wait, timeout=None):
```
tildedave: The `wait` argument is only used for the non-greenlet (non-gevent, non-eventlet) flow.  Sort of confusing to have it around here.

tildedave:
I pushed some new unit tests (test/backends/test_asynchronous.py) that should cover these scenarios for gevent, eventlet, and default.  Let me know if there's anything different I should do there.  Thanks for looking if you get a chance.

celery/backends/asynchronous.py
```diff
@@ -93,28 +92,36 @@ def stop(self):
...
 class geventDrainer(greenletDrainer):
 
-    @cached_property
-    def spawn(self):
-        from gevent import spawn
-        return spawn
+    def spawn(self, func):
+        from gevent import spawn, sleep
+        g = spawn(func)
+        sleep(0)
```
tildedave: I found that I needed to sleep(0) in scenarios where we weren't monkeypatching gevent to yield control back to the calling thread.

tildedave:
@auvipy Thanks - looks like importing eventlet was doing some DNS monkeypatching which broke something else.  Pushed a fix.

tildedave:
Thanks @auvipy !

auvipy:
thanks for your work! would appreciate if you tell your company to donate for celery https://opencollective.com/celery

PR #6005, opened 2020-03-18
Add uptime to the stats inspect command

jpays:
## Description

Being able to see the uptime of a worker remotely would be a nice addition.
This code adds a "uptime" key to the stats inspect command.

Kind regards
Jerome

### DIFF
--- a/celery/worker/worker.py
+++ b/celery/worker/worker.py
@@ -17,6 +17,8 @@ from __future__ import absolute_import, unicode_literals
 import os
 import sys
 
+from datetime import datetime
+
 from billiard import cpu_count
 from kombu.utils.compat import detect_environment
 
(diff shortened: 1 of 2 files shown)


celery/worker/worker.py
```diff
@@ -296,9 +299,11 @@ def _maybe_reload_module(self, module, force_reload=False, reloader=None):
             return reload_from_cwd(sys.modules[module], reloader)
 
     def info(self):
+        uptime = datetime.utcnow() - self.startup_time
```
drsantos20: hi @jpays there is some existing test for this function?

jpays:
Hi, 

I found some tests of the function "stats", but they do not seem to check the format of the return structure. 

Hence, adding a new field does not change the existing tests.

Sorry I'm not too familiar yet with the unit test section

Jerome

PR #6006, opened 2020-03-19
Fix backend utf-8 encoding in s3 backend

shuklaabhi:
Celery backend uses utf-8 to deserialize results,
which would fail for some serializations like pickle.

## Description
We support multiple serialization formats in the backend like JSON, pickle, msgpack and yaml.
UTF-8 decoding is hardcoded in s3 irrespective of serialization format defined in the configuration, which would lead to error in cases someone uses pickle or other non-utf-8 formats.

### DIFF
--- a/celery/backends/s3.py
+++ b/celery/backends/s3.py
@@ -61,7 +61,8 @@ class S3Backend(KeyValueStoreBackend):
         s3_object = self._get_s3_object(key)
         try:
             s3_object.load()
-            return s3_object.get()['Body'].read().decode('utf-8')
+            data = s3_object.get()['Body'].read()
+            return data if self.content_encoding == 'binary' else data.decode('utf-8')
         except botocore.exceptions.ClientError as error:
             if error.response['Error']['Code'] == "404":
                 return None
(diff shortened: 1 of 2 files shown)


shuklaabhi:
> would you mind adding a failing test case for this? to check that this change wont regress

Sure, I. would add testcases by tomorrow.

shuklaabhi:
def get(self, key): in backends return bytes in cases of binary serialization, and string in case of nonbinary serialization. Since pickle bytes can be decoded to UTF-8 encoding, which makes sense. 
You can check the same from the Redis test cases I have put down below, while s3 getObject return bytes only, so in order the follow the same contract I am using conditional encoding. 

```
    def test_get_set_bin_result(self):
        self.app.conf.result_serializer = 'pickle'
        self.b = self.Backend(expires=512, app=self.app)
        self.b.store_result('foo', 'bar', states.SUCCESS)
        value = self.b.get(self.b.get_key_for_task('foo'))
        print(type(value)) ## A byte array

        value = self.b.get_result('foo')
        assert value == 'baar'

    def test_get_set_bin_result2(self):
        self.b = self.Backend(expires=512, app=self.app)
        self.b.store_result('foo', 'bar', states.SUCCESS)
        value = self.b.get(self.b.get_key_for_task('foo'))
        print(type(value)) ## str

        value = self.b.get_result('foo')
        assert value == 'baar'
```

shuklaabhi:
@auvipy It's an issue with pytest-sugar. https://github.com/Teemu/pytest-sugar/issues/187

Issue #6023, opened 2020-04-06
Memory leak  in Exception Task with Prefork(worker)


heedong-jung:
# Checklist

- [v ] I have verified that the issue exists against the `master` branch of Celery.
- [ ] This has already been asked to the [discussion group](https://groups.google.com/forum/#!forum/celery-users) first.
- [v ] I have read the relevant section in the
  [contribution guide](http://docs.celeryproject.org/en/latest/contributing.html#other-bugs)
  on reporting bugs.
- [ v] I have checked the [issues list](https://github.com/celery/celery/issues?q=is%3Aissue+label%3A%22Issue+Type%3A+Bug+Report%22+-label%3A%22Category%3A+Documentation%22)
  for similar or identical bug reports.
- [ v] I have checked the [pull requests list](https://github.com/celery/celery/pulls?q=is%3Apr+label%3A%22PR+Type%3A+Bugfix%22+-label%3A%22Category%3A+Documentation%22)
  for existing proposed fixes.
- [v ] I have checked the [commit log](https://github.com/celery/celery/commits/master)
  to find out if the bug was already fixed in the master branch.
- [ ] I have included all related issues and possible duplicate issues
  in this issue (If there are none, check this box anyway).

## Mandatory Debugging Information

- [ v] I have included the output of ``celery -A proj report`` in the issue.
    (if you are not able to do this, then at least specify the Celery
     version affected).
- [ v] I have verified that the issue exists against the `master` branch of Celery.
- [ v] I have included the contents of ``pip freeze`` in the issue.
- [ v] I have included all the versions of all the external dependencies required
  to reproduce this bug.

## Optional Debugging Information

- [v ] I have tried reproducing the issue on more than one Python version
  and/or implementation.
- [ ] I have tried reproducing the issue on more than one message broker and/or
  result backend.
- [ ] I have tried reproducing the issue on more than one version of the message
  broker and/or result backend.
- [ ] I have tried reproducing the issue on more than one operating system.
- [ ] I have tried reproducing the issue on more than one workers pool.
- [ ] I have tried reproducing the issue with autoscaling, retries,
  ETA/Countdown & rate limits disabled.
- [ ] I have tried reproducing the issue after downgrading
  and/or upgrading Celery and its dependencies.

#### Related Issues

- 
#### Possible Duplicates

- None

## Environment & Settings

**Celery version**: 4.3.0 , 4.4.2

<details>
<summary><b><code>celery report</code> Output:</b></summary>
<p>

[...]

PR #6024, opened 2020-04-06
Clean TraceBack to reduce memory leaks for exception task

heedong-jung:

## Description

Fixes: #6023

It seems that a frame leak is occurring in the use of the e value returned by 'Execption as e'. Del tb is not enough. So we clean up the traceback in the last part that can be explicit.

### DIFF
--- a/celery/backends/base.py
+++ b/celery/backends/base.py
@@ -256,13 +256,18 @@ class Backend(object):
             self.mark_as_failure(task_id, exc, exception_info.traceback)
             return exception_info
         finally:
-            while tb is not None:
-                try:
-                    tb.tb_frame.clear()
-                except RuntimeError:
-                    # Ignore the exception raised if the frame is still executing.
-                    pass
-                tb = tb.tb_next
+            if sys.version_info >= (3, 4, 0):
+                while tb is not None:
+                    try:
+                        tb.tb_frame.clear()
+                    except RuntimeError:
+                        # Ignore the exception raised if the frame is still executing.
+                        pass
+                    tb = tb.tb_next
+
+            elif (2, 7, 0) <= sys.version_info < (3, 0, 0):
+                sys.exc_clear()
+
             del tb
 
     def prepare_exception(self, exc, serializer=None):
--- a/celery/backends/base.py
+++ b/celery/backends/base.py
@@ -260,6 +260,7 @@ class Backend(object):
                 while tb is not None:
                     try:
                         tb.tb_frame.clear()
+                        tb.tb_frame.f_locals
                     except RuntimeError:
                         # Ignore the exception raised if the frame is still executing.
                         pass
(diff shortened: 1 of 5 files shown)


celery/backends/base.py
```diff
@@ -256,6 +256,19 @@ def fail_from_current_stack(self, task_id, exc=None):
             self.mark_as_failure(task_id, exc, exception_info.traceback)
             return exception_info
         finally:
+            if sys.version_info >= (3, 4, 0):
```
auvipy, line 259: 3.5?

t/unit/tasks/test_trace.py
```diff
@@ -150,23 +152,79 @@ def add(x, y):
...
+            frame_str_temp = str(inspect.currentframe().__repr__)
+            frame_list.append(frame_str_temp)
+            local_value = 1214
+            raise KeyError('foo')
+        try:
+            raise_dummy()
+        except KeyError as exc:
+            traceback_clear(exc)
+
+            if sys.version_info >= (3, 4, 0):
```
auvipy, line 170: 3.5

t/unit/tasks/test_trace.py
```diff
@@ -150,23 +152,79 @@ def add(x, y):
...
+                    tb_ = tb_.tb_next
+            elif (2, 7, 0) <= sys.version_info < (3, 0, 0):
+                sys.exc_clear.assert_called()
+
+        try:
+            raise_dummy()
+        except KeyError as exc:
+            traceback_clear(str(exc))
+
+            if sys.version_info >= (3, 4, 0):
```
auvipy, line 198: 3.4

celery/backends/base.py
heedong-jung: This property is from 3.4 and above.  ok  it will change this to 3.5

celery/backends/base.py
```diff
@@ -256,6 +256,19 @@ def fail_from_current_stack(self, task_id, exc=None):
             self.mark_as_failure(task_id, exc, exception_info.traceback)
             return exception_info
         finally:
+            if sys.version_info >= (3, 5, 0):
+                while tb is not None:
+                    try:
+                        tb.tb_frame.clear()
```
asfaltboy, line 262: Sorry I'm a little "late to the party" here, but if I may, I want to forward a couple of question for @heedong-jung (perhaps for a future improvement):

1. Could we use [`traceback.clear_frames()`][1], introduced in python 3.4, to avoid manually iterating over and clearing the frame data above?

2. Are we not able to add an integration test, to more explicitly check for memory leaks over task execution?

3. Would it be possible to reuse the code in `traceback_clear`, eeither by allowing passing a traceback object, or perhaps by splitting into separate "extract traceback from exception" and "clear traceback frames"?

4. Do you believe it's related to (or resolves) issue #6058 ?

I believe that with the above information, I can start working on improvements where needed.

[1]: https://docs.python.org/3.5/library/traceback.html#traceback.clear_frames

auvipy: you can come with a PR with appropriate tests

heedong-jung: clear_frames() is not suitable for solving this problem.

It is related to the memory leak you mentioned.

PR #6027, opened 2020-04-07
Add internal_error signal

ldragan:

## Description

There is no special signal for an out of body error which can be the
result of a bad result backend.

We want to add a new signal `task_internal_error` which will call the signal handler in case of an out of body error.

### DIFF
--- a/celery/app/trace.py
+++ b/celery/app/trace.py
@@ -522,6 +522,7 @@ def build_tracer(name, task, loader=None, hostname=None, store_errors=True,
         except MemoryError:
             raise
         except Exception as exc:
+            _signal_internal_error(task, uuid, args, kwargs, request, exc)
             if eager:
                 raise
             R = report_internal_error(task, exc)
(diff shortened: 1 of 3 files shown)


celery/app/trace.py
```diff
@@ -540,9 +541,31 @@ def trace_task(task, uuid, args, kwargs, request=None, **opts):
             task.__trace__ = build_tracer(task.name, task, **opts)
         return task.__trace__(uuid, args, kwargs, request)
     except Exception as exc:
+        _signal_internal_error(task, uuid, args, kwargs, request, exc)
         return trace_ok_t(report_internal_error(task, exc), None, 0.0, None)
 
 
+def _signal_internal_error(task, uuid, args, kwargs, request, exc):
```
auvipy, line 548: this need proper test coverage. and entry to changelog

thedrow:
It's a shame you didn't use a different branch other than master.
I can't push the required changes directly to your branch because of that.

Please create a branch on your fork when contributing in the future.

PR #6049, opened 2020-04-20
Add task_internal_error signal

thedrow:

## Description

There is no special signal for an out of body error which can be the
result of a bad result backend.

We want to add a new signal task_internal_error which will call the signal handler in case of an out of body error.

Superceeds #6027


thedrow:
@auvipy Where is the new changelog?

auvipy:
https://github.com/celery/celery/blob/master/Changelog.rst

thedrow:
I just rebased.

PR #6032, opened 2020-04-09
Avoid PyModulelevel, deprecated in Sphinx 4

lpsinger:
Use `PyFunction` instead of `PyModulelevel` to avoid this deprecation warning:

    RemovedInSphinx40Warning: PyModulelevel is deprecated.
    Please check the implementation of
    <class 'celery.contrib.sphinx.TaskDirective'>

This replacement is one of the options listed in the Sphinx docs (https://www.sphinx-doc.org/en/master/extdev/deprecated.html).

PR #6051, opened 2020-04-20
Fix the build and all documentation warnings.

thedrow:

## Description

I finally upgraded our theme to 2.0.
As a result we've upgraded Sphinx to 2.0.
Work to upgrade Sphinx to 3.0 will proceed in a different PR.

This upgrade also fixes our build issues caused by #6032.
We don't support Sphinx 1.x as a result of that patch.

I've also included the missing 4.3 changelog to our history.


thedrow:
Building the docs with Sphinx 3 raises some more warnings.
I've included a quick fix which allows us to build with both versions but for now I'm not going to upgrade.

thedrow:
For some reason the unit tests on Python 2 still fail.
However, this PR still leaves the build in a better state than it was before.

Please review and merge ASAP.

mchataigner:
Hi, I dig a bit on why python 2.7 tests were failing and it's because in the `tox.ini`, there is a require on doc.txt:
```
-r{toxinidir}/requirements/docs.txt
```
but as sphinx has been updated to 2+, there is no version available  for python 2.7 hence it does not download all requirements or maybe the virtualenv is corrupted.

I tried locally to comment this line out of `tox.ini` file. Now tests are running but there are some test errors due to:
```
ValueError: unsupported pickle protocol: 4
```
So I guess pickle for Python 2.7 does not support this protocol.

Easy fix => remove support for python 2.7
Longer fix => revert changes related to pickle protocol update

mchataigner:
and using Python 3.8, I get an error as well
```
                                                                                                                                                                      
self = <celery.contrib.testing.app.Trap object at 0x7f10c1b3ca00>                                                                                                     
name = '__func__'                                                                                                                                                     
                                                                                                                                                                      
    def __getattr__(self, name):                                                                                                                                      
        # Workaround to allow unittest.mock to patch this object                                                                                                      
        # in Python 3.8 and above.
        if name == '_is_coroutine':
            return None
        print(name)
>       raise RuntimeError('Test depends on current_app')
E       RuntimeError: Test depends on current_app

celery/contrib/testing/app.py:38: RuntimeError
```
on `t/unit/security/test_security.py::test_security.test_setup_security__default_app`

auvipy:
> Hi, I dig a bit on why python 2.7 tests were failing and it's because in the `tox.ini`, there is a require on doc.txt:
> 
> ```
> -r{toxinidir}/requirements/docs.txt
> ```
> 
> but as sphinx has been updated to 2+, there is no version available for python 2.7 hence it does not download all requirements or maybe the virtualenv is corrupted.
> 
> I tried locally to comment this line out of `tox.ini` file. Now tests are running but there are some test errors due to:
> 
> ```
> ValueError: unsupported pickle protocol: 4
> ```
> 
> So I guess pickle for Python 2.7 does not support this protocol.
> 
> Easy fix => remove support for python 2.7
> Longer fix => revert changes related to pickle protocol update

will drop python 2.7 in celery 4.5.

Issue #6017, opened 2020-03-31
Celery Multi creates pid and log files in the wrong directory


maxmalysh:
**Here is what [docs](http://docs.celeryproject.org/en/latest/userguide/daemonizing.html) say**:

- `CELERYD_PID_FILE` — Full path to the PID file. Default is `/var/run/celery/%n.pid`
- `CELERYD_LOG_FILE` — Full path to the worker log file. Default is `/var/log/celery/%n%I.log` Note: Using %I is important when using the prefork pool as having multiple processes share the same log file will lead to race conditions.

**In reality**, `celery multi` creates logs and pid files in the working directory instead of `/var/run` and `/var/log`:
```
--logfile=default%I.log --pidfile=default.pid
```

**My systemd service**:
```
# /etc/systemd/system/celery.service

[Unit]
Description=Celery Service
After=network.target

[Service]
Type=forking
User=myuser
Restart=always
EnvironmentFile=-/etc/conf.d/celery
WorkingDirectory=/home/myuser/myproject

ExecStart=/bin/bash -c 'source /etc/environment; \
  /home/myuser/venv/bin/celery multi start ${CELERYD_NODES} \
  -A ${CELERY_APP} ${CELERYD_OPTS}'

ExecStop=/bin/bash -c 'source /etc/environment; \
  /home/myuser/venv/bin/celery multi stopwait ${CELERYD_NODES}'

ExecReload=/bin/bash -c 'source /etc/environment; \
  /home/myuser/venv/bin/celery multi restart ${CELERYD_NODES} \
  -A ${CELERY_APP} --loglevel="${CELERYD_LOG_LEVEL}" ${CELERYD_OPTS}'

[Install]
WantedBy=multi-user.target
```

**My `/etc/conf.d/celery`:**
```
# /etc/conf.d/celery
CELERYD_NODES="default slow"
CELERY_APP="pool.celery:app"
CELERYD_MULTI="multi"
CELERYD_OPTS="-Q:1 celery -Q:2 slow --time-limit=15 --concurrency=1 -O fair -E"
```

**Systemd status report:**
```
root@pool1:~# sudo systemctl status celery
● celery.service - Celery Service
   Loaded: loaded (/etc/systemd/system/celery.service; enabled; vendor preset: enabled)
   Active: active (running) since Tue 2020-03-31 13:38:02 UTC; 1min 23s ago
[...]

PR #6055, opened 2020-04-22
[Fixes Issue #6017] --> Added Multi default logfiles and pidfiles paths

hanchau:
[Description]:
--> Changed the default paths for log files & pid files to be '/var/log/celery' and '/var/run/celery'
--> Handled by creating the respective paths if not exist.
--> Used os.makedir(path,if_exists=True)

[Unit Test Added]:
--> .travis.yml - config updated with 'before install'.
--> t/unit/apps/test_multi.py - Changed the default log files & pid files paths wherever required.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -81,6 +81,10 @@ matrix:
     env: TOXENV=pypy3
 
 before_install:
+    - sudo mkdir -p /var/log/celery
+    - sudo mkdir -p /var/run/celery
+    - sudo chown travis /var/log/celery
+    - sudo chown travis /var/run/celery
     - sudo apt install libcurl4-openssl-dev libssl-dev gnutls-dev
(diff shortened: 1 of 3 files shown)


hanchau:
1. The changes are just setting the default logs and pid files paths
2. If user has configured the paths in multi config then
   - it will be backward compatible
3. if not then
   - the old logs and pid files would be in /var/... and the new logs and pid files would be in current 
      directory.
4. Conclusion: It won't break celery but if the user is processing logs then he/she has to reconfigure his/her application.

mchataigner:
This has a side effect of requiring write access /var/run during local tests.
Could you update `t/unit/apps/test_multi.py` unit tests so that it does not try to create files and directories in this system protected directory?
Thanks.

auvipy:
would you mind sending the fix?

PR #6070, opened 2020-05-03
Allow relative paths in the filesystem backend

watercrossing:
## Description

Celery does not currently allow relative paths for the filesystem backend, as urls are required to start with a `/`. This removes this requirement.

### DIFF
--- a/celery/backends/filesystem.py
+++ b/celery/backends/filesystem.py
@@ -61,10 +61,10 @@ class FilesystemBackend(KeyValueStoreBackend):
     def _find_path(self, url):
         if not url:
             raise ImproperlyConfigured(E_NO_PATH_SET)
-        if url.startswith('file://'):
-            return url[7:]
         if url.startswith('file://localhost/'):
             return url[16:]
+        if url.startswith('file://'):
+            return url[7:]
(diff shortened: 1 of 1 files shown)


celery/backends/filesystem.py
```diff
@@ -61,10 +61,10 @@ def __init__(self, url=None, open=open, unlink=os.unlink, sep=os.sep,
     def _find_path(self, url):
         if not url:
             raise ImproperlyConfigured(E_NO_PATH_SET)
-        if url.startswith('file:///'):
-            return url[7:]
         if url.startswith('file://localhost/'):
             return url[16:]
+        if url.startswith('file://'):
```
drsantos20, line 66: Im seeing just a ordering of if statements change, do we have any tests to cover this @watercrossing ?

watercrossing: I have not had time to look into celery's test setup. The re-ordering of if statements is necessary to preserve current behaviour; otherwise any `url` starting with `'file://localhost/' would not get the `'localhost'` striped.

auvipy: please also add tests for the change :)

PR #6079, opened 2020-05-09
🐛 Correctly handle configuring the serializer for always_eager mode.

coderanger:
options['serializer'] will always exist, because it is initialized from an mattrgetter. Even if unset, it will be present in the options dict with a value of None.


coderanger:
Test failures look unrelated?

PR #6081, opened 2020-05-09
Remove doubling of prefetch_count increase when prefetch_multiplier gt 1

nadflinn:

## Description

This is a follow up to: https://github.com/celery/celery/pull/6069

In the linked PR one of the changes was to update the prefetch_count (number of tasks a worker pre-fetches or "claims") when the autoscale range is adjusted.  In order to calculate the amount of the increase the diff in processes was multiplied by the consumer prefetch_multiplier.  I overlooked that the value passed to `consumer._update_prefetch_count` is [subsequently multiplied by the consumer prefetch_multiplier](https://github.com/celery/celery/blob/master/celery/worker/consumer/consumer.py#L264-L267) so we end up with a doubling of the intended increase of prefetch_count.

The implications of this aren't dramatic because for those who set prefetch multiplier = 1 to avoid prefetching, there would still be no prefetching because there is no double counting when the factor is 1.  For those that did want prefetching AND adjusted their autoscaling settings via remote control command then their prefetch_count would be greater than intended.  This PR will correct this.

### DIFF
--- a/celery/worker/autoscale.py
+++ b/celery/worker/autoscale.py
@@ -153,7 +153,7 @@ class Autoscaler(bgThread):
         diff = new_max - self.max_concurrency
         if diff:
             self.worker.consumer._update_prefetch_count(
-                diff * self.worker.consumer.prefetch_multiplier
+                diff
             )
 
     def info(self):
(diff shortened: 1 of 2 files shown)


nadflinn:
@auvipy a quick follow up fix for [previous autoscale PR](https://github.com/celery/celery/pull/6069), when you have a chance.  Thank you!

auvipy:
it would be great if you could find out some other edge cases of autoscalar :)

PR #5894, opened 2019-12-29
Add encode to meta task in base.py

Anakael:
Fix bug with impossibility to load None from task meta.

Issue #5663, opened 2019-08-01
Completed tasks restore on shutdown causing duplicates for recursive tasks 


gmjosack:
# Checklist

- [x] I have read the relevant section in the
  [contribution guide](http://docs.celeryproject.org/en/latest/contributing.html#other-bugs)
  on reporting bugs.
- [x] I have checked the [issues list](https://github.com/celery/celery/issues?q=is%3Aissue+label%3A%22Issue+Type%3A+Bug+Report%22+-label%3A%22Category%3A+Documentation%22)
  for similar or identical bug reports.
- [x] I have checked the [pull requests list](https://github.com/celery/celery/pulls?q=is%3Apr+label%3A%22PR+Type%3A+Bugfix%22+-label%3A%22Category%3A+Documentation%22)
  for existing proposed fixes.
- [x] I have checked the [commit log](https://github.com/celery/celery/commits/master)
  to find out if the bug was already fixed in the master branch.
- [x] I have included all related issues and possible duplicate issues
  in this issue (If there are none, check this box anyway).

## Mandatory Debugging Information

- [x] I have included the output of ``celery -A proj report`` in the issue.
    (if you are not able to do this, then at least specify the Celery
     version affected).
- [x] I have verified that the issue exists against the `master` branch of Celery.
- [x] I have included the contents of ``pip freeze`` in the issue.
- [x] I have included all the versions of all the external dependencies required
  to reproduce this bug.

## Optional Debugging Information

- [ ] I have tried reproducing the issue on more than one Python version
  and/or implementation.
- [ ] I have tried reproducing the issue on more than one message broker and/or
  result backend.
- [ ] I have tried reproducing the issue on more than one version of the message
  broker and/or result backend.
- [ ] I have tried reproducing the issue on more than one operating system.
- [ ] I have tried reproducing the issue on more than one workers pool.
- [ ] I have tried reproducing the issue with autoscaling, retries,
  ETA/Countdown & rate limits disabled.
- [x] I have tried reproducing the issue after downgrading
  and/or upgrading Celery and its dependencies.

#### Related Issues

- None

#### Possible Duplicates

- None

## Environment & Settings

**4.3.0 (rhubarb)**:

<details>
<summary><b><code>celery report</code> Output:</b></summary>
<p>

```
$ celery -A tasks report

software -> celery:4.3.0 (rhubarb) kombu:4.6.3 py:2.7.15rc1
            billiard:3.6.0.0 redis:3.2.1
[...]

PR #6090, opened 2020-05-13
Fix potential race condition in distributed mode leading to duplicated tasks

mchataigner:
## Description
In some circumstances like a network partitioning, some tasks might
be duplicated. Sometimes, this lead to a race condition where a lost
task overwrites the result of the last successful task in the backend.

In order to avoid this race condition we prevent updating the result if
it's already in successful state.

This fix has been done for KV backends only and therefore won't work
with other backends.

### DIFF
--- a/celery/backends/base.py
+++ b/celery/backends/base.py
@@ -768,6 +768,17 @@ class BaseKeyValueStoreBackend(Backend):
                                      traceback=traceback, request=request)
         meta['task_id'] = bytes_to_str(task_id)
 
+        # Retrieve metadata from the backend, if the status
+        # is a success then we ignore any following update to the state.
+        # This solves a task deduplication issue because of network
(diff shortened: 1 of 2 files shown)


auvipy:
any related /existing issues to link?

auvipy:
if you can point related issues/pr it would be really helpful

mchataigner:
It may be related to https://github.com/celery/celery/issues/5663 but I'm not sure it's the correct fix for this issue.
This is mainly a workaround when using KeyValueStoreBackend based backend.

It will avoid workers storing any state if the backend already his the tasks marked as successful.
Now the question is how two workers ended up working on the same task. I did not dig deep enough on how shutdown and fail-over work so I don't know.

auvipy:
can you please review this PR as well https://github.com/celery/celery/pull/5894/files?

PR #6096, opened 2020-05-14
Use ubuntu bionic for python <= 3.6 in Travis

mchataigner:


auvipy:
can you check here why python 2.7 is failing? https://github.com/celery/celery/pull/6095

mchataigner:
By curiosity, why do you keep python 2.7 as it's EOL since January, 1st 2020?

auvipy:
will drop it in celery 4.5

auvipy:
should we try xenial for python 2.7?

mchataigner:
Pytest > 4.6.x does not support python < 3.5
But here the problem is about wrong arguments passed to pytest.
Command specified for unit tests in tox.ini seems not compatible with python 2.7.
I don't know which version of pytest it's using though.

It was working with bionic and python 2.7 on this patch https://travis-ci.org/github/celery/celery/builds/667597080 so it's not related to distribution.

The first build with error on python 2.7 was https://travis-ci.org/github/celery/celery/builds/677296325

.travis.yml
```diff
@@ -22,6 +22,18 @@ env:
 matrix:
   fast_finish: true
   include:
+  - python: 2.7
+    stage: test
```
auvipy, line 26: stage is not needed

auvipy:
apparently this PR is not needed as only python 2.7 parsing command failing on master

PR #6100, opened 2020-05-15
drop py33 check

auvipy:


auvipy:
will come with a combined effort of cleaning these stuff

PR #6102, opened 2020-05-15
do not load docs.txt requirements for python 2.7

mchataigner:
As it requires Sphinx >= 2.0.0 and there is no such version compatible
with python 2.7


tox.ini
```diff
@@ -14,10 +14,10 @@ envlist =
 deps=
     -r{toxinidir}/requirements/default.txt
     -r{toxinidir}/requirements/test.txt
-    -r{toxinidir}/requirements/docs.txt
```
auvipy, line 17: good catch!

mchataigner:
Cassandra integration test is fixed.
flake8 and pydocstyle linter fixed.
I need help for the remaining linter configcheck.

mchataigner:
Cassandra integration test is fixed.
flake8, configcheck and pydocstyle linter fixed.
Travis CI green!

PR #6104, opened 2020-05-15
Fix windows build

mchataigner:
those libraries depends on native libraries libcouchbase and libmemcached
that are not installed on Appveyor.
As only unit tests runs on Appveyor, it should be fine


mchataigner:
Only python 2.7 on windows left to fix.
It fails due to an error when installing pycurl. It requires curl-dir option from pip.
For some reason it works on python 3+.

auvipy:
we can drop python 2.7 from win build :)

mchataigner:
I don't know why codecov is reporting 0%…

auvipy:
probably not related to this PR

Issue #4661, opened 2018-04-12
Task.retry() always raises RuntimeError if task_always_eager and task_eager_propagates are set


lpsinger:
`Task.retry()` always raises a `RuntimeError` if `task_always_eager` and `task_eager_propagates` are set to `True`. This is counterintuitive. I would expect this example to work:

```python
from celery import Celery

app = Celery(__name__,
             config_source=dict(task_always_eager=True,
                                task_eager_propagates=True))

@app.task(bind=True, default_retry_delay=1, max_retries=None, shared=False)
def foo(self, i):
    if i <= 0:
        return 'Task succeeded!'
    else:
        self.retry((i - 1,))


foo.delay(3)
```

But instead it raises this error:

```
Traceback (most recent call last):
    ...
    raise RuntimeError(E_WOULDBLOCK)
RuntimeError: Never call result.get() within a task!
See http://docs.celeryq.org/en/latest/userguide/tasks.html#task-synchronous-subtasks
```

This can be reproduced on master and with 4.2.0rc2.

javiermatos:
I upgraded celery from version 4.1.1 to 4.2.0 and I have that error when `task_always_eager` and `task_eager_propagates` are `True`. I am not using retry.

I created a decorator to make synchronous calls:

```python
def sync_task(func):
    def func_wrapper(*args, **kwargs):
        promise = func.apply_async(args, kwargs, serializer='dill')  # with dill serializer we allow olt_manager models
        return promise.get(timeout=settings.CELERY_GET_TIMEOUT)
    return func_wrapper
```
and after that I use it the following way:

```python
@sync_task
@shared_task
def myfunction(...):
    ...
```

I need to make tasks synchronous as they deal with an external hardware and I am using Celery as a queue with just one processor to ensure requests are processed in order and one at a time.

auvipy:
so this is a regression

joekohlsdorf:
Also happens on 4.2.1.

untitaker:
since this is marked with the 4.3 milestone, I want to note that this still happens with 4.3. Please let me know if you require additional information or a clean repro case.

auvipy:
re-milestoning for 4.5

jarussi:
I had the same issue, but I only use always_eager for tests. I made a workaround with mock that allows you to run your tests on version `4.3.x`

```python
from unittest import mock

with mock.patch(
    'celery.app.task.denied_join_result'
) as mocked_task_join_will_block:
[...]

PR #6109, opened 2020-05-18
Change eager retry behaviour

mchataigner:
Even with raise self.retry, it should return the eventual value
or MaxRetriesExceededError.

If return value of eager apply is Retry exception, retry eagerly
the task signature.

### DIFF
--- a/celery/app/task.py
+++ b/celery/app/task.py
@@ -709,15 +709,14 @@ class Task(object):
                 ), task_args=S.args, task_kwargs=S.kwargs
             )
 
-        ret = Retry(exc=exc, when=eta or countdown)
+        ret = Retry(exc=exc, when=eta or countdown, is_eager=is_eager, sig=S)
 
         if is_eager:
             # if task was executed eagerly using apply(),
-            # then the retry must also be executed eagerly.
-            retry_ret = S.apply().get()
+            # then the retry must also be executed eagerly in apply method
             if throw:
                 raise ret
-            return retry_ret
+            return ret
 
         try:
             S.apply_async()
@@ -777,6 +776,8 @@ class Task(object):
         retval = ret.retval
         if isinstance(retval, ExceptionInfo):
             retval, tb = retval.exception, retval.traceback
+        if isinstance(retval, Retry) and retval.sig is not None:
+            return retval.sig.apply(retries=retries + 1)
         state = states.SUCCESS if ret.info is None else ret.info.state
         return EagerResult(task_id, retval, state, traceback=tb)
 
--- a/celery/exceptions.py
+++ b/celery/exceptions.py
@@ -137,7 +137,7 @@ class Retry(TaskPredicate):
     #: :class:`~datetime.datetime`.
     when = None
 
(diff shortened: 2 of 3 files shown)


auvipy:
And flake8

mchataigner:
A unit test is enough IMHO as I added a test with eager retry task.
Fixed flake8

auvipy:
ok

auvipy:
any related open issues to close?

fazlerabby7:
> any related open issues to close?

@auvipy I think this one: https://github.com/celery/celery/issues/4661

auvipy:
> > any related open issues to close?
> 
> @auvipy I think this one: #4661

thnaks mr noob :D great to see you here :D

PabloCastellano:
This change has made our travis builds fail, check out https://github.com/openwisp/openwisp-firmware-upgrader/issues/78.

Did it introduce an unnoticed bug or is there something wrong in our code that we need to fix? I wouldn't expect a backwards-incompatible change like this to be released in a minor version and the release notes don't contain any migration path.

Our celery task is defined here: https://github.com/openwisp/openwisp-firmware-upgrader/blob/88b3bd01a521eb9af79704d74f58837631179c71/openwisp_firmware_upgrader/tasks.py#L13-L31

auvipy:
what about celery==4.4.4?

PabloCastellano:
@auvipy Same for 4.4.4 and master

auvipy:
there shouldn't be bc breakable changes. did you try installing the future package on your build? there are possible regressions in the new releases which is not unusual for a project of this scale.

PabloCastellano:
The future package issue is only an issue in 4.4.4 if I'm not mistaken but in any case, the future package is installed so it's not the cause

auvipy:
can you share the trackback?

auvipy:
> Even with raise self.retry, it should return the eventual value
> or MaxRetriesExceededError.
> 
> If return value of eager apply is Retry exception, retry eagerly
> the task signature.

can you check, please?

PabloCastellano:
The traceback is available in Travis: https://travis-ci.org/github/openwisp/openwisp-firmware-upgrader/jobs/693522196#L850

auvipy:
what abot this https://github.com/paramiko/paramiko/issues/1438?

PabloCastellano:
@auvipy That was working perfectly with celery==4.4.2

mchataigner:
@PabloCastellano I replied on google groups, but I will put my answer here as well:


So after some investigation in OpenWisp, it turns out that you have a couple of settings for your tests (in settings.py):
`CELERY_TASK_ALWAYS_EAGER = True`
`CELERY_TASK_EAGER_PROPAGATES = True`

This forces all tasks to be eager.

However,
This commit has been merged https://github.com/celery/celery/commit/26f24faed68c4a0a7053fb9ce2db382e5eaf0151
changing the behaviour of `_ensure_not_eager`

previously it was raising a `RuntimeError`, now it's a simple warning.
In your tests, you expect a `RuntimeError`.

So I guess, you may update your tests.

PR #6113, opened 2020-05-21
Protocol version option for task

MardanovTimur:
## Description

Added optional option `protocol` for applying task with different protocol versions
```python
@shared_task(bind=True, protocol=1)
def debug_task(self):
    # this task will execute with protocol = 1 (but for all other tasks protocol will be `protocol_version`)
    pass
```

P.S.: why? gocelery (lib for golang) supports only **1** celery protocol version. But `flower` application normally supports only **2** protocol version

### DIFF
--- a/celery/app/amqp.py
+++ b/celery/app/amqp.py
@@ -259,6 +259,9 @@ class AMQP(object):
     def create_task_message(self):
         return self.task_protocols[self.app.conf.task_protocol]
 
+    def create_task_message_with_protocol(self, protocol):
+        return self.task_protocols[protocol]
+
     @cached_property
     def send_task_message(self):
         return self._create_task_sender()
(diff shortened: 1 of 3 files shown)


auvipy:
you need to improve the test coverage to avoid regression

MardanovTimur:
> you need to improve the test coverage to avoid regression

Okay, sorry for slow response, I will do it a bit later.

auvipy:
ping

thedrow:
Have we discussed why we need this?

auvipy:
nope not yet

thedrow:
Why shouldn't gocelery add support for protocol version 2 instead?
I don't think we should support this, especially since we're planning further protocol changes.

PR #6115, opened 2020-05-24
Avoid race condition in elasticsearch backend

mchataigner:
if a task is retried, the task retry may work concurrently to current task.
store_result may come out of order.
it may cause a non ready state (Retry) to override a ready state (Success, Failure).
If this happens, it will block indefinitely pending any chord depending on this task.

this change makes document updates safe for concurrent writes.

https://www.elastic.co/guide/en/elasticsearch/reference/current/optimistic-concurrency-control.html

### DIFF
--- a/celery/backends/elasticsearch.py
+++ b/celery/backends/elasticsearch.py
@@ -4,6 +4,7 @@ from __future__ import absolute_import, unicode_literals
 
 from datetime import datetime
 
+from celery import states
 from kombu.utils.encoding import bytes_to_str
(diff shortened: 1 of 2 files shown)


mchataigner:
by the way, is there a way to reject/retry messages if the backend is unreachable?
Because currently, if my backend is dead/down, messages get acked on RMQ but results are not written. As a result any pending chord will have a chord_unlock task continuously retrying.

I'm thinking of a way to requeue or an exponential backoff in the backend until it's available again.

PR #6116, opened 2020-05-24
backends base get_many function now passes READY_STATES

lironhl:
## Description

The `get_many` function in the `backends/base.py` file previously received `READY_STATES` argument, but only considered it, when looking into tasks that are already in cache.
The `get_many` function now also considers the `READY_STATES` arg when fetching new results.

Also added a test case for this behavior.


auvipy:
can you link the related issue?

lironhl:
@auvipy an issue about this didn't exist, should I create one?

PR #6121, opened 2020-05-27
Fix random distribution of jitter for exponential backoff

mchataigner:
random.randrange should be called with the actual so that all numbers
have equivalent probability, otherwise maximum value does have a way
higher probability of occuring.


mchataigner:
Do you have any advices on how to test random distribution for this exponential backoff?

auvipy:
any test that demonstrates changing this will work well. in that case integration test is a good idea

mchataigner:
Done.

auvipy:
thank you

PR #6122, opened 2020-05-27
Add retry on recoverable exception for the backend

mchataigner:
acks.late makes celery acknowledge messages only after processing and
storing result on the backend.

However, in case of backend unreachable, it will shadow a Retry
exception and put the task as failed in the backend not retrying the
task and acknoledging it on the broker.

With this new result_backend_always_retry setting, if the backend
exception is recoverable (to be defined per backend implementation),
it will retry the backend operation with an exponential backoff.

In addition, make ES retry storing updates in a better way

if existing value in the backend is success, then do nothing.
if it is a ready status, then update it only if new value is a ready status as well.
else update it.

This way, a SUCCESS cannot be overriden so that we do not loose
results but any ready state other than success (FAILURE, REVOKED) can
be overriden by another ready status (i.e. a SUCCESS)

### DIFF
--- a/celery/app/defaults.py
+++ b/celery/app/defaults.py
@@ -217,6 +217,10 @@ NAMESPACES = Namespace(
         serializer=Option('json'),
         backend_transport_options=Option({}, type='dict'),
         chord_join_timeout=Option(3.0, type='float'),
+        backend_max_sleep_between_retries_ms=Option(10000, type='int'),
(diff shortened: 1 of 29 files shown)


auvipy:
flake8 issues need to be fixed

mchataigner:
Done.

PR #6134, opened 2020-06-01
Use Django DB max age connection setting (fixes #4116)

### LINKED ISSUE #4116: Django celery fixup doesn't respect Django settings for PostgreSQL connections

whtsky:

## Description

Fixes #4116

#4292 makes Celery respect Django's `CONN_MAX_AGE` settings, but was reverted in #5515. 

This PR tries to respect Django's `CONN_MAX_AGE` while force close db connections after fork.

### DIFF
--- a/celery/fixups/django.py
+++ b/celery/fixups/django.py
@@ -151,7 +151,7 @@ class DjangoWorkerFixup(object):
                 self._maybe_close_db_fd(c.connection)
 
         # use the _ version to avoid DB_REUSE preventing the conn.close() call
-        self._close_database()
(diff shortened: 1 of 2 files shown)


auvipy:
can you comment on #6216

Issue #6132, opened 2020-05-31
Memory leaks


Ixiodor:
# Checklist

- [x] I have verified that the issue exists against the `master` branch of Celery.
- [x] This has already been asked to the [discussion group](https://groups.google.com/forum/#!forum/celery-users) first.
- [x] I have read the relevant section in the
  [contribution guide](http://docs.celeryproject.org/en/latest/contributing.html#other-bugs)
  on reporting bugs.
- [ ] I have checked the [issues list](https://github.com/celery/celery/issues?q=is%3Aissue+label%3A%22Issue+Type%3A+Bug+Report%22+-label%3A%22Category%3A+Documentation%22)
  for similar or identical bug reports.
- [ ] I have checked the [pull requests list](https://github.com/celery/celery/pulls?q=is%3Apr+label%3A%22PR+Type%3A+Bugfix%22+-label%3A%22Category%3A+Documentation%22)
  for existing proposed fixes.
- [ ] I have checked the [commit log](https://github.com/celery/celery/commits/master)
  to find out if the bug was already fixed in the master branch.
- [ ] I have included all related issues and possible duplicate issues
  in this issue (If there are none, check this box anyway).

## Mandatory Debugging Information

- [ ] I have included the output of ``celery -A proj report`` in the issue.
    (if you are not able to do this, then at least specify the Celery
     version affected).
- [x] I have verified that the issue exists against the `master` branch of Celery.
- [ ] I have included the contents of ``pip freeze`` in the issue.
- [ ] I have included all the versions of all the external dependencies required
  to reproduce this bug.

## Optional Debugging Information

- [ ] I have tried reproducing the issue on more than one Python version
  and/or implementation.
- [ ] I have tried reproducing the issue on more than one message broker and/or
  result backend.
- [ ] I have tried reproducing the issue on more than one version of the message
  broker and/or result backend.
- [ ] I have tried reproducing the issue on more than one operating system.
- [ ] I have tried reproducing the issue on more than one workers pool.
- [ ] I have tried reproducing the issue with autoscaling, retries,
  ETA/Countdown & rate limits disabled.
- [ ] I have tried reproducing the issue after downgrading
  and/or upgrading Celery and its dependencies.

#### Related Issues

- None

#### Possible Duplicates

- None

## Environment & Settings

**Celery version**: celery==4.4.2

<details>
<summary><b><code>celery report</code> Output:</b></summary>
<p>

```
[...]

Issue #6135, opened 2020-06-01
Retry args change BUG


Ixiodor:
# Checklist

- [x] I have verified that the issue exists against the `master` branch of Celery.
- [x] This has already been asked to the [discussion group](https://groups.google.com/forum/#!forum/celery-users) first.
- [x] I have read the relevant section in the
  [contribution guide](http://docs.celeryproject.org/en/latest/contributing.html#other-bugs)
  on reporting bugs.
- [ ] I have checked the [issues list](https://github.com/celery/celery/issues?q=is%3Aissue+label%3A%22Issue+Type%3A+Bug+Report%22+-label%3A%22Category%3A+Documentation%22)
  for similar or identical bug reports.
- [ ] I have checked the [pull requests list](https://github.com/celery/celery/pulls?q=is%3Apr+label%3A%22PR+Type%3A+Bugfix%22+-label%3A%22Category%3A+Documentation%22)
  for existing proposed fixes.
- [ ] I have checked the [commit log](https://github.com/celery/celery/commits/master)
  to find out if the bug was already fixed in the master branch.
- [ ] I have included all related issues and possible duplicate issues
  in this issue (If there are none, check this box anyway).

## Mandatory Debugging Information

- [ ] I have included the output of ``celery -A proj report`` in the issue.
    (if you are not able to do this, then at least specify the Celery
     version affected).
- [x] I have verified that the issue exists against the `master` branch of Celery.
- [ ] I have included the contents of ``pip freeze`` in the issue.
- [ ] I have included all the versions of all the external dependencies required
  to reproduce this bug.

## Optional Debugging Information

- [ ] I have tried reproducing the issue on more than one Python version
  and/or implementation.
- [ ] I have tried reproducing the issue on more than one message broker and/or
  result backend.
- [ ] I have tried reproducing the issue on more than one version of the message
  broker and/or result backend.
- [ ] I have tried reproducing the issue on more than one operating system.
- [ ] I have tried reproducing the issue on more than one workers pool.
- [ ] I have tried reproducing the issue with autoscaling, retries,
  ETA/Countdown & rate limits disabled.
- [ ] I have tried reproducing the issue after downgrading
  and/or upgrading Celery and its dependencies.

#### Related Issues

- None

#### Possible Duplicates

- None

## Environment & Settings

**Celery version**: celery==4.4.2

<details>
<summary><b><code>celery report</code> Output:</b></summary>
<p>

```
[...]

PR #6137, opened 2020-06-01
BUG retry

Ixiodor:
Bug solve for retry?


Ixiodor:
> [2020-06-01 14:29:54,698: INFO/MainProcess] Connected to amqp://remote_worker:**@127.0.0.1:5672//
> [2020-06-01 14:29:54,736: INFO/MainProcess] mingle: searching for neighbors
> [2020-06-01 14:29:55,786: INFO/MainProcess] mingle: all alone
> [2020-06-01 14:29:55,807: WARNING/MainProcess] /home/ubuntu/.local/lib/python3.7/site-packages/celery/fixups/django.py:203: UserWarning: Using settings.DEBUG leads to a memory
>             leak, never use this setting in production environments!
>   leak, never use this setting in production environments!''')
> [2020-06-01 14:29:55,808: INFO/MainProcess] celery@testroom ready.
> [2020-06-01 14:30:02,428: INFO/MainProcess] Received task: api_v3.tests.execute[a56bdab9-401a-4cc8-9801-dc8ed3dae764]  
> [2020-06-01 14:30:02,788: WARNING/ForkPoolWorker-1] started
> [2020-06-01 14:30:02,789: WARNING/ForkPoolWorker-1] retry
> [2020-06-01 14:30:02,820: INFO/MainProcess] Received task: api_v3.tests.execute[a56bdab9-401a-4cc8-9801-dc8ed3dae764]  ETA:[2020-06-01 14:30:12.789893+00:00] 
> [2020-06-01 14:30:02,880: INFO/ForkPoolWorker-1] Task api_v3.tests.execute[a56bdab9-401a-4cc8-9801-dc8ed3dae764] succeeded in 0.45013424299941107s: None
> [2020-06-01 14:30:15,056: WARNING/ForkPoolWorker-1] started
> [2020-06-01 14:30:15,056: WARNING/ForkPoolWorker-1] ended
> [2020-06-01 14:30:15,116: INFO/ForkPoolWorker-1] Task api_v3.tests.execute[a56bdab9-401a-4cc8-9801-dc8ed3dae764] succeeded in 0.35210886299864796s: None
> 
result after the fix, to be comparated to the one in issure report

celery/app/task.py
```diff
@@ -714,6 +714,8 @@ def retry(self, args=None, kwargs=None, exc=None, throw=True,
         if is_eager:
             # if task was executed eagerly using apply(),
             # then the retry must also be executed eagerly in apply method
+            # This is needed for eaged
```
auvipy, line 717: this need tests as well

Ixiodor: this one got removed from someone that didn't checked that eager wasn't working...i just re-introduced

auvipy: for that reason, a unit test is needed to verify this change

Ixiodor: i don't wanna discuss, thanks for your work. But where is the unit test for that new introduced BUG? I updated the 4.4.3 and now i have a BUG more.

Ixiodor: @auvipy this is the fixed regression

celery/app/task.py
```diff
@@ -714,6 +714,8 @@ def retry(self, args=None, kwargs=None, exc=None, throw=True,
         if is_eager:
             # if task was executed eagerly using apply(),
             # then the retry must also be executed eagerly in apply method
+            # This is needed for eaged 
+            S.apply()
```
Ixiodor, line 718: @auvipy this is the fixed regression

auvipy: can you also check the related unit test?

Ixiodor: My problem is that i dosn't test things directly on unit test/git soulution, but debugging my "work code" entering in celery. I have to setup the env to test/make unit tests

Ixiodor:
do you think it's possibile to release an "urgent hotfix"? Cause this totally broke the eager and patching retry behaviour could be good too :/
We can connect to try to get an unit test cause requires me too much time to get in

celery/app/task.py
mchataigner: could you share how you do retry your tasks?
for eager retries, you may either use:
raise self.retry(...)
or
return self.retry(..., throw=False)

mchataigner: Actually, it works with both raise/return throw True/False.
#6138

Ixiodor: You can find in issure report the whole code to test the BUG #6135

mchataigner:
If you have urgent production issue I suggest you to pin celery version to 4.4.2 until this gets fixed.
By the way, your patch is breaking tests for eager retry.

Ixiodor:
4.4.2 have that BUG too...

Ixiodor:
Ok, we can revert the changes on task.py seems.
4.4.1 is affected to the same BUG too, I'll manually patch for now.

Ixiodor:
@mchataigner sorry, I'm doing things rapdly and I'm doing wrong things.
Eager dosn't work as i said.

mchataigner:
Could you share the skeleton of your task? it would help to understand your problem.

Ixiodor:
```
@app_cluster.task(bind=True, autoretry_for=(Exception,), max_retries=3,
                  default_retry_delay=10)
def execute(self, param_a, param_b=None, **kwargs):
    print("started")
    if param_b is None:
        param_b = "filled"
        print("retry")
        self.retry(exc=Exception("i have filled now"), args=[param_a, param_b], kwargs=kwargs)
    print("ended")

def test_celery(self):
    sig = execute.si("something")
    t = sig.delay()
    t = 0
```

run this in eager=true in 4.4.3

mchataigner:
Ok, I'm able to reproduce locally. I will update you asap with a patch.

Ixiodor:
Did you even noticed the split of taks?

mchataigner:
So, it seems that putting autoretry_for with Exception is too broad as Celery is managing some states with Exceptions.
But then, there is the autoretry_for mechanism that comes in play. It catches the Retry exception (the one with the correct parameters) and then messes with your new args and throw another Retry exception.
So quick fix for you, specify a concrete Exception in autoretry_for. Actually, you don't need autoretry_for in your case as you are calling self.retry directly, and putting and nested Exception in retry is optional.

So, yes, there is a bug in autoretry_for, not related to my patch concerning eager retry.
But, good news, there is a workaround:
```
@app_cluster.task(bind=True, max_retries=3,
                  default_retry_delay=10)
def execute(self, param_a, param_b=None, **kwargs):
    print("started")
    if param_b is None:
        param_b = "filled"
        print("retry")
        raise self.retry(args=[param_a, param_b], kwargs=kwargs)
    print("ended")

def test_celery(self):
    sig = execute.si("something")
    t = sig.delay()
    t = 0
```

Ixiodor:
Well no...this was an example code. I need to autoretry for all Exception(General) in my real code.
And the BUG about the code you introduced is inside eager=true, the second task, the one with new parameters, dosn't get executed.

mchataigner:
Yes, that's what I pushed here https://github.com/celery/celery/pull/6138
But if you're in a hurry, I would suggest not using autoretry_for and explicitly put a try except around your code:

```
@app_cluster.task(bind=True, max_retries=3,
                  default_retry_delay=10)
def execute(self, param_a, param_b=None, **kwargs
    print("started")
    try:
        if param_b is None:
            param_b = "filled"
            print("retry")
            raise self.retry(args=[param_a, param_b], kwargs=kwargs)
    except Retry:
        raise
    except Exception as exc:
        raise self.retry(exc=exc, args=[param_a, param_b], kwargs=kwargs)
    print("ended")

def test_celery(self):
    sig = execute.si("something")
    t = sig.delay()
    t = 0
```

autoretry_for is made as syntactic suggar for simple code.

Ixiodor:
Oh, ok. I got it now. Thanks.
I'll check if it's all fine for now.

Ixiodor:
SEEMS to fix #6132 too

auvipy:
thanks for raising this! should be fixed by another PR.

celery/app/task.py
PabloCastellano: I'm having the same issue. A minor release shouldn't contain a change like this IMO. Or at least the release notes should have a section about the migration path

Issue #6155, opened 2020-06-04
Celery 4.4.4 raises TypeError: set() takes 3 positional arguments but 4 were given


robvdl:
This doesn't happen on 4.4.3, only on 4.4.4.

Here is a traceback:

```
celery_default_1      | [2020-06-04 19:51:47,723: ERROR/MainProcess] Pool callback raised exception: TypeError('set() takes 3 positional arguments but 4 were given',)
celery_default_1      | Traceback (most recent call last):
celery_default_1      |   File "/usr/share/dashboard/venv/lib/python3.6/site-packages/billiard/pool.py", line 1796, in safe_apply_callback
celery_default_1      |     fun(*args, **kwargs)
celery_default_1      |   File "/usr/share/dashboard/venv/lib/python3.6/site-packages/celery/worker/request.py", line 528, in on_failure
celery_default_1      |     store_result=self.store_errors,
celery_default_1      |   File "/usr/share/dashboard/venv/lib/python3.6/site-packages/celery/backends/base.py", line 169, in mark_as_failure
celery_default_1      |     traceback=traceback, request=request)
celery_default_1      |   File "/usr/share/dashboard/venv/lib/python3.6/site-packages/celery/backends/base.py", line 443, in store_result
celery_default_1      |     request=request, **kwargs)
celery_default_1      |   File "/usr/share/dashboard/venv/lib/python3.6/site-packages/celery/backends/base.py", line 858, in _store_result
celery_default_1      |     self.set(self.get_key_for_task(task_id), self.encode(meta), state)
celery_default_1      | TypeError: set() takes 3 positional arguments but 4 were given
```

I am running on Python 3.6 on Bionic.

For now I've stuck to 4.4.3 as 4.4.4 keeps throwing this exception.

# Checklist

- [ ] I have verified that the issue exists against the `master` branch of Celery.
- [ ] This has already been asked to the [discussion group](https://groups.google.com/forum/#!forum/celery-users) first.
- [ ] I have read the relevant section in the
  [contribution guide](http://docs.celeryproject.org/en/latest/contributing.html#other-bugs)
  on reporting bugs.
- [x] I have checked the [issues list](https://github.com/celery/celery/issues?q=is%3Aissue+label%3A%22Issue+Type%3A+Bug+Report%22+-label%3A%22Category%3A+Documentation%22)
  for similar or identical bug reports.
[...]

PR #6156, opened 2020-06-04
Customize the retry interval of chord_unlock tasks

mchataigner:
By default chord_unlock tasks have interval set to 1.
This patch adds a way to configure this default value for all chords.

### DIFF
--- a/celery/app/builtins.py
+++ b/celery/app/builtins.py
@@ -48,7 +48,7 @@ def add_unlock_chord_task(app):
     from celery.result import allow_join_result, result_from_tuple
 
     @app.task(name='celery.chord_unlock', max_retries=None, shared=False,
-              default_retry_delay=1, ignore_result=True, lazy=False, bind=True)
+              default_retry_delay=app.conf.result_chord_retry_interval, ignore_result=True, lazy=False, bind=True)
     def unlock_chord(self, group_id, callback, interval=None,
                      max_retries=None, result=None,
                      Result=app.AsyncResult, GroupResult=app.GroupResult,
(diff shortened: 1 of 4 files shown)


auvipy:
Besides, can you chime in here https://github.com/celery/celery/issues/6155

auvipy:
can't we unit test this?

mchataigner:
I will work on it later today.

Ixiodor:
@mchataigner there is no way you'll check even the chord_unlock that want to be written in DB?
https://groups.google.com/forum/#!topic/celery-users/WHiLtdcj0M0

mchataigner:
@Ixiodor, I don't get your point, I did not touch ignore_result parameter.

Ixiodor:
@mchataigner there is a BUG in 4.4.4 around chord_unlock when an Exception is triggered inside a task,. The exception will be passed to the chord, the cord re-raise it and tries to write this task sate on DB with null values

mchataigner:
I have another understanding of ignore_result parameter.
For me it's here to avoid saving group result as chord result in the backend. But it still fetches the backend to get result status. and if there is an exception it will bubble up unless you set propagate to False.

Ixiodor:
Actually it writes to DB as "task_id=null" and "task_name=null" and i think this is ivalid in each case

mchataigner:
Thanks for the traceback. I will dig a bit.
Some thoughts, it seems you are calling a chord without a body otherwise it will try to save the error on the callback task. As your traceback is mentioning an exception due to having a null task_id, it's my first guess.

Ixiodor:
my chords are:
`        chords.append(chord("SOME TASK", body=bridge.si(), immutable=True))`
bridge is a dummy task to bypass another "feature" about chords upgrades,
all this chords become:
`    item = chain(*chords).on_error(error_job)
     item.delay()`

mchataigner:
Added a test to cover this new code.

PR #6163, opened 2020-06-07
chord: merge init options with run options

mchataigner:
This way, it's possible to define chord options at creations such as interval that will be used when called.


mchataigner:
I increased coverage.
I don't think integration tests are required for this one.

PR #6173, opened 2020-06-16
put back KeyValueStoreBackend.set method without state

mchataigner:
It turns out it was breaking some other projects.
wrapping set method with _set_with_state, this way it will not break existing Backend.
while enabling this feature for other Backend.

Currently, only ElasticsearchBackend supports this feature.

It protects concurrent update to corrupt state in the backend.
Existing success cannot be overriden, nor a ready state by a non ready state.
i.e. a Retry state cannot override a Success or Failure.
As a result, chord_unlock task will not loop forever due to missing ready state on the backend.

Fixes https://github.com/celery/celery/issues/6155
Fixes https://github.com/celery/django-celery-results/issues/150

### DIFF
--- a/celery/backends/arangodb.py
+++ b/celery/backends/arangodb.py
@@ -144,7 +144,7 @@ class ArangoDbBackend(KeyValueStoreBackend):
             logging.error(err)
             return None
 
-    def set(self, key, value, state):
+    def set(self, key, value):
         """Insert a doc with value into task attribute and _key as key."""
         try:
             logging.debug(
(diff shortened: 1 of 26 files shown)


mchataigner:
Fixed flake8 and pydocstyle.

mchataigner:
@auvipy I still hesitate for visibility, should set_with_state be private or public method?

mchataigner:
@auvipy thanks for merging this one.
Do you think it's worth doing a new celery release?

auvipy:
yes

Issue #6174, opened 2020-06-16
Allow specifying hostname to contrib.testing start_worker()


thenewguy:
# Checklist

- [ x ] I have checked the [issues list](https://github.com/celery/celery/issues?utf8=%E2%9C%93&q=is%3Aissue+label%3A%22Issue+Type%3A+Feature+Request%22+)
  for similar or identical feature requests.
- [ x ] I have checked the [pull requests list](https://github.com/celery/celery/pulls?utf8=%E2%9C%93&q=is%3Apr+label%3A%22PR+Type%3A+Feature%22+)
  for existing proposed implementations of this feature.
- [ x ] I have checked the [commit log](https://github.com/celery/celery/commits/master)
  to find out if the if the same feature was already implemented in the
  master branch.
- [ x ] I have included all related issues and possible duplicate issues
  in this issue (If there are none, check this box anyway).

I did not see anything related to this when searching for start_worker and hostname

#### Related Issues

- None

#### Possible Duplicates

- None

# Brief Summary

It is currently not possible to adjust the hostname in start_worker per https://github.com/celery/celery/blob/master/celery/contrib/testing/worker.py#L114

# Design

It would be helpful if a hostname could be specified or a token could be passed so that multiple workers can be started for running test cases.

## Architectural Considerations

None

## Proposed Behavior

Allow passing a hostname explicitly or either a token passed to anon_nodename() - i.e. worker index

## Diagrams

N/A

## Alternatives

None

auvipy:
are you up for contribution?

thenewguy:
@auvipy sure thing

PR #6177, opened 2020-06-19
Fix upstream 6174

thenewguy:
fixed upstream #6174

### DIFF
--- a/celery/contrib/testing/worker.py
+++ b/celery/contrib/testing/worker.py
@@ -94,6 +94,7 @@ def _start_worker_thread(app,
                          logfile=None,
                          WorkController=TestWorkController,
                          perform_ping_check=True,
+                         hostname=None,
                          **kwargs):
     # type: (Celery, int, str, Union[str, int], str, Any, **Any) -> Iterable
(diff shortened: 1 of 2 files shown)


thenewguy:
@auvipy any tips here?  The solution isn't immediately obvious to me

thedrow:
I think I encountered this problem before.
If you start a new worker, you're changing the root logger.

```python
@pytest.fixture(autouse=True)
    from _pytest.logging import LogCaptureHandler
    root = logging.getLogger()
    rootlevel = root.level
    roothandlers = [
        x for x in root.handlers if not isinstance(x, LogCaptureHandler)]

    yield

    this = request.node.name
        raise RuntimeError(CASE_LOG_LEVEL_EFFECT.format(this))
    newhandlers = [x for x in root_now.handlers if not isinstance(
        x, LogCaptureHandler)]
    if newhandlers != roothandlers:
        raise RuntimeError(CASE_LOG_HANDLER_EFFECT.format(this))
```

We need to either change this check or ensure that we're reusing the same handlers whenever we start a new test app.

PR #6189, opened 2020-06-25
Preserve callbacks when replacing a task with a chain

thedrow:

## Description

When replacing a task in the middle of a chain with a chain that has `link` callbacks, they are not executed anymore.

Example:
```python
@task
def add(a, b):
  return a + b

@task(bind=True)
def callback(*args, **kwargs):
  print("callback")

@task
def replace_with_chain(self, *args, **kwargs):
  c = chain(add.s(4, 5), add.s(1))
  c.link(callback.s)

  return self.replace(c)

chain(add.s(1,2), replace_with_chain.s(), add.s(3))
```

### DIFF
--- a/celery/app/task.py
+++ b/celery/app/task.py
@@ -886,8 +886,8 @@ class Task:
             for t in reversed(self.request.chain):
                 sig |= signature(t, app=self.app)
             sig.set(
-                link=options['link'],
-                link_error=options['link_error']
+                link=options.get('link', None),
+                link_error=options.get('link_error', None)
             )
 
         sig.set(
--- a/celery/app/task.py
+++ b/celery/app/task.py
@@ -882,8 +882,13 @@ class Task:
             )
 
         if self.request.chain:
+            options = sig.options.copy()
             for t in reversed(self.request.chain):
                 sig |= signature(t, app=self.app)
+            sig.set(
+                link=options['link'],
+                link_error=options['link_error']
+            )
 
         sig.set(
             chord=chord,
--- a/celery/app/task.py
+++ b/celery/app/task.py
@@ -883,6 +883,7 @@ class Task:
 
         if self.request.chain:
             options = sig.options.copy()
+            sig.freeze(self.request.id)
             for t in reversed(self.request.chain):
                 sig |= signature(t, app=self.app)
             sig.set(
--- a/celery/app/task.py
+++ b/celery/app/task.py
@@ -882,14 +882,19 @@ class Task:
             )
 
         if self.request.chain:
-            options = sig.options.copy()
+            # We need to freeze the new signature with the current task's ID to
+            # ensure that we don't disassociate the new chain from the existing
+            # task IDs which would break previously constructed results
+            # objects.
             sig.freeze(self.request.id)
+            if "link" in sig.options:
+                final_task_links = sig.tasks[-1].options.setdefault("link", [])
+                final_task_links.extend(maybe_list(sig.options["link"]))
+            # Construct the new remainder of the task by chaining the signature
+            # we're being replaced by with signatures constructed from the
+            # chain elements in the current request.
             for t in reversed(self.request.chain):
                 sig |= signature(t, app=self.app)
-            sig.set(
-                link=options.get('link', None),
-                link_error=options.get('link_error', None)
-            )
 
         sig.set(
             chord=chord,
--- a/t/integration/test_canvas.py
+++ b/t/integration/test_canvas.py
@@ -414,6 +414,7 @@ class test_chain:
         res = c()
         assert res.get(timeout=TIMEOUT) == [8, 8]
 
+    @flaky
     def test_nested_chain_group_lone(self, manager):
         """
         Test that a lone group in a chain completes.
--- a/t/integration/test_canvas.py
+++ b/t/integration/test_canvas.py
@@ -465,7 +465,18 @@ class test_chain:
 
         assert res.get(timeout=TIMEOUT) == 13
 
-        redis_messages = list(redis_connection.lrange('redis-echo', 0, -1))
+        redis_messages = []
+        for _ in range(10):
+            redis_messages = list(
+                redis_connection.lrange('redis-echo', 0, -1)
+            )
+
+            # Avoid race condition in CI where the test finishes
+            # before the link callback is called
+            if not redis_messages:
+                sleep(0.5)
+            else:
+                break
 
         assert redis_messages == [b'link called']
 
@@ -484,7 +495,18 @@ class test_chain:
         with pytest.raises(ValueError):
             res.get(timeout=TIMEOUT)
 
-        redis_messages = list(redis_connection.lrange('redis-echo', 0, -1))
+        redis_messages = []
+        for _ in range(10):
+            redis_messages = list(
+                redis_connection.lrange('redis-echo', 0, -1)
+            )
+
+            # Avoid race condition in CI where the test finishes
+            # before the link_error callback is called
+            if not redis_messages:
+                sleep(0.5)
+            else:
+                break
 
         assert redis_messages == [b'link_error called']
 
--- a/t/integration/test_canvas.py
+++ b/t/integration/test_canvas.py
@@ -17,7 +17,7 @@ from .tasks import (ExpectedException, add, add_chord_to_chord, add_replaced,
                     delayed_sum_with_soft_guard, fail, identity, ids,
                     print_unicode, raise_error, redis_echo, retry_once,
                     return_exception, return_priority, second_order_replace1,
-                    tsum)
+                    tsum, replace_with_chain, replace_with_chain_which_raises)
 
 RETRYABLE_EXCEPTIONS = (OSError, ConnectionError, TimeoutError)
 
@@ -451,6 +451,43 @@ class test_chain:
         )
         res = sig.delay()
         assert res.get(timeout=TIMEOUT) == [42, 42]
+
+    def test_chain_replaced_with_a_chain_and_a_callback(self, manager):
+        if not manager.app.conf.result_backend.startswith('redis'):
+            raise pytest.skip('Requires redis result backend.')
+
+        redis_connection = get_redis_connection()
+        redis_connection.delete('redis-echo')
+
+        c = chain(add.s(1, 2), replace_with_chain.s(), add.s(3))
+        res = c.delay()
+
+        assert res.get(timeout=TIMEOUT) == 13
+
+        redis_messages = list(redis_connection.lrange('redis-echo', 0, -1))
+
+        assert redis_messages == [b'link called']
+
+        redis_connection.delete('redis-echo')
+
+    def test_chain_replaced_with_a_chain_and_an_error_callback(self, manager):
+        if not manager.app.conf.result_backend.startswith('redis'):
+            raise pytest.skip('Requires redis result backend.')
+
+        redis_connection = get_redis_connection()
+        redis_connection.delete('redis-echo')
+
+        c = chain(add.s(1, 2), replace_with_chain_which_raises.s(), add.s(3))
+        res = c.delay()
+
+        with pytest.raises(ValueError):
+            res.get(timeout=TIMEOUT)
+
+        redis_messages = list(redis_connection.lrange('redis-echo', 0, -1))
+
+        assert redis_messages == [b'link_error called']
+
+        redis_connection.delete('redis-echo')
 
 
 class test_result_set:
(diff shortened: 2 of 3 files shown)


thedrow:
I haven't found a way to properly test this yet.
Any idea?

celery/app/task.py
```diff
@@ -885,8 +885,13 @@ def replace(self, sig):
             )
 
         if self.request.chain:
+            options = sig.options.copy()
             for t in reversed(self.request.chain):
                 sig |= signature(t, app=self.app)
+            sig.set(
+                link=options['link'],
+                link_error=options['link_error']
```
thedrow, line 893:
```suggestion
                link=options.get('link', None),
                link_error=options.get('link_error', None)
```

It turns out both `link` and `link_error` may not be present.

auvipy (dismissed):

thedrow:
@maybe-sybr Do you have a moment to look into this?
Locally, `test_chain.test_chain_replaced_with_a_chain_and_a_callback` that is failing in CI isn't failing for me and the `test_chain_replaced_with_a_chain_and_an_error_callback` test is hanging until timeout.
Any leads would be greatly appreciated.

maybe-sybr:
So IIUC, this is taking any `link`/`link_error` signature which is set for the signature being passed to `replace()`, and tacking it onto the end of the original chain which has been reconstructed in the loop on L886-887? Comments below are based on that reading of the diff.

I can only point my finger at perhaps a latency issue for `test_chain_replaced_with_a_chain_and_a_callback()` since we could arguably resolve the result for the final task in the chain prior to the final linked task having been executed, thereby racing the linked redis_echo task. This definitely seems like it could cause the test to fail on a resource constrained box. In fact, I can reproduce the error if I stick a `time.sleep(0.1)` at the start of the `redis_echo()` task. Perhaps something as simple as a short sleep in between getting the result back and attempting to find the redis-echo message would help confirm that? If that works, we could add a retry N times loop to make it behave nicely for merge.

I can confirm the same breakage you're observing locally for `test_chain_replaced_with_a_chain_and_an_error_callback()`. I believe it's most likely being caused by the chain reconstruction which appears to not be maintaining the ID of the task being replaced. When the replacement signature fails, it marks the subsequent tasks as being complete and the actor waiting on the original (client/test function side) result object notices this - however the test then tries to resolve the result of the child task of the chain which is still the original middle task in the chain which we replaced. Because that task never completed, the test sits there waiting and periodically checking the redis backend to see if that key gets set. It never does and we eventually time out. We can see that in this test output I made it spit out with some hacking on the test and trace.py:
```
--------------------------------------------------------- Captured stdout call ---------------------------------------------------------
AWAITING (('c900acac-2392-4d40-b831-dd2bd85bfb08', (('440b4ddf-abdc-4531-8daf-2841343699c7', (('f7567454-8389-47e7-9b0f-806c9f121a28', None), None)), None)), None)
NEW SIG IS (('c900acac-2392-4d40-b831-dd2bd85bfb08', (('a97f55c6-f55f-4972-8b48-283907f9d961', (('cb30011b-d1d8-44bf-bfa4-4d39e2f69964', None), None)), None)), None)
```

I think the fix for `test_chain_replaced_with_a_chain_and_an_error_callback()` would be to make the signature passed to replace have its ID set to the task_id in `self.request`. We can see above that it should be `440b4ddf-...` but it ended up being `a97f55c6-...` and there's a mismatch between what result key will be set and what the test is going to wait for.

celery/app/task.py
```diff
@@ -882,8 +882,13 @@ def replace(self, sig):
             )
 
         if self.request.chain:
+            options = sig.options.copy()
```
maybe-sybr, line 885: This appears to fix `test_chain_replaced_with_a_chain_and_an_error_callback()` AFAICT
```suggestion
            options = sig.options.copy()
            sig.freeze(self.request.id)
```

celery/app/task.py
```diff
@@ -882,8 +882,13 @@ def replace(self, sig):
...
             for t in reversed(self.request.chain):
                 sig |= signature(t, app=self.app)
+            sig.set(
+                link=options.get('link', None),
```
maybe-sybr, line 889: What happens if we already have a `link` or `link_error` for the rest of this chain? Will we overwrite it, and would that be surprising?

thedrow: I honestly have no idea.

thedrow:
Now `test_nested_chain_group_lone` is failing.

thedrow:
It is passing locally for me though.

celery/app/task.py
thedrow: Would you like to try to come up with such a test case?

maybe-sybr: Sure thing. Sorry, this was on my list but I hadn't gotten around to it over the weekend. I'll see what I can do after work today now that you've pushed that fix for the test which raced the callback.

thedrow:
@maybe-sybr I can wait for another test case until Sunday.
If you can get to it by then, it'll be swell.

maybe-sybr:
> @maybe-sybr I can wait for another test case until Sunday.
> If you can get to it by then, it'll be swell.

@thedrow - I'm working on this now. Will push some changes shortly or post back and resolve my comment above if nothing needs to change.

maybe-sybr:
Okay, I have a solid answer for what's really broken here. It appears that the issues isn't in `task.replace()` at all, the fix we've been looking at would simply address the symptom. The root cause is that `Signature.__or__()` is busted and doesn't retain linked tasks as one might expect. Check out this snippet on current master:

```python
import celery

app = celery.Celery()

@app.task
def foo():
    pass

@app.task
def bar():
    pass

@app.task
def baz():
    pass

print("FIRST CHAIN")
c = foo.s() | bar.s()
c.link(baz.s("SUCCESS"))
c.link_error(baz.s("ERROR"))
print(c.tasks, c.options)
print("TASKS")
for t in c.tasks:
    print(t, t.options)

print("CHAINED CHAIN")
c2 = c | foo.s()
print(c2.tasks, c2.options)
print("TASKS")
for t in c2.tasks:
    print(t, t.options)
```
->
```
FIRST CHAIN
(__main__.foo(), __main__.bar()) {'link': [__main__.baz('SUCCESS')], 'link_error': [__main__.baz('ERROR')]}
TASKS
__main__.foo() {}
__main__.bar() {}
CHAINED CHAIN
[__main__.foo(), __main__.bar(), __main__.foo()] {}
TASKS
__main__.foo() {'link_error': [__main__.baz('ERROR')]}
__main__.bar() {'link_error': [__main__.baz('ERROR')]}
__main__.foo() {}
```

There's a special case in `Signature.unchain_tasks()` which jams any errback signature onto all tasks being unchained which works out fine because if any one of them fails, we won't ever run the subsequent ones and so the `link_error`ed signature will only run once.

https://github.com/celery/celery/blob/8c5e9888ae10288ae1b2113bdce6a4a41c47354b/celery/canvas.py#L632-L639

However, we don't do anything to respect `link`ed tasks during all of this and so we end up losing them to the ether if we use `|` to join chains together. I believe this is true for chains on either side of the vertical bar operator since `unchain_tasks()` is used throughout the `Signature.__or__()` definition when chains are encountered.

I think some of this is partially due to a difference in expectations with respect to what each type of canvas signature does when signatures get `link`ed or `link_error`ed to them. `group`s and `chord`s will push the linked signature down to their 0th task and `body` respectively, while `chain`s simply leave it attached to themselves, rather than attaching it to any particular element of their `tasks`.

This behaviour might however be because it could be considered ambiguous when the linked task should run in a chain constructed like so? :

```python
c = foo.s() | bar.s()
c.link(cb.s())
c |= baz.s()
c.link(cb2.s())
```

Depending on how you interpret the mutation properties of chains (and signature objects more broadly), that could be read as equivalent to either of:

```python
c = foo.s() | bar.s(options={"link": cb.s()}) | baz.s(options={"link": cb2.s()})
# OR
c = foo.s() | bar.s() | baz.s(options={"link": [cb.s(), cb2.s()]})
```

In addition to all of this, there are some surprising difference in behaviours when we `apply()` vs `apply_async()` chains with call/errbacks. In particular, the `_chain.apply()` implementation passes down the chain's options to the encapsulated tasks to be run which means the call/errbacks of the encapsulated tasks can be lost if the chain has some set, and callbacks for the chain will be called multiple times for each encapsulated task rather than at the end of the chain.

I'm going to make a few issues to track some of this stuff since I think there are some pretty confusing inconsistencies with call/errbacks and canvas composition which needs to be fixed and/or documented as we move forward.

 ---

Now, how this related back to this PR:

### 1
This changeset only fixes a symptom of task chaining using the vertical bar operator being busted, but that seems sensible for inclusion in 5.0.2.
### 2
Because it's only `link`ed signatures that we lose when reconstructing the tail of the chain we're replacing, we don't need to do anything about the `link_error` signatures AFAICT. We can change the diff on `app/task.py` to avoid this and rely on the second new integration test to validate the behaviour
### 3
In answer to my original review question, we do manage to run a signature `link`ed to a chain within which we replace some task with a chain which also has a `link`ed signature. This is because by that point the signature `link`ed to the *external* chain has been pushed down to the final task in the chain as part of the async apply/freezing process (IIUC).

This means that we end up with the `link`ed signature from the *internal* chain which we replaced the task with being `link`ed to the end of the newly reconstructed chain (with the diff in this PR, previously we'd lose that `link`ed signature). We also have the signature originally `link`ed to the *external* chain `link`ed to the final task of the newly reconstructed chain. When we eventually `apply()` or `delay` the reconstructed signature, this pushes the signature which was `link`ed to the *internal* replacement chain down to the final task in the reconstructed chain, alongside the signature `link`ed to the *external* chain.

Both get run happily, however the signature from the *internal* replacement signature is now run at the wrong time and will receive the wrong arguments (retval from the final task instead of retval from the intermediate task which used to be the final task of the replacement chain). We can fix this by attaching the *internal* chain's signature to the final task prior to chaining with the remainder of the external chain. I'll review comment that now to make it clear what needed to be changed and will also make the change myself.

wow, what an adventure

celery/app/task.py
maybe-sybr: I have a new test case for this and have explained what I worked out in the main thread. Will resolve this and replace with a comment suggesting the diff I'm working on now

celery/app/task.py
```diff
@@ -882,8 +882,14 @@ def replace(self, sig):
...
 
         if self.request.chain:
+            options = sig.options.copy()
+            sig.freeze(self.request.id)
             for t in reversed(self.request.chain):
                 sig |= signature(t, app=self.app)
+            sig.set(
+                link=options.get('link', None),
+                link_error=options.get('link_error', None)
+            )
```
maybe-sybr, line 892: We shouldn't need to do this for `link_error` because unchaining which occurs when we compose chains with the vertical bar operator duplicates `link_error` errbacks onto all of a chain's encapsulated tasks. We do need to ensure the `link`ed callback is retained (the issue originally identified) but pushing it to the end of the newly reconstructed chain means it'll be run at the wrong time and with the wrong argument(s). This should be roughly:

```suggestion
            sig.freeze(self.request.id)
            if "link" in sig.options:
              sig.tasks[-1].options.setdefault("link", []).append(sig.options["link"])
            for t in reversed(self.request.chain):
                sig |= signature(t, app=self.app)
```

maybe-sybr (changes requested):

celery/app/task.py
maybe-sybr: The new diff makes this change and there are a couple of new tests to hit the cases I mentioned in previous comments.

maybe-sybr:
Just pushed a minor fixed to remove a dead line I missed. We'll want to autosquash or squashmerge, happy for either.

auvipy (approved):
looks great to me

t/integration/test_canvas.py
```diff
@@ -452,6 +453,138 @@ def test_nested_chain_group_last(self, manager):
...
+            identity.s('Hello '),
+            replace_with_chain_which_raises.s(link_msg=link_msg),
+            add.s(' will never be seen :(')
+        )
+        res = c.delay()
+
+        with pytest.raises(ValueError):
+            res.get(timeout=TIMEOUT)
+
+        expected_msgs = {link_msg, }
```
auvipy, line 505: only {link_msg, } ?

maybe-sybr: Yep, I left it like this so that we could refactor these tests with paramaterize and maybe subtest on a future pass through the file. If they all expect to check against a set of messages we should be able to handle that in a parameterized test a bit easier.

maybe-sybr:
I've just repushed with my fixups squashed down and a fix for that lint failure. I'll throw this back to @thedrow now - ping me if we need anything else on this :)

auvipy (approved):
LGTM

thedrow:
While looking into this I thought about another use-case that is unrelated to this PR.
If we replace something with an empty chain, an error occurs and it is not clear from it what went wrong.
I'll file an issue about it.

thedrow:
Thank you guys for all your hard work.

Issue #6200, opened 2020-06-30
Chain of a Chain with .apply() - fails to pass the right arguments


AbdealiLoKo:
# Checklist

* [x] I have verified that the issue exists against the `master` branch of Celery.
* [ ] This has already been asked to the [discussion group](https://groups.google.com/forum/#!forum/celery-users) first.
* [x] I have read the relevant section in the
  [contribution guide](http://docs.celeryproject.org/en/latest/contributing.html#other-bugs)
  on reporting bugs.
* [x] I have checked the [issues list](https://github.com/celery/celery/issues?q=is%3Aissue+label%3A%22Issue+Type%3A+Bug+Report%22+-label%3A%22Category%3A+Documentation%22)
  for similar or identical bug reports.
* [ ] I have checked the [pull requests list](https://github.com/celery/celery/pulls?q=is%3Apr+label%3A%22PR+Type%3A+Bugfix%22+-label%3A%22Category%3A+Documentation%22)
  for existing proposed fixes.
* [ ] I have checked the [commit log](https://github.com/celery/celery/commits/master)
  to find out if the bug was already fixed in the master branch.
* [ ] I have included all related issues and possible duplicate issues
  in this issue (If there are none, check this box anyway).

## Mandatory Debugging Information
* [x] I have included the output of ``celery -A proj report`` in the issue.
    (if you are not able to do this, then at least specify the Celery
     version affected).
* [x] I have verified that the issue exists against the `master` branch of Celery.
* [ ] I have included the contents of ``pip freeze`` in the issue.
* [x] I have included all the versions of all the external dependencies required
  to reproduce this bug.

## Optional Debugging Information

* [x] I have tried reproducing the issue on more than one Python version
  and/or implementation.
* [x] I have tried reproducing the issue on more than one message broker and/or
  result backend.
* [x] I have tried reproducing the issue on more than one version of the message
  broker and/or result backend.
* [x] I have tried reproducing the issue on more than one operating system.
* [ ] I have tried reproducing the issue on more than one workers pool.
* [ ] I have tried reproducing the issue with autoscaling, retries,
  ETA/Countdown & rate limits disabled.
* [x] I have tried reproducing the issue after downgrading
  and/or upgrading Celery and its dependencies.

#### Related Issues
* None

#### Possible Duplicates
* None

## Environment & Settings

**Celery version**:

<details>
<summary><b><code>celery report</code> Output:</b>
/summary>
<p>

```
[...]

PR #6201, opened 2020-06-30
test_canvas: Add test for chain-in-chain

AbdealiJK:
Add test case for the issue where a chain in a chain does not
work when using .apply(). This works fine with .apply_async().

The inner chain should have only 1 item.

## Description

Related to: https://github.com/celery/celery/issues/6200

### DIFF
--- a/t/integration/test_canvas.py
+++ b/t/integration/test_canvas.py
@@ -1158,6 +1158,25 @@ class test_chord:
         res1 = c1.apply(args=(1,))
         assert res1.get(timeout=TIMEOUT) == [1, 1]
 
+    @pytest.mark.xfail(reason="Issue #6200")
+    def test_chain_in_chain_with_args(self):
+        try:
+            manager.app.backend.ensure_chords_allowed()
+        except NotImplementedError as e:
+            raise pytest.skip(e.args[0])
+
+        c1 = chain(  # NOTE: This chain should have only 1 chain inside it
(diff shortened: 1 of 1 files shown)


AbdealiJK:
Rebase done :+1:

auvipy:
please check flake8 and pydocs errors

t/integration/test_canvas.py
```diff
@@ -952,6 +953,26 @@ def test_chord_in_chain_with_args(self, manager):
...
+
+    @pytest.mark.xfail(reason="Issue #6200")
+    def test_chain_in_chain_with_args(self):
+        try:
+            manager.app.backend.ensure_chords_allowed()
+        except NotImplementedError as e:
+            raise pytest.skip(e.args[0])
+
+        c1 = chain(  # NOTE: This chain should have only 1 chain inside it
+            chain(
```
thedrow, line 965: So this doesn't fold into one chain anymore?

AbdealiJK:
Seems like I missed the email notifications for this PR - I've updated the fix the flake8 issues.

t/integration/test_canvas.py
AbdealiJK: Not too sure - but considering it works for redis - I guess it is fine. It just seems to be an issue with the `.apply()`

thedrow (approved):

PR #6205, opened 2020-07-01
[ES backend] add 401 as safe for retry

mchataigner:
if authentication mechanism is temporarily unavailable ES API may return 401 Unauthorized.
Resulting in lost result because it was not retried in this case.


mchataigner:
I don't understand why tests are not passing on Travis. Locally I'm able to correctly run all tests in python 2.7 and 3.5. :/

mchataigner:
trying to fix tests here => https://github.com/celery/celery/pull/6207

PR #6206, opened 2020-07-01
treat internal errors as failure

mchataigner:
this way, task may be rejected and not acknowledged in case of
unrecoverable error by the backend or any other celery components.

Handle result correctly in request direct execution outside a pool.


mchataigner:
same issue here with tests on travis with python 2.7 and 3.5. on my local machine everything is green. :/

auvipy:
avoid those errors

mchataigner:
trying to solve those flaky tests here https://github.com/celery/celery/pull/6207

mchataigner:
I updated tests.

mchataigner:
Fixed flake8

PR #6207, opened 2020-07-01
Trying to fix flaky tests in ci

mchataigner:


mchataigner:
tests are failing on travis due to some rate limiting on docker hub…

PR #6211, opened 2020-07-02
Explicit call to Drainer.stop()

mchataigner:
It may help resolving flakyness on windows build.


mchataigner:
it seems a bad idea.

PR #6212, opened 2020-07-02
Drainer tests, put a lower constraint on number of intervals

mchataigner:
liveness should iterate 10 times per interval while drain_events only
once. However, as it may use thread that may be scheduled out of
order, we may end up in some situation where liveness and drain_events
were called the same amount of time.

Lowering the constraint from < to <= to avoid failing the tests.


auvipy:
thanks a lot!

PR #6218, opened 2020-07-10
Preserve order of group results with Redis result backend

### LINKED ISSUE #3781: Group tasks' results order getting messed up when using redis

maybe-sybr:
## Description

This PR is an alternate to #4858 authored by @lpsinger and fixes #3781. It ensures
that the order of results from a canvas group task wrapper matches the order of the
individual encapsulated tasks when using Redis as a results backend.

### DIFF
--- a/celery/backends/redis.py
+++ b/celery/backends/redis.py
@@ -425,8 +425,8 @@ class RedisBackend(BaseKeyValueStoreBackend, AsyncBackendMixin):
         result = self.encode_result(result, state)
         with client.pipeline() as pipe:
             pipeline = pipe \
-                .zadd(jkey, group_index,
-                      self.encode([1, tid, state, result])) \
+                .zadd(jkey,
+                      {self.encode([1, tid, state, result]): group_index}) \
                 .zcount(jkey, '-inf', '+inf') \
                 .get(tkey)
 
--- a/celery/backends/redis.py
+++ b/celery/backends/redis.py
@@ -413,9 +413,11 @@ class RedisBackend(BaseKeyValueStoreBackend, AsyncBackendMixin):
     def on_chord_part_return(self, request, state, result,
                              propagate=None, **kwargs):
         app = self.app
-        tid, gid = request.id, request.group
+        tid, gid, group_index = request.id, request.group, request.group_index
         if not gid or not tid:
             return
+        if group_index is None:
+            group_index = '+inf'
 
         client = self.client
         jkey = self.get_key_for_group(gid, '.j')
@@ -423,8 +425,9 @@ class RedisBackend(BaseKeyValueStoreBackend, AsyncBackendMixin):
         result = self.encode_result(result, state)
         with client.pipeline() as pipe:
             pipeline = pipe \
-                .rpush(jkey, self.encode([1, tid, state, result])) \
-                .llen(jkey) \
+                .zadd(jkey, group_index,
+                      self.encode([1, tid, state, result])) \
+                .zcount(jkey, '-inf', '+inf') \
                 .get(tkey)
 
             if self.expires is not None:
@@ -443,7 +446,7 @@ class RedisBackend(BaseKeyValueStoreBackend, AsyncBackendMixin):
                 decode, unpack = self.decode, self._unpack_chord_result
                 with client.pipeline() as pipe:
                     resl, = pipe \
-                        .lrange(jkey, 0, total) \
+                        .zrangebyscore(jkey, '-inf', '+inf') \
                         .execute()
                 try:
                     callback.delay([unpack(tup, decode) for tup in resl])
--- a/celery/backends/redis.py
+++ b/celery/backends/redis.py
@@ -446,7 +446,7 @@ class RedisBackend(BaseKeyValueStoreBackend, AsyncBackendMixin):
                 decode, unpack = self.decode, self._unpack_chord_result
                 with client.pipeline() as pipe:
                     resl, = pipe \
-                        .zrangebyscore(jkey, '-inf', '+inf') \
+                        .zrange(jkey, 0, -1) \
                         .execute()
                 try:
                     callback.delay([unpack(tup, decode) for tup in resl])
--- a/celery/app/amqp.py
+++ b/celery/app/amqp.py
@@ -305,7 +305,7 @@ class AMQP(object):
         )
 
     def as_task_v2(self, task_id, name, args=None, kwargs=None,
-                   countdown=None, eta=None, group_id=None,
+                   countdown=None, eta=None, group_id=None, group_index=None,
                    expires=None, retries=0, chord=None,
                    callbacks=None, errbacks=None, reply_to=None,
                    time_limit=None, soft_time_limit=None,
@@ -363,6 +363,7 @@ class AMQP(object):
                 'eta': eta,
                 'expires': expires,
                 'group': group_id,
+                'group_index': group_index,
                 'retries': retries,
                 'timelimit': [time_limit, soft_time_limit],
                 'root_id': root_id,
@@ -397,7 +398,7 @@ class AMQP(object):
         )
 
     def as_task_v1(self, task_id, name, args=None, kwargs=None,
-                   countdown=None, eta=None, group_id=None,
+                   countdown=None, eta=None, group_id=None, group_index=None,
                    expires=None, retries=0,
                    chord=None, callbacks=None, errbacks=None, reply_to=None,
                    time_limit=None, soft_time_limit=None,
@@ -442,6 +443,7 @@ class AMQP(object):
                 'args': args,
                 'kwargs': kwargs,
                 'group': group_id,
+                'group_index': group_index,
                 'retries': retries,
                 'eta': eta,
                 'expires': expires,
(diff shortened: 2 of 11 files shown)


maybe-sybr:
flakes seem happier than the previous PR on my machine. I'll see how CI here goes before I dig into getting the full test suite running locally.

I've stripped a commit from the previous PR which changed the version of flake8 used since it wasn't relevant to this changeset and was probably super outdated anyway. Otherwise, it's pretty much the same as #4858 except for some resolving the conflict in `celery/backends/redis.py`.

maybe-sybr:
@georgepsarakis - this is a refresh of #4858 where you raised a question about the changes from using a List to a SortedSet in the Redis backend wrt backward compatibility. Can you revisit this? ref: https://github.com/celery/celery/pull/4858#issuecomment-403697120

maybe-sybr:
>  1 failed
>         - t/unit/backends/test_redis.py:591 test_RedisBackend.test_on_chord_part_return

Looks like we didn't change the args the mock accepts for `zadd()` - I think the fixup commit solves that. I am getting some other unit test failures on my box in that file though which is a bit concerning. We'll see how CI goes this time I guess.

maybe-sybr:
This changeset should be good now. I had to correct the behaviour of the sorted set mocks in the backend test suite since they didn't match what Redis would actually do an also returned the score as well as the desired value in a 2-tuple.

I've left a fixup commit on the end which should be squashed down onto the first commit in this branch prior to merge. The mock fixups are in their own commit to make it clear what has changed since the previous MR to get things working. Let me know if you would like any further changes prior to merge, @auvipy. I'll keep an eye open for this CI run but I'm hoping it'll go green now.

auvipy:
E           redis.exceptions.ResponseError: WRONGTYPE Operation against a key holding the wrong kind of value

.tox/3.8-integration-redis/lib/python3.8/site-packages/redis/connection.py:756: ResponseError

maybe-sybr:
Interesting. IIUC that's blowing up because the integration test is trying to access an element of the result object (or maybe the whole object) as a list. Should be a fairly straightforward fix, I'll try to get to it tomorrow.

maybe-sybr:
Looks like we need a test for `group.freeze(group_index=...)` and another which sanity tests that the `group_index()` property exists. That'll fix up the drops in coverage and make codecov happy.

auvipy:
please do so when you have time.

celery/backends/redis.py
```diff
@@ -413,18 +413,21 @@ def apply_chord(self, header_result, body, **kwargs):
...
 
         client = self.client
         jkey = self.get_key_for_group(gid, '.j')
         tkey = self.get_key_for_group(gid, '.t')
         result = self.encode_result(result, state)
         with client.pipeline() as pipe:
             pipeline = pipe \
-                .rpush(jkey, self.encode([1, tid, state, result])) \
-                .llen(jkey) \
+                .zadd(jkey,
```
georgepsarakis, line 428: @maybe-sybr unfortunately at this point (and Line 449) we break backwards compatibility and exceptions will be raised. I think we need to handle both structures for the transition. What do you think?

maybe-sybr: For my purposes, I haven't run long term deployments where the Redis instance outlives the Celery one(s). It certainly seems feasible but would either:

 * require an extra type check against the backend before we actually build up the right pipeline for the observed type, or
 * construct a list or zset based pipeline depending on whether we think the group is zset based or not (perhaps if `group_index is not None`?)

Any preference on the approach?

maybe-sybr: Thinking about this a bit more, I don't really see how we can reasonably maintain backward compatibility here. Only workers interact with the results backend and use this code, so IIUC we're talking about situations where there's a progressive upgrade of celery workers in a cluster, with some running old code and others running the new code. Or am I misunderstanding and does result resolution on any part of the system also hit the result backend?

The problem with the mixed worker node situation is that old code will unconditionally use operations for lists while new code should optimistically use operations for sorted sets and fall back to lists if it has to interact with older workers. That'll be racy and break if a worker running old code happens to get beaten to the results key by one running new code. I suppose there could be a decision made on something like feature flags (or a naive version check) passed between workers when they mingle but I'm not familiar enough with the clustering protocol to know if that's a reasonable idea.

celery/backends/redis.py
```diff
@@ -413,18 +413,21 @@ def apply_chord(self, header_result, body, **kwargs):
     def on_chord_part_return(self, request, state, result,
                              propagate=None, **kwargs):
         app = self.app
-        tid, gid = request.id, request.group
+        tid, gid, group_index = request.id, request.group, request.group_index
```
maybe-sybr, line 416: There is some breakage when we run a worker on this branch but issue tasks from an app (using the example app code from #4858) using celery master:

```
[2020-07-13 09:14:51,351: ERROR/MainProcess] Pool callback raised exception: AttributeError("'Context' object has no attribute 'group_index'")
  File "/dev/celery/venv/lib64/python3.8/site-packages/billiard/pool.py", line 1796, in safe_apply_callback
    fun(*args, **kwargs)
  File "/dev/celery/celery/worker/request.py", line 531, in on_failure
    self.task.backend.mark_as_failure(
  File "/dev/celery/celery/backends/base.py", line 172, in mark_as_failure
    self.on_chord_part_return(request, state, exc)
  File "/dev/celery/celery/backends/redis.py", line 416, in on_chord_part_return
    tid, gid = request.id, request.group
AttributeError: 'Context' object has no attribute 'group_index'
```

From what I can tell, when running a worker node on master and a task issuer on this branch, the behaviour simply falls back to having the results be randomly ordered based on order of completion. That seems like a fine graceful degradation.

maybe-sybr: This is fixed by adding the `Context.group_index` attribute and handling passing of `group_index` when uplifting to chords. That should also fix the missing coverage on the `Request.group_index` property method.

maybe-sybr:
@auvipy This changeset is green in CI now but @georgepsarakis ' question regarding backward compatibility needs to be resolved prior to merge. Let me know if there are any other merge-prep changes you'd like to see.

maybe-sybr:
I think the Appveyor build may have died/timed out. The red job ran for an hour and seems t have truncated output.

auvipy:
need not to worry about AppVeyor now

celery/backends/redis.py
maybe-sybr: ping @georgepsarakis - any input on this? AFAICT running code which issues tasks on master and a worker on this branch works fine, and vice versa (with the results being unordered since the worker uses a list to store the group results). So I think what I said in the previous comment about mixed code in clusters of workers is where breakage would occur, and need some input on whether we need some sort of behavioural switch based on something sourced from the mingling process.

auvipy:
>           res.children[0].children[0].result[0])

E       TypeError: 'NoneType' object is not subscriptable

maybe-sybr:
That's interesting @auvipy - I think that test passed on effectively the same changeset previously.

Build 8047 (red):
 * https://travis-ci.org/github/celery/celery/builds/707534893
 * https://github.com/celery/celery/commit/e88df677dd8de7854e28a00a4c0e5a48f74734a0

Build 8046 (green):
 * https://travis-ci.org/github/celery/celery/builds/707508045
 * https://github.com/celery/celery/commit/66de942a80049cf9f334caf02c13ceb29a357e74

Seems like the test might be racey.

auvipy:
I restarted the build

maybe-sybr:
I managed to hit the error running the chord integration tests in `t/integration/test_canvas.py` with a redis backend. I'll submit a separate PR if I work out how to fix it since it doesn't seem like it's relevant to this changeset.

auvipy:
all the tests passing now here! please submit another PR for the unrelated error

auvipy:
I will wait for clarification or suggestion about keeping BC from George

thedrow:
@auvipy Did we address @georgepsarakis's concerns?

maybe-sybr:
@thedrow - I've made no changes to address the back-compat concerns. Awaiting a reply from @georgepsarakis since I'm not sure how we can go about solving the issue (per my comments in that thread) given my current understanding of when problems would manifest (multiple workers running mixed versions of celery IIUC).

thedrow:
Maybe we can introduce the new behavior in a setting.
That way, everyone who wants it can migrate.

PR #6223, opened 2020-07-12
FIX: -A and --args should behave the same.

### LINKED ISSUE #4558: Unlike what doc states, --app does not behave like -A
## Steps to reproduce

* Try to run `celery --app foo *anything*`.
* Look at the "celery usage" output instead of the "*anything*" expected output.

## Expected behavior

"-A foo" and "--app foo" to behave the same way.

## Actual behavior

"-A foo" works, while "--app foo" just outputs the "celery usage" screen

## Environment

```
software -> celery:4.1.0 (latentcall) kombu:4.1.0 py:3.6.3
settings -> transport:amqp results:django-db
```

JulienPalard:
Closes #4558

The added test should fail like this, without this patch:

    AssertionError: assert 't.unit.bin.test_celery.APP' == 'worker'

## Description

This is a try to solve #4558 (which just bite me).

As celery argument parsing is a bit convoluted and as it's the really first time I contribute to it and even read it, please review carefully.

The way I understand it, is that `_relocate_args_from_start` is a bit dangerous, as it moves single dashed commands as always having a value, and doubly dashed options as never having one.

I bypassed `_relocate_args_from_start` for `-A` / `--app` easily as it was already parsed and stored in the object.

But I did not fixed the issue in `_relocate_args_from_start`, typically `-l debug worker -c 3` is rewritten as `worker -c 3 -l debug`, which work, but `--loglevel debug worker -c 3` is rewritten as `debug worker -c 3 --level`, which fails.

### DIFF
--- a/celery/bin/base.py
+++ b/celery/bin/base.py
@@ -514,7 +514,7 @@ class Command(object):
         else:
             self.app = Celery(fixups=[])
 
-        self._handle_user_preload_options(remaining_options)
+        self._handle_user_preload_options(argv)
 
         return remaining_options
 
--- a/celery/bin/base.py
+++ b/celery/bin/base.py
@@ -474,7 +474,7 @@ class Command(object):
         return parser
 
     def setup_app_from_commandline(self, argv):
-        preload_options = self.parse_preload_options(argv)
+        preload_options, remaining_options = self.parse_preload_options(argv)
         quiet = preload_options.get('quiet')
         if quiet is not None:
             self.quiet = quiet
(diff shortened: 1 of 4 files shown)


auvipy:
>   ???

E   AssertionError: assert 'args' == 'worker'

E     - args

E     + worker

t/unit/bin/test_celery.py:222: AssertionError

JulienPalard:
Yes I'll look at the CI, I only ran pytest at home. But I'm going to sleep now :p

auvipy:
its OK good night :D

JulienPalard:
> > ???
> 
> E AssertionError: assert 'args' == 'worker'
> 
> E - args
> 
> E + worker
> 
> t/unit/bin/test_celery.py:222: AssertionError

It was due to me using `mock.call_args.args` which was introduced in 3.8 only, fixed.

JulienPalard:
While I'm proofreading `_relocate_args_from_start`:

```python
# if there are no more arguments then the last arg in rest'
# must be the command.
[rest.pop()] + rest
```

is suspicious, and it's outside of the `while` statement, `rest` is dropped after that, so I think it can be removed entierly.

JulienPalard:
(LGTM bot is spamming random things :()

auvipy:
sadly we are getting reports for a regression for this. if you can help to find out a fix we need to revert this otherwise

JulienPalard:
Yes just revert it, I'm mostly offline until september.

auvipy:
took into in when you are back in september

PR #6230, opened 2020-07-15
test: Fix rare failures in chord error test

maybe-sybr:
## Description

This change fixes a rare failure in the canvas integration tests observed in builds for #6218 

The failure can be reliably reproduced by changing the 3 lambdas in this changeset to a single one like `lambda: True`. This change is a bit odd but works fine and should be more reliable than the sleep which was there previously.

### DIFF
--- a/t/integration/test_canvas.py
+++ b/t/integration/test_canvas.py
@@ -812,7 +812,6 @@ class test_chord:
     def test_chord_on_error(self, manager):
         from celery import states
         from .tasks import ExpectedException
-        import time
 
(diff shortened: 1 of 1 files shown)


maybe-sybr:
We spin on the checks now which makes the test run faster but might be more aggressive than we'd like in CI environments where workers may have reduced resources. I can easily change the `pass` to a small sleep to avoid spinning so hard.

maybe-sybr:
New changeset removes that unused import

PR #6233, opened 2020-07-17
Class-based tasks autoretry

### LINKED ISSUE #6117: Celery class based task is not retrying
I defined a class-based celery task defined as follows:

```python
class MyTask(Task):  
    name = "MY_TASK"
    autoretry_for = (Exception, )
    retry_kwargs = {'max_retries': 5}
    retry_backoff = True
    retry_backoff_max = 700
    retry_jitter = False

    def run(self, *args, **kwargs):
        pass
        # do something
        # return some value

    def on_failure(self, exc, task_id, args, kwargs, einfo):
        # can extend failure method if needed
        super().on_failure(exc, task_id, args, kwargs, einfo)


app.tasks.register(MyTask())
```
### Issue
My task is executing perfectly. But when an exception occurs, the task is not retried

### Expected behaviour
The task should retry

art-vasilyev:
## Description

Currently auto-retry functionality is added to tasks in `Celery._task_from_fun` method and not applied to class based tasks.  Added applying auto-retry behaviour when class based task is registered in tasks registry.

Fixes #6117

### DIFF
--- a/celery/app/autoretry.py
+++ b/celery/app/autoretry.py
@@ -1,0 +1,50 @@
+from vine.utils import wraps
+
+from celery.exceptions import Ignore, Retry
+from celery.utils.time import get_exponential_backoff_interval
+
+
+def add_autoretry_behaviour(task, **options):
+    autoretry_for = tuple(
+        options.get('autoretry_for',
+                    getattr(task, 'autoretry_for', ()))
+    )
+    retry_kwargs = options.get(
+        'retry_kwargs', getattr(task, 'retry_kwargs', {})
+    )
+    retry_backoff = int(
+        options.get('retry_backoff',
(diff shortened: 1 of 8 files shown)


art-vasilyev:
Sure, I'll add them soon

auvipy:
celery/app/autoretry.py:1 at module level:

        D100: Missing docstring in public module

celery/app/autoretry.py:7 in public function `add_autoretry_behaviour`:

        D103: Missing docstring in public function

art-vasilyev:
I'm still not sure that `TaskRegistry.register` is right place for modifying task behaviour, bu can't find where else it can be done.

auvipy:
time out

art-vasilyev:
Yes, hard to debug without working development docker image, looks like need to fix it first

PR #6235, opened 2020-07-21
Let MongoClient handle the authentication to the mongodb backend

wojcikk2903:
Authenticate per database is deprecated:
https://api.mongodb.com/python/current/api/pymongo/database.html#pymongo.database.Database.authenticate

Calling 'authenticate' on a database object causes an issue when trying to
use the same database for both storing the results and as authSource.

### DIFF
--- a/celery/backends/mongodb.py
+++ b/celery/backends/mongodb.py
@@ -269,14 +269,6 @@ class MongoBackend(BaseBackend):
     def _get_database(self):
         conn = self._get_connection()
         db = conn[self.database_name]
-        if self.user and self.password:
-            source = self.options.get(
-                'authsource',
-                self.database_name or 'admin'
-            )
-            if not db.authenticate(self.user, self.password, source=source):
-                raise ImproperlyConfigured(
-                    'Invalid MongoDB username or password.')
         return db
 
     @cached_property
--- a/t/unit/backends/test_mongodb.py
+++ b/t/unit/backends/test_mongodb.py
@@ -275,8 +275,7 @@ class test_MongoBackend:
 
         assert database is mock_database
         assert self.backend.__dict__['database'] is mock_database
-        mock_database.authenticate.assert_called_once_with(
-            MONGODB_USER, MONGODB_PASSWORD, source=self.backend.database_name)
+        mock_database.authenticate.assert_not_called()
 
     @patch('celery.backends.mongodb.MongoBackend._get_connection')
     def test_get_database_no_existing_no_auth(self, mock_get_connection):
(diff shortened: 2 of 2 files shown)


auvipy:
this is enforced from which version?

wojcikk2903:
This is only deprecated for now, since pymongo 3.5. But it causes problems today as it uses a single MongoClient instance to authenticate multiple times. The error we see while trying to write the result is "Another user is already authenticated". The problem disappears if we let the authentication to the client instance.

celery/backends/mongodb.py
```diff
@@ -269,14 +269,6 @@ def __reduce__(self, args=(), kwargs=None):
     def _get_database(self):
         conn = self._get_connection()
         db = conn[self.database_name]
-        if self.user and self.password:
```
thedrow, line 272: Don't we need to check the client's version first or are we dropping support for py-mongo<3.5?

georgepsarakis:
@thedrow I think authentication parameters have already been passed to the [client instance](https://api.mongodb.com/python/current/api/pymongo/mongo_client.html#pymongo.mongo_client.MongoClient) when the connection is created [here](https://github.com/celery/celery/blob/1b32dd6b6ce70af39a8bd4d869dd0d4166806866/celery/backends/mongodb.py#L165).

If the call to `authenticate` was redundant for previous versions of `pymongo` perhaps it can be removed overall. Otherwise, we need to detect the pymongo version and skip the call only for versions after the deprecation. In order to get the pymongo version:

```python
>>> import pymongo
>>> pymongo.version
'3.11.0'
```

I noticed in the documentation that this might be a serious issue with MongoDB server 3.6+, is that right @wojcikk2903 ?

thedrow:
> @thedrow I think authentication parameters have already been passed to the [client instance](https://api.mongodb.com/python/current/api/pymongo/mongo_client.html#pymongo.mongo_client.MongoClient) when the connection is created [here](https://github.com/celery/celery/blob/1b32dd6b6ce70af39a8bd4d869dd0d4166806866/celery/backends/mongodb.py#L165).
> 
> If the call to `authenticate` was redundant for previous versions of `pymongo` perhaps it can be removed overall. Otherwise, we need to detect the pymongo version and skip the call only for versions after the deprecation. In order to get the pymongo version:
> 
> ```python
> >>> import pymongo
> >>> pymongo.version
> '3.11.0'
> ```
> 
> I noticed in the documentation that this might be a serious issue with MongoDB server 3.6+, is that right @wojcikk2903 ?

Since different databases could have different authentication credentials I think that the call was not redundant.
I'm not 100% sure though.

wojcikk2903:
> I noticed in the documentation that this might be a serious issue with MongoDB server 3.6+, is that right @wojcikk2903 ?

@georgepsarakis indeed, writing result ends in an error because `Database` allows only for a single call to authenticate and it happens that the mongodb backend does it several times in case when the result database and `authSource` database are the same.
As for multiple databases, the documentation encourages using a new `MongoClient` instances when creating a connection to a different database:  [link](https://api.mongodb.com/python/current/api/pymongo/database.html#pymongo.database.Database.authenticate)

I will confirm the behavior for pymongo >= 3.3.0 and get back to you.

wojcikk2903:
I've verified that the authentication mechanism doesn't work for pymongo==3.3.0 (with or without the reviewed change, there's a mismatch in kwargs).  With pymongo==3.5.0 the authentication works correctly regardless of the authSource.

thedrow:
> I've verified that the authentication mechanism doesn't work for pymongo==3.3.0 (with or without the reviewed change, there's a mismatch in kwargs). With pymongo==3.5.0 the authentication works correctly regardless of the authSource.

In that case, we can bump the minimum required version of pymongo right?

auvipy (commented):
i guess we can close this?

thedrow:
Yes we can close this.
@wojcikk2903 Thanks for the contribution. Sorry we couldn't use it.

PR #6245, opened 2020-07-28
improv: Make use of ordered sets in Redis opt-in

maybe-sybr:
## Description

This should ensure that there is no breakage between workers with the
code from #6218 and those without, unless the cluster owner specifically
opts into the new behaviour.

Amends #6218

### DIFF
--- a/celery/backends/redis.py
+++ b/celery/backends/redis.py
@@ -410,6 +410,13 @@ class RedisBackend(BaseKeyValueStoreBackend, AsyncBackendMixin):
         # this flag.
         pass
 
+    @cached_property
+    def _chord_zset(self):
+        transport_options = self.app.conf.get(
+            'result_backend_transport_options', {}
+        )
+        return transport_options.get('result_chord_ordered', False)
+
     def on_chord_part_return(self, request, state, result,
                              propagate=None, **kwargs):
         app = self.app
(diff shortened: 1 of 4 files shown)


maybe-sybr:
This change makes the behaviour merged into master in #6218 opt-in and adds unit tests for the behaviour.

I'm a bit unhappy with the unit test code duplication but I didn't spot any test function parameterisation magic which I could use. LMK if there's a better way to do this since I just need to repeat those tests with `{no config, false config, true config}` and confirm the mock calls based on the state.

I've added some documentation for the new behaviour and specifically mention that 4.4.6 and below have the old behaviour, and 4.4.7 will introduce the opt-in behaviour. TBH I think it might be better to revert #6218 and stage that change and this one for 4.5. That would make it easier to justify the change and since it's not breaking between workers now, we shouldn't need to wait until 5.0. I think it would also be sensible to change the default behaviour to opt-out of the ordered set in 5.0 and mark the code path as deprecated for removal in 6.0. I'm not sure what your processes around feature pruning are, so happy to defer on any of this.

ping @auvipy @georgepsarakis @thedrow

maybe-sybr:
@auvipy - Looks like the integration test environment might be unhealthy, causing the travis build to die. Please ping me if you notice that this change breaks things when you re-run it :)

auvipy:
I won't mind adding the fix for a buggy behavior in a minor release as it reduces bug in the existing code.

thedrow:
You can use [`@pytest.mark.parametrize`](https://docs.pytest.org/en/stable/parametrize.html) I think.

thedrow:
> I've added some documentation for the new behaviour and specifically mention that 4.4.6 and below have the old behaviour, and 4.4.7 will introduce the opt-in behaviour. TBH I think it might be better to revert #6218 and stage that change and this one for 4.5. That would make it easier to justify the change and since it's not breaking between workers now, we shouldn't need to wait until 5.0. I think it would also be sensible to change the default behaviour to opt-out of the ordered set in 5.0 and mark the code path as deprecated for removal in 6.0. I'm not sure what your processes around feature pruning are, so happy to defer on any of this.

I agree.
Let's open an issue about that?

graingert:
@maybe-sybr is this related? https://github.com/celery/celery/pull/6263

Issue #6244, opened 2020-07-27
REMAP_SIGTERM=SIGQUIT not working


sayanarijit:
# Checklist

- [x] I have verified that the issue exists against the `master` branch of Celery.
- [x] This has already been asked to the [discussion group](https://groups.google.com/forum/#!forum/celery-users) first.
- [x] I have read the relevant section in the
  [contribution guide](http://docs.celeryproject.org/en/latest/contributing.html#other-bugs)
  on reporting bugs.
- [x] I have checked the [issues list](https://github.com/celery/celery/issues?q=is%3Aissue+label%3A%22Issue+Type%3A+Bug+Report%22+-label%3A%22Category%3A+Documentation%22)
  for similar or identical bug reports.
- [x] I have checked the [pull requests list](https://github.com/celery/celery/pulls?q=is%3Apr+label%3A%22PR+Type%3A+Bugfix%22+-label%3A%22Category%3A+Documentation%22)
  for existing proposed fixes.
- [x] I have checked the [commit log](https://github.com/celery/celery/commits/master)
  to find out if the bug was already fixed in the master branch.
- [x] I have included all related issues and possible duplicate issues
  in this issue (If there are none, check this box anyway).

## Mandatory Debugging Information

- [x] I have included the output of ``celery -A proj report`` in the issue.
    (if you are not able to do this, then at least specify the Celery
     version affected).
- [x] I have verified that the issue exists against the `master` branch of Celery.
- [x] I have included the contents of ``pip freeze`` in the issue.
- [x] I have included all the versions of all the external dependencies required
  to reproduce this bug.

## Optional Debugging Information

- [ ] I have tried reproducing the issue on more than one Python version
  and/or implementation.
- [ ] I have tried reproducing the issue on more than one message broker and/or
  result backend.
- [ ] I have tried reproducing the issue on more than one version of the message
  broker and/or result backend.
- [ ] I have tried reproducing the issue on more than one operating system.
- [ ] I have tried reproducing the issue on more than one workers pool.
- [ ] I have tried reproducing the issue with autoscaling, retries,
  ETA/Countdown & rate limits disabled.
- [ ] I have tried reproducing the issue after downgrading
  and/or upgrading Celery and its dependencies.

#### Related Issues

- https://github.com/celery/celery/issues/2839
- https://github.com/celery/celery/issues/4550

#### Possible Duplicates

- None

## Environment & Settings

**Celery version**:

<details>
<summary><b><code>celery report</code> Output:</b></summary>
[...]

PR #6249, opened 2020-07-28
Fix REMAP_SIGTERM=SIGQUIT not working

sayanarijit:
Heroku should do cold shutdown SIGTERM when we remap SIGTERM to SIGQUIT.

Heroku article: https://devcenter.heroku.com/articles/celery-heroku#using-remap_sigterm

Ref: https://github.com/celery/celery/issues/6244


t/unit/bin/test_worker.py
```diff
@@ -8,6 +8,7 @@
 from billiard.process import current_process
 from case import Mock, mock, patch, skip
 from kombu import Exchange, Queue
+from six.moves import reload_module
```
auvipy, line 11: isn't it available in the five module?

sayanarijit: Right. Updated.

sayanarijit: Looks like five isn't there in 2.7...

auvipy:
can you check the import error?

t/unit/bin/test_worker.py
thedrow: No. Use five.
I use it later to find code to remove for Celery 5.

sayanarijit:
> can you check the import error?

Import error happens in Py2.7 because celery.five is missing in there. Any suggestion how to fix it without using six?

thedrow:
> > can you check the import error?
> 
> Import error happens in Py2.7 because celery.five is missing in there. Any suggestion how to fix it without using six?

That's... odd.

You should be able to import it from Python 2.7. Try locally.

sayanarijit:
Ok `from celery.five import ...` is the way to import it...

PR #6252, opened 2020-07-28
use raise with context instead of reraise

graingert:

### DIFF
--- a/celery/backends/base.py
+++ b/celery/backends/base.py
@@ -28,7 +28,7 @@ from celery._state import get_current_task
 from celery.exceptions import (ChordError, ImproperlyConfigured,
                                NotRegistered, TaskRevokedError, TimeoutError,
                                BackendGetMetaError, BackendStoreError)
-from celery.five import PY3, items, reraise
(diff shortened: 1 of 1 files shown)


celery/backends/base.py
```diff
@@ -453,10 +454,8 @@ def store_result(self, task_id, result, state,
                             self.max_sleep_between_retries_ms, True) / 1000
                         self._sleep(sleep_amount)
                     else:
-                        reraise(
-                            BackendStoreError,
+                        raise_with_context(
                             BackendStoreError("failed to store result on the backend", task_id=task_id, state=state),
-                            traceback,
```
graingert, line 459: @auvipy I've been looking at this code, and reraise-ing this traceback is not correct.

I think the intention was to include the context of the failed request from the retry - not the failure of the actual task/result.

graingert: there's already a utility to re-raise an exception from another context in serialization

graingert:
lint fixes moved to https://github.com/celery/celery/pull/6253

PR #6257, opened 2020-07-29
pin pytest versions to semver range appropriate for python_version + upgrade pytest to v6 on py3

graingert:


auvipy:
some tests might need an update in latest pytest

graingert:
> some tests might need an update in latest pytest

@auvipy only 1!

PR #6260, opened 2020-07-29
apply flake8 to all py files

graingert:


thedrow:
I'm not so sure we should do this.
What's the reasoning behind it?

graingert:
@thedrow we don't want a syntax error in `setup.py,examples/*,docs/*,extra/*` and it looks like people have been manually applying flake8 to these modules anyway, so it's good to enforce it

PR #6261, opened 2020-07-29
remove pytest-rerunfailures

graingert:
we should use @pytest.mark.flaky on the tests that need it instead


thedrow:
Wait why would we want to remove this plugin?

graingert:
It seems redundant to have this and flakey, maybe we should remove flakey
instead?

thedrow:
I thought I removed our own `flaky` marker implementation...
pytest-rerunfailures have their own `flaky` marker.

graingert:
@thedrow yep you're right, I just misread the code

PR #6268, opened 2020-07-30
switch to official amazon/dynamodb-local, use httping/expect to verify service statuses

graingert:
https://github.com/dwmkerr/docker-dynamodb/pull/18

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -86,7 +86,7 @@ before_install:
     - |
           if [[ "$TOXENV" == *rabbitmq ]]; then
               docker run -d -p 5672:5672 -p 15672:15672 rabbitmq:3.8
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -103,13 +103,13 @@ before_install:
           fi
     - |
           if [[ "$TOXENV" == *dynamodb ]]; then
-              docker run -d -p 8000:8000 dwmkerr/dynamodb:38 -inMemory
-              while ! nc -zv 127.0.0.1 8000; do sleep 10; done
```
graingert, line 107: so it turns out this nc doesn't work with docker - because `nc -zv` will pass with docker listening and not the actual service.

I now use httping/expect to actually communicate with the services to see if they are up yet

graingert:
@auvipy ok this one is ready for review

PR #6270, opened 2020-08-01
configure 300 second timeout to prevent CI hanging

graingert:

### DIFF
--- a/celery/backends/redis.py
+++ b/celery/backends/redis.py
@@ -429,21 +429,13 @@ class RedisBackend(BaseKeyValueStoreBackend, AsyncBackendMixin):
         jkey = self.get_key_for_group(gid, '.j')
         tkey = self.get_key_for_group(gid, '.t')
         result = self.encode_result(result, state)
+        encoded = self.encode([1, tid, state, result])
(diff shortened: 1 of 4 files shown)


graingert:
I don't think https://github.com/celery/celery/pull/6255 actually did anything as when `@pytest.mark.flaky` restarts a test it prints the message in the console so should not trigger travis killing the build

graingert:
@auvipy those are lint errors on the default branch

.travis.yml
```diff
@@ -10,9 +10,9 @@ python:
 os:
     - linux
 stages:
+  - lint
```
auvipy, line 13: thanks but can you revert this?

graingert: @auvipy I've done that, why though?

auvipy: we used to run lint first for many years. but later we find out that, many initial contributors/comparatively irregular contributors do not get the benefit of running their code through the unit & integration test for lint errors and the don't get the benefit of early feedback loop on their actual code changes. so we come to the conclusion that we should let the unit and integration tests first.

PR #6271, opened 2020-08-01
create github action for windows

graingert:
an action to potentially replace appveyor, which seems a bit unreliable on cloning


graingert:
@auvipy sadly I think it needs merging into the default branch before it will run the actions on PRs

graingert:
ah you can see the windows status here: https://github.com/graingert/celery/pull/1/checks?check_run_id=935657081

auvipy:
this is better than appveyor

graingert:
> this is better than appveyor

Yeah, I think so

graingert:
@auvipy I can remove appveyor in a follow-up

.github/workflows/windows.yml
```diff
@@ -0,0 +1,36 @@
...
+          python -m pip install --upgrade pip setuptools wheel
+          python -m pip install --requirement requirements/pkgutils.txt
+
+      - name: Test
+        run: tox -vv -e ${{ matrix.python-version }}-unit
+        
+      - name: Build
+        run: python setup.py bdist_wheel
+
+      - uses: actions/upload-artifact@v2
```
graingert, line 33: @auvipy do we need this step? celery ships universal wheels so it seems redundant

graingert:
@auvipy also you get 40 concurrent builds on decent hardware, rather than 4 potatoes

thedrow:
While this is good work, we haven't had the discussion on where to switch our CI infrastructure to.

I'm not very happy with Travis-CI or AppVeyor as of late. I think none of us were. 😐 
I considered Azure Pipelines as a potential candidate as well.

Concurrency wise, Github Actions provides 20 concurrent jobs but limits the number of API requests to 1000 API requests per hour.
I guess that's enough for now.

I agree that this is a step in the right direction but to make this permanent in all of our repositories we need a CEP.

See https://github.com/celery/ceps/issues/15. I'm willing to shepherd this effort.

graingert:
@thedrow it feels like Github Actions is the spiritual successor to azure pipelines: https://docs.github.com/en/actions/migrating-to-github-actions/migrating-from-azure-pipelines-to-github-actions

graingert:
> but to make this permanent in all of our repositories we need a CEP.

I've only made this a Windows job to celery/celery because only the AppVeyor celery/celery job is impacting me directly - I do think it's worth switching wholesale but I think this PR is a step forward either way.

.github/workflows/windows.yml
```diff
@@ -0,0 +1,36 @@
+name: Windows
+on: [push, pull_request]
+jobs:
+  build:
+    runs-on: windows-2019
+    strategy:
+      matrix:
+        python-version: [3.5, 3.6, 3.7, 3.8]
```
auvipy, line 8: python 3.6++ only

.github/workflows/windows.yml
auvipy: seems redundant

auvipy (changes requested):
should we try this as test purpose?

.github/workflows/windows.yml
graingert:
```suggestion
        python-version: [3.6, 3.7, 3.8, 3.9]
```

thedrow:
pycurl hasn't released a wheel to that version. They haven't been consistent with binary releases.

@graingert Would you mind setting it up so that if a wheel is not present in CI, it will still build the extension?

.github/workflows/windows.yml
```diff
@@ -0,0 +1,36 @@
+name: Windows
+on: [push, pull_request]
+jobs:
+  build:
+    runs-on: windows-2019
+    strategy:
+      matrix:
+        python-version: [3.6, 3.7, 3.8, 3.9]
```
graingert, line 8:
```suggestion
        python-version: [3.6, 3.7, 3.8]
```

graingert:
> @graingert Would you mind setting it up so that if a wheel is not present in CI, it will still build the extension?

looks like a huge pain to do. I think removing pycurl and using a modern async client (httpx) is the only way here

graingert:
I'll xfail all these remaining failures and rebase away my mistakes shortly

graingert:
@auvipy I made an attempt to work around pycurl here and support py3.9, but it drifted too much from the original scope/goal of this PR.

This PR is actually ready for review and I don't plan to touch it again unless you request changes.

I'm continuing the py3.9+ work here: https://github.com/celery/celery/pull/6868

.github/workflows/python-package.yml
```diff
@@ -20,15 +20,24 @@ on:
...
-    continue-on-error: ${{ startsWith(matrix.python-version, '3.10.0-beta.') }}
+        os: ["ubuntu-20.04", "windows-2019"]
+        exclude:
+          - os: windows-2019
+            python-version: "pypy3"
+          - os: windows-2019
+            python-version: "3.10.0-beta.4"
+          - os: windows-2019
+            python-version: "3.9"
+    continue-on-error: ${{ startsWith(matrix.python-version, '3.10.0-beta') }}
```
graingert, line 36:
```suggestion
    continue-on-error: ${{ startsWith(matrix.python-version, '3.10.0-beta.') }}
```

thedrow:
> 
> 
> > @graingert Would you mind setting it up so that if a wheel is not present in CI, it will still build the extension?
> 
> looks like a huge pain to do. I think removing pycurl and using a modern async client (httpx) is the only way here

If you feel like refactoring our SQS broker, I'll walk you through it. I'd love to get rid of that dependency.

thedrow (approved):

PR #6272, opened 2020-08-01
pytest-timeout fixes - apply only to flaky tests, ignore pytest_timeout in lingering thread check

graingert:

### DIFF
--- a/t/unit/worker/test_worker.py
+++ b/t/unit/worker/test_worker.py
@@ -793,7 +793,6 @@ class test_WorkController(ConsumerCase):
         assert worker.autoscaler
 
     @skip.if_win32()
-    @pytest.mark.nothreads_not_lingering
     @mock.sleepdeprived(module=autoscale)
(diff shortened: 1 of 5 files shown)


t/unit/worker/test_worker.py
```diff
@@ -793,7 +793,6 @@ def test_with_autoscaler(self):
         assert worker.autoscaler
 
     @skip.if_win32()
-    @pytest.mark.nothreads_not_lingering
```
graingert, line 796: seems like it was introduced here but I can't see where the mark is processed https://github.com/celery/celery/commit/19f63e849b0ce2f9ea0f8c9c1696ca6a0e8338fa#diff-b7f2202289ddf07518813ae257f11290R795

graingert: looks like there was an attempt to remove this previously: https://github.com/celery/celery/pull/5729

PR #6282, opened 2020-08-03
lower max-line-length to 117

graingert:
Anything above 120 is almost universally considered egregious, [117 is now the longest line in the codebase below that limit](https://github.com/graingert/celery/blob/be5e8c0f33997b140547d54fb0084a1653178fc9/celery/backends/base.py#L458)

I think we could safely lower the line length further as that line would look much nicer black-style:
```python
                        raise_with_context(
                            BackendStoreError(
                                "failed to store result on the backend",
                                task_id=task_id,
                                state=state,
                            ),
                        )
```

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -12,7 +12,7 @@ all_files = 1
 [flake8]
 # classes can be lowercase, arguments and variables can be uppercase
 # whenever it makes the code more readable.
-max-line-length = 128
+max-line-length = 125
 extend-ignore =
     D102,  # Missing docstring in public method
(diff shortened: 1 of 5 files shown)


graingert:
@auvipy do you have a target line length? I think the black 88 limit is reasonable

auvipy:
I would prefer 88 in the near future if other team members agree. Django uses 117-120 I guess?

graingert:
Django are moving to 88 https://github.com/django/deps/blob/master/accepted/0008-black.rst#specification

PR #6298, opened 2020-08-12
Retry after race during schema creation in database backend

### LINKED ISSUE #6296: Database backend race condition during table creation
# Checklist

  on reporting bugs.
  for existing proposed fixes.

     version affected).
- [ ] I have included the contents of ``pip freeze`` in the issue. **N/A**
  to reproduce this bug. **N/A**

  and/or implementation.
  result backend.
  broker and/or result backend.

#### Related Issues

- https://github.com/celery/celery/issues/4653

#### Possible Duplicates

- None

## Environment & Settings

**Celery version**: 4.4.7 (cliffs)

<details>
<p>

```
software -> celery:4.4.7 (cliffs) kombu:4.6.11 py:3.8.2
            billiard:3.6.3.0 py-amqp:2.6.1
            kernel version:16.7.0 imp:CPython
settings -> transport:amqp results:db+postgresql://<redacted>

include: ['redacted']
accept_content: ['redacted-custom']
database_table_names: {
 'group': 'celery_group', 'task': 'celery_task'}
result_serializer: 'redacted-custom'
task_serializer: 'redacted-custom'
task_track_started: True
broker_url: 'amqp://<redacted>'
result_backend: 'db+postgresql://<redacted>'
```

</p>
</details>

# Steps to Reproduce

When celery uses a database result backend, the following line can be called multiple times from different processes:

https://github.com/celery/celery/blob/9a6c2923e859b6993227605610255bd632c1ae68/celery/backends/database/session.py#L56

This is a race condition because SQLAlchemy first checks if the tables/sequences exist and then tries to create them. It causes errors like this (at least on PostgreSQL):

```
  File "/usr/local/lib/python3.7/site-packages/redacted.py", line 168, in _redacted
    result = async_result.get()
  File "/usr/local/lib/python3.7/site-packages/celery/result.py", line 226, in get
  File "/usr/local/lib/python3.7/site-packages/celery/result.py", line 342, in maybe_throw
    self.throw(value, self._to_remote_traceback(tb))
  File "/usr/local/lib/python3.7/site-packages/celery/result.py", line 335, in throw
    self.on_ready.throw(*args, **kwargs)
  File "/usr/local/lib/python3.7/site-packages/vine/promises.py", line 244, in throw
    reraise(type(exc), exc, tb)
  File "/usr/local/lib/python3.7/site-packages/vine/five.py", line 195, in reraise
    raise value
Exception: <class 'sqlalchemy.exc.IntegrityError'>(('(psycopg2.errors.UniqueViolation) duplicate key value violates unique constraint "pg_type_typname_nsp_index"\nDETAIL:  Key (typname, typnamespace)=(taskset_id_sequence, 2200) already exists.\n',))
```

One workaround is  to force the table creation ahead of time as was proposed by a user in the issue I linked: https://github.com/celery/celery/issues/4653#issuecomment-400029147.

I think Celery should handle this itself. A possible solution would catch `IntegrityError` and try again until `create_all` succeeds. (Perhaps with a limited number of retries and with sleeps compared to this snippet):

```python
    def prepare_models(self, engine):
        from sqlalchemy.exc import IntegrityError
        if not self.prepared:
            while True:
                try:
                    ResultModelBase.metadata.create_all(engine)
                except IntegrityError:
                    continue
                else:
                    break
            self.prepared = True
```

This example doesn't use celery at all, but shows that calling create_all in multiple processes can cause the error. It's a race condition, so you might need to try it multiple times or play around with the number of processes:

<details>
<p>

Requires a local postgres, and this database must be created:

```
createdb racetest
```

```python
from concurrent.futures import ProcessPoolExecutor, as_completed

from sqlalchemy import Column, Integer, Table, MetaData, create_engine

metadata = MetaData()

tbl1 = Table('tbl1', metadata, Column('id', Integer, primary_key=True))


def create_all(url):
    engine = create_engine(url)
    metadata.create_all(bind=engine)


def main():
    url = 'postgresql:///racetest'
    engine = create_engine(url)
    
    # Make sure schema is empty before we start
    metadata.drop_all(bind=engine)

    with ProcessPoolExecutor(max_workers=50) as executor:
        futures = []
        for _ in range(50):
            future = executor.submit(create_all, url)
            futures.append(future)

        for fut in as_completed(futures):
            fut.result()


if __name__ == '__main__':
    main()

```

</p>
</details>

RazerM:
*Note*: Before submitting this pull request, please review our [contributing guidelines](https://docs.celeryproject.org/en/master/contributing.html).

## Description

Fixes #6296

This race condition does not commonly present, since the schema creation only needs to happen once per database. It's more likely to appear in e.g. a test suite that uses a new database each time.

For context of the sleep times I chose, the schema creation takes ~50 ms on my laptop.

I did a simulated test run of 50 concurrent calls to `MetaData.create_all` repeated 200 times and the number of retries was:

- 0 retries: 8717x
- 1 retry:   1279x
- 2 retries  4x

### DIFF
--- a/celery/backends/database/session.py
+++ b/celery/backends/database/session.py
@@ -14,6 +14,8 @@ ResultModelBase = declarative_base()
 
 __all__ = ('SessionManager',)
 
+PREPARE_MODELS_MAX_RETRIES = 10
+
 
 def _after_fork_cleanup_session(session):
     session._after_fork()
@@ -59,13 +61,12 @@ class SessionManager:
             # create them, which is a race condition. If it raises an error
             # in one iteration, the next may pass all the existence checks
             # and the call will succeed.
-            max_retries = 10
             retries = 0
             while True:
                 try:
                     ResultModelBase.metadata.create_all(engine)
                 except DatabaseError:
-                    if retries < max_retries:
+                    if retries < PREPARE_MODELS_MAX_RETRIES:
                         sleep_amount_ms = get_exponential_backoff_interval(
                             10, retries, 1000, True
                         )
--- a/celery/backends/database/session.py
+++ b/celery/backends/database/session.py
@@ -1,9 +1,14 @@
 """SQLAlchemy session."""
+import time
+
 from kombu.utils.compat import register_after_fork
 from sqlalchemy import create_engine
+from sqlalchemy.exc import DatabaseError
 from sqlalchemy.ext.declarative import declarative_base
 from sqlalchemy.orm import sessionmaker
 from sqlalchemy.pool import NullPool
+
+from celery.utils.time import get_exponential_backoff_interval
 
 ResultModelBase = declarative_base()
 
@@ -50,7 +55,26 @@ class SessionManager:
 
     def prepare_models(self, engine):
         if not self.prepared:
-            ResultModelBase.metadata.create_all(engine)
+            # SQLAlchemy will check if the items exist before trying to
+            # create them, which is a race condition. If it raises an error
+            # in one iteration, the next may pass all the existence checks
+            # and the call will succeed.
+            max_retries = 10
+            retries = 0
+            while True:
+                try:
+                    ResultModelBase.metadata.create_all(engine)
+                except DatabaseError:
+                    if retries < max_retries:
+                        sleep_amount_ms = get_exponential_backoff_interval(
+                            10, retries, 1000, True
+                        )
+                        time.sleep(sleep_amount_ms / 1000)
+                        retries += 1
+                    else:
+                        raise
+                else:
+                    break
             self.prepared = True
 
     def session_factory(self, dburi, **kwargs):
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -277,3 +277,4 @@ Kyle Johnson, 2019/09/23
 Dipankar Achinta, 2019/10/24
 Sardorbek Imomaliev, 2020/01/24
 Maksym Shalenyi, 2020/07/30
+Frazer McLean, 2020/09/29
--- a/t/unit/backends/test_database.py
+++ b/t/unit/backends/test_database.py
@@ -13,7 +13,8 @@ pytest.importorskip('sqlalchemy')
 from celery.backends.database import (DatabaseBackend, retry, session,  # noqa
                                       session_cleanup)
 from celery.backends.database.models import Task, TaskSet  # noqa
-from celery.backends.database.session import SessionManager  # noqa
+from celery.backends.database.session import (  # noqa
+    PREPARE_MODELS_MAX_RETRIES, ResultModelBase, SessionManager)
 from t import skip  # noqa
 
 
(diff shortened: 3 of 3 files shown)


RazerM:
The CI failure was unrelated to my changes. I've rebased onto the latest master to see if it passes now.

auvipy (changes requested):
can you add unit test for the suggested changes?

RazerM:
Given that it's a race condition I'm not sure how I would make a test for it. (see the Minimally Reproducible Test Case in #6296).

Would you like a test which ensures that the loop does eventually terminate if `ResultModelBase.metadata.create_all(engine)` is mocked to always raise a `DatabaseError`?

auvipy:
> Given that it's a race condition I'm not sure how I would make a test for it. (see the Minimally Reproducible Test Case in #6296).
> 
> Would you like a test which ensures that the loop does eventually terminate if `ResultModelBase.metadata.create_all(engine)` is mocked to always raise a `DatabaseError`?

that would be a very good started point

celery/backends/database/session.py
```diff
@@ -50,7 +55,26 @@ def create_session(self, dburi, short_lived_sessions=False, **kwargs):
...
-            ResultModelBase.metadata.create_all(engine)
+            # SQLAlchemy will check if the items exist before trying to
+            # create them, which is a race condition. If it raises an error
+            # in one iteration, the next may pass all the existence checks
+            # and the call will succeed.
+            max_retries = 10
+            retries = 0
+            while True:
+                try:
+                    ResultModelBase.metadata.create_all(engine)
```
georgepsarakis, line 66: So subsequent calls to `.create_all` will succeed, even if the schema has been already created?

celery/backends/database/session.py
```diff
@@ -50,7 +55,26 @@ def create_session(self, dburi, short_lived_sessions=False, **kwargs):
 
     def prepare_models(self, engine):
         if not self.prepared:
-            ResultModelBase.metadata.create_all(engine)
+            # SQLAlchemy will check if the items exist before trying to
+            # create them, which is a race condition. If it raises an error
+            # in one iteration, the next may pass all the existence checks
+            # and the call will succeed.
+            max_retries = 10
```
georgepsarakis, line 62: Should we define this threshold as a constant?

georgepsarakis (commented):
Apart from some minor questions, changes seem fine.

celery/backends/database/session.py
RazerM: Yes, which is currently what happens since each celery worker calls `.create_all` unconditionally. See the checkfirst arg [in the docs](https://docs.sqlalchemy.org/en/13/core/metadata.html#sqlalchemy.schema.MetaData.create_all).

celery/backends/database/session.py
RazerM: I've made this a constant so that the test can see it

RazerM:
I have added the suggested test

georgepsarakis (approved):
Nice work! Thank you for your contribution!

georgepsarakis:
@RazerM just noticed some failures in the Travis builds for the new test case: https://travis-ci.org/github/celery/celery/jobs/731370177

Would you mind having a look?

RazerM:
@georgepsarakis Thanks! Looks like I missed an import when I rebased onto master. Should be fixed now

auvipy (approved):

Issue #6285, opened 2020-08-04
Basic Django config files break upgrading from 4.4.6 to 4.4.7


chrisconlan:
The basic tutorial setup outlined here: https://docs.celeryproject.org/en/latest/django/first-steps-with-django.html now fails to locate set the `broker` or `CELERY_BROKER_URL` parameter. Celery just ignores the settings and defaults to searching for a host at `amqp://guest:**@127.0.0.1:5672//:`, which is not my host of choice.

I'm ignoring the issues template because this is a very basic regression that the last committor can fix quickly.

pg1671:
4.4.7 broke my django site.  Just reverted to 4.4.6.

Did not investigate further but jobs were not being processed. Assume it is related to this issue.

rdmrocha:
Another confirmation for this issue. Django3 project stoped being able to connect by ignoring settings.

yoch:
Same issue for me (no Django, but settings was ignored as well).

kptac:
Same here. Reverting back to 4.4.6

thedrow:
@auvipy Please address this.
Did we do something I'm not aware of?

dratchkov:
The symptom in my case is that messages are received, but rejected because the called task is not registered.
What was really confusing is that 'flower', sees the list of tasks.

But yes, 4.4.6 resolves this.

nkuttler:
The pre 4.0 style config doesn't seem to be supported any more in 4.4.7, despite the docs saying otherwise. We also reverted.

gargantuanprism:
confirmed, but not with django. if i have env vars set in my shell, then run `celery worker` interactively, it's fine. as soon as i add `--detach`, it seems to ignore my broker URL env var. downgraded to 4.4.6 and it works again.

fnk0c:
Same here. Whenever I add `-D` it simple ignores configuration file. Also when running `celery worker --help` it ignores the `--help` and starts Celery.

auvipy:
> @auvipy Please address this.
> Did we do something I'm not aware of?

I dont think so but will look into it on next Monday Hopefully.

auvipy:
> The basic tutorial setup outlined here: https://docs.celeryproject.org/en/latest/django/first-steps-with-django.html now fails to locate set the `broker` or `CELERY_BROKER_URL` parameter. Celery just ignores the settings and defaults to searching for a host at `amqp://guest:**@127.0.0.1:5672//:`, which is not my host of choice.
> 
[...]

PR #6305, opened 2020-08-17
CeleryCommand: Add regression test

garw:
Adds a test to check that the application parameter
handed to celery multi is then passed on to the actual
implementation of the multi command.

## Description

This is a regression test to check for the in issue #6285.


auvipy:
this is no longer valid in celery 5.

PR #6308, opened 2020-08-19
fixed flake8 errors in utils/encoding

auvipy:


thedrow:
That module is deprecated and was removed.
The other change which slided into this PR does not change anything as the build is passing.

PR #6316, opened 2020-08-27
Remove code deemed deprecated for 4.0 and removed for 5.0.

clokep:
This was found be grepping for `@deprecated`, it just nukes code that says it is going to be removed in 5.0.

I'm unsure if I'm targeting the right branch here, but let me know if I'm not.


clokep:
@thedrow this is another example of stuff that was meant to be removed for 5.0.

I think the codecov failures here are bogus since we're removing code...

auvipy (approved):

clokep:
Thanks @auvipy!

Issue #6225, opened 2020-07-13
Can not work in gevent/eventlet model


Seumi:
# Checklist

- [x] I have verified that the issue exists against the `master` branch of Celery.

- [ ] This has already been asked to the [discussion group](https://groups.google.com/forum/#!forum/celery-users) first.
- [x] I have read the relevant section in the
  [contribution guide](http://docs.celeryproject.org/en/latest/contributing.html#other-bugs)
  on reporting bugs.
- [x] I have checked the [issues list](https://github.com/celery/celery/issues?q=is%3Aissue+label%3A%22Issue+Type%3A+Bug+Report%22+-label%3A%22Category%3A+Documentation%22)
  for similar or identical bug reports.
- [x] I have checked the [pull requests list](https://github.com/celery/celery/pulls?q=is%3Apr+label%3A%22PR+Type%3A+Bugfix%22+-label%3A%22Category%3A+Documentation%22)
  for existing proposed fixes.
- [x] I have checked the [commit log](https://github.com/celery/celery/commits/master)
  to find out if the bug was already fixed in the master branch.
- [x] I have included all related issues and possible duplicate issues
  in this issue (If there are none, check this box anyway).

## Mandatory Debugging Information

- [x] I have included the output of ``celery -A proj report`` in the issue.
    (if you are not able to do this, then at least specify the Celery
     version affected).
- [x] I have verified that the issue exists against the `master` branch of Celery.
- [x] I have included the contents of ``pip freeze`` in the issue.
- [x] I have included all the versions of all the external dependencies required
  to reproduce this bug.

## Optional Debugging Information

- [x] I have tried reproducing the issue on more than one Python version
  and/or implementation.
- [x] I have tried reproducing the issue on more than one message broker and/or
  result backend.
- [x] I have tried reproducing the issue on more than one version of the message
  broker and/or result backend.
- [x] I have tried reproducing the issue on more than one operating system.
- [x] I have tried reproducing the issue on more than one workers pool.
- [x] I have tried reproducing the issue with autoscaling, retries,
  ETA/Countdown & rate limits disabled.
- [x] I have tried reproducing the issue after downgrading
  and/or upgrading Celery and its dependencies.

#### Related Issues

- #4758
- #6111 

#### Possible Duplicates

- None

## Environment & Settings

**Celery version**: 4.4.6

<details>
<summary><b><code>celery report</code> Output:</b></summary>
<p>

```
[...]

PR #5718, opened 2019-09-08
Refactor CLI to use Click instead of our custom argparse based framework

thedrow:
This work is based on our decision in celery/ceps#19 to replace our custom implementation of the CLI using argparse with Click.
This change is pending the approval of celery/ceps#19 and of celery/ceps#25.

### **TODO:**
- [ ] Unit Tests
- [x] [Preload Options](https://docs.celeryproject.org/en/latest/userguide/extending.html#preload-options)
- [x] `celery worker` command
  - [x] Accept all previous arguments
  - [x] Detach main process
  - [x] Drop privileges
  - [x] Configure worker from command line
  - [x] Extract defaults from app configuration
  - [x] User options
  - [x] Documentation
- [x] `celery events` command
  - [x] Accept all previous arguments
  - [x] User options
  - [x] Documentation
- [ ] `celery beat` command
  - [x] Accept all previous arguments
  - [x] Detach main process
  - [x] Drop privileges
  - [x] Configure beat from command line
  - [ ] Extract defaults from app configuration
  - [x] User options
  - [x] Documentation
- [x] `celery shell` command
  - [x] Accept all previous arguments
  - [x] Documentation
- [x] `celery multi` command
  - [x] Accept all previous arguments
  - [x] Documentation
- [x] `celery amqp` command
  - [x] Accept all previous arguments
  - [x] Documentation
- [x] `celery status` command
  - [x] Accept all previous arguments
  - [x] Documentation
- [x] `celery inspect active` command
  - [x] Accept all previous arguments
  - [x] Documentation
- [x] `celery inspect active_queues` command
  - [x] Accept all previous arguments
  - [x] Documentation
- [x] `celery inspect clock` command
  - [x] Accept all previous arguments
  - [x] Documentation
- [x] `celery inspect conf` command
  - [x] Accept all previous arguments
  - [x] Documentation
- [x] `celery inspect memdump` command
  - [x] Accept all previous arguments
  - [x] Documentation
- [x] `celery inspect memsample` command
  - [x] Accept all previous arguments
  - [x] Documentation
- [x] `celery inspect objgraph` command
  - [x] Accept all previous arguments
  - [x] Documentation
- [x] `celery inspect ping` command
  - [x] Accept all previous arguments
  - [x] Documentation
- [x] `celery inspect query_task` command
  - [x] Accept all previous arguments
  - [x] Documentation
- [x] `celery inspect registered` command
  - [x] Accept all previous arguments
  - [x] Documentation
- [x] `celery inspect report` command
[...]

PR #6322, opened 2020-08-30
calling eventlib/gevent patching

avikam:
## Description

eventlet gPidbox is not running in 5.0.0rc1, as described here: [#6225](https://github.com/celery/celery/issues/6225)

Following the move to Click ([https://github.com/celery/celery/pull/5718](5718)), I noticed that the patching code was commented out.
I'm not sure if the proposed way is a valid way to port the import patching, or the maintainers had other design in mind, but this fix is working for me.

### DIFF
--- a/celery/bin/base.py
+++ b/celery/bin/base.py
@@ -47,8 +47,7 @@ class CLIContext:
     def ERROR(self):
         return self.style("ERROR", fg="red", bold=True)
 
-    @property
-    @functools.lru_cache(None)
+    @cached_property
     def app(self):
         return self._app() or get_current_app()
 
--- a/celery/bin/base.py
+++ b/celery/bin/base.py
@@ -48,9 +48,9 @@ class CLIContext:
 
     @cached_property
     def app(self):
-        """Instantiate an app if it wasn't already and return it. This lazy
+        """Instantiate the app if it wasn't already and return it. This lazy
         approach gives opportunity for further initializations before the app
-        is loaded."""
+        is imported."""
         return self._app_factory() or get_current_app()
 
     def style(self, message=None, **kwargs):
--- a/celery/bin/base.py
+++ b/celery/bin/base.py
@@ -1,5 +1,4 @@
 """Click customizations for Celery."""
-import functools
 import json
 from collections import OrderedDict
 from pprint import pformat
@@ -33,7 +32,7 @@ class CLIContext:
 
     def __init__(self, app, no_color, workdir, quiet=False):
         """Initialize the CLI context."""
-        self._app = app
+        self._app_factory = app
         self.no_color = no_color
         self.quiet = quiet
         self.workdir = workdir
@@ -49,7 +48,10 @@ class CLIContext:
 
     @cached_property
     def app(self):
-        return self._app() or get_current_app()
+        """Instantiate an app if it wasn't already and return it. This lazy
+        approach gives opportunity for further initializations before the app
+        is loaded."""
+        return self._app_factory() or get_current_app()
 
     def style(self, message=None, **kwargs):
         if self.no_color:
--- a/celery/bin/base.py
+++ b/celery/bin/base.py
@@ -1,4 +1,5 @@
 """Click customizations for Celery."""
+import functools
 import json
 from collections import OrderedDict
 from pprint import pformat
@@ -32,7 +33,7 @@ class CLIContext:
 
     def __init__(self, app, no_color, workdir, quiet=False):
         """Initialize the CLI context."""
-        self.app = app or get_current_app()
+        self._app = app
         self.no_color = no_color
         self.quiet = quiet
         self.workdir = workdir
@@ -45,6 +46,11 @@ class CLIContext:
     @cached_property
     def ERROR(self):
         return self.style("ERROR", fg="red", bold=True)
+
+    @property
+    @functools.lru_cache(None)
+    def app(self):
+        return self._app() or get_current_app()
 
     def style(self, message=None, **kwargs):
         if self.no_color:
--- a/celery/bin/celery.py
+++ b/celery/bin/celery.py
@@ -31,10 +31,12 @@ class App(ParamType):
     name = "application"
 
     def convert(self, value, param, ctx):
-        try:
-            return find_app(value)
-        except (ModuleNotFoundError, AttributeError) as e:
-            self.fail(str(e))
+        def lazy():
+            try:
+                return find_app(value)
+            except (ModuleNotFoundError, AttributeError) as e:
+                self.fail(str(e))
+        return lazy
 
 
 APP = App()
@@ -107,9 +109,9 @@ def celery(ctx, app, broker, result_backend, loader, config, workdir,
     ctx.obj = CLIContext(app=app, no_color=no_color, workdir=workdir, quiet=quiet)
 
     # User options
-    worker.params.extend(ctx.obj.app.user_options.get('worker', []))
-    beat.params.extend(ctx.obj.app.user_options.get('beat', []))
-    events.params.extend(ctx.obj.app.user_options.get('events', []))
+    # worker.params.extend(ctx.obj.app.user_options.get('worker', []))
+    # beat.params.extend(ctx.obj.app.user_options.get('beat', []))
+    # events.params.extend(ctx.obj.app.user_options.get('events', []))
 
 
 @celery.command(cls=CeleryCommand)
--- a/celery/bin/celery.py
+++ b/celery/bin/celery.py
@@ -31,12 +31,12 @@ class App(ParamType):
     name = "application"
 
     def convert(self, value, param, ctx):
-        def lazy():
+        def factory():
             try:
                 return find_app(value)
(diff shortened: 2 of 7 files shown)


avikam:
looks like it's too late, i got a max-recursion error trying to open an ssl connection.
drafting for now

auvipy:
seems lazy app passing

avikam:
> seems lazy app passing

yes, now i'm thinking how can I port the `user_options` feature, since it requires instantiating the app on `celery` command which might be too early. (https://github.com/celery/celery/blob/master/celery/bin/celery.py#L110)
any idea how can I read this configuration without importing the entire app?

avikam:
This is still a draft, @auvipy, @thedrow - i'm happy to hear your feedback.

The idea of this change is to instantiate the app lazily so that eventlet/gevent will patch the standard library before the app is imported. To achieve that I used a cached property on the click-context.
However inside the main `celery` command (before `bin.worker` is evaluated), there is a reference to the app, which adds additional user options to worker+beat+event commands (side note - the documentation mentions [argparse](https://docs.celeryproject.org/en/stable/userguide/extending.html#adding-new-command-line-options) which I assume should be changed to Click?)
So to circumvent this reference, I manually parse the residual args that complies with user options on the worker command.
There is more work left to do, for example to project these user options on the help page, but I think it's achievable bc the app isn't really needed when celery is invoked with `--help`. 

I was able to run a celery worker with this app:
```
from celery import Celery
from celery.bin.base import CeleryOption

app = Celery('tasks', backend='redis://localhost', broker='redis://localhost')

app.user_options['worker'].add(CeleryOption(('-z', '--hello'), help_group="User Options"))
app.user_options['worker'].add(CeleryOption(('--flag', ), is_flag=True, help_group="User Options"))
```
and with this commandline:
`celery -A myapp worker -P eventlet -c 10 broker_url=redis:// --hello world --flag`

thedrow:
Oh I completely forgot about gevent/eventlet patching.
It's good that you noticed.

The documentation wasn't updated yet. I'll read the entire issue first before I respond more thoroughly.

celery/bin/base.py
```diff
@@ -46,6 +47,11 @@ def OK(self):
     def ERROR(self):
         return self.style("ERROR", fg="red", bold=True)
 
+    @property
+    @functools.lru_cache(None)
```
thedrow, line 51: You can use `@cached_property` here.
We have a backport for Python versions which lack it.

celery/bin/base.py
```diff
@@ -46,6 +47,11 @@ def OK(self):
...
+    def app(self):
+        return self._app() or get_current_app()
```
thedrow, line 53: Please provide a docstring which explains what happens here.

celery/bin/celery.py
```diff
@@ -107,9 +109,9 @@ def celery(ctx, app, broker, result_backend, loader, config, workdir,
     ctx.obj = CLIContext(app=app, no_color=no_color, workdir=workdir, quiet=quiet)
 
     # User options
-    worker.params.extend(ctx.obj.app.user_options.get('worker', []))
-    beat.params.extend(ctx.obj.app.user_options.get('beat', []))
-    events.params.extend(ctx.obj.app.user_options.get('events', []))
+    # worker.params.extend(ctx.obj.app.user_options.get('worker', []))
```
thedrow, line 112: We need this to support preload options.

celery/bin/worker.py
```diff
@@ -17,6 +17,39 @@
...
+        import gevent.monkey
+        import gevent.signal
+
+        gevent.monkey.patch_all()
+        if gevent.version_info[0] == 0:  # pragma: no cover
+            # Signals aren't working in gevent versions <1.0,
+            # and aren't monkey patched by patch_all()
+            import signal
+
+            signal.signal = gevent.signal
```
thedrow, line 41:
```suggestion

```

I think that at this point we can drop support for `gevent<1.0.0`.

celery/bin/worker.py
```diff
@@ -42,6 +75,8 @@ def __init__(self):
     def convert(self, value, param, ctx):
         # Pools like eventlet/gevent needs to patch libs as early
         # as possible.
+        maybe_patch_concurrency(value)
```
thedrow, line 78: I'm pretty sure that's the wrong hook for this.

celery/bin/worker.py
```diff
@@ -131,7 +169,6 @@ def detach(path, argv, logfile=None, pidfile=None, uid=None,
               '--statedb',
               cls=CeleryOption,
               type=click.Path(),
-              callback=lambda ctx, _, value: value or ctx.obj.app.conf.worker_state_db,
```
thedrow, line 134: This cannot be removed. The default value must be derived from the app if it is not supplied.

celery/bin/worker.py
```diff
@@ -152,7 +189,6 @@ def detach(path, argv, logfile=None, pidfile=None, uid=None,
 @click.option('--prefetch-multiplier',
               type=int,
               metavar="<prefetch multiplier>",
-              callback=lambda ctx, _, value: value or ctx.obj.app.conf.worker_prefetch_multiplier,
```
thedrow, line 155: This cannot be removed. The default value must be derived from the app if it is not supplied.

celery/bin/worker.py
```diff
@@ -161,7 +197,6 @@ def detach(path, argv, logfile=None, pidfile=None, uid=None,
               '--concurrency',
               type=int,
               metavar="<concurrency>",
-              callback=lambda ctx, _, value: value or ctx.obj.app.conf.worker_concurrency,
```
thedrow, line 164: This cannot be removed. The default value must be derived from the app if it is not supplied.

thedrow (changes requested):
Some issues here are minor, some are major.

This is a release blocker for 5.0.0 GA so we need to urgently resolve this.

celery/bin/worker.py
avikam: It's not removed, only moved below, so the app will not be loaded too soon. See the call to worker.start in this function

celery/bin/worker.py
thedrow: Master dropped support for `gevent<1.0.0`.

celery/bin/worker.py
thedrow: Let's address the other issues first and see if we can find the way to write this code as Click intended it to be written.

celery/bin/worker.py
avikam: I'm open for suggestions.
My reasoning was to make it as close as possible to `-P` evaluation which is the only indication that patching is needed.

A possibility is to move this call to the command function itself, but then we can't reference the app in any of the `click.option`s (bc it will cause the app to be imported and the patching to be too late)

celery/bin/celery.py
avikam: a compensation for that is done here:
https://github.com/celery/celery/pull/6322/files#diff-a6e64fe5a7c24a0e587e8c92e83b87b5R321

(of course, it is only a draft as it needs to be applied for beat and events as well. i wanted to get an approval for this idea before pursuing it)

there is a downside to this approach, but i'm not sure how else can to get the use options only after the patching

thedrow:
```patch
diff --git a/celery/__init__.py b/celery/__init__.py
index 03a69a6b8..f2e406a9a 100644
--- a/celery/__init__.py
+++ b/celery/__init__.py
@@ -112,12 +112,6 @@ def _patch_gevent():
     import gevent.signal
 
     gevent.monkey.patch_all()
-    if gevent.version_info[0] == 0:  # pragma: no cover
-        # Signals aren't working in gevent versions <1.0,
-        # and aren't monkey patched by patch_all()
-        import signal
-
-        signal.signal = gevent.signal
 
 
 def maybe_patch_concurrency(argv=None, short_opts=None,
diff --git a/celery/__main__.py b/celery/__main__.py
index b0557b185..e865ea4bd 100644
--- a/celery/__main__.py
+++ b/celery/__main__.py
@@ -2,15 +2,15 @@
 
 import sys
 
-# from . import maybe_patch_concurrency
+from . import maybe_patch_concurrency
 
 __all__ = ('main',)
 
 
 def main():
     """Entrypoint to the ``celery`` umbrella command."""
-    # if 'multi' not in sys.argv:
-    #     maybe_patch_concurrency()
+    if 'multi' not in sys.argv:
+        maybe_patch_concurrency()
     from celery.bin.celery import main as _main
     sys.exit(_main())
``` 

Does this patch also work? Can you try to `git apply` this on master?
It is much simpler and reuses the old code.

avikam:
yep, i thought the idea was to leave all argv responsibility to click, but this code works.
closing.

thedrow:
> yep, i thought the idea was to leave all argv responsibility to click, but this code works.
> closing.

Thank you for bringing this to our attention. :smiley_cat: 
Support for gevent and eventlet is going away in 6.0 anyway as we migrate to an async event loop which does not support them.
I'd rather not touch whatever works.

I'm going to release a new RC in a few moments.
Please try it out :)

PR #6326, opened 2020-09-01
Update class-based task retry docs

SergeyKosarchuk:
## Description
Updated Task Retrying docs.

Option `retry_kwargs` is not supported by class-based tasks. But you can change `max_retries` option with corresponding attribute.


auvipy (approved):
I checked the code and verified the changes.

PR #6328, opened 2020-09-02
Travis CI: Test Python 3.9 release candidate 1

cclauss:

## Description

Travis CI: Test Python 3.9 release candidate 1
https://www.python.org/downloads/release/python-390rc1/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ env:
   global:
   - PYTHONUNBUFFERED=yes
   - CELERY_TOX_PARALLEL=
-  matrix:
+  jobs:
   - MATRIX_TOXENV=unit
 
 jobs:
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,12 +5,16 @@ python:
   - '3.6'
   - '3.7'
   - '3.8'
+  - '3.9-dev'
 os:
     - linux
 stages:
   - test
   - integration
   - lint
+services:
+  - redis
+  - docker
 env:
   global:
   - PYTHONUNBUFFERED=yes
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -113,21 +123,18 @@ before_install:
           wget -qO - https://packages.couchbase.com/ubuntu/couchbase.key | sudo apt-key add -
           sudo apt-add-repository -y 'deb http://packages.couchbase.com/ubuntu bionic bionic/main'
           sudo apt-get update && sudo apt-get install -y libcouchbase-dev
+          install: pip --disable-pip-version-check install --upgrade-strategy eager -U tox | cat
```
thedrow, line 126: Why did you move this to before_install?

thedrow (changes requested):

thedrow:
It seems like we have a genuine test failure to investigate.

cclauss:
Is the pypy3 failure caused by something in this PR?

cclauss:
@untitaker FYI

.travis.yml
cclauss: That is the order in which things execute when the job is run.

thedrow:
There's no need to allow failures for PyPy.
I've rebased your commits and it should now run as expected.

thedrow:
This failure is very strange.

```python
    def test_append_to_empty_chain(self):
        x = chain()
        x = x | self.add.s(1, 1)
        x = x | self.add.s(1)
        x.freeze()
        tasks, _ = x._frozen
        assert len(tasks) == 2

        assert x.apply().get() == 3
```
passes the test but
```python
    def test_append_to_empty_chain(self):
        x = chain()
        x |= self.add.s(1, 1)
        x |= self.add.s(1)
        x.freeze()
        tasks, _ = x._frozen
        assert len(tasks) == 2

        assert x.apply().get() == 3
```
does not.
```

thedrow:
I figured out why.
The `Signature` class inherits from `dict` which in Python 3.9 has the merge operator `|` and `|=` is the in place version of that operator.
We override `__or__` but not `__ior__` since previously `__ior__` would use the default implementation.

thedrow:
Thank you so much!

PR #6331, opened 2020-09-03
Fix examples of starting a worker in comments

kigawas:
Gotcha: 
Now you have to use
```
celery --app=proj worker
```

### DIFF
--- a/celery/bin/worker.py
+++ b/celery/bin/worker.py
@@ -273,10 +273,10 @@ def worker(ctx, hostname=None, pool_cls=None, app=None, uid=None, gid=None,
 
     Examples
     --------
-    $ celery worker --app=proj -l info
-    $ celery worker -A proj -l info -Q hipri,lopri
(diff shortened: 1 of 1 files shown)


qiaocci:
These command was supported in celery 4.4.7.
Should we port it to celery 5.x?

kigawas:
> These command was supported in celery 4.4.7.
> Should we port it to celery 5.x?

I don't think so since `--app=proj` does not belong to `celery worker` command

thedrow:
> These command was supported in celery 4.4.7.
> Should we port it to celery 5.x?

This is indeed a breaking change. I haven't found a way to implement this behavior with Click without resorting to hacks.
I don't believe it is necessary anyway,

thedrow (approved):
Thank you for this.
I'm working on the documentation myself so this helps :)

georgepsarakis (approved):

PR #6333, opened 2020-09-03
add warning against use of sort key on dynamodb table, closes #6332

### LINKED ISSUE #6332: Celery >= 4.4.3 incompatible with DynamoDB table with sort key as results backend
# Checklist

  for existing proposed fixes.

#### Related Issues

- PR: https://github.com/celery/celery/pull/5275 (status is open, but feature is already present since Celery v4.4.0)

#### Possible Duplicates

- None

# Description

First off, documentation is not especially wrong on the use of DynamoDB as a results backend. However, it is incomplete in a significant way. Because I manage the DynamoDB table through IaC I did not realize that defining a sort key would break compatibilty with Celery. 

When you use a DynamoDB table with a [sort key](https://aws.amazon.com/blogs/database/using-sort-keys-to-organize-data-in-amazon-dynamodb/) you can not use Celery v4.4 versions >= v4.4.3. The reason is commit c711047036819cbe003dcfbd81fc35bb0b4af5d3 which introduces a query of existing Celery result status. When you have a sort key defined for your DynamoDB table, the primary key is no longer _just_ the partition key ('id'), but a combination of the partition and sort keys ('id' and 'timestamp' in my case).

# Suggestions

It would be fun to add sort key support to the DynamoDB results backend, but I fear it will not be possible. For example, in my case I will need to know the value of the 'timestamp' in addition to the 'id' of the result item we need to retrieve. I don't believe Celery has knowledge of the 'timestamp' we are looking for at the time the query from commit c711047036819cbe003dcfbd81fc35bb0b4af5d3 is performed.

So my suggestion is simple: Let's add a line in the documentation about DynamoDB use as results backend that warns us that you cannot use a sort key on the table. Luckily, you could still add a [global secondary index (GSI)](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/GSI.html) if you want to perform queries based on something other than the partition key. A GSI will not break the primary key schema for the queries from Celery.

martin-paulus:

### DIFF
--- a/docs/userguide/configuration.rst
+++ b/docs/userguide/configuration.rst
@@ -1675,6 +1675,13 @@ AWS DynamoDB backend settings
     See :ref:`bundles` for information on combining multiple extension
     requirements.
 
+.. warning::
+
+    The Dynamodb backend is not compatible with tables that have a sort key defined.
+
+    If you want to query the results table based on something other than the partition key,
+    please define a global secondary index (GSI) instead.
+
 This backend requires the :setting:`result_backend`
 setting to be set to a DynamoDB URL::
 


auvipy (approved):

thedrow:
We should backport this.

Issue #5947, opened 2020-01-31
chord of chains with groups: body duplication and invalid task execution order


the-owl:
# Checklist
- [x] I have verified that the issue exists against the `master` branch of Celery.
- [ ] This has already been asked to the [discussion group](https://groups.google.com/forum/#!forum/celery-users) first.
- [x] I have read the relevant section in the
  [contribution guide](http://docs.celeryproject.org/en/latest/contributing.html#other-bugs)
  on reporting bugs.
- [x] I have checked the [issues list](https://github.com/celery/celery/issues?q=is%3Aissue+label%3A%22Issue+Type%3A+Bug+Report%22+-label%3A%22Category%3A+Documentation%22)
  for similar or identical bug reports.
- [x] I have checked the [pull requests list](https://github.com/celery/celery/pulls?q=is%3Apr+label%3A%22PR+Type%3A+Bugfix%22+-label%3A%22Category%3A+Documentation%22)
  for existing proposed fixes.
- [x] I have checked the [commit log](https://github.com/celery/celery/commits/master)
  to find out if the bug was already fixed in the master branch.
- [x] I have included all related issues and possible duplicate issues
  in this issue (If there are none, check this box anyway).

## Mandatory Debugging Information

- [x] I have included the output of ``celery -A proj report`` in the issue.
    (if you are not able to do this, then at least specify the Celery
     version affected).
- [x] I have verified that the issue exists against the `master` branch of Celery.
- [x] I have included the contents of ``pip freeze`` in the issue.
- [x] I have included all the versions of all the external dependencies required
  to reproduce this bug.

## Optional Debugging Information
- [ ] I have tried reproducing the issue on more than one Python version
  and/or implementation.
- [ ] I have tried reproducing the issue on more than one message broker and/or
  result backend.
- [ ] I have tried reproducing the issue on more than one version of the message
  broker and/or result backend.
- [ ] I have tried reproducing the issue on more than one operating system.
- [ ] I have tried reproducing the issue on more than one workers pool.
- [ ] I have tried reproducing the issue with autoscaling, retries,
  ETA/Countdown & rate limits disabled.
- [x] I have tried reproducing the issue after downgrading
  and/or upgrading Celery and its dependencies.

#### Related Issues

- None

#### Possible Duplicates

- None

## Environment & Settings

**Celery version**: 4.4.0

<details>
<summary><b><code>celery report</code> Output:</b></summary>
<p>

```
[...]

PR #5984, opened 2020-03-01
Fix chords with chained groups (#5947)

the-owl:
## Description

Fixes #5947

PR #6342, opened 2020-09-08
Deeply deserialize groups

### LINKED ISSUE #6341: Chords contained in a group raise `AttributeError`s during freezing if passed through a backend
When calling `self.replace()` with a signature reconstructed from a serialized dictionary (e.g. after having been passed through a backend), if that signature is (or contains) a group which itself contains a chord, celery explodes after attempting to treat deeply nested dictionaries as signature objects. See below for a minimal repro I got together today.

My gut feel is that instantiating a signature from a dictionary may not be recursing down through the structure far enough and it leaves some of the encapsulated tasks as dicts. I've also noticed that groups containing a group also break in the same way, but I think that's because there's an internal promotion to a chord happening somewhere.

# Checklist

  on reporting bugs.
  for existing proposed fixes.

     version affected).
  to reproduce this bug.

  and/or implementation.
  result backend.
  broker and/or result backend.

#### Related Issues
 - #4015

#### Possible Duplicates

- None

## Environment & Settings

**Celery version**:

<details>
<p>

```
```

</p>
</details>

# Steps to Reproduce

## Required Dependencies

### Python Packages

<details>
<p>

```
amqp==5.0.1
billiard==3.6.3.0
# Editable install with no version control (celery==5.0.0rc3)
-e /home/maybe/tmp/capp/venv/lib/python3.8/site-packages
click==7.1.2
click-didyoumean==0.0.3
click-repl==0.1.6
future==0.18.2
kombu==5.0.2
prompt-toolkit==3.0.7
pytz==2020.1
redis==3.5.3
six==1.15.0
vine==5.0.0
wcwidth==0.2.5
```

</p>
</details>

### Other Dependencies

<details>
<p>
N/A
</p>
</details>

<details>
<p>

```python
import celery

app = celery.Celery("app", backend="redis://")

@app.task
def foo(*_):
    return 42

@app.task(bind=True)
def replace_with(self, sig):
    assert isinstance(sig, dict)
    sig = celery.Signature.from_dict(sig)
    raise self.replace(sig)

if __name__ == "__main__":
    sig = celery.group(
        celery.group(foo.s()),
    )
    res = sig.delay()
    print(res.get())

    sig.freeze()
    res = replace_with.delay(sig)
    print(res.get())
```

</p>
</details>

# Expected Behavior
It shouldn't explode. Presumably tasks within the group/chord should be signatures rather than dicts.

# Actual Behavior

Stack trace in the worker output:
```
[2020-09-08 12:44:05,453: DEBUG/MainProcess] Task accepted: app.replace_with[dcea02fd-23a3-404a-9fdd-b213eb51c0d1] pid:453431
[2020-09-08 12:44:05,457: ERROR/ForkPoolWorker-8] Task app.replace_with[dcea02fd-23a3-404a-9fdd-b213eb51c0d1] raised unexpected: AttributeError("'dict' object has no attribute '_app'")
  File "/home/maybe/tmp/capp/venv/lib64/python3.8/site-packages/kombu/utils/objects.py", line 41, in __get__
    return obj.__dict__[self.__name__]
KeyError: 'app'

  File "/home/maybe/tmp/capp/venv/lib64/python3.8/site-packages/celery/app/trace.py", line 409, in trace_task
  File "/home/maybe/tmp/capp/venv/lib64/python3.8/site-packages/celery/app/trace.py", line 701, in __protected_call__
    return self.run(*args, **kwargs)
  File "/home/maybe/tmp/capp/app.py", line 13, in replace_with
    raise self.replace(sig)
  File "/home/maybe/tmp/capp/venv/lib64/python3.8/site-packages/celery/app/task.py", line 894, in replace
    sig.freeze(self.request.id)
  File "/home/maybe/tmp/capp/venv/lib64/python3.8/site-packages/celery/canvas.py", line 1302, in freeze
    self.tasks = group(self.tasks, app=self.app)
  File "/home/maybe/tmp/capp/venv/lib64/python3.8/site-packages/kombu/utils/objects.py", line 43, in __get__
    value = obj.__dict__[self.__name__] = self.__get(obj)
  File "/home/maybe/tmp/capp/venv/lib64/python3.8/site-packages/celery/canvas.py", line 1456, in app
    return self._get_app(self.body)
  File "/home/maybe/tmp/capp/venv/lib64/python3.8/site-packages/celery/canvas.py", line 1466, in _get_app
    app = tasks[0]._app
AttributeError: 'dict' object has no attribute '_app'
```

maybe-sybr:
## Description
Fixes #6341

### DIFF
--- a/celery/canvas.py
+++ b/celery/canvas.py
@@ -1047,8 +1047,16 @@ class group(Signature):
 
     @classmethod
     def from_dict(cls, d, app=None):
+        # We need to mutate the `kwargs` element in place to avoid confusing
+        # `freeze()` implementations which end up here and expect to be able to
+        # access elements from that dictionary later and refer to objects
+        # canonicalized here
+        orig_tasks = d["kwargs"]["tasks"]
+        d["kwargs"]["tasks"] = rebuilt_tasks = type(orig_tasks)((
+            maybe_signature(task, app=app) for task in orig_tasks
+        ))
         return _upgrade(
-            d, group(d['kwargs']['tasks'], app=app, **d['options']),
+            d, group(rebuilt_tasks, app=app, **d['options']),
         )
 
     def __init__(self, *tasks, **options):
--- a/t/integration/tasks.py
+++ b/t/integration/tasks.py
@@ -1,6 +1,6 @@
 from time import sleep
 
-from celery import Task, chain, chord, group, shared_task
+from celery import Signature, Task, chain, chord, group, shared_task
 from celery.exceptions import SoftTimeLimitExceeded
 from celery.utils.log import get_task_logger
 
@@ -244,3 +244,67 @@ class ClassBasedAutoRetryTask(Task):
         if self.request.retries:
             return self.request.retries
         raise ValueError()
+
+
+# The signatures returned by these tasks wouldn't actually run because the
+# arguments wouldn't be fulfilled - we never actually delay them so it's fine
+@shared_task
+def return_nested_signature_chain_chain():
+    return chain(chain([add.s()]))
+
+
+@shared_task
+def return_nested_signature_chain_group():
+    return chain(group([add.s()]))
+
+
+@shared_task
+def return_nested_signature_chain_chord():
+    return chain(chord([add.s()], add.s()))
+
+
+@shared_task
+def return_nested_signature_group_chain():
+    return group(chain([add.s()]))
+
+
+@shared_task
+def return_nested_signature_group_group():
+    return group(group([add.s()]))
+
+
+@shared_task
+def return_nested_signature_group_chord():
+    return group(chord([add.s()], add.s()))
+
+
+@shared_task
+def return_nested_signature_chord_chain():
+    return chord(chain([add.s()]), add.s())
+
+
+@shared_task
+def return_nested_signature_chord_group():
+    return chord(group([add.s()]), add.s())
+
+
+@shared_task
+def return_nested_signature_chord_chord():
+    return chord(chord([add.s()], add.s()), add.s())
+
+
+@shared_task
+def rebuild_signature(sig_dict):
+    sig_obj = Signature.from_dict(sig_dict)
+
+    def _recurse(sig):
+        if not isinstance(sig, Signature):
+            raise TypeError("{!r} is not a signature object".format(sig))
+        # Most canvas types have a `tasks` attribute
+        if isinstance(sig, (chain, group, chord)):
+            for task in sig.tasks:
+                _recurse(task)
+        # `chord`s also have a `body` attribute
+        if isinstance(sig, chord):
+            _recurse(sig.body)
+    _recurse(sig_obj)
--- a/t/unit/tasks/test_canvas.py
+++ b/t/unit/tasks/test_canvas.py
@@ -2,6 +2,7 @@ import json
 from unittest.mock import MagicMock, Mock, call, patch, sentinel
 
 import pytest
+import pytest_subtests  # noqa: F401
 
 from celery._state import _task_stack
 from celery.canvas import (Signature, _chain, _maybe_group, chain, chord,
@@ -1005,22 +1006,35 @@ class test_chord(CanvasCase):
         x.kwargs['body'] = None
         assert 'without body' in repr(x)
 
-    def test_freeze_tasks_body_is_group(self):
-        # Confirm that `group index` is passed from a chord to elements of its
-        # body when the chord itself is encapsulated in a group
+    def test_freeze_tasks_body_is_group(self, subtests):
+        # Confirm that `group index` values counting up from 0 are set for
+        # elements of a chord's body when the chord is encapsulated in a group
         body_elem = self.add.s()
-        chord_body = group([body_elem])
+        chord_body = group([body_elem] * 42)
         chord_obj = chord(self.add.s(), body=chord_body)
         top_group = group([chord_obj])
         # We expect the body to be the signature we passed in before we freeze
-        (embedded_body_elem, ) = chord_obj.body.tasks
-        assert embedded_body_elem is body_elem
-        assert embedded_body_elem.options == dict()
-        # When we freeze the chord, its body will be clones and options set
+        with subtests.test(msg="Validate body tasks are retained"):
+            assert all(
+                embedded_body_elem is body_elem
+                for embedded_body_elem in chord_obj.body.tasks
+            )
+        # We also expect the body to have no initial options - since all of the
+        # embedded body elements are confirmed to be `body_elem` this is valid
+        assert body_elem.options == {}
+        # When we freeze the chord, its body will be cloned and options set
         top_group.freeze()
-        (embedded_body_elem, ) = chord_obj.body.tasks
-        assert embedded_body_elem is not body_elem
-        assert embedded_body_elem.options["group_index"] == 0   # 0th task
+        with subtests.test(
+            msg="Validate body group indicies count from 0 after freezing"
+        ):
+            assert all(
+                embedded_body_elem is not body_elem
+                for embedded_body_elem in chord_obj.body.tasks
+            )
+            assert all(
+                embedded_body_elem.options["group_index"] == i
+                for i, embedded_body_elem in enumerate(chord_obj.body.tasks)
+            )
 
     def test_freeze_tasks_is_not_group(self):
         x = chord([self.add.s(2, 2)], body=self.add.s(), app=self.app)
--- a/t/unit/tasks/test_canvas.py
+++ b/t/unit/tasks/test_canvas.py
@@ -694,6 +694,32 @@ class test_group(CanvasCase):
         x['args'] = None
         assert group.from_dict(dict(x))
 
+    @pytest.mark.xfail(reason="#6341")
+    def test_from_dict_deep_deserialize(self):
+        original_group = group([self.add.s(1, 2)] * 42)
+        serialized_group = json.loads(json.dumps(original_group))
+        deserialized_group = group.from_dict(serialized_group)
+        assert all(
+            isinstance(child_task, Signature)
+            for child_task in deserialized_group.tasks
+        )
+
+    @pytest.mark.xfail(reason="#6341")
+    def test_from_dict_deeper_deserialize(self):
+        inner_group = group([self.add.s(1, 2)] * 42)
+        outer_group = group([inner_group] * 42)
+        serialized_group = json.loads(json.dumps(outer_group))
+        deserialized_group = group.from_dict(serialized_group)
+        assert all(
+            isinstance(child_task, Signature)
+            for child_task in deserialized_group.tasks
+        )
+        assert all(
+            isinstance(grandchild_task, Signature)
+            for child_task in deserialized_group.tasks
+            for grandchild_task in child_task.tasks
+        )
+
     def test_call_empty_group(self):
         x = group(app=self.app)
         assert not len(x())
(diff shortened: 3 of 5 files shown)


maybe-sybr:
Probably needs tests like the one that #4015 added but I need to switch off this for the moment

celery/canvas.py
```diff
@@ -1272,6 +1273,7 @@ class chord(Signature):
     def from_dict(cls, d, app=None):
         options = d.copy()
         args, options['kwargs'] = cls._unpack_args(**options['kwargs'])
+        args = [maybe_signature(task) for task in args]
```
maybe-sybr, line 1276: My more complex failure case hits this code somewhere but it doesn't seem to be required to fix my breakage (just the change to groups). In any case, this seems like a legit change to make.

thedrow: Could you please include the more complicated cases as tests in our integration suite?
You can mark them as `xfail` for now.

thedrow (changes requested):
This PR requires integration tests.

Please include the original test case in #6341 as a test in our integration suite.

celery/canvas.py
maybe-sybr: Turns out this isn't strictly necessary because the `_maybe_group()` and `_maybe_signature()` in `chord.__init__()` cause nested `dicts` to get deserialized properly.

https://github.com/celery/celery/blob/9b5f6f5531f45525e904cef114d653a93f1c9635/celery/canvas.py#L1289-L1298

It's slightly different behaviour to the other `from_dict()` code paths since `chain.from_dict()` and (now) `group.from_dict()` deserialize their children prior to instantiating new objects. I've removed this from the diff, but let me know if you'd like it restored to make all three behave in the same way. (/ping @thedrow )

maybe-sybr:
> This PR requires integration tests.
> 
> Please include the original test case in #6341 as a test in our integration suite.

@thedrow - sorry I took so long to get back to you on this. I've pushed up some unit tests which expose where the bug was actually manifesting (and importantly, where it wasn't - ie. in `chord`s). Per my other comment, I've removed the change to `chord.from_dict()` because the bug didn't manifest there , but it might still be nice to have that in the changeset so that all of the `from_dict()` implementations do the same thing, rather than relying on a side-effect of `chord.__init__()` which might change in future. Let me know what you think on that one.

I'm also happy to add an integration test based on the test case I posted in #6341 but I'm not sure how to go about that since I need to define some tasks to run with results being passed via some backend so that the serialization happens. The rough flow of the integration test would be:
1. for a task to create a group signature containing another group, and return that object.
1. another task to get the result of the first, and the run `Signature.from_dict()` on it
1. verification in the second task that the returned Signature object has indeed been deeply deseriaized

maybe-sybr:
I noticed the test causing CI breakage [0] failing on my box too but thought it was just flakey. It's fine on master so definitely caused by my change - I'll sort it out now

[0] https://travis-ci.org/github/celery/celery/jobs/727590711

maybe-sybr:
@thedrow - worked it out. Turns out the `group.freeze()` implementation relies on the sub-tasks it passes down for recursive freezing (via `group._freeze_unroll()`) being mutated in-place. By making a copy of the sub-tasks of a group encapsulated in a group (or a chord) when that sub-group made its way to `group.from_dict()` via `Signature.clone()`, it turns out I was creating a canonical version of the tasks in the sub-group, which the top level group wouldn't know about later. The frozen groupresult object looks like it was correct, but the top level group task wouldn't have been mutated like that test I broke assumed. Other bits of the codebase also seem to assume that things should be mutated in place if possible, so I've made an amendment to this diff to meet that assumption.

That's a bit complicated, so let me know if you want me to explain things any better. The tl;dr is that we didn't mutate a thing when we should have, and that broke assumptions. IMO the freezing logic could stand to be reworked a bit to get rid of some of these implicit mutations of built-in types like maps and lists, since they're a bit hard to follow - but that's certainly not the point here and I've just rolled with the current code expectations.

This should be good to merge now hopefully. I'll check in on CI as it runs though to make sure it's happy. Per my previous comment, still not sure about adding an integration test - let me know what you'd like to see in that space, if anything. Thanks!

thedrow:
I've restarted the build to see if the error is consistent.

As for the integration tests, you should include your test case as a test in our suite.
If you have more complicated cases, please include them as well.

maybe-sybr:
@thedrow - I think that CI failure is related to another odd behaviour I've observed with chains containing groups. I've spent some time tracing it and it looks like the change in 01dd66c (#5984 which fixed #5947 ) is causing this since it seems to cause an off-by-one in the calculated size of a chord body and the redis backend is never able to complete the chord.

The off by one reliably happens with my change to `group.from_dict()` whereas it doesn't happen on master for the same signature structure. Presumably because calling `maybe_signature()` if `group.from_dict()` is upgrading some nested signature structure into a chord and getting confused along the way. Commenting out the lines changed in #5984 makes the the broken test pass but obviously breaks the test added in that PR but obviously also breaks the test which was added. I'm continuing to try to work out what the right fix is here.

maybe-sybr:
@thedrow - I think I've fixed the problems and have pushed up anther PR (#6354) which blocks this one (since this is based in that one). We can merge this one independently if we mark the failing test with an xfail which the other PR can resolve later. Let me know what you'd prefer.

auvipy (changes requested):
the pytest is causing errors in integration tests. can you try to supress that?

maybe-sybr:
per my comment on #6354 , I think those failures are on master and might be environmental rather than to do with actual code changes in celery :)

maybe-sybr:
I've rebased this on top of the changeset for #6354 and hopefully it will also go green in CI. Please don't merge this until the other PR is merged :)

auvipy (changes requested):
It will be easier to review after the prerequisite PR is merged

maybe-sybr:
I've looked back over this and I think it may be possible to merge it independently. The changeset didn't really depend on anything in #6354 but I was under the impression that some of the integration testing would. I've rebased just the commits relevant to #6341 on top of master, so we'll see if CI is happy with it soon enough. In the meantime, you're welcome to review the diff again @thedrow and @auvipy

maybe-sybr:
Ah, yes. That's the test I remember failing and sending me onto the chord counting tangent. I'll leave this as is for now so others can review the substantive diff, and rebase once #6354 gets merged.

auvipy (changes requested):
please rebase

t/unit/tasks/test_canvas.py
```diff
@@ -816,21 +833,21 @@ def test_repr(self):
...
+            assert embedded_body_elem is body_elem
+            assert not embedded_body_elem.options
+        # When we freeze the chord, its body will be cloned and options set
         top_group.freeze()
-        (embedded_body_elem, ) = chord_obj.body.tasks
-        assert embedded_body_elem is not body_elem
-        assert embedded_body_elem.options["group_index"] == 0   # 0th task
+        for i, embedded_body_elem in enumerate(chord_obj.body.tasks):
+            assert embedded_body_elem is not body_elem
+            assert embedded_body_elem.options["group_index"] == i
```
thedrow, line 850: I think that this loop for example is a great opportunity to introduce [`pytest-subtests`](https://github.com/pytest-dev/pytest-subtests/) to our codebase.

There are other opportunities in this patch to use it.

t/unit/tasks/test_canvas.py
```diff
@@ -860,6 +877,73 @@ def chord_add():
...
+        assert isinstance(rebuilt_chord, chord)
+
+    def test_from_dict_deep_deserialize(self):
+        header = body = self.add.s(1, 2)
+        original_chord = chord(header=header, body=body)
+        serialized_chord = json.loads(json.dumps(original_chord))
+        deserialized_chord = chord.from_dict(serialized_chord)
+        assert isinstance(deserialized_chord, chord)
+        for task in deserialized_chord.tasks:
+            assert isinstance(task, Signature)
```
thedrow, line 899:
```suggestion
        assert all(isinstance(task, Signature) for task in deserialized_chord.tasks)
```

That applies to all other loops of this type.

thedrow (changes requested):

t/unit/tasks/test_canvas.py
maybe-sybr: I've pivoted to using `all(...)` for the other loopy checks in this diff now, so this should be the only place subtests might be relevant. I'll do it here to support the first commit and introduce the subtest tooling so that it can be used elsewhere in the suite.

t/unit/tasks/test_canvas.py
maybe-sybr: Done, this will land in my next push. I've used the trope throughout this test suite and updated two tests unrelated to this diff to match.

t/unit/tasks/test_canvas.py
maybe-sybr: Oh, I actually see what you mean. We could use subtests to check, e.g., that the body of a deeply deserialized chord is correct, separate from the checks on the header! That's pretty cool, I'll do that too.

maybe-sybr: Done, check out the new diff and let me know what you think

t/unit/tasks/test_canvas.py
```diff
@@ -1050,6 +1083,117 @@ def chord_add():
...
+        rebuilt_chord = chord.from_dict(dict(original_chord))
+        assert isinstance(rebuilt_chord, chord)
+
+    def test_from_dict_deep_deserialize(self, subtests):
+        header = body = self.add.s(1, 2)
+        original_chord = chord(header=header, body=body)
+        serialized_chord = json.loads(json.dumps(original_chord))
+        deserialized_chord = chord.from_dict(serialized_chord)
+        with subtests.test(msg="Verify chord is deserialized"):
+            assert isinstance(deserialized_chord, chord)
```
maybe-sybr, line 1104: I was unsure if this line should be part of the main test or a subtest. It's a prerequisite for the other subtests, so I feel like it might be better to take this out of a subtest block (and similarly in the tests below).

requirements/test-ci-base.txt
```diff
@@ -1,5 +1,6 @@
 pytest-cov
 pytest-travis-fold
+pytest-subtests
```
thedrow, line 3: Isn't this a duplicate requirement?
It's already specified in `test.txt`.

t/integration/tasks.py
```diff
@@ -1,6 +1,6 @@
 from time import sleep
 
-from celery import Task, chain, chord, group, shared_task
+from celery import Signature, Task, chain, chord, group, shared_task
```
thedrow, line 3: Please run isort on this file.

t/unit/tasks/test_canvas.py
thedrow: Generally speaking if we follow a variation of the `AAA` pattern, every assert should be in a different subtest.

I'm not going to enforce that for the time being.

thedrow (changes requested):
Just a few more minor things.

Overall this looks great!

t/unit/tasks/test_canvas.py
maybe-sybr: All good, sounds like I've done the right thing then for these tests with multiple assertions. `test_freeze_tasks_body_is_group()` has a single sanity assertion in the main body of the test but I'm okay with that structure since the rest of the test is nonsense if that is untrue.

requirements/test-ci-base.txt
maybe-sybr: Oh, I didn't realise that it only needed to be in one. I'll fix this.

t/integration/tasks.py
maybe-sybr: `isort` seems fine with this file, however the import in `t/integration/test_canvas.py` needed to be moved. I'll amend and push in a moment.

maybe-sybr:
Addressed those minor comments, sanity checked the `celery/` and `t/` trees with isort and it seems happy. Should be good to go, @thedrow - lmk if you see anything else!

auvipy (approved):
great job!

Issue #6327, opened 2020-09-02
Remove configuration compatibility shims from 3.x


clokep:
Celery 4.0 introduced new configuration setting names (see [the documentation](https://docs.celeryproject.org/en/latest/userguide/configuration.html#new-lowercase-settings)), but maintained compatibility with the old names. This seems to be a source of confusion (and I've witnessed numerous issues filed that are due to configuration problems). 5.0 seems like a good time to simplify this and have only a single standard form for these configuration variables.

I believe the code lives around [here]https://github.com/celery/celery/blob/5b86b35c81ea5a1fbfd439861f4fee6813148d16/celery/app/defaults.py#L76-L341).

thedrow:
I gave this a shot.
I'll ping you on the PR on Sunday since I have a test that fails who looks like it shouldn't.

thedrow:
I have a branch that removes support for the old setting names.
However, we have a couple of issues with implementing this, both technical and operational issues.

- While we have declared these settings deprecated in the documentation and in the release notes of 4.0, we have not prompted our users with a deprecation warning.
- The settings are set up before the logging subsystem is set up, which makes it harder to notify the user that the settings he is using are deprecated.
- Django users, a significant audience for Celery, still use the all-caps settings. Should we force them to upgrade? Should we support uppercase settings too?

I'm not so sure anymore we should do this in 5.0.
Instead we should notify the user that he is using deprecated setting names.

What do you think?

auvipy:
> I have a branch that removes support for the old setting names.
> However, we have a couple of issues with implementing this, both technical and operational issues.
> 
>     * While we have declared these settings deprecated in the documentation and in the release notes of 4.0, we have not prompted our users with a deprecation warning.
> 
>     * The settings are set up before the logging subsystem is set up, which makes it harder to notify the user that the settings he is using are deprecated.
> 
>     * Django users, a significant audience for Celery, still use the all-caps settings. Should we force them to upgrade? Should we support uppercase settings too?
> 
> 
> I'm not so sure anymore we should do this in 5.0.
> Instead we should notify the user that he is using deprecated setting names.
> 
> What do you think?

we should not remove them in v5 for easier migration

clokep:
> Django users, a significant audience for Celery, still use the all-caps settings. Should we force them to upgrade? Should we support uppercase settings too?

[...]

PR #6353, opened 2020-09-17
Warn when deprecated settings are used

thedrow:

## Description

We deprecated settings in 4.0 but did not warn about the deprecation.
This PR enables displaying deprecation pending warnings and deprecation warnings to our users.
It also prints which settings are deprecated and what should be used instead and instructs the users to run `celery upgrade settings`.

Ref #6327.

### DIFF
--- a/celery/app/log.py
+++ b/celery/app/log.py
@@ -9,12 +9,14 @@ related compatibility fixes, and so on.
 import logging
 import os
 import sys
+import warnings
 from logging.handlers import WatchedFileHandler
 
 from kombu.utils.encoding import set_default_encoding_file
 
 from celery import signals
 from celery._state import get_current_task
+from celery.exceptions import CDeprecationWarning, CPendingDeprecationWarning
 from celery.local import class_property
 from celery.platforms import isatty
 from celery.utils.log import (ColorFormatter, LoggingProxy, get_logger,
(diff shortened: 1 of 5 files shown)


thedrow:
@clokep Feel free to review this.

celery/app/utils.py
```diff
@@ -190,6 +195,20 @@ def humanize(self, with_defaults=False, censored=True):
...
 
+    def maybe_warn_deprecated_settings(self):
+        # TODO: Remove this method in Celery 6.0
+        if self.deprecated_settings:
+            from celery.utils import deprecated
+            from celery.app.defaults import _TO_NEW_KEY
+            for setting in self.deprecated_settings:
+                deprecated.warn(description=f'The {setting!r} setting',
+                                removal='6.0.0',
+                                alternative=f'Use the {_TO_NEW_KEY[setting]} instead')
```
clokep, line 206: This is going to be pretty noisy (people who haven't changed one setting probably have many that aren't changed), but I don't see a better way to do it.

celery/app/log.py
```diff
@@ -70,6 +72,9 @@ def setup(self, loglevel=None, logfile=None, redirect_stdouts=False,
             CELERY_LOG_LEVEL=str(loglevel) if loglevel else '',
             CELERY_LOG_FILE=str(logfile) if logfile else '',
         )
+        warnings.filterwarnings('always', category=CDeprecationWarning)
+        warnings.filterwarnings('always', category=CPendingDeprecationWarning)
+        logging.captureWarnings(True)
```
clokep, line 77: I'm not 100% sure what this change is doing -- I gather it is somehow making it so that we show these warnings to users, but why isn't that happening anyway? Is it since warnings aren't shown unless you use `python -W`?

clokep (approved):
Seems pretty reasonable! We should probably update [the documentation](https://github.com/celery/celery/blame/14a3524253f7769118fcb23de12b6707f38a1701/docs/userguide/configuration.rst#L50-L55) to say that this backwards compatibility will be removed in 6.0.

celery/app/log.py
thedrow: Yes, that's exactly why.

clokep: Neat. Thanks for confirming! I think this looks good then. 👍

Issue #1509, opened 2013-08-15
Groups in a chord throw exception


malnahlawi:
```
from celery import Celery

celery = Celery()
celery.config_from_object({
    'CELERY_RESULT_BACKEND': 'redis://localhost:6379'
})
from celery import chord, group, task

@task
def test():
    return True

c = chord([group(test.si(), test.si()), group(test.si(), test.si())])(group(test.si()))
c.get()
```

The tasks succeed.  The groups throw the following exception:

```
Traceback (most recent call last):
  File "/Library/Python/2.7/site-packages/celery-3.0.21-py2.7.egg/celery/task/trace.py", line 277, in trace_task
    on_chord_part_return(task)
  File "/Library/Python/2.7/site-packages/celery-3.0.21-py2.7.egg/celery/backends/base.py", line 503, in  on_chord_part_return
    if val >= len(deps):
TypeError: object of type 'NoneType' has no len()
```

Granted, the two groups can be rewritten as a single group.  it makes it a little harder for composition though.  If every application has to rewrite, it would be nice for the framework to do so.

PR #6354, opened 2020-09-18
Fix chord element counting

maybe-sybr:
This change fixes chord accounting in the redis backend by ensuring that
group IDs are not overwritten by chained callbacks.

Amends the changeset from #5984

### DIFF
--- a/celery/backends/redis.py
+++ b/celery/backends/redis.py
@@ -13,6 +13,7 @@ from celery import states
 from celery._state import task_join_will_block
 from celery.canvas import maybe_signature
 from celery.exceptions import ChordError, ImproperlyConfigured
+from celery.result import GroupResult, allow_join_result
 from celery.utils.functional import dictfilter
 from celery.utils.log import get_logger
 from celery.utils.time import humanize_seconds
@@ -401,12 +402,14 @@ class RedisBackend(BaseKeyValueStoreBackend, AsyncBackendMixin):
         return retval
 
     def apply_chord(self, header_result, body, **kwargs):
-        # Overrides this to avoid calling GroupResult.save
-        # pylint: disable=method-hidden
-        # Note that KeyValueStoreBackend.__init__ sets self.apply_chord
-        # if the implements_incr attr is set.  Redis backend doesn't set
-        # this flag.
-        pass
+        # If any of the child results of this chord are complex (ie. group
+        # results themselves), we need to save `header_result` to ensure that
+        # the expected structure is retained when we finish the chord and pass
+        # the results onward to the body in `on_chord_part_return()`. We don't
+        # do this is all cases to retain an optimisation in the common case
+        # where a chord header is comprised of simple result objects.
+        if any(isinstance(nr, GroupResult) for nr in header_result.results):
+            header_result.save(backend=self)
 
     @cached_property
     def _chord_zset(self):
@@ -449,20 +452,38 @@ class RedisBackend(BaseKeyValueStoreBackend, AsyncBackendMixin):
             callback = maybe_signature(request.chord, app=app)
             total = callback['chord_size'] + totaldiff
             if readycount == total:
-                decode, unpack = self.decode, self._unpack_chord_result
-                with client.pipeline() as pipe:
-                    if self._chord_zset:
-                        pipeline = pipe.zrange(jkey, 0, -1)
-                    else:
-                        pipeline = pipe.lrange(jkey, 0, total)
-                    resl, = pipeline.execute()
-                try:
-                    callback.delay([unpack(tup, decode) for tup in resl])
+                header_result = GroupResult.restore(gid)
+                if header_result is not None:
+                    # If we manage to restore a `GroupResult`, then it must
+                    # have been complex and saved by `apply_chord()` earlier.
+                    #
+                    # Before we can join the `GroupResult`, it needs to be
+                    # manually marked as ready to avoid blocking
+                    header_result.on_ready()
+                    # We'll `join()` it to get the results and ensure they are
+                    # structured as intended rather than the flattened version
+                    # we'd construct without any other information.
+                    join_func = (
+                        header_result.join_native
+                        if header_result.supports_native_join
+                        else header_result.join
+                    )
+                    with allow_join_result():
+                        resl = join_func(timeout=3.0, propagate=True)
+                else:
+                    # Otherwise simply extract and decode the results we
+                    # stashed along the way, which should be faster for large
+                    # numbers of simple results in the chord header.
+                    decode, unpack = self.decode, self._unpack_chord_result
                     with client.pipeline() as pipe:
-                        _, _ = pipe \
-                            .delete(jkey) \
-                            .delete(tkey) \
-                            .execute()
+                        if self._chord_zset:
+                            pipeline = pipe.zrange(jkey, 0, -1)
+                        else:
+                            pipeline = pipe.lrange(jkey, 0, total)
+                        resl, = pipeline.execute()
+                    resl = [unpack(tup, decode) for tup in resl]
+                try:
+                    callback.delay(resl)
                 except Exception as exc:  # pylint: disable=broad-except
                     logger.exception(
                         'Chord callback for %r raised: %r', request.group, exc)
@@ -470,6 +491,12 @@ class RedisBackend(BaseKeyValueStoreBackend, AsyncBackendMixin):
                         callback,
                         ChordError(f'Callback error: {exc!r}'),
                     )
+                finally:
+                    with client.pipeline() as pipe:
+                        _, _ = pipe \
+                            .delete(jkey) \
+                            .delete(tkey) \
+                            .execute()
         except ChordError as exc:
             logger.exception('Chord %r raised: %r', request.group, exc)
             return self.chord_error_from_stack(callback, exc)
--- a/celery/canvas.py
+++ b/celery/canvas.py
@@ -1383,21 +1383,30 @@ class chord(Signature):
             args=(tasks.apply(args, kwargs).get(propagate=propagate),),
         )
 
-    def _traverse_tasks(self, tasks, value=None):
-        stack = deque(tasks)
-        while stack:
-            task = stack.popleft()
-            if isinstance(task, group):
-                stack.extend(task.tasks)
-            elif isinstance(task, _chain) and isinstance(task.tasks[-1], group):
-                stack.extend(task.tasks[-1].tasks)
-            else:
-                yield task if value is None else value
+    @classmethod
+    def __descend(cls, sig_obj):
+        # Sometimes serialized signatures might make their way here
+        if not isinstance(sig_obj, Signature) and isinstance(sig_obj, dict):
+            sig_obj = Signature.from_dict(sig_obj)
+        if isinstance(sig_obj, group):
+            # Each task in a group counts toward this chord
+            subtasks = getattr(sig_obj.tasks, "tasks", sig_obj.tasks)
+            return sum(cls.__descend(task) for task in subtasks)
+        elif isinstance(sig_obj, _chain):
+            # The last element in a chain counts toward this chord
+            return cls.__descend(sig_obj.tasks[-1])
+        elif isinstance(sig_obj, chord):
+            # The child chord's body counts toward this chord
+            return cls.__descend(sig_obj.body)
+        elif isinstance(sig_obj, Signature):
+            # Each simple signature counts as 1 completion for this chord
+            return 1
+        # Any other types are assumed to be iterables of simple signatures
+        return len(sig_obj)
 
     def __length_hint__(self):
-        tasks = (self.tasks.tasks if isinstance(self.tasks, group)
-                 else self.tasks)
-        return sum(self._traverse_tasks(tasks, 1))
+        tasks = getattr(self.tasks, "tasks", self.tasks)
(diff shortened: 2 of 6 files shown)


celery/canvas.py
```diff
@@ -227,8 +227,7 @@ def _merge(self, args=None, kwargs=None, options=None, force=False):
         options = options if options else {}
         if self.immutable and not force:
             return (self.args, self.kwargs,
-                    dict(self.options,
-                         **options) if options else self.options)
+                    dict(options, **self.options) if options else self.options)
```
maybe-sybr, line 230: I should point this bit out specifically. It was previously expected that any options passed when freezing a task would override ones already set in the task object being frozen. This change flips that logic such that options already set on the signature will be used rather than any of the ones passed in. This change is critical for making sure group IDs flow down to chained tasks, but I'm a bit concerned about such a radical change. None of the tests seem to break AFAICT, but it would be good to get some thought from the maintainers here.

maybe-sybr:
Ah, I think the CI failures were a result of me slightly misunderstanding the base `Signature` type. Should be fixed now :crossed_fingers:

maybe-sybr:
Hmm, interestingly a group result in a chord doesn't behave like I might expect in redis. Not sure if it's my fault or not just yet:
```python
    def test_nested_chord_group_chain_group_tail(self, manager):
        """
        Sanity check that a deeply nested group is completed as expected.
    
        Groups at the end of chains nested in chords have had issues and this
        simple test sanity check that such a tsk structure can be completed.
        """
        try:
            manager.app.backend.ensure_chords_allowed()
        except NotImplementedError as e:
            raise pytest.skip(e.args[0])
    
        sig = chord(group(chain(
            identity.s(42),     # -> 42
            group(
                identity.s(),   # -> 42
                identity.s(),   # -> 42
            ),                  # [42, 42]
        )), identity.s())       # [[42, 42]]
        res = sig.delay()
>       assert res.get(timeout=TIMEOUT) == [[42, 42]]
E       assert [42, 42] == [[42, 42]]
E         At index 0 diff: 42 != [42, 42]
E         Left contains one more item: 42
E         Full diff:
E         - [[42, 42]]
E         ? -        -
E         + [42, 42]
```

It failed the other way in the previous CI pipeline because I had it asserting `res.get(timeout=TIMEOUT) == [42, 42]` without the encapsulating list :/

maybe-sybr:
It also looks like I may have broken a test case for the cache backend:
https://travis-ci.org/github/celery/celery/jobs/728244102

It expects a chorderror but it must not be getting raised as expected...

auvipy (changes requested):
File "/home/travis/build/celery/celery/.tox/3.8-integration-rabbitmq/lib/python3.8/site-packages/pytest_rerunfailures.py", line 7, in <module>
pytest is throwing the error in integration tests
    from _pytest.resultlog import ResultLog

ModuleNotFoundError: No module named '_pytest.resultlog'

maybe-sybr:
@auvipy - I think this got broken for master/everything (https://travis-ci.org/github/celery/celery/builds/730832672) :( perhaps some test deps got updated and broke it?

celery/canvas.py
auvipy: does this break existing codes?

auvipy (changes requested):
can you try to pin some test deps to run the integration tests?

thedrow:
I can't restart the build for some reason.
This does not reproduce locally for me.

auvipy:
Redis & cache integration tests are failing now

celery/canvas.py
maybe-sybr: I can't confidently say that it doesn't, but I'd need to spend a bit of time with the failing tests to be sure that this is (or isn't :crossed_fingers: ) the cause.

auvipy (changes requested):
can you re check this failure? >       assert res.get(timeout=TIMEOUT) == [[42, 42]]

E       assert [42, 42] == [[42, 42]]

E         At index 0 diff: 42 != [42, 42]

E         Left contains one more item: 42

E         Full diff:

E         - [[42, 42]]

E         ? -        -

E         + [42, 42]

t/integration/test_canvas.py:1015: AssertionError

maybe-sybr:
Hey @auvipy . Sorry this PR hasn't gotten much love recently, I've been a bit busy!

wrt that test failure, I am now under the impression that the (AFAICT undocumented?) behaviour that groups included in groups and chord get unrolled as part of the fix for #1509 many years ago might be blame for some of this. I have added some new tests which enumerate the behaviour as I understand it and should make it more clear what we need to fix prior to merging this.

I've also found a couple of other issues in my adventures testing this PR, and I'm about to push up some new commits which fix those. One is closely related to chord counting, the other is about nested groups in chains and the promises used to fulfil them. Let me know if you'd like the promises on to be split into its own PR (it's arguably separate from this PR).

maybe-sybr:
The latest build shows that the redis and key-value backends behave differently. Perhaps I've misunderstood unrolling and it doesn't happen with chords? I'm adding some more intermediate test cases locally, but if someone could chime in meanwhile to explain unrolling (if anyone knows :o ) it'd be helpful.

auvipy:
sounds great. rabbit & redis integrations tests are now passing while others are failing

celery/canvas.py
maybe-sybr: AFAICT this doesn't break anything. I could potentially change it to look more like this if you'd like:

```python
new_options = dict(self.options, **{k: v for k, v in options.items() if k not in {"group"}})
```

That would ensure that we only retain specific options from the original task which we knew need to be retained across a freeze. Thoughts, @auvipy ?

maybe-sybr:
@auvipy - I believe CI will go green with the new changeset and is hopefully ready to merge! I'll keep an eye on the CI pipelines to see how they go.

The commit on the tip of this branch is what addresses the chord header result structure issue we were seeing with the redis backend. The tl;dr is that the implementation merge for v3.1 made an optimisation which assumed all results contributing to a chord were top-level siblings. We know this isn't true because nested groups don't have an encapsulating task so each of their results need to contribute to the chord, rather than some top level "group aggregation" task.

I've left the original optimisation there since it was probably made for performance reasons in situations with large chords - but if the chord header is "complex", we save it and restore it similar to what all the other backends which support chords do. One extra little quirk is that the restored group result needs to have its promise manually fulfilled since redis is an async backend.

celery/canvas.py
auvipy: sounds good to me

auvipy:
> Hey @auvipy . Sorry this PR hasn't gotten much love recently, I've been a bit busy!
> 
> wrt that test failure, I am now under the impression that the (AFAICT undocumented?) behaviour that groups included in groups and chord get unrolled as part of the fix for #1509 many years ago might be blame for some of this. I have added some new tests which enumerate the behaviour as I understand it and should make it more clear what we need to fix prior to merging this.
> 
> I've also found a couple of other issues in my adventures testing this PR, and I'm about to push up some new commits which fix those. One is closely related to chord counting, the other is about nested groups in chains and the promises used to fulfil them. Let me know if you'd like the promises on to be split into its own PR (it's arguably separate from this PR).

that would be easier to review, your observations seem right to me

celery/canvas.py
```diff
@@ -1364,21 +1371,30 @@ def apply(self, args=None, kwargs=None,
...
-                stack.extend(task.tasks)
-            elif isinstance(task, _chain) and isinstance(task.tasks[-1], group):
-                stack.extend(task.tasks[-1].tasks)
-            else:
-                yield task if value is None else value
+    @classmethod
+    def __descend(cls, sig_obj):
+        # Sometimes serialized signatures might make their way here
+        if not isinstance(sig_obj, Signature) and isinstance(sig_obj, dict):
+            sig_obj = Signature.from_dict(sig_obj)
```
auvipy, line 1378: line #L1378 is not covered by tests yet

celery/canvas.py
```diff
@@ -663,8 +662,16 @@ def run(self, args=None, kwargs=None, group_id=None, chord=None,
...
-            first_task.apply_async(**options)
-            return results[0]
+            real_result = first_task.apply_async(**options)
+            # If we only have a single task, it may be important that we pass
+            # the real result object rather than the one obtained via freezing.
+            # e.g. For `GroupResult`s, we need to pass back the result object
+            # which will actually have its promise fulfilled by the subtasks,
+            # something that will never occur for the frozen result.
+            if not tasks:
+                return real_result
```
auvipy, line 672: line #L672 is not covered by tests yet

celery/backends/redis.py
```diff
@@ -401,12 +402,14 @@ def _unpack_chord_result(self, tup, decode,
...
-        # this flag.
-        pass
+        # If any of the child results of this chord are complex (ie. group
+        # results themselves), we need to save `header_result` to ensure that
+        # the expected structure is retained when we finish the chord and pass
+        # the results onward to the body in `on_chord_part_return()`. We don't
+        # do this is all cases to retain an optimisation in the common case
+        # where a chord header is comprised of simple result objects.
+        if any(isinstance(nr, GroupResult) for nr in header_result.results):
+            header_result.save(backend=self)
```
auvipy, line 412: line #L412 is not covered by tests yet

celery/backends/redis.py
```diff
@@ -449,27 +452,47 @@ def on_chord_part_return(self, request, state, result,
...
-                decode, unpack = self.decode, self._unpack_chord_result
-                with client.pipeline() as pipe:
-                    if self._chord_zset:
-                        pipeline = pipe.zrange(jkey, 0, -1)
-                    else:
-                        pipeline = pipe.lrange(jkey, 0, total)
-                    resl, = pipeline.execute()
-                try:
-                    callback.delay([unpack(tup, decode) for tup in resl])
+                hr = GroupResult.restore(gid)
```
auvipy, line 455: full form of hr variable would be more readable

celery/canvas.py
auvipy, line 230: some doc string would be great

celery/canvas.py
```diff
@@ -1364,21 +1371,30 @@ def apply(self, args=None, kwargs=None,
...
+            # The last element in a chain counts toward this chord
+            return cls.__descend(sig_obj.tasks[-1])
+        elif isinstance(sig_obj, chord):
+            # The child chord's body counts toward this chord
+            return cls.__descend(sig_obj.body)
+        elif isinstance(sig_obj, Signature):
+            # Each simple signature counts as 1 completion for this chord
+            return 1
+        # Any other types are assumed to be iterables of simple signatures
+        return len(sig_obj)
```
auvipy, line 1393: line #L1393 is not covered by tests yet

auvipy (changes requested):
so far great effort! I have added some of my inputs

auvipy:
would be really nice to have 100% test coverage for such a big & important change

auvipy:
splitting multiple commits in a single PR is also good.

thedrow:
I'm confused. Is #6411 part of this PR as well?

maybe-sybr:
> I'm confused. Is #6411 part of this PR as well?

I'm in the middle of splitting the change up for merge in #6411 out of this PR since it's independent. I'll have a new changeset up here shortly once I sanity check a couple of things, @thedrow .

celery/canvas.py
maybe-sybr: I'm pretty sure this also needs to be done for the mutable case on the lines below. Even if the signature is mutable, allowing the group_id to be overridden would break completion so we can't allow it to be changed. That's actually kind of nice since it means I can make a `new_options` regardless of the signature's mutability and build it like I laid out in my previous comment. I'll have some changes and unit tests for this to push soon.

celery/canvas.py
maybe-sybr: This has been pulled out into #6411 - it'll disappear from the diff when I next push

maybe-sybr:
The change from #6411 has been stripped out of this PR now. I believe I've addressed all of the coverage commentary and have left extra commentary where requested. I'll wait for the coverage report to come back before I resolve some of those review comments.

auvipy:
needs rebase

maybe-sybr:
Rebased on top of master

celery/canvas.py
maybe-sybr: Missed this line - my test was actually validating deserializing in `__init__()` by accident!

celery/backends/redis.py
maybe-sybr: This is now `header_result`

celery/canvas.py
maybe-sybr: This has some more explanation now.

celery/canvas.py
maybe-sybr: This is now covered

celery/backends/redis.py
maybe-sybr: Now covered

celery/canvas.py
maybe-sybr: This should be covered now

maybe-sybr:
This new changeset should address the last of the missed coverage :crossed_fingers:

celery/canvas.py
maybe-sybr: This is done

auvipy (approved):

maybe-sybr:
Thanks @auvipy

PR #6367, opened 2020-09-25
Update daemonizing.rst

mfonville:
Fix daemonizing documentation for issue #6363 to put `multi` before `-A`

PR #6376, opened 2020-09-30
Revert "Update daemonizing.rst"

thedrow:
Reverts celery/celery#6367

@auvipy @mfonville I'm reverting this as this is the wrong fix.
The right fix would be to figure out why the unit file does not work as expected.


thedrow:
LOL at my dysgraphia.
@mfonville I read your fix all wrong :facepalm:

mfonville:
Ok, completely understand!

PR #6411, opened 2020-10-14
fix: Pass back real result for single task chains

maybe-sybr:
## Description

When chains are delayed, they are first frozen as part of preparation
which causes the sub-tasks to also be frozen. Afterward, the final (0th
since we reverse the tasks/result order when freezing) result object
from the freezing process would be passed back to the caller. This
caused problems in signaling completion of groups contained in chains
because the group relies on a promise which is fulfilled by a barrier
linked to each of its applied subtasks. By constructing two
`GroupResult` objects (one during freezing, one when the chain sub-tasks
are applied), this resulted in there being two promises; only one of
which would actually be fulfilled by the group subtasks.

This change ensures that in the special case where a chain has a single
task, we pass back the result object constructed when the task was
actually applied. When that single child is a group which does not get
unrolled (ie. contains more than one child itself), this ensures that we
pass back a `GroupResult` object which will actually be fulfilled. The
caller can then await the result confidently!

### DIFF
--- a/celery/canvas.py
+++ b/celery/canvas.py
@@ -652,19 +652,27 @@ class _chain(Signature):
         args = (tuple(args) + tuple(self.args)
                 if args and not self.immutable else self.args)
 
-        tasks, results = self.prepare_steps(
+        tasks, results_from_prepare = self.prepare_steps(
             args, kwargs, self.tasks, root_id, parent_id, link_error, app,
             task_id, group_id, chord,
         )
 
-        if results:
+        if results_from_prepare:
             if link:
                 tasks[0].extend_list_option('link', link)
             first_task = tasks.pop()
             options = _prepare_chain_from_options(options, tasks, use_link)
 
-            first_task.apply_async(**options)
-            return results[0]
+            result_from_apply = first_task.apply_async(**options)
+            # If we only have a single task, it may be important that we pass
+            # the real result object rather than the one obtained via freezing.
+            # e.g. For `GroupResult`s, we need to pass back the result object
+            # which will actually have its promise fulfilled by the subtasks,
+            # something that will never occur for the frozen result.
+            if not tasks:
+                return result_from_apply
+            else:
+                return results_from_prepare[0]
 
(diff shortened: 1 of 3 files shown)


maybe-sybr:
This change is the one I mentioned I would pull out of #6354 - FYI @auvipy . I believe the changeset here will also address your coverage comment on that PR.

maybe-sybr:
Ah, I need to pull one of the integrations tests across from that PR as well. This is the change that makes it not time out!

celery/canvas.py
```diff
@@ -663,8 +663,16 @@ def run(self, args=None, kwargs=None, group_id=None, chord=None,
             first_task = tasks.pop()
             options = _prepare_chain_from_options(options, tasks, use_link)
 
-            first_task.apply_async(**options)
-            return results[0]
+            real_result = first_task.apply_async(**options)
```
thedrow, line 666: The term `real result` is a little bit unclear.
Can we find a better name for it?

thedrow (changes requested):
Can we add an integration test for this?

maybe-sybr:
> Can we add an integration test for this?

Yep, `test_nested_chain_group_lone()` is the one which demonstrates the fix. The other is just a sanity check for simple groups since there wasn't one in the suite.

celery/canvas.py
maybe-sybr: Hmm, perhaps `concrete_result` or `result_from_apply`? The comment block below this would explain why we'd be using `result_from_apply` adequately.

auvipy (approved):
apart from the better naming, looks good

maybe-sybr:
I went with `result_from_apply` and also changed the `results` returned from `self.prepare_steps()` to be `results_from_prepare`. That should make it more obvious, and with the block comment explaining why we should return the one `_from_apply` I think it's nice and readable :)

auvipy (approved):

maybe-sybr:
Thanks @auvipy and @thedrow :)

Issue #6409, opened 2020-10-13
Travis doesn't run pypy3 unit tests


maybe-sybr:
I noticed that the pypy3 job in travis doesn't seem to be doing anything except building and installing into the toxenv. Is that on purpose? Check out [0] as an example job. I think the TOXENV isn't being set properly - it's just "pypy3" instead of "pypy3-unit"

[0] https://travis-ci.org/github/celery/celery/jobs/735267307

maybe-sybr:
Oh, I picked the doco report since this isn't really a code bug.

thedrow:
Yes you are correct. The TOXENV is incorrect.

PR #6417, opened 2020-10-15
ci: Fix pypy3 CI

maybe-sybr:
This changeset fixes the TOXENV for pypy3 unit tests in travis and adds
rabbitmq integration tests against pypy3 as well like we do for Python 3.9.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -64,8 +64,9 @@ jobs:
       - TOXENV=flake8,apicheck,configcheck,bandit
       - CELERY_TOX_PARALLEL='--parallel --parallel-live'
     stage: lint
+
   - python: pypy3.6-7.3.1
-    env: TOXENV=pypy3
+    env: TOXENV=pypy3-unit
     stage: test
 
 before_install:


maybe-sybr:
I've put in a speculative integration test against a rabbitmq backend to see if it works. It's not really necessary to fix #6409 but can't hurt if it works (except chew CI time, so happy for it to be vetoed based on that as well).

maybe-sybr:
Looks like the unit tests were fine but one of my integration tests from #6354 timed out. Probably needs a bit of digging to work out if it's actually broken or just flaky on pypy3 for some reason. I don't have the time to dig into it right now so I'd be fine with stripping the integration test job out of this changeset and just fixing the unit tests for now. LMK what you'd like to do and I can amend the branch.

maybe-sybr:
@auvipy I've stripped the commit which added the breaking integration test for now. I can look at that at some later point, for now getting pypy3 unit tests running is a good start.

auvipy (approved):

Issue #4737, opened 2018-05-16
ImportError on GreenletExit on celery/concurrency/gevent.py


joaomg:
## Checklist

- [x] I have included the output of ``celery -A proj report`` in the issue.
      (if you are not able to do this, then at least specify the Celery
       version affected).
- [x] I have verified that the issue exists against the `master` branch of Celery.

## Steps to reproduce
I'm using Python 3.5.2 with celery 4.1.0 and gevent 1.3.0
When launching a worker, it returns an ImportError on GreenletExit

celery worker --app=superset.sql_lab:celery_app --pool=gevent -Ofair

## Expected behavior
The worker should launch.

## Actual behavior
Celery is unable to launch the worker, it return this error:

/home/joao/Superset/lib/python3.5/site-packages/celery/__init__.py:113: MonkeyPatchWarning: Monkey-patching ssl after ssl has already been imported may lead to errors, including RecursionError on Python 3.6. Please monkey-patch earlier. See https://github.com/gevent/gevent/issues/1016
  monkey.patch_all()
Loaded your LOCAL configuration at [/home/joao/Superset/superset_config.py]
/home/joaog/Superset/lib/python3.5/site-packages/flask_cache/jinja2ext.py:33: ExtDeprecationWarning: Importing flask.ext.cache is deprecated, use flask_cache instead.
  from flask.ext.cache import make_template_fragment_key
Traceback (most recent call last):
  File "/home/joao/superset/bin/celery", line 11, in <module>
    load_entry_point('celery==4.1.0', 'console_scripts', 'celery')()
...
...
...
File "/home/joao/Superset/lib/python3.5/site-packages/celery/concurrency/gevent.py", line 34, in __init__
    from gevent.greenlet import Greenlet, GreenletExit
ImportError: cannot import name 'GreenletExit'

It appears the GreenletExit should be imported like this:
from gevent import GreenletExit

arjun27:
Same issue here. I fixed it by manually installing `gevent==1.2.2`

joaomg:
Confirmed! By downgrading to 1.2.2 I'm able to launch the celery worker.
@arjun27 thanks for the hint.

auvipy:
So it;s not a celery issue? or celery need to be patched for new version of gevent?

joaomg:
In order to use gevent 1.3 there's a need to patch celery/concurrency/gevent.py:

from gevent.greenlet import Greenlet
from gevent import GreenletExit

[...]

PR #6431, opened 2020-10-21
Backport gevent import fix into 3.1 branch

DrMeers:
cherry-picked 022f447dd into 3.1 branch so the gevent import is fixed. See #4737, #4754


auvipy (approved):

auvipy:
But that branch is not maintained

PR #6432, opened 2020-10-22
Flush worker prints, notably the banner

thomas-riccardi:

## Description

In some cases (kubernetes, root) the banner is only printed at the end of the process execution, instead of at the beginning.

I have no idea why it only sometimes doesn't flush, but it's in the same scenario as in #4611: kubernetes, root.

### DIFF
--- a/celery/apps/worker.py
+++ b/celery/apps/worker.py
@@ -79,7 +79,7 @@ def active_thread_count():
 
 
 def safe_say(msg):
-    print(f'\n{msg}', file=sys.__stderr__)
+    print(f'\n{msg}', file=sys.__stderr__, flush=True)
 
 
 class Worker(WorkController):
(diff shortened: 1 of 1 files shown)


thomas-riccardi:
I'm not sure which 4.x branch is still maintained; I can provide a PR for that too (using python 2 compatible `sys.__stdout__.flush()`).

auvipy (approved):
https://github.com/celery/celery/tree/v4.5

auvipy:
can you add some regression tests as well?

thomas-riccardi:
@auvipy I'm not sure what could be tested really: mocking flush() and somehow just construct a worker and call emit_banner()?

If you think it's useful could you give me some pointers in the tests where Workers are already instantiated and for which `emit_banner` can be called?

I'll prepare a PR for v4.5.

PR #6433, opened 2020-10-22
Fix Exception throw by Retry(exc parameter)

Ixiodor:
## Description
When Retry has paramater "max_retries" set, that value is valid only for "Retry Exception" and not for exc launched from "Retry Exception".
So using Retry with max_retries = 0, exc exception still uses the "task max_retries" making use of "max_retries" useless.
An override is an override. 
max_retries init value to -1 is for not preventing "None" as value for "infinite retry".

Actually i changed the behaviour you wanted to apply. You wanted to use max_retries as an additional check outside the normal "max_retries". But there is a double exception begin executed and the second one dosn't follow the caller(Retry) parameters.

### DIFF
--- a/celery/app/task.py
+++ b/celery/app/task.py
@@ -612,7 +612,7 @@ class Task:
     subtask_from_request = signature_from_request  # XXX compat
 
     def retry(self, args=None, kwargs=None, exc=None, throw=True,
-              eta=None, countdown=None, max_retries=None, **options):
+              eta=None, countdown=None, max_retries=-1, **options):
         """Retry the task, adding it to the back of the queue.
 
         Example:
@@ -675,7 +675,8 @@ class Task:
         """
         request = self.request
         retries = request.retries + 1
-        max_retries = self.max_retries if max_retries is None else max_retries
+        if max_retries is None or max_retries >= 0:
+            self.max_retries = max_retries
 
         # Not in worker or emulated by (apply/always_eager),
         # so just raise the original exception.
@@ -694,7 +695,7 @@ class Task:
             **options
         )
 
-        if max_retries is not None and retries > max_retries:
+        if self.max_retries is not None and retries > self.max_retries:
             if exc:
                 # On Py3: will augment any current exception with
                 # the exc' argument provided (raise exc from orig)


celery/app/task.py
```diff
@@ -612,7 +612,7 @@ def signature_from_request(self, request=None, args=None, kwargs=None,
     subtask_from_request = signature_from_request  # XXX compat
 
     def retry(self, args=None, kwargs=None, exc=None, throw=True,
-              eta=None, countdown=None, max_retries=None, **options):
+              eta=None, countdown=None, max_retries=-1, **options):
```
auvipy, line 615: is max_retires=-1 actually needed?

celery/app/task.py
```diff
@@ -675,7 +675,8 @@ def retry(self, args=None, kwargs=None, exc=None, throw=True,
         """
         request = self.request
         retries = request.retries + 1
-        max_retries = self.max_retries if max_retries is None else max_retries
+        if max_retries is None or max_retries >= 0:
```
auvipy, line 678: this change seem right but needs regression tests

auvipy (changes requested):
regression tests and entry to the changelog is needed as well

celery/app/task.py
Ixiodor: Needed cause "None" is used for "unlimited retries". Elsewhere each retry without "max_retries" set will be converted to "unlimited retries" indirectly

Ixiodor:
@mchataigner Can you help on that for tests?

mchataigner:
@Ixiodor do you have a code sample so that I may understand better the issue here?

Ixiodor:
@mchataigner 
Use somenthing like that:
```
import random

@app.task(name="TestTask", bind=True, autoretry_for=(Exception,),
                  max_retries=10)
def foo(self):
    number = random.randint(0, 200)
    if number % 2 == 0: # I wanna retry in that case
        print(str(number) + " : going to retry")
        self.retry(exc=Exception('Trying again...'))
    else: # I wanna fail elsewhere
        print(str(number) + " : going to fail")
        self.retry(exc=Exception('You failed'), max_retries=0)

foo.si().delay()
```

I wanna retry in some cases but i wanna fail in other cases, i wanna fail forcing no more retries. Retry is an Exception, so it get caught by autoretry_for=(Exception,), Retry will check for max_retries and  max_retries is 0, so it should stop retry and launch the exc exception, but exc exception will be caught again by autoretry_for=(Exception,) and in this case you can't override max_retries, so the decorator(or default) will apply.

Ixiodor:
I found a trick:

```
    else: # I wanna fail elsewhere
        print(str(number) + " : going to fail")
        self.retry(exc=Retry('You failed'), max_retries=0)
```
That works cause Retry is protected from autoretry_for. Should we create another protected Exception? (Somenthing like BypassException)

Ixiodor:
Made a PR that seems better than this.
#6436

PR #6436, opened 2020-10-24
Fix max_retries override on self.retry

Ixiodor:
## Description
This seems a better patch for #6433 

This one FIX base idea from max_retries on self.retry and lets exception from self.retry escape from autoretry_for.
ProtectedException can be used even alone in a task to make it fail on spot.
`        raise ProtectedException(exc=Exception('You failed'))`
Doing this makes the task fail bypassing all. (This BUG raised cause i wanted to make a task fail on purpose from a trigger)

### DIFF
--- a/celery/app/autoretry.py
+++ b/celery/app/autoretry.py
@@ -1,7 +1,7 @@
 """Tasks auto-retry functionality."""
 from vine.utils import wraps
 
-from celery.exceptions import Ignore, Retry
+from celery.exceptions import Ignore, Retry, ProtectedException
 from celery.utils.time import get_exponential_backoff_interval
 
 
@@ -38,6 +38,8 @@ def add_autoretry_behaviour(task, **options):
                 raise
             except Retry:
                 raise
+            except ProtectedException as exc:
+                raise exc.incapsulated
             except autoretry_for as exc:
                 if retry_backoff:
                     retry_kwargs['countdown'] = \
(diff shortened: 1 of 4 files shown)


auvipy (changes requested):
1 unit test is failing. could you please recheck?

auvipy (changes requested):
can you add an integration test for this?

Ixiodor:
There is another BUG.
Explicit raised Exceptions are raised as "ProtectedException". Trying to fix.

Ixiodor:
I don't really like how autoretry_for works now. There is an useless layer(autoretry.py). I think it's better to call always retry(task.py) and handle here autoretry_for. 
What happens now is:
Calling self.retry => retry => autoretry_for
Calling raise => autoretry_for => retry

Ixiodor:
Someone wrote some tests with value returned instead of value expected...I'm fixing all wrong tests pulled out by solving that bug

Ixiodor:
Less intrusive change.

auvipy:
@safwanrahman can you try this patch?

auvipy:
@mchataigner

thedrow:
Please address the lint errors.

auvipy (approved):
not sure lint error is related

thedrow:
It was. I fixed it myself.

Ixiodor:
@thedrow wich error happened in lint?

Issue #6445, opened 2020-10-29
workdir option no longer works after upgrade to 5.0.1


Miraliu:
# Checklist

- [x] I have verified that the issue exists against the `master` branch of Celery.
- [x] This has already been asked to the [discussion group](https://groups.google.com/forum/#!forum/celery-users) first.
- [x] I have read the relevant section in the
  [contribution guide](http://docs.celeryproject.org/en/latest/contributing.html#other-bugs)
  on reporting bugs.
- [x] I have checked the [issues list](https://github.com/celery/celery/issues?q=is%3Aissue+label%3A%22Issue+Type%3A+Bug+Report%22+-label%3A%22Category%3A+Documentation%22)
  for similar or identical bug reports.
- [x] I have checked the [pull requests list](https://github.com/celery/celery/pulls?q=is%3Apr+label%3A%22PR+Type%3A+Bugfix%22+-label%3A%22Category%3A+Documentation%22)
  for existing proposed fixes.
- [x] I have checked the [commit log](https://github.com/celery/celery/commits/master)
  to find out if the bug was already fixed in the master branch.
- [x] I have included all related issues and possible duplicate issues
  in this issue (If there are none, check this box anyway).

## Mandatory Debugging Information

- [] I have included the output of ``celery -A proj report`` in the issue.
    (if you are not able to do this, then at least specify the Celery
     version affected).
- [x] I have verified that the issue exists against the `master` branch of Celery.
- [x] I have included the contents of ``pip freeze`` in the issue.
- [x] I have included all the versions of all the external dependencies required
  to reproduce this bug.

## Optional Debugging Information

- [ ] I have tried reproducing the issue on more than one Python version
  and/or implementation.
- [ ] I have tried reproducing the issue on more than one message broker and/or
  result backend.
- [ ] I have tried reproducing the issue on more than one version of the message
  broker and/or result backend.
- [ ] I have tried reproducing the issue on more than one operating system.
- [ ] I have tried reproducing the issue on more than one workers pool.
- [ ] I have tried reproducing the issue with autoscaling, retries,
  ETA/Countdown & rate limits disabled.
- [x] I have tried reproducing the issue after downgrading
  and/or upgrading Celery and its dependencies.

#### Related Issues

- None

#### Possible Duplicates

- None

## Environment & Settings

**Celery version**:

<details>
<summary><b><code>celery report</code> Output:</b></summary>
<p>

software -> celery:5.0.1 (singularity) kombu:5.0.2 py:3.8.6
[...]

PR #6446, opened 2020-10-30
fix: Move to workdir before app import

maybe-sybr:

### DIFF
--- a/celery/bin/celery.py
+++ b/celery/bin/celery.py
@@ -46,6 +46,13 @@ class App(ParamType):
     name = "application"
 
     def convert(self, value, param, ctx):
+        # We need to ensure that we can find an app importable from the
+        # specified working directory (partially works around #6445). This
+        # `chdir()` has a problem where the `workdir` argument may not have
+        # been processed yet since this is a processing time type coercion,
+        # rather than one which occurs after all arguments have been processed.
+        if ctx is not None and "workdir" in ctx.params:
+            os.chdir(ctx.params["workdir"])
         try:
             return find_app(value)
         except ModuleNotFoundError as e:


thedrow:
If there's no other choice, we'll merge this.
The workaround looks good.

thedrow (approved):

maybe-sybr:
@thedrow - when do you plan to call 5.0.2 done? I'll either unmark this as a draft or put up a more convincing PR before then.

auvipy:
> @thedrow - when do you plan to call 5.0.2 done? I'll either unmark this as a draft or put up a more convincing PR before then.

you can try to improve the PR in the meantime

thedrow:
We have a couple of very urgent fixes in 5.0.2 so I'd rather release sooner than later.
I'll work on the changelog today and if I don't have a better fix I'll merge this one. We'll keep #6445 open until we have a better fix.

auvipy:
that would be practical to merge and release faster

maybe-sybr:
I'll de-draft this so we can get it in immediately for 5.0.3, and take #6445 to look into the arg ordering issue which @Miraliu described more seriously.

celery/bin/celery.py
```diff
@@ -46,6 +46,13 @@ class App(ParamType):
     name = "application"
 
     def convert(self, value, param, ctx):
+        # We need to ensure that we can find an app importable from the
+        # specified working directory (partially works around #6445). This
+        # `chdir()` has a problem where the `workdir` argument may not have
+        # been processed yet since this is a processing time type coercion,
+        # rather than one which occurs after all arguments have been processed.
+        if ctx is not None and "workdir" in ctx.params:
```
maybe-sybr, line 54: > it needs to be able to deal with param and context being None. This can be the case when the object is used with prompt inputs.

From https://click.palletsprojects.com/en/7.x/api/?highlight=convert#click.ParamType.convert

maybe-sybr:
It turns out the eager option fix was very easy and appears to behave as expected for arguments regardless of their ordering. I've put up #6457 and will close this for the moment to avoid confusion. If we're unsure of that PR, we can re-open this one.

PR #6469, opened 2020-11-08
Fix passing queues into purge command

last-partizan:
In current wersion calling `celery --app my.celery_app purge -Q queue_name` is failing with following trace:

```
    names = (queues or set(app.amqp.queues.keys())) - exclude_queues
TypeError: unsupported operand type(s) for -: 'list' and 'list'
```

Becouse code is expecting set and `queues` is actually a list.

Here is a fix.

### DIFF
--- a/celery/bin/purge.py
+++ b/celery/bin/purge.py
@@ -32,10 +32,10 @@ def purge(ctx, force, queues, exclude_queues):
 
         There's no undo operation for this command.
     """
-    queues = queues or set()
-    exclude_queues = exclude_queues or set()
     app = ctx.obj.app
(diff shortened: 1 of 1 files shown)


thedrow (approved):
LGTM.
Thanks!

jaoxford:
Just what I needed. Was unable to purge a specific queue name on 5.0.2. And the default queue name was 'celery' so it was impossible.

Thanks @last-partizan !

thedrow:
@jaoxford I'm very pleased you are happy with this fix but every time you comment, we get a ping and we're already overflowing.
This takes extra work to go through all the issues.
Please use reactions to show gratitude going forward.
Thank you.

PR #6505, opened 2020-11-29
Fix inconsistency in documentation for `link_error`

### LINKED ISSUE #4099: link_error argument of apply_async doesn't work as documented
## Checklist
```
software -> celery:4.0.2 (latentcall) kombu:4.0.2 py:3.5.3
            billiard:3.5.0.2 redis:2.10.5
settings -> transport:redis results:redis://localhost/
```

The documentation about linking tasks at http://docs.celeryproject.org/en/master/userguide/calling.html#linking-callbacks-errbacks doesn't work. It also doesn't match the documentation at http://docs.celeryproject.org/en/master/userguide/canvas.html#chains => "Here's an example errback..." which does work.

## Steps to reproduce
Follow the code example for linking tasks at http://docs.celeryproject.org/en/master/userguide/calling.html#linking-callbacks-errbacks

## Expected behavior
Prints the error and traceback

## Actual behavior
produces a `RuntimeError: Never call result.get() within a task!` exception.

henribru:
## Description

- Makes the documentation for `link_error` consistent between the canvas and calling pages. I've just copied over the text from the canvas page, but let me know if it should be adjusted to better fit the context of the calling page. Fixes #4099
- Fix an undefined variable in the existing new-style errback example

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -278,3 +278,4 @@ Dipankar Achinta, 2019/10/24
 Sardorbek Imomaliev, 2020/01/24
 Maksym Shalenyi, 2020/07/30
 Frazer McLean, 2020/09/29
+Henrik Bruåsdal, 2020/11/29
--- a/docs/userguide/calling.rst
+++ b/docs/userguide/calling.rst
@@ -135,23 +135,18 @@ task that adds 16 to the previous result, forming the expression
 
 
 You can also cause a callback to be applied if task raises an exception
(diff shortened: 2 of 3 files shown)


maybe-sybr (approved):
lgtm. This aligns the documentation in `calling.rst` with the changes @ask made in ae3f36f7 when the new errback signatures were introduced for 4.0. The snippets in both files now appear to be correct, referring to `request.id` correctly.

Thanks for the contribution!

Issue #6007, opened 2020-03-20
The `celery_enable_logging` fixture doesn't do anything


ktchen14:
# Checklist
* [x] I have verified that the issue exists against the `master` branch of Celery.
* [x] This has already been asked to the [discussion group](https://groups.google.com/forum/#!forum/celery-users) first.
* [x] I have read the relevant section in the
  [contribution guide](http://docs.celeryproject.org/en/latest/contributing.html#other-bugs)
  on reporting bugs.
* [x] I have checked the [issues list](https://github.com/celery/celery/issues?q=is%3Aissue+label%3A%22Issue+Type%3A+Bug+Report%22+-label%3A%22Category%3A+Documentation%22)
  for similar or identical bug reports.
* [x] I have checked the [pull requests list](https://github.com/celery/celery/pulls?q=is%3Apr+label%3A%22PR+Type%3A+Bugfix%22+-label%3A%22Category%3A+Documentation%22)
  for existing proposed fixes.
* [x] I have checked the [commit log](https://github.com/celery/celery/commits/master)
  to find out if the bug was already fixed in the master branch.
* [x] I have included all related issues and possible duplicate issues
  in this issue (If there are none, check this box anyway).

## Mandatory Debugging Information
* [x] I have included the output of ``celery -A proj report`` in the issue.
    (if you are not able to do this, then at least specify the Celery
     version affected).
* [x] I have verified that the issue exists against the `master` branch of Celery.
* [x] I have included the contents of ``pip freeze`` in the issue.
* [x] I have included all the versions of all the external dependencies required
  to reproduce this bug.

## Environment & Settings
**Celery version**: All

# Steps to Reproduce
## Required Dependencies
* **Minimal Python Version**: N/A
* **Minimal Celery Version**: N/A
* **Minimal Kombu Version**: N/A
* **Minimal Broker Version**: N/A
* **Minimal Result Backend Version**: N/A
* **Minimal OS and/or Kernel Version**: N/A
* **Minimal Broker Client Version**: N/A
* **Minimal Result Backend Client Version**: N/A

### Python Packages
<details>
<summary><b><code>pip freeze</code> Output:</b></summary>
<p>

```
pytest
```

</p>
</details>

## Minimally Reproducible Test Case
<details>
<p>

```python
from celery.signals import setup_logging

@setup_logging.connect
def on_setup_logging(sender, *args, **kwargs):
    print("Hello")
```

```python
import pytest

@pytest.fixture
def celery_enable_logging():
[...]

PR #6526, opened 2020-12-07
Make `setup_default_app` actually set the default app

### LINKED ISSUE #6525: setup_default_app doesn't use app argument (or the current_app fixture does nothing)
# Checklist
* [x] I have verified that the issue exists against the `master` branch of Celery.
* [ ] This has already been asked to the [discussion group](https://groups.google.com/forum/#!forum/celery-users) first.
* [ ] I have read the relevant section in the
  on reporting bugs.
* [x] I have checked the [issues list](https://github.com/celery/celery/issues?q=is%3Aissue+label%3A%22Issue+Type%3A+Bug+Report%22+-label%3A%22Category%3A+Documentation%22)
* [x] I have checked the [pull requests list](https://github.com/celery/celery/pulls?q=is%3Apr+label%3A%22PR+Type%3A+Bugfix%22+-label%3A%22Category%3A+Documentation%22)
  for existing proposed fixes.

* Version: Celery 4.4.7, 5.0.2, and `master`

# Problem
I was trying to set up testing for a project using Celery using py.test and while wondering why `celery_app` didn't seem to have an effect (accessing `celery.current_app` printed out a decidedly non-test app), I think I may have found a bug in the implementation of `setup_default_app` (which is used by the `celery_app` fixture).

If you look at the implementation below (which is the same in 4.4.7, https://github.com/celery/celery/blob/d50c6cfc448e9e31384190b23f285254cb44e8a2/celery/contrib/testing/app.py), the `app` argument is actually never assigned to the current state globals. I believe this problem has existed for a long while, effectively making `setup_default_app` a no-op unless the trap is enabled.

https://github.com/celery/celery/blob/f50cf7d9944558167b85c14d73e8f790da251730/celery/contrib/testing/app.py#L68-L109

The `celery_app` pytest fixture attempts to use `setup_default_app()`: https://github.com/celery/celery/blob/c3e041050ae252be79d9b4ae400ec0c5b2831d14/celery/contrib/pytest.py#L40-L41

but when the trap argument is False, no change is done to the ambient current app setting.

https://github.com/celery/celery/issues/6007 may be vaguely related to this.

akx:
Fixes #6525
Refs #6007

## Description
This PR:

* refactors the `set_trap()` context manager into a more reusable internal `_set_app()` context manager
* has `setup_default_app()` actually use that context manager when not using a trap app

This seems like the correct thing to do (after all, otherwise `setup_default_app()` doesn't really use `app`), but I'm afraid people who have assumed `celery_app` _doesn't_ set the app as the default app could be upset by the change.

### DIFF
--- a/celery/contrib/testing/app.py
+++ b/celery/contrib/testing/app.py
@@ -2,6 +2,7 @@
 import weakref
 from contextlib import contextmanager
 from copy import deepcopy
+from types import SimpleNamespace
 
 from kombu.utils.imports import symbol_by_name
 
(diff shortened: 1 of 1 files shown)


auvipy (changes requested):
hey, thanks for tackling this! I would like to see unit & integration tests for these changes. Also, it would be great if we could keep the BC & mention it in the changelog/docs as well

thedrow:
I think this change is good but it definitely needs test coverage.
@akx Do you have time to do that?

thedrow:
@akk ping?

akx:
@thedrow Sorry, I don't think I have the bandwidth for that. With this and other issues with Celery I'm looking into switching to Huey or a bespoke solution instead.

thedrow:
That's a shame.
I'm sorry you're having such a hard time.

auvipy:
unit tests seems passing, but codecov report not generated, have to double check before merge.

thedrow:
I'm closing this since it is going to be resolved in 5.3 with #7149.

PR #6528, opened 2020-12-07
Fixes #3773 if we get a broken pipe exit

### LINKED ISSUE #3773: Couldn't ack, reason: BrokenPipeError(32, 'Broken pipe')
## Checklist

       version affected).
```
software -> celery:4.0.2 (latentcall) kombu:4.0.2 py:3.4.3
```

## Steps to reproduce
Upgrade to Celery 4.0.2.

## Expected behavior
No error.

## Actual behavior
Getting this error from my workers:

```
[2017-01-19 04:07:57,411: CRITICAL/MainProcess] Couldn't ack 461, reason:BrokenPipeError(32, 'Broken pipe')
  File "/opt/app/current/venv/lib/python3.4/site-packages/kombu/message.py", line 130, in ack_log_error
    self.ack(multiple=multiple)
  File "/opt/app/current/venv/lib/python3.4/site-packages/kombu/message.py", line 125, in ack
    self.channel.basic_ack(self.delivery_tag, multiple=multiple)
  File "/opt/app/current/venv/lib/python3.4/site-packages/amqp/channel.py", line 1408, in basic_ack
    spec.Basic.Ack, argsig, (delivery_tag, multiple),
  File "/opt/app/current/venv/lib/python3.4/site-packages/amqp/abstract_channel.py", line 64, in send_method
    conn.frame_writer(1, self.channel_id, sig, args, content)
  File "/opt/app/current/venv/lib/python3.4/site-packages/amqp/method_framing.py", line 174, in write_frame
    write(view[:offset])
  File "/opt/app/current/venv/lib/python3.4/site-packages/amqp/transport.py", line 269, in write
    self._write(s)
BrokenPipeError: [Errno 32] Broken pipe
```

This causes the worker to reset it's connection to RabbitMQ every few seconds, and sometimes leads to hitting the max connection limit allowed on the RabbitMQ server.

jagguli:

## Description
This pull request fixes #3773, 

This makes celery worker exit if it gets a broken pipe error

During our testing and in production we hit this error quite frequently


thedrow:
What happens after the worker exits? Does the task execute again?
Is a new worker spawned?

jagguli:
nope, it just exits, you need to configure your service manager ie, supervisor/k8s/systemd to respawn the worker

auvipy:
> nope, it just exits, you need to configure your service manager ie, supervisor/k8s/systemd to respawn the worker

can you share your service mnager config as part of this PR?

thedrow:
> > nope, it just exits, you need to configure your service manager ie, supervisor/k8s/systemd to respawn the worker
> 
> can you share your service mnager config as part of this PR?

Why? Why does it make sense to just exit?
Assuming that the user has a process manager is not a good idea.
What if the user is encountering this error during development and the worker just exits?

This PR as it stands is just a workaround.
What happens if instead we close the connection and reopen it?

thedrow:
I'd be happy to get more eyes on this.

maybe-sybr:
Commenting with a rough understanding of the logic here, please do correct me if anything sounds wrong. I'm pretty sure that this is a sledgehammer workaround which we won't be able to merge, Luckily, it'd be something which is simple enough to monkey patch in as a workaround if anyone wanted to do so and were confident in their process management respawning workers.

From what I can tell, we should probably be spending effort trying to get #5581 merged which appears to be all about better handling of flakey/dropped connections to the broker via the heartbeat mechanism. That wouldn't directly fix this issue but would cause us to recognise the issue and respawn consumers when we do, which (IIUC) should allow the worker to resume consuming tasks.

jagguli:
I run several celery processes in a k8s ha cluster, so letting processes fail is the easiest and cleanest solution. Now I have chosen this path because of my recent erlang enlightnenment  'let it crash' :grimacing: but more practically because it works for me. Only thing I'm having trouble with is detecting if a worker is in the hung state.

I have obsrved this happening when we suddenly have a spike of messages, but sometimes thats not the case, so its kind of a heisenbug. I think something is happening at the gevent level, when its under resource constraints.

So I guess the first step would be to be able to reliably replicate it. There has been a lot of discussion on the original issue report https://github.com/celery/celery/issues/3773. And I implemented the patch by @bremac https://github.com/celery/celery/commit/385a60df09201a17ad646c71eb1c00255d0a4431?diff=unified 
But this didnt seem to work for me, because the issue started again (I will need to re-test it, could have been because of a version mismatch). Thats what is in the PRs. 
 
This is definitely a gevent thing, just need to setup a testcase to replicate this reliably, maybe I need to look into load testing celery ?

thedrow:
I'm confused. You have both eventlet and gevent in that traceback.
How is that possible?

jagguli:
Yea that is weird, not sure whats going on there...

jagguli:
FWIW here's the command line

```               readOnly: false
          command:
            - /usr/local/bin/celery
          args:
            - --app=cloud.celeryapp:app
            - worker
            - --uid=1
            - -O
            - fair
            - --queues=ha.celery,_user_stats,ha.pointzi,ha.push
            - --loglevel=DEBUG
            - --concurrency=150
            - --max-tasks-per-child=2500
            - --pool=gevent
            - --soft-time-limit=1200
            - --time-limit=3600
            - --without-gossip
            - --without-mingle
            - --heartbeat-interval=50
#- --without-heartbeat

          env:
             - name: DJANGO_SETTINGS_MODULE
               value: cloud.settings
             - name: PYTHONPATH
               value: /opt/streethawk/cloud
             - name: C_FORCE_ROOT
               value: "true"
             - name: GEVENT_MONITOR_THREAD_ENABLE
               value: "0"
             - name: GEVENT_MAX_BLOCKING_TIME
               value: "0.5"
             - name: POSTGRES_PASSWORD
               valueFrom:
                 secretKeyRef:
                   name: postgres
                   key: password
```

jagguli:
I've removed eventlet and upgraded to the latest version of gevent.

Let me see if that solves this issue.

auvipy:
> I've removed eventlet and upgraded to the latest version of gevent (20.9.0).
> 
> Let me see if that solves this issue.

what happens if you use https://github.com/gevent/gevent/releases/tag/21.1.2 ?

yshaban:
Would love to see this merged along with the following:
https://github.com/celery/kombu/pull/1231
https://github.com/celery/celery/pull/5581

This patch really does work!

thedrow:
See our comments above. It doesn't work.
Simply exiting celery is not good enough.

auvipy (changes requested):
needs rebase and the tests to pass

thedrow:
> needs rebase and the tests to pass
> 
> =========================== short test summary info ============================ FAILED t/unit/worker/test_loops.py::test_asynloop::test_on_task_received - As... FAILED t/unit/worker/test_loops.py::test_asynloop::test_on_task_received_executes_on_task_message = 2 failed, 2629 passed, 17 skipped, 3 xfailed, 37 warnings in 78.77s (0:01:18) =

But why merge this if this is the wrong fix? You can't just exit.

SuperMasterBlasterLaser:
I think exit option should be as some kind of worker argument. The problem is that celery workers that are being run by kubertenetes just hang and consume CPU with RAM without doing anything. When exit occures k8s just recreate this pod.

PR #6532, opened 2020-12-08
Change deprecated from collections import Mapping/MutableMapping to from collections.abc import ...

frantisekz:

## Description

'from collections import Mapping' is deprecated since Python 3.3 and will be removed in Python 3.10 .

Since celery supports only Python >= 3.6 according to tox.ini and this form of import should work since Python 3.3, I think it's okay to make the change without try/except ImportError .


auvipy (approved):

thedrow:
Thanks!

PR #6536, opened 2020-12-09
Allow pycurl==7.43.0.6 to be used too

bobbyrenwick:

## Description

The improved SSL backend detection released in pycurl 7.43.0.6 makes life a lot easier but you can't currently install `pycurl==7.43.0.6` and `celery[sqs]` at the same time because you end up with a ResolutionImpossible error.

```
The conflict is caused by:
    The user requested pycurl==7.43.0.6
    celery[sqs] 4.4.7 depends on pycurl==7.43.0.5; extra == "sqs"
```

Please ignore the 4.4.7, it's still the same on v5.0.4

### DIFF
--- a/requirements/extras/sqs.txt
+++ b/requirements/extras/sqs.txt
@@ -1,2 +1,2 @@
 boto3>=1.9.125
-pycurl==7.43.0.5  # Latest version with wheel built (for appveyor)
+pycurl>=7.43.0.5,<7.44.0  # Latest version with wheel built (for appveyor)
--- a/requirements/test-ci-default.txt
+++ b/requirements/test-ci-default.txt
@@ -20,4 +20,4 @@
 -r extras/azureblockblob.txt
 
 # SQS dependencies other than boto
-pycurl==7.43.0.5  # Latest version with wheel built (for appveyor)
+pycurl==7.43.0.6  # Latest version with wheel built (for appveyor)


auvipy:
did you run tests with the upgraded dependency?

bobbyrenwick:
I'm not sure why the tests timed out. I have been using `celery[sqs]==4.4.7` together with `pycurl==7.43.0.6` without issue in production for a couple of months but the latest versions of pip don't allow me to do that any more. I've temporarily fixed my issue by pinning pip to an earlier version, but would like to get that up to date when possible.

I've bumped the pycurl version you use for CI. Please let me know if there's anything else I can do.

bobbyrenwick:
Looks like there's no wheel for `7.43.0.6`. I wonder why there was one for 0.5 but not for this release?

maybe-sybr:
Given the comment in the requirements file and the history of changes to versions which have published wheels, I think we'll have to close this PR since it seems like 7.43.0.6 only has an sdist as you've noted, @bobbyrenwick .

When you say the SSL backend detection makes life easier, how do you mean that? Perhaps there's something else we can do to cater to appveyor/windows distributions if the benefit is worthwhile. For now I'll close this since I don't think it's mergable, but we can definitely continue the discussion here or create a new issue when we have any ideas.

alext:
Hi, would it be possible to revisit this?

At the moment, pycurl 7.43.0.5 does not work with the current version of curl shipped with Homebrew on the Mac.  This is because it is compiled with multiple SSL backends (Apple Secure Transport and openssl) which this version of pycurl doesn't handle, leading to a runtime error because the detected SSL backend doesn't match the compile-time backend.

pycurl 7.43.0.6 adds support for multiple SSL backends (https://github.com/pycurl/pycurl/pull/639) and therefore works with the Homebrew version of libcurl.

As things stand it doesn't appear to be possible to use celery[sqs] on a Mac with curl provided by Homebrew.

thedrow:
You can avoid using the extra and install the `pycurl` version that works for you.

PR #5721, opened 2019-09-10
Fix type error in S3 backend caused by key / task ID

MrSaints:
The S3 backend for Celery was introduced in cd6c7444cadbdaaccca14a986e7e1b3330c33024. The key received by `def get(self, key)`, and `def set(self, key, value)` however, is of type `bytes` causing one of the following exceptions in Python 3.7:

```
TypeError                                 Traceback (most recent call last)
<ipython-input-14-d851513b5988> in <module>
----> 1 result.ready()

/usr/local/lib/python3.7/site-packages/celery/result.py in ready(self)
    311         for retry then :const:`False` is returned.
    312         """
--> 313         return self.state in self.backend.READY_STATES
    314 
    315     def successful(self):

/usr/local/lib/python3.7/site-packages/celery/result.py in state(self)
    471                 then contains the tasks return value.
    472         """
--> 473         return self._get_task_meta()['status']
    474     status = state  # XXX compat
    475 

/usr/local/lib/python3.7/site-packages/celery/result.py in _get_task_meta(self)
    410     def _get_task_meta(self):
    411         if self._cache is None:
--> 412             return self._maybe_set_cache(self.backend.get_task_meta(self.id))
    413         return self._cache
    414 

/usr/local/lib/python3.7/site-packages/celery/backends/base.py in get_task_meta(self, task_id, cache)
    384                 pass
    385 
--> 386         meta = self._get_task_meta_for(task_id)
    387         if cache and meta.get('status') == states.SUCCESS:
    388             self._cache[task_id] = meta

/usr/local/lib/python3.7/site-packages/celery/backends/base.py in _get_task_meta_for(self, task_id)
    725     def _get_task_meta_for(self, task_id):
    726         """Get task meta-data for a task by id."""
--> 727         meta = self.get(self.get_key_for_task(task_id))
    728         if not meta:
    729             return {'status': states.PENDING, 'result': None}

/usr/local/lib/python3.7/site-packages/celery/backends/s3.py in get(self, key)
     58 
     59     def get(self, key):
---> 60         s3_object = self._get_s3_object(key)
     61         try:
     62             s3_object.load()

/usr/local/lib/python3.7/site-packages/celery/backends/s3.py in _get_s3_object(self, key)
     54 
     55     def _get_s3_object(self, key):
---> 56         key_bucket_path = self.base_path + key if self.base_path else key
     57         return self._s3_resource.Object(self.bucket_name, key_bucket_path)
     58 

TypeError: can only concatenate str (not "bytes") to str
[...]

PR #6537, opened 2020-12-10
Fix type error in S3 backend

razorx89:
Fixes a `TypeError` when deleting a task result with S3 backend. This links to #5721, which already solved this problem for `.get()` and `.set()`, but not for `.delete()`


auvipy (approved):

auvipy:
thanks, both of you

PR #6550, opened 2020-12-16
Update celerybeat

clichedmoog:

## Description

I would upgrade init.d script to systemd for the demonization in the future.
But our team still using init.d and updated the celery version recently,
And found celerybeat not starting using the init.d script,
So this is a simple change by celery 5.x command format. I've changed the minimum portion only.


extra/generic-init.d/celerybeat
```diff
@@ -110,7 +110,7 @@ DEFAULT_USER="celery"
 DEFAULT_PID_FILE="/var/run/celery/beat.pid"
```
auvipy, line 110: are you beaten by line 110 and 111? or they are working well

auvipy (approved):

Issue #6307, opened 2020-08-18
celery multi - preload options unrecognized


maconfr:
Preload options are not unrecognized anymore since celery v4.4.4

The example at [Preload options](https://docs.celeryproject.org/en/stable/userguide/extending.html#preload-options) works with celery v4.4.3 but not with >=4.4.4.

```python3
# app.py
from celery import Celery
from celery import signals
from celery.bin import Option

app = Celery()


def add_preload_options(parser):
    parser.add_argument(
        "-Z", "--template", default="default", help="Configuration template to use.",
    )


app.user_options["preload"].add(add_preload_options)


@signals.user_preload_options.connect
def on_preload_parsed(options, **kwargs):
    print(options["template"])
```

Here is the output with v4.4.3:

```bash
$ venv/bin/celery \
> multi start worker1 \
> -A app \
> --template=foo \
> --concurrency=1
celery multi v4.4.3 (cliffs)
> Starting nodes...
foo
	> worker1@d113: OK
$ venv/bin/celery -A app report
default

software -> celery:4.4.3 (cliffs) kombu:4.6.11 py:3.7.3
            billiard:3.6.3.0 py-amqp:2.6.1
platform -> system:Linux arch:64bit, ELF
            kernel version:4.19.0-10-amd64 imp:CPython
loader   -> celery.loaders.app.AppLoader
settings -> transport:amqp results:disabled
```

This is the output with v4.4.7:

```bash
$ venv/bin/celery \
> multi start worker1 \
> -A app \
> --template=foo \
> --concurrency=1
celery multi v4.4.7 (cliffs)
> Starting nodes...
usage: /tmp/celery_preload/venv/lib/python3.7/site-packages/celery/bin/celery.py worker [options] 
/tmp/celery_preload/venv/lib/python3.7/site-packages/celery/bin/celery.py: error: unrecognized arguments: --template=foo
	> worker1@d113: * Child terminated with exit code 2
FAILED
$ venv/bin/celery -A app report
default

software -> celery:4.4.7 (cliffs) kombu:4.6.11 py:3.7.3
            billiard:3.6.3.0 py-amqp:2.6.1
platform -> system:Linux arch:64bit, ELF
            kernel version:4.19.0-10-amd64 imp:CPython
loader   -> celery.loaders.app.AppLoader
settings -> transport:amqp results:disabled
```

auvipy:
can you help to find out the commit that could cause the potential regression?

maconfr:
[...]

PR #6553, opened 2020-12-21
Allow custom args to pass to cluster start

JackBurdick:
Though preload options were fixed by https://github.com/celery/celery/issues/6307, `multi` start will not accept the custom arguments as expected.

## Description

The `multi` command will not accept custom preload args on start -- if used, the following traceback is given:

example call

```bash
celery --app=celery_app --workdir=/path/to/project multi start worker_1 worker_2 --pidfile='/var/run/celery/%n.pid' --logfile='/var/log/celery/%n.log' --loglevel='INFO' -P solo -c 1 -Q:main 'collect' -P:main solo -c:main 1  -Q:worker_2 'q_1' -P:worker_2 solo -c:worker_2 1 -Z '<custom_arg_string>' -l debug
```
where `-Z` is the custom argument

```bash
  File "/home/pi/aero/bin/celery", line 10, in <module>
    sys.exit(main())
  File "/home/pi/aero/lib/python3.7/site-packages/celery/__main__.py", line 15, in main
    sys.exit(_main())
  File "/home/pi/aero/lib/python3.7/site-packages/celery/bin/celery.py", line 213, in main
    return celery(auto_envvar_prefix="CELERY")
  File "/home/pi/aero/lib/python3.7/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/home/pi/aero/lib/python3.7/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/home/pi/aero/lib/python3.7/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/pi/aero/lib/python3.7/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/pi/aero/lib/python3.7/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/home/pi/aero/lib/python3.7/site-packages/click/decorators.py", line 21, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/home/pi/aero/lib/python3.7/site-packages/celery/bin/base.py", line 132, in caller
    return f(ctx, *args, **kwargs)
TypeError: multi() got an unexpected keyword argument '<some_preload_arg>'
```

By allowing **kwargs in the function definition (from the MR), the app will then started as expected.

NOTE: I'm new-ish to celery and my proposed change should be looked over carefully. Also, I would have expected to have to pass **kwargs into `cluster.start(**kwargs)` as well, but even without the workers start as expected with the correct options (as far as I can tell).

Happy to raise this as an issue instead if easier. Thanks

### DIFF
--- a/celery/bin/multi.py
+++ b/celery/bin/multi.py
@@ -288,7 +288,7 @@ class MultiTool(TermLogger):
 
     @splash
     @using_cluster
-    def start(self, cluster, **kwargs):
+    def start(self, cluster):
         self.note('> Starting nodes...')
         return int(any(cluster.start()))
 
@@ -469,7 +469,7 @@ class MultiTool(TermLogger):
 )
 @click.pass_context
 @handle_preload_options
-def multi(ctx):
+def multi(ctx, **kwargs):
     """Start multiple worker instances."""
     cmd = MultiTool(quiet=ctx.obj.quiet, no_color=ctx.obj.no_color)
     # In 4.x, celery multi ignores the global --app option.
--- a/celery/bin/multi.py
+++ b/celery/bin/multi.py
@@ -288,7 +288,7 @@ class MultiTool(TermLogger):
 
     @splash
     @using_cluster
-    def start(self, cluster):
+    def start(self, cluster, **kwargs):
         self.note('> Starting nodes...')
         return int(any(cluster.start()))
 


celery/bin/multi.py
```diff
@@ -288,7 +288,7 @@ def _handle_reserved_options(self, argv):
 
     @splash
     @using_cluster
-    def start(self, cluster):
+    def start(self, cluster, **kwargs):
```
auvipy, line 291: thanks for the PR, can you please add some test for the **kwargs part?

auvipy (changes requested):

celery/bin/multi.py
thedrow: No need for CLI tests right now.
I'll work on them later.

thedrow (changes requested):
While this patch causes `celery multi` to no longer crash it won't really do anything with the preload options.
You need to dig deeper into the code to ensure that the `kwargs` are passed to each node.

JackBurdick:
Thanks for checking!

Yeah, I was surprised my workers started as expected (at least as far as I can tell) w/out passing **kwargs through and chalked it up to not understanding how celery worked fully.

I have a busy couple days but will try and track this down further -- thank you again.

thedrow:
@JackBurdick This is a friendly ping to check this PR out.

JackBurdick:
Hi @thedrow -- really sorry, forgot about this as my device was "working".. If it's ok I will investigate this again this weekend, but I'm not sure I'm qualified to fix this beyond what might only be a bandaid fix.

thedrow:
An analysis which would lead to any fix would be nice to have.
I'll make sure that we merge the right fix.

JackBurdick:
OK -- I've updated the commit to be what is presently working on my devices.

NOTE: I've only tried the `start` command.

I would really appreciate someone to confirm this works for them as well.

If this change is not present, my workers do not start (at all). However, with the change present all my workers start as I would expect them to.

If I inspect my process I see all the expected arguments to each worker and they behave according to the settings I am applying in my custom argument (in this case a config specifying options for each worker).

auvipy (approved):

celery/bin/multi.py
```diff
@@ -469,7 +469,7 @@ def DOWN(self):
 )
 @click.pass_context
 @handle_preload_options
-def multi(ctx):
+def multi(ctx, **kwargs):
```
thedrow, line 472: This won't do anything but accept additional arguments.
You aren't passing it anywhere 🤷 .

Can you provide a test case where this works as intended?

thedrow (changes requested):

JackBurdick:
Hey, I'm sorry, I was not able to track down an explanation for this and am going to close out the MR. I could share an example of this working/not working, but I'm unsure of how best to capture this in a test at this point.

Realistically, I should have opened as an issue in the first place rather than an MR.

My best guess is that the individual nodes must be starting from somewhere else (maybe?) and that the current error is only a result of this function failing since it's receiving extra args it's not expecting (which the **kwargs are allowing)

PR #6554, opened 2020-12-22
Added integration tests for control.inspect() and task revoking

matusvalo:
## Description

This PR adds 
* integration tests for https://docs.celeryproject.org/en/stable/reference/celery.app.control.html#celery.app.control.Inspect
* Integration test for revoking a task

### DIFF
--- a/t/integration/test_inspect.py
+++ b/t/integration/test_inspect.py
@@ -1,4 +1,5 @@
 import os
+import re
 from datetime import datetime, timedelta
 from unittest.mock import ANY
 from time import sleep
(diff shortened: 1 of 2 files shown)


auvipy (dismissed):

t/integration/test_inspect.py
```diff
@@ -47,10 +48,17 @@ def test_clock(self, inspect):
     @flaky
     def test_registered(self, inspect):
         """Tests listing registered tasks"""
+        # TODO: We can check also the exact values of the registered methods
```
auvipy, line 51: we can also check the exact values of......

matusvalo: Yes we can but it is tricky. To do it properly we need load separate set of tasks in this test and this needs more work in loading worker which I wanted to avoid right now...

matusvalo:
Any other comments or can be merged this PR?

auvipy (approved):

auvipy:
thanks!

Issue #5890, opened 2019-12-24
4.4 silently changes the Inspect API


ShaheedHaque:
# Checklist

- [x] I have checked the [issues list](https://github.com/celery/celery/issues?utf8=%E2%9C%93&q=is%3Aissue+label%3A%22Category%3A+Documentation%22+)
  for similar or identical bug reports.
- [x] I have checked the [pull requests list](https://github.com/celery/celery/pulls?q=is%3Apr+label%3A%22Category%3A+Documentation%22)
  for existing proposed fixes.
- [x] I have checked the [commit log](https://github.com/celery/celery/commits/master)
  to find out if the bug was already fixed in the master branch.
- [x] I have included all related issues and possible duplicate issues in this issue
      (If there are none, check this box anyway).

#### Related Issues

- None

#### Possible Duplicates

- None

# Description

Prior to Celery 4.4.0, the celery.app.control.Inspect.scheduled() API used to return "args" and "kwargs" as strings which could be parsed using ast.literal_eval. Now, they seem to be a list and dict respectively:

 {'id': 'd82058c7-12b0-4914-8381-2fef9ad874a9', 'name': '...', 'args': ['time', 'sleep', 5], 'kwargs': {}, 'type': '...', }

# Suggestions

I assume this is a deliberate design change; it would have been helpful to add this breaking change to the release notes!

auvipy:
I admit the release notes are not complete as I released the 4.4.0 quite hastily before the holiday season starts. thanks for the report

kitsuyui:
I think this issue is related to: https://groups.google.com/g/celery-users/c/1F61iEj-YFI/m/N7d64AalDAAJ

https://github.com/celery/celery/blob/8492b75c579564c2af5c2be75fe4b2118ebd0cd1/celery/worker/request.py#L558-L559

args and kwargs are passed directly. So correctly working when using JSON serialization.
But when using pickle serialization, it causes error.

```
kombu.exceptions.EncodeError: Object of type **** is not JSON serializable
```

## Reproduction

`celery -A repro worker` and `python repro.py` in separated two shells.

```python
import time

from celery import Celery


app = Celery('example')
app.conf.update(
    backend_url='redis://localhost:6379',
    broker_url='redis://localhost:6379',
    result_backend='redis://localhost:6379',
    task_serializer='pickle',
    accept_content=['pickle', 'json'],
)


@app.task(name='task1')
def task1(*args, **kwargs):
    print('Start', args, kwargs)
[...]

PR #6567, opened 2020-12-28
Update request info inspect api for making pickle serialization compatibility

kitsuyui:
## Description

A proposal to fix https://github.com/celery/celery/issues/5890#issuecomment-751657642
Use argsrepr and kwargsrepr to make pickle serialization compatibility.
Inspections don't have to be passed args and kwargs directly.


kitsuyui:
I think this changes breaks backward compatibility. But it should not be a big problem.
Because `request.info()` is mainly used for just inspection and monitoring, I think.

auvipy (changes requested):
could you please add unit test for this? as you were able to reproduce it

auvipy:
And the breaking change should be mentioned in the doc as well

kitsuyui:
Okay, I will do it.
> could you please add unit test for this? as you were able to reproduce it

auvipy (approved):

auvipy:
@ShaheedHaque would you mind checking this pr?

ShaheedHaque:
@auvipy Just to confirm: I assume that trying to `inspect active` while a job is running in my setup is what you had in mind? Or is there some other config change I need?

matusvalo:
Hi All. I will review the PR soon.

matusvalo:
I have merged master and fixed integration tests.

matusvalo (approved):
LGTM. Documentation also needs updating? If yes please update also docs.

thedrow:
> I think this changing breaks backward compatibility. But it should not be a big problem.
> Because `request.info()` is mainly used for just inspection and monitoring, I think.

This can break monitoring systems, possibly including flower.

thedrow (changes requested):
We need to find a way to do this properly so it will work with json and pickle.

matusvalo:
@thedrow problem is mostly following:
1. call/response of tasks serializer is configurable. Json is preferred bug pickle can be used.
2. inspect/control calls/responses are serialized as json and I am not aware of changing this behavior.

So the problem occurs when you are using pickle serializer for tasks and you pass objects which are not serializable via json. This will cause that sending tasks will be working but inspect/controll calls will be broken.

Possible solution is to use the same serializer for inspect/control as for tasks is used.

thedrow:
Thank you for the contribution. We went ahead with another solution in #6757.
You are welcome to contribute again at any time :)

PR #6572, opened 2020-12-31
feat: thread task pool support revoke

yywing:

### DIFF
--- a/celery/utils/thread_pool.py
+++ b/celery/utils/thread_pool.py
@@ -1,0 +1,93 @@
+# copy from https://github.com/intuited/terminable_thread
+"""Code based on [http://sebulba.wikispaces.com/recipe+thread2].
+`terminable_thread` provides a subclass of `threading.Thread`,
+  adding the facility to raise exceptions in the context of the given thread.
+See the `README` for information on problematic issues, licensing,
+  and other shenanigans.
+    >>> import time
+    >>>
+    >>> def f():
+    ...     try:
+    ...         while True:
+    ...             time.sleep(0.1)
+    ...     finally:
+    ...         print "outta here"
+    ...
+    >>> t = Thread(target = f)
+    >>> t.start()
+    >>> t.isAlive()
+    True
+    >>> t.terminate()
+    >>> t.join()
+    outta here
+    >>> t.isAlive()
+    False
+"""
+# This program is free software.
+# It comes without any warranty, to the extent permitted by applicable law.
+# You can redistribute it and/or modify it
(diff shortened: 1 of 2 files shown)


celery/utils/thread_pool.py
```diff
@@ -0,0 +1,93 @@
...
+from concurrent.futures import ThreadPoolExecutor as _T
+
+from celery.exceptions import WorkerTerminate
+
+
+def _async_raise(tid, exctype):
+    """raises the exception, performs cleanup if needed"""
+    if not inspect.isclass(exctype):
+        raise TypeError("Only types can be raised (not instances)")
+    res = ctypes.pythonapi.PyThreadState_SetAsyncExc(ctypes.c_long(tid),
```
yywing, line 47: TODO: Changed in version 3.7: The type of the id parameter changed from long to unsigned long.

thedrow: Please add a condition here and a TODO comment

celery/utils/thread_pool.py
```diff
@@ -0,0 +1,93 @@
...
+        self.raise_exc(WorkerTerminate)
+
+
+class ThreadPoolExecutor(_T):
+    """
+    1. Don't catch WorkerTerminate exception, unless you know what you do
+    2. The exception will be raised only when execution returns to the python code.
+    """
+    def _adjust_thread_count(self):
+        with mock.patch("threading.Thread", new=Thread):
```
thedrow, line 92: I don't really like this in production code.

thedrow (changes requested):

matusvalo:
I am skeptical about this PR. This PR kills a thread via triggering exception in other thread via Dark Magic. This can work when python code is executed during revoking. But in case when the code is blocked in I/O or is executing some C code the **revoking simply won't work**. E.g. revoking a task that is blocked by executing SQL query (e.g. due Database lock) will fail.

I am not aware about any universal way how to kill/terminate a thread. Hence, supporting revoking thread task pool can cause that for some users confusion when it simply fails but only sometimes.

auvipy:
> I am skeptical about this PR. This PR kills a thread via triggering exception in other thread via Dark Magic. This can work when python code is executed during revoking. But in case when the code is blocked in I/O or is executing some C code the **revoking simply won't work**. E.g. revoking a task that is blocked by executing SQL query (e.g. due Database lock) will fail.
> 
> I am not aware about any universal way how to kill/terminate a thread. Hence, supporting revoking thread task pool can cause that for some users confusion when it simply fails but only sometimes.

I think we should close this PR.

matusvalo:
Closed due limitations of killing a thread.

PR #6580, opened 2021-01-06
Upgrade AzureBlockBlob storage backend to use Azure blob storage library v12

JanusAsmussen:

## Description
This PR upgrades the AzureBlockBlob storage backend to use Azure blob storage library v12. This replaces the currently used SDK (Azure blob storage library v2.1) which is considered a legacy version by Microsoft.

### DIFF
--- a/celery/backends/azureblockblob.py
+++ b/celery/backends/azureblockblob.py
@@ -8,13 +8,11 @@ from celery.utils.log import get_logger
 from .base import KeyValueStoreBackend
 
 try:
-    from azure import storage as azurestorage
-    from azure.common import AzureMissingResourceHttpError
-    from azure.storage.blob import BlockBlobService
-    from azure.storage.common.retry import ExponentialRetry
-except ImportError:  # pragma: no cover
-    azurestorage = BlockBlobService = ExponentialRetry = \
-        AzureMissingResourceHttpError = None  # noqa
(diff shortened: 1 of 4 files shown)


auvipy:
could you please check if there are any other things that need to be updated? integration test/tox etc?

JanusAsmussen:
Hi,

Both unit & integration tests related to the AzureBlockBlob storage backend are passing when using the environment set up with Docker and docker-compose. However, I had to change the azurite docker image used when spinning up the environment (It is now using the official Microsoft image). I've committed this change to the remote branch.

Regarding tox: I do not believe anything should be changed.

auvipy (changes requested):
can you check the following alerts?
This pull request introduces 17 alerts and fixes 3 when merging 0493f66 into 2c6f46d - view on LGTM.com

new alerts:

    12 for Wrong number of arguments in a call
    3 for Module is imported with 'import' and 'import from'
    1 for __eq__ not overridden when adding attributes
    1 for Use of the return value of a procedure

JanusAsmussen:
Commit 0493f66 changes a single string in a yaml file. I would believe these warnings are caused by 2c6f46d?

matusvalo (approved):
LGTM for me.

matusvalo:
@auvipy do you have still objections?

auvipy (approved):

PR #6602, opened 2021-01-18
exit celery with non zero exit value if failing

tned73:

## Description

There is an issue with version 5.x for the celery command. If the command is failing, the exit value is still 0. This PR will fix this and exits with the intended exit value.

that version series.

be sure to link to that issue via (Fixes https://github.com/celery/celery/issues/6601) for example.
-->

### DIFF
--- a/celery/bin/control.py
+++ b/celery/bin/control.py
@@ -6,6 +6,7 @@ from kombu.utils.json import dumps
 
 from celery.bin.base import (COMMA_SEPARATED_LIST, CeleryCommand,
                              CeleryOption, handle_preload_options)
+from celery.exceptions import CeleryCommandException
 from celery.platforms import EX_UNAVAILABLE
 from celery.utils import text
 from celery.worker.control import Panel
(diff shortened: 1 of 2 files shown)


thedrow (changes requested):
We don't need a specialized exception.
We can simply use [`ctx.fail()`](https://click.palletsprojects.com/en/7.x/api/#click.Context.fail).

tned73:
If I look at `ctx.fail()` it has a message parameter and not a exit value parameter. Using this will solve a part of the issue. It will probably exit with value 1 and not the 69 if the `celery inspect ping` fails as version 4.x does.
for my purposes it is not an issue.

please advice on direction to solve this issue

thedrow:
Huh.
That API is very surprising.

I think we're good to go on this one.

thedrow (approved):
LGTM

thedrow (approved):
LGTM

auvipy (approved):
Looks good. should we merge this as is or need unit test?

thedrow:
There are no unit tests for the CLI at the time being, unfortunately.
I don't have time to write them, currently.

PR #6611, opened 2021-01-22
New errors

auvipy:


thedrow:
Note that the build fails as you probably did something incorrectly.
Instead of a large PR, could we split these exception type replacements to different PRs so we'll know exactly what's the effect of each replacement?

auvipy:
> Note that the build fails as you probably did something incorrectly.
> Instead of a large PR, could we split these exception type replacements to different PRs so we'll know exactly what's the effect of each replacement?

ok

thedrow:
I'm closing this PR for now. Let's do this for 5.1?

PR #6576, opened 2021-01-04
start chord header tasks as soon as possible

mjhoffman65:
Fixes #3021 . Previously, header tasks were not submitted to the broker for
execution until the entire generator was consumed.

Continuation of https://github.com/celery/celery/pull/6112 and much of this was originally pulled from https://github.com/celery/celery/pull/3043.

Update: this was split into 2 PRs per below review: https://github.com/celery/celery/pull/6589 should be merged first.  This has been rebased on top of that branch.

PR #6613, opened 2021-01-22
fixes Github Action unit tests

mjhoffman65:
Sets `PYTHONPATH` so tests will run against the code in this repo instead of against the celery installed by pip.

See https://github.com/celery/celery/pull/6576#pullrequestreview-574472666


auvipy (approved):
thanks

PR #6619, opened 2021-01-27
Safeguard against schedule entry without kwargs

auvipy:
Fixes ##5889

### DIFF
--- a/celery/beat.py
+++ b/celery/beat.py
@@ -381,7 +381,7 @@ class Scheduler:
 
         try:
             entry_args = [v() if isinstance(v, BeatLazyFunc) else v for v in (entry.args or [])]
-            entry_kwargs = {k: v() if isinstance(v, BeatLazyFunc) else v for k, v in (entry.kwargs.items() or {})}
+            entry_kwargs = {k: v() if isinstance(v, BeatLazyFunc) else v for k, v in (entry.kwargs or {}).items()}
             if task:
                 return task.apply_async(entry_args, entry_kwargs,
                                         producer=producer,
--- a/celery/beat.py
+++ b/celery/beat.py
@@ -381,7 +381,7 @@ class Scheduler:
 
         try:
             entry_args = [v() if isinstance(v, BeatLazyFunc) else v for v in (entry.args or [])]
-            entry_kwargs = {k: v() if isinstance(v, BeatLazyFunc) else v for k, v in entry.kwargs.items()}
+            entry_kwargs = {k: v() if isinstance(v, BeatLazyFunc) else v for k, v in (entry.kwargs.items() or {})}
             if task:
                 return task.apply_async(entry_args, entry_kwargs,
                                         producer=producer,
--- a/celery/beat.py
+++ b/celery/beat.py
@@ -203,6 +203,24 @@ class ScheduleEntry:
         return not self == other
 
 
+def _evaluate_entry_args(entry_args):
+    if not entry_args:
+        return []
+    return [
+        v() if isinstance(v, BeatLazyFunc) else v
+        for v in entry_args.args
+    ]
+
+
+def _evaluate_entry_kwargs(entry_kwargs):
+    if not entry_kwargs:
+        return {}
+    return {
+        k: v() if isinstance(v, BeatLazyFunc) else v
+        for k, v in entry_kwargs.items()
+    }
+
+
 class Scheduler:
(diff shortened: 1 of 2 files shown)


celery/beat.py
```diff
@@ -381,7 +381,7 @@ def apply_async(self, entry, producer=None, advance=True, **kwargs):
 
         try:
             entry_args = [v() if isinstance(v, BeatLazyFunc) else v for v in (entry.args or [])]
-            entry_kwargs = {k: v() if isinstance(v, BeatLazyFunc) else v for k, v in entry.kwargs.items()}
+            entry_kwargs = {k: v() if isinstance(v, BeatLazyFunc) else v for k, v in (entry.kwargs.items() or {})}
```
thedrow, line 384: How can this be right?
`kwargs` is `None` so it has no `items()`.
You should check if `kwargs` is `None` and if it is call `items()` on an empty dictionary.

```suggestion
            entry_kwargs = {k: v() if isinstance(v, BeatLazyFunc) else v for k, v in (entry.kwargs or {}).items()}
```

thedrow (changes requested):
This also needs tests.

celery/beat.py
auvipy: I wasn't fully sure so wanted more eyes for some suggestion O:) thanks

celery/beat.py
```diff
@@ -381,7 +381,7 @@ def apply_async(self, entry, producer=None, advance=True, **kwargs):
 
         try:
             entry_args = [v() if isinstance(v, BeatLazyFunc) else v for v in (entry.args or [])]
-            entry_kwargs = {k: v() if isinstance(v, BeatLazyFunc) else v for k, v in entry.kwargs.items()}
+            entry_kwargs = {k: v() if isinstance(v, BeatLazyFunc) else v for k, v in (entry.kwargs or {}).items()}
```
xirdneh, line 384: **[suggestion]**: Refactor this function for better readability. Avoid doing an extra boolean operation inside a comprehension. Avoid calling `items()` on an empty dict.

```python
def apply_async(self, entry, producer=None, advance=True, **kwargs):
    # Update time-stamps and run counts before we actually execute,
    # so we have that done if an exception is raised (doesn't schedule
    # forever.)
    entry = self.reserve(entry) if advance else entry
    task = self.app.tasks.get(entry.task)

    try:
        entry_args = self._evaluate_entry_args(entry.args)
        entry_kwargs = self._evaluate_entry_kwargs(entry.kwargs)

        self.send_task(entry, entry_args, entry_kwargs, task, producer)
        reraise(SchedulingError, SchedulingError(
            "Couldn't apply scheduled task {0.name}: {exc}".format(
                entry, exc=exc)), sys.exc_info()[2])
    finally:
        self._tasks_since_sync += 1
        if self.should_sync():
            self._do_sync()

def _evaluate_entry_args(entry_args):
    if not entry_args:
        return []
    return [v() if isinstance(v, BeatLazyFunc) else v for v in entry_args]

def _evaluate_entry_kwargs(entry_kwargs):
    if not entry_kwargs:
        return {}
    return {k: v() if isinstance(v, BeatLazyFunc) else v for k, v in entry_kwargs}

# Repurpose existing function
def send_task(self, entry, entry_args, entry_kwargs, task, producer):
        if task:
            return task.apply_async(
                entry_args, entry_kwargs,
                producer=producer,
                **entry.options
            )
        else:
            return self.app.send_task(
                entry.task,
                entry_args,
                entry_kwargs,
                producer=producer,
                **entry.options
            )

xirdneh (changes requested):
This also needs a test.

celery/beat.py
thedrow: I agree.
@xirdneh Care to fix this?
@auvipy What about the test?

thedrow (approved):

thedrow:
Codecov is wrong and is reporting on unrelated files.

xirdneh:
If this only happens on `advance=False` then it looks like the problem is somewhere before this code.
This might not fix the root cause but it certainly fixes this specific problem.
The issue is that this  problem where `entry.kwargs == None` might show up somewhere else.

Should we create an issue to make sure that `entry` has correct values?

thedrow:
> Should we create an issue to make sure that `entry` has correct values?

Yes we should.

thedrow (approved):

PR #6630, opened 2021-02-11
Handle Terminated exception (fixes #2727)

### LINKED ISSUE #2727: raise Terminated(15, ) when terminating task
[2015-07-27 12:52:38,386: ERROR/MainProcess] Task myapp.tasks.test [150a75dc-592e-4c0c-b686-1cc5aaf5a168] raised unexpected: Terminated(15,)
  File "/home/blake/projects/venv/myapp/lib/python2.7/site-packages/billiard/pool.py", line 1674, in _set_terminated
    raise Terminated(-(signum or 0))
Terminated: 15

packages installed:

amqp==1.4.6
anyjson==0.3.3
beautifulsoup4==4.4.0
billiard==3.3.0.20
celery==3.1.18
Django==1.7.7
django-celery==3.1.16
django-pipeline==1.5.2
futures==3.0.3
kombu==3.0.25
pytz==2015.4
redis==2.10.3
requests==2.7.0
wheel==0.24.0

I'm starting worker with:

celery -A myapp worker --loglevel=INFO --concurrency=10 

i'm revoking with:

from celery.task.control import revoke
revoke(task_id, terminate=True, signal='SIGKILL')

is this the wrong approach?

pakallis:
## Description

I am opening this pull request to propose a fix for #2727.

The worker won't anymore raise a Terminated exception when `revoke(task_id, terminate=True)` is called, but instead will ack
the message and terminate the task successfully.

Is the desired behavior to not throw the `Terminated` exception and acknowledge the message?

If yes, I will fix the broken tests. Also, feel free to suggest if there are other tests that need to be written.

### DIFF
--- a/celery/worker/request.py
+++ b/celery/worker/request.py
@@ -489,7 +489,7 @@ class Request:
             raise MemoryError(f'Process got: {exc_info.exception}')
         elif isinstance(exc_info.exception, Reject):
             return self.reject(requeue=exc_info.exception.requeue)
-        elif isinstance(exc_info.exception, Ignore):
+        elif isinstance(exc_info.exception, (Ignore, Terminated)):
             return self.acknowledge()
 
         exc = exc_info.exception
@@ -518,11 +518,7 @@ class Request:
 
         # These are special cases where the process would not have had time
         # to write the result.
-        if isinstance(exc, Terminated):
-            self._announce_revoked(
-                'terminated', True, str(exc), False)
-            send_failed_event = False  # already sent revoked event
-        elif not requeue and (isinstance(exc, WorkerLostError) or not return_ok):
+        if not requeue and (isinstance(exc, WorkerLostError) or not return_ok):
             # only mark as failure if task has not been requeued
             self.task.backend.mark_as_failure(
                 self.id, exc, request=self._context,
--- a/t/unit/worker/test_request.py
+++ b/t/unit/worker/test_request.py
@@ -254,7 +254,7 @@ class test_Request(RequestCase):
         req.on_retry(Mock())
         req.on_ack.assert_called_with(req_logger, req.connection_errors)
 
-    def test_on_failure_Terminated(self):
+    def test_on_failure_Terminated_acknowledges(self):
         einfo = None
         try:
             raise Terminated('9')
@@ -263,10 +263,7 @@ class test_Request(RequestCase):
         assert einfo is not None
         req = self.get_request(self.add.s(2, 2))
         req.on_failure(einfo)
-        req.eventer.send.assert_called_with(
-            'task-revoked',
-            uuid=req.id, terminated=True, signum='9', expired=False,
-        )
+        req.on_ack.assert_called_with(req_logger, req.connection_errors)
 
     def test_on_failure_propagates_MemoryError(self):
         einfo = None


celery/worker/request.py
```diff
@@ -517,11 +517,7 @@ def on_failure(self, exc_info, send_failed_event=True, return_ok=False):
 
         # These are special cases where the process would not have had time
```
auvipy, line 518: should we change the comments as well?

pakallis: I think it wouldn't hurt to leave the comments intact as they have to do with the `WorkerLostError` and `return_ok == False` special cases

pakallis:
Is there any issue with the build system?

I am getting failures with the appveyor. They are probably not related with this PR.

https://ci.appveyor.com/project/ask/celery/builds/37726692/job/xs0y53qqo9ivrdc2

auvipy:
appveyor is not stable. github actions are working with unit tests

auvipy:
can you try integration test somethinglike this locally from tests.tasks import simple_task
"""
@task(bind=True)
def simple_task(self):
    print("Going to sleep")
    time.sleep(5)
"""

_task = simple_task.si()
result = _task.apply_async()

result.revoke(terminate=True)

celery/worker/request.py
```diff
@@ -518,11 +518,7 @@ def on_failure(self, exc_info, send_failed_event=True, return_ok=False):
...
         # to write the result.
-        if isinstance(exc, Terminated):
-            self._announce_revoked(
```
thedrow, line 522: Why did you remove this line?

celery/worker/request.py
```diff
@@ -489,7 +489,7 @@ def on_failure(self, exc_info, send_failed_event=True, return_ok=False):
             raise MemoryError(f'Process got: {exc_info.exception}')
         elif isinstance(exc_info.exception, Reject):
             return self.reject(requeue=exc_info.exception.requeue)
-        elif isinstance(exc_info.exception, Ignore):
+        elif isinstance(exc_info.exception, (Ignore, Terminated)):
             return self.acknowledge()
```
thedrow, line 493: This is probably not the right behavior.
We should reject revoked messages so monitoring systems would catch that these tasks were revoked.

I thought that this is what this feature does.

thedrow (changes requested):
Interesting and possibly in the right direction but there's still more to do to get there.

pakallis:
@thedrow Thanks for taking the time to review this. 

I spent more time reviewing this code and the current proposal won't solve the issue.

The `Terminated` exception is already handled [here](https://github.com/celery/celery/blob/master/celery/worker/request.py#L521), but we log the `Task handler raised error` error a few lines [below](https://github.com/celery/celery/blob/master/celery/worker/request.py#L539).

Turns out that this might be only a logging issue (we log an error instead of info) which caused some confusion in #2727.
I think we have to pause this PR until we decide what the proper fix is.

auvipy:
> @thedrow Thanks for taking the time to review this.
> 
> I spent more time reviewing this code and the current proposal won't solve the issue.
> 
> The `Terminated` exception is already handled [here](https://github.com/celery/celery/blob/master/celery/worker/request.py#L521), but we log the `Task handler raised error` error a few lines [below](https://github.com/celery/celery/blob/master/celery/worker/request.py#L539).
> 
> Turns out that this might be only a logging issue (we log an error instead of info) which caused some confusion in #2727, as the `Terminated` exception was already handled previously, and this PR might not be needed.
> 
> I am going to close this unless you have a clear idea of how to proceed with the issue.

at least some documentation would be great describing the issue

pakallis:
> > @thedrow Thanks for taking the time to review this.
> > I spent more time reviewing this code and the current proposal won't solve the issue.
> > The `Terminated` exception is already handled [here](https://github.com/celery/celery/blob/master/celery/worker/request.py#L521), but we log the `Task handler raised error` error a few lines [below](https://github.com/celery/celery/blob/master/celery/worker/request.py#L539).
> > Turns out that this might be only a logging issue (we log an error instead of info) which caused some confusion in #2727, as the `Terminated` exception was already handled previously, and this PR might not be needed.
> > I am going to close this unless you have a clear idea of how to proceed with the issue.
> 
> at least some documentation would be great describing the issue

Sure, I have included the documentation in the original issue.
Another approach here is to not log an error when the process is terminated and instead display an info message as this is not exactly an exceptional case (the process has been terminated by someone so it is not something unexpected).
If you are ok with that, I could adjust the log level message to `info` in this pull request

auvipy:
> > at least some documentation would be great describing the issue
> 
> Sure, I have included the documentation in the original issue.
> Another approach here is to not log an error when the process is terminated and instead display an info message as this is not exactly an exceptional case (the process has been terminated by someone so it is not something unexpected).
> If you are ok with that, I could adjust the log level message to `info` in this pull request

@matusvalo I saw you assigned the issue, what do you think about this approach?

thedrow:
I still think that we should reject the task.
If you revoke a task, where is it acked?

pakallis:
> I still think that we should reject the task.
> If you revoke a task, where is it acked?

Assuming `acks_late` is True, message acknowledgement happens in this line:

https://github.com/celery/celery/blob/master/celery/worker/request.py#L512

The `ack` depends on whether `acks_on_failure_or_timeout` is True or False (by default is True), so the default behavior is to `ack` the message, even if it is revoked.

Quoting the documentation:
```
acks_on_failure_or_timeout = True
When enabled messages for this task will be acknowledged even if it fails or times out.
```

bramverhelst:
hi, as I'm invested in this fix, I'll ask a question. @pakallis, the initial problem (for which I posted a small example in #2727) was not so much the log message being displayed or not, it was that the entire celery process was canceled on the worker machine in case the task got revoked (if already being handled by a worker machine)

This essentially turns the worker machine into a 'zombie' state where it exists without being able to receive tasks from the task pool. It requires manual intervention (e.g. restarting celery service, possibly fixing some issues in the broker) to get the entire system setup functional again.

So my question is: will this PR fix that issue or is it mainly geared towards not showing the log line?

thedrow:
> 
> 
> > I still think that we should reject the task.
> > If you revoke a task, where is it acked?
> 
> @thedrow Assuming `acks_late` is True, message acknowledgement happens in this line:
> 
> https://github.com/celery/celery/blob/master/celery/worker/request.py#L512
> 
> The `ack` depends on whether `acks_on_failure_or_timeout` is True or False (by default is True), so the default behavior is to `ack` the message, even if `revoke` is called, with `terminate` `False` or `True`.
> 
> Quoting the documentation:
> 
> ```
> acks_on_failure_or_timeout = True
> When enabled messages for this task will be acknowledged even if it fails or times out.
> ```

Ok, but you get where I'm coming from right?

matusvalo:
I had a look in the issue #2727 and this PR and these are my thoughts: There are two aspects we need to handle:

1. worker is terminated from outside - e.g. OS/User is sending SIGTERM signal to worker. In this case,  the current behavior is correct and current PR is introducing bug (for `ack_late=True`) since the user/OS terminating worker is causing acknowledging the task which is not correct semantics - the task should be reexecuted after worker restart.
2. User is sending revoke via celery with `terminate=True`. This will cause that master process sends signal to the worker to signal to terminate.  Based on documentation, the master process should send the signal to worker signal and acknowledge the task (to ignore it). I checked the master branch and the code is following this semantics - the process is terminated and task is acknowledged.

Hence the only problem is the misleading error message for logs. But this is very difficult to understand from worker standpoint whether it is killed by external entity or by master proces of celery...

thedrow:
@matusvalo based on your comments, I think we should close this PR.

pakallis:
I am going to close this as we agree there is not something to be done here.
Thank you very much for your help.

PR #6632, opened 2021-02-11
Allow heartbeats to be sent in tests

danihodovic:
I'm writing a Prometheus exporter at
https://github.com/danihodovic/celery-exporter.

In order to test the worker events: worker-heartbeat, worker-online,
worker-offline I need to be able to enable heartbeats for the testing
worker.

The change allows for overriding the default value of heartbeat = False.

### DIFF
--- a/celery/contrib/testing/worker.py
+++ b/celery/contrib/testing/worker.py
@@ -119,7 +119,7 @@ def _start_worker_thread(app,
         logfile=logfile,
         # not allowed to override TestWorkController.on_consumer_ready
         ready_callback=None,
-        without_heartbeat=True,
+        without_heartbeat=kwargs.pop("without_heartbeat", True),
         without_mingle=True,
         without_gossip=True,
         **kwargs)
--- a/docs/userguide/testing.rst
+++ b/docs/userguide/testing.rst
@@ -194,6 +194,20 @@ Example:
     def test_other(celery_worker):
         ...
 
+Heartbeats are disabled by default which means that the test worker doesn't
+send events for ``worker-online``, ``worker-offline`` and ``worker-heartbeat``.
+To enable heartbeats modify the :func:`celery_worker_parameters` fixture:
+
+.. code-block:: python
+
+    # Put this in your conftest.py
+    @pytest.fixture(scope="session")
+    def celery_worker_parameters():
+        return {"without_heartbeat": False}
+        ...
+
+
+
 Session scope
 ^^^^^^^^^^^^^
 


celery/contrib/testing/worker.py
```diff
@@ -119,7 +119,7 @@ def _start_worker_thread(app,
         logfile=logfile,
         # not allowed to override TestWorkController.on_consumer_ready
         ready_callback=None,
-        without_heartbeat=True,
+        without_heartbeat=kwargs.pop("without_heartbeat", True),
```
auvipy, line 122: this isn't going to break the existing API write? shouldn't this overridable feature mentioned in the docs? and can we provide some sort of unit tests?

danihodovic: The default value is False, so this should be backwards incompatible.

auvipy: can you add documentation for this?

thedrow (changes requested):
This solution means you'll need to overwrite the `worker` fixture to override the value of `without_heartbeat`.
This isn't the most comfortable API.

We should change https://github.com/celery/celery/blob/master/celery/contrib/pytest.py#L145 instead to return `{without_heartbeat: True}` and ensure it is passed onwards.

danihodovic:
> This solution means you'll need to overwrite the celery_worker/celery_session_worker fixtures to override the value of without_heartbeat. This isn't the most comfortable API.

No, you can override the fixture `celery_worker_parameters` to enable the heartbeat to be sent.
https://github.com/danihodovic/celery-exporter/blob/master/conftest.py#L16-L18

celery/contrib/testing/worker.py
danihodovic: Yes, probably next week.

thedrow:
> > We should change https://github.com/celery/celery/blob/master/celery/contrib/pytest.py#L145 instead to return {"without_heartbeat": True} and ensure it is passed onwards.
> 
> The question is if this breaks existing behavior for users that don't use the pytest fixture? By default the value is False - https://github.com/celery/celery/blob/7288147d65a32b726869ed887d99e4bfd8c070e2/celery/worker/consumer/heart.py#L22.
> 
> WorkController behavior would silently change and heartsbeats would be enabled.

In that case we still need to check if `celery_worker_parameters` contains the `without_heartbeat` key and if it does not, we need to add it ourselves.

> > This solution means you'll need to overwrite the celery_worker/celery_session_worker fixtures to override the value of without_heartbeat.
> 
> No, you can override `celery_worker_parameters` as seen here:
> https://github.com/danihodovic/celery-exporter/blob/e5430516de22b9f095da4d8d099169ec518b6670/conftest.py#L15-L18

I meant the solution proposed in this PR, not the solution I'm proposing.

danihodovic:
> In that case we still need to check if celery_worker_parameters contains the without_heartbeat key and if it does not, we need to add it ourselves.

Hmm... I'm confused because that's what this PR does https://github.com/celery/celery/pull/6632/files#diff-fee53e249a978fd480b1cade781b4e997b18492ab090d76d0377786d99128b92R122

`kwargs` in `_start_worker_thread` is passed in from `celery_worker_parameters`

- https://github.com/celery/celery/blob/b6376da1dafaa8fe59cab333c2f4b862975a227a/celery/contrib/testing/worker.py#L99
- https://github.com/celery/celery/blob/b6376da1dafaa8fe59cab333c2f4b862975a227a/celery/contrib/pytest.py#L196

thedrow:
Gottcha 🤦 
The lack of documentation (that's our fault) confused me.

danihodovic:
> Gottcha facepalm
> The lack of documentation (that's our fault) confused me.

No worries, happy we cleared it up. I'll be adding some docs later this week.

auvipy (approved):

auvipy (approved):

auvipy:
thanks!

PR #6633, opened 2021-02-12
Small refactor in exception handling of `on_failure`

pakallis:
## Description

- Use the previously declared variable `exc` to avoid duplication in `if` statements

### DIFF
--- a/celery/worker/request.py
+++ b/celery/worker/request.py
@@ -485,16 +485,15 @@ class Request:
     def on_failure(self, exc_info, send_failed_event=True, return_ok=False):
         """Handler called if the task raised an exception."""
         task_ready(self)
-        if isinstance(exc_info.exception, MemoryError):
-            raise MemoryError(f'Process got: {exc_info.exception}')
-        elif isinstance(exc_info.exception, Reject):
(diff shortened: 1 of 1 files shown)


celery/worker/request.py
```diff
@@ -485,16 +485,15 @@ def on_retry(self, exc_info):
...
-            raise MemoryError(f'Process got: {exc_info.exception}')
-        elif isinstance(exc_info.exception, Reject):
-            return self.reject(requeue=exc_info.exception.requeue)
-        elif isinstance(exc_info.exception, Ignore):
-            return self.acknowledge()
-
         exc = exc_info.exception
 
-        if isinstance(exc, Retry):
+        if isinstance(exc, MemoryError):
```
auvipy, line 490: can you explain in comment what the refactor is doing? also can you adjust the unit test?

pakallis: There is no change in logic here and everything should work as previously.
The only change I did was to move the `if isinstance(exc, Retry)` on the upper part with the other if statements so that it is more readable and use `exc` which is assigned to `exc_info.exception` for the checks.

The tests should pass, as there is not change in logic, except you have something else in mind.

thedrow (approved):

Issue #6489, opened 2020-11-18
pypy3 unit tests sometimes fail due to an attribute we expect to be an `Event` being a `bool` instead


maybe-sybr:
We've seen this CI failure sporadically since we fixed the pypy3 CI job. It possible an attribute in a timer is being mocked/assigned incorrectly, or there's just some difference in how pypy3 implements `Timer` objects (I believe the complaint is about a `threading.Timer()` object).

# Checklist

- [x] I have verified that the issue exists against the `master` branch of Celery.
- [ ] This has already been asked to the [discussion group](https://groups.google.com/forum/#!forum/celery-users) first.
- [ ] I have read the relevant section in the
  [contribution guide](http://docs.celeryproject.org/en/latest/contributing.html#other-bugs)
  on reporting bugs.
- [ ] I have checked the [issues list](https://github.com/celery/celery/issues?q=is%3Aissue+label%3A%22Issue+Type%3A+Bug+Report%22+-label%3A%22Category%3A+Documentation%22)
  for similar or identical bug reports.
- [ ] I have checked the [pull requests list](https://github.com/celery/celery/pulls?q=is%3Apr+label%3A%22PR+Type%3A+Bugfix%22+-label%3A%22Category%3A+Documentation%22)
  for existing proposed fixes.
- [ ] I have checked the [commit log](https://github.com/celery/celery/commits/master)
  to find out if the bug was already fixed in the master branch.
- [ ] I have included all related issues and possible duplicate issues
  in this issue (If there are none, check this box anyway).

## Mandatory Debugging Information

- [ ] I have included the output of ``celery -A proj report`` in the issue.
    (if you are not able to do this, then at least specify the Celery
     version affected).
- [ ] I have verified that the issue exists against the `master` branch of Celery.
- [ ] I have included the contents of ``pip freeze`` in the issue.
- [ ] I have included all the versions of all the external dependencies required
  to reproduce this bug.

## Optional Debugging Information

- [ ] I have tried reproducing the issue on more than one Python version
  and/or implementation.
- [ ] I have tried reproducing the issue on more than one message broker and/or
  result backend.
- [ ] I have tried reproducing the issue on more than one version of the message
  broker and/or result backend.
- [ ] I have tried reproducing the issue on more than one operating system.
- [ ] I have tried reproducing the issue on more than one workers pool.
- [ ] I have tried reproducing the issue with autoscaling, retries,
  ETA/Countdown & rate limits disabled.
[...]

PR #6635, opened 2021-02-18
Fixes build for PyPy3

mjhoffman65:
@auvipy, I took at stab at fixing the PyPy failures.  My first step is to try to install all the requirements and recreate the unit test failure in the docker environment.  I ran into issues installing the python requirements and am wondering if more work needs to be done before we can expect celery to run with PyPy.  Looks like at least a couple of the dependencies do not work with PyPy since they rely on a C backend.

couchbase: https://docs.couchbase.com/python-sdk/current/hello-world/start-using-sdk.html#pypy-support
ephem: https://github.com/celery/django-celery-beat/issues/69#issue-247235755 (Note we may be able to replace this with skyfield.  See [here](https://rhodesmill.org/pyephem/) for more.)

This PR illustrates the same error I'm getting in the docker environment: https://github.com/celery/celery/runs/1930241932?check_suite_focus=true

How should I move forward?

### DIFF
--- a/celery/utils/timer2.py
+++ b/celery/utils/timer2.py
@@ -49,8 +49,12 @@ class Timer(threading.Thread):
         self.on_start = on_start
         self.on_tick = on_tick or self.on_tick
         threading.Thread.__init__(self)
-        self._is_shutdown = threading.Event()
-        self._is_stopped = threading.Event()
+        # `_is_stopped` is likely to be an attribute on `Thread` objects so we
+        # double underscore these names to avoid shadowing anything and
+        # potentially getting confused by the superclass turning these into
+        # something other than an `Event` instance (e.g. a `bool`)
+        self.__is_shutdown = threading.Event()
+        self.__is_stopped = threading.Event()
         self.mutex = threading.Lock()
         self.not_empty = threading.Condition(self.mutex)
         self.daemon = True
@@ -71,7 +75,7 @@ class Timer(threading.Thread):
             self.running = True
             self.scheduler = iter(self.schedule)
 
-            while not self._is_shutdown.isSet():
+            while not self.__is_shutdown.isSet():
                 delay = self._next_entry()
                 if delay:
                     if self.on_tick:
@@ -80,7 +84,7 @@ class Timer(threading.Thread):
                         break
                     sleep(delay)
             try:
-                self._is_stopped.set()
+                self.__is_stopped.set()
             except TypeError:  # pragma: no cover
                 # we lost the race at interpreter shutdown,
                 # so gc collected built-in modules.
@@ -91,9 +95,9 @@ class Timer(threading.Thread):
             os._exit(1)
 
     def stop(self):
-        self._is_shutdown.set()
+        self.__is_shutdown.set()
         if self.running:
-            self._is_stopped.wait()
+            self.__is_stopped.wait()
             self.join(THREAD_TIMEOUT_MAX)
             self.running = False
 
--- a/t/unit/backends/test_mongodb.py
+++ b/t/unit/backends/test_mongodb.py
@@ -45,7 +45,6 @@ class test_MongoBackend:
         self.patching('celery.backends.mongodb.MongoBackend.encode')
         self.patching('celery.backends.mongodb.MongoBackend.decode')
         self.patching('celery.backends.mongodb.Binary')
-        self.patching('datetime.datetime')
         self.backend = MongoBackend(app=self.app, url=self.default_url)
 
     def test_init_no_mongodb(self, patching):
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -48,18 +48,16 @@ jobs:
           ${{ matrix.python-version }}-v1-${{ hashFiles('**/setup.py') }}
         restore-keys: |
           ${{ matrix.python-version }}-v1-
-    - name: Install dependencies
-      run: |
-        python -m pip install --upgrade pip
-        python -m pip install -r requirements/dev.txt \
-            -r requirements/test.txt \
-            -r requirements/test-ci-default.txt \
-            -r requirements/docs.txt \
-            -r requirements/test-integration.txt \
-            -r requirements/pkgutils.txt
-    - name: Run Unit test with pytest
-      run: |
-        PYTHONPATH=. pytest -xv --cov=celery --cov-report=xml --cov-report term t/unit
+
+    - name: Install tox
+      run: python -m pip install tox
+    - name: >
+        Run tox for
+        "${{ matrix.python-version }}-unit"
+      timeout-minutes: 15
+      run: >
+        tox --verbose --verbose -e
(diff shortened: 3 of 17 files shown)


auvipy:
I think we should stop couchbase testing and eventually remove it from the build. I am not sure how much popular it is.

thedrow:
> 
> 
> I think we should stop couchbase testing and eventually remove it from the build. I am not sure how much popular it is.

Since we're getting PRs for it, it is in use.

thedrow:
What we can do is specify that couchebase cannot be installed in pypy using markers.

thedrow:
We now see actual errors.
Are you working on fixing those?

mjhoffman65:
> What we can do is specify that couchebase cannot be installed in pypy using markers.



> We now see actual errors.
> Are you working on fixing those?



> We now see actual errors.
> Are you working on fixing those?
Yep!  Working now to fix them.

mjhoffman65:
@auvipy @thedrow This is ready for review.  I have some in-progress Dockerfile changes so it's easy to run the unit tests with pypy3 via docker as well, but that can go separately.

t/unit/backends/test_mongodb.py
```diff
@@ -45,7 +45,6 @@ def setup(self):
         self.patching('celery.backends.mongodb.MongoBackend.encode')
         self.patching('celery.backends.mongodb.MongoBackend.decode')
         self.patching('celery.backends.mongodb.Binary')
-        self.patching('datetime.datetime')
```
thedrow, line 48: Why did you remove this?

thedrow:
What about that event is a boolean failure that pops up every now and then?
@auvipy You know what I'm talking about right? Can you find a build log where that happens?
@mjhoffman65 If you repeatedly run the test suite multiple times, does it always pass?

t/unit/backends/test_mongodb.py
mjhoffman65: Removing that line fixes this error when running the MongoDB backend tests.
image

mjhoffman65:
> What about that event is a boolean failure that pops up every now and then?
> @auvipy You know what I'm talking about right? Can you find a build log where that happens?
> @mjhoffman65 If you repeatedly run the test suite multiple times, does it always pass?

I've ran it 4 times in docker and so far it has always passed.  I'm not sure how to trigger reruns with the github actions, but I haven't seen that error pop up yet.  I'm assuming it's this one: https://github.com/celery/celery/issues/6489

t/unit/backends/test_mongodb.py
thedrow: Why does it pass on CPython?

thedrow:
Yes, exactly.

mjhoffman65:
Was able to recreate #6489.  Will reopen after I fix it.

thedrow:
> Was able to recreate #6489. Will reopen after I fix it.

There's no need to close this PR. Just mark it as a draft if it's not ready for merge.

celery/utils/timer2.py
```diff
@@ -93,8 +93,9 @@ def run(self):
     def stop(self):
         self._is_shutdown.set()
         if self.running:
-            self._is_stopped.wait()
-            self.join(THREAD_TIMEOUT_MAX)
+            if not self._is_stopped:
```
thedrow, line 96: Why is this a boolean on PyPy?

thedrow:
It seems like the failing test is not related to your change.
Would you mind investigating or should I open an issue?

celery/utils/timer2.py
thedrow: @mjhoffman65 Ping?

thedrow:
I've rebased this PR but I'm still waiting for your response @mjhoffman65.

thedrow:
@celery/core-developers Can anyone complete this PR?

maybe-sybr:
Those dockerfile changes @mjhoffman65 mentioned might make life easier, but I don't see a ref to them. In any case, I've got a pypy3 interpreter on my box so I'll see if I can clean this up.

celery/utils/timer2.py
maybe-sybr: I'm suspicious about this change. If this is falsey, we're going to try an `.wait()` on it which seems odd since that probably implies that it's some garbage object. I don't think `Event`s should be treated like this, and if it is a `boolean` in the `pypy3` case, then we're going to do `bool.wait()` which seems wrong.

maybe-sybr:
Rebased this to hopefully fix the CI failures on that delivery info test, also with the correct fix for #6489

t/unit/backends/test_mongodb.py
maybe-sybr: `tox.ini` appears to need a change to run these tests.
```diff
diff --git a/tox.ini b/tox.ini
index f62ea3cdf..51cf5d020 100644
--- a/tox.ini
+++ b/tox.ini
@@ -31,7 +31,7 @@ deps=
     3.6,3.7,3.8,3.9: -r{toxinidir}/requirements/test-ci-default.txt
     3.5,3.6,3.7,3.8,3.9: -r{toxinidir}/requirements/docs.txt
     3.6,3.7,3.8,3.9: -r{toxinidir}/requirements/docs.txt
-    pypy3: -r{toxinidir}/requirements/test-ci-base.txt
+    pypy3: -r{toxinidir}/requirements/test-ci-default.txt
 
     integration: -r{toxinidir}/requirements/test-integration.txt
 
```

maybe-sybr: This patch was originally added back in the depths of time: 3ea8824d9b It seems to be a vestigial thing, so I think it'll be fine to remove it.

The breakage is because pypy3 appears to be a little less duck-type-y with its datetime subtraction implementation. We're giving it a `Mock` that looks like a `datetime` object, but isn't probably isn't using a `spec=` to pass `isinstance()` checks.

maybe-sybr: Leaving this for @thedrow to resolve

thedrow:
Why is the build failing?

maybe-sybr:
It's still broken :( I de-drafted this thinking it would run through and go green and now can't work out how to put it back :grimacing:

maybe-sybr:
Freshened this on top of master now that 5.1 is released. I've also made some changes to the workflow to run unit tests using tox. I think that might fix up some of the weirdness around the missing dependencies. We'll see...

maybe-sybr:
Freshened on master and also with a few extra fixes for the same kind of `Thread._is_stopped` shadowing in the `bgThread` class which is probably broken in a race-y way similar to `Timer` but only touched in two tests so we probably never noticed that it was different to the timer failure.

maybe-sybr:
Huh, weird, it's trying to run flake8 under action environments without python 3.9 installed :/

auvipy (approved):

thedrow (approved):

thedrow:
Thanks to everyone who helped us stablize the build :)

Issue #6639, opened 2021-02-20
Possibility to override the retry policy when a task is scheduled


glumia:
# Checklist
- [x] I have checked the [issues list](https://github.com/celery/celery/issues?q=is%3Aissue+label%3A%22Issue+Type%3A+Enhancement%22+-label%3A%22Category%3A+Documentation%22)
  for similar or identical enhancement to an existing feature.
- [x] I have checked the [pull requests list](https://github.com/celery/celery/pulls?q=is%3Apr+label%3A%22Issue+Type%3A+Enhancement%22+-label%3A%22Category%3A+Documentation%22)
  for existing proposed enhancements.
- [x] I have checked the [commit log](https://github.com/celery/celery/commits/master)
  to find out if the if the same enhancement was already implemented in the
  master branch.
- [x] I have included all related issues and possible duplicate issues in this issue
      (If there are none, check this box anyway).

#### Related Issues

- None

#### Possible Duplicates

- None

# Brief Summary
Currently there's no way to specify a retry policy for a task at the moment it's scheduled. 

If one has a task and wants to handle its failures differently in different places (eg. when scheduled from one point retry every 30 seconds for a max of 4 retries, when scheduled from another one retry 5 times with exponential delay between attempts) the only way he has to do so is to factor out task's logic into a function and declare one task wrapper for each place where he wants a different retry policy.

It would be much better if we had the possibility to override task's retry settings (`autoretry_for`, `default_retry_delay`, `retry_kwargs`) when we schedule it.

# Design

## Architectural Considerations

None

## Proposed Behavior
The existing behaviour of the `autoretry_for`, `default_retry_delay`, `retry_kwargs` options should not be modified.

The change should affect only the behaviour of the `app.task` decorator.

Example of a wrapper that implements this behaviour (just hacked together, could be certainly improved):
```python
from celery import Celery as _Celery
from celery.task import Task as _Task

class Celery(_Celery):
    def task(self, *args, **opts):
        autoretry_for_opts = opts.pop("autoretry_for", set())
        default_retry_delay = opts.pop("default_retry_delay", _Task.default_retry_delay)
        retry_kwargs = opts.pop("retry_kwargs", {})
        max_retries = retry_kwargs.get("max_retries", _Task.max_retries)
        countdown = retry_kwargs.get("countdown")

        default_delays = max_retries * (
            [countdown] if countdown else [default_retry_delay]
        )

        celery_decorator = super(Celery, self).task

        def retry_policy_decorator(func):
[...]

PR #6640, opened 2021-02-21
Draft: Add possibility to override autoretry options

glumia:
## Description
Draft implementation for #6639


celery/app/autoretry.py
```diff
@@ -1,35 +1,56 @@
...
         'retry_jitter', getattr(task, 'retry_jitter', True)
     )
 
-    if autoretry_for and not hasattr(task, '_orig_run'):
+    if _autoretry_for and not hasattr(task, '_orig_run'):
 
         @wraps(task.run)
-        def run(*args, **kwargs):
+        def run(*args, autoretry_opts=None, **kwargs):
+            if autoretry_opts is None:
```
auvipy, line 47: if autoretry_opts is None: or if autoretry_opts is not None:?

glumia: `if autoretry_opts is None`, it's just to not set `{}` as default argument because it's mutable. 

Even if it would be safe to do in this case it's better to avoid to do so.

thedrow:
Please use draft pull requests instead of adding "Darft" in the title.

PR #6644, opened 2021-02-24
Celery running in eager mode does not allow to use `Task.replace`

andreycizov:
## Description
In eager mode:

```
@task(bind=True)
def task(self):
     self.replace(task_b.s())

@task
def task_b():
     pass
```

Raises

```
  File "/usr/local/lib/python3.6/unittest/mock.py", line 1183, in patched
    return func(*args, **keywargs)
  File "/deploy/app/iwoca_common/tests/test_logging.py", line 288, in test_celery_replace
    fun_replace.delay()
  File "/usr/local/lib/python3.6/site-packages/celery/app/task.py", line 421, in delay
    return self.apply_async(args, kwargs)
  File "/usr/local/lib/python3.6/site-packages/celery/app/task.py", line 559, in apply_async
    link=link, link_error=link_error, **options)
  File "/usr/local/lib/python3.6/site-packages/celery/app/task.py", line 771, in apply
    ret = tracer(task_id, args, kwargs, request)
  File "/usr/local/lib/python3.6/site-packages/celery/app/trace.py", line 426, in trace_task
    I, R, state, retval = on_error(task_request, exc, uuid)
  File "/usr/local/lib/python3.6/site-packages/celery/app/trace.py", line 409, in trace_task
  File "/deploy/app/iwoca_common/logging.py", line 152, in wrapper
    return task_logger(*args, **kwargs)
  File "/deploy/app/iwoca_common/logging.py", line 78, in __call__
    rtn = self.decorated(*args, **kwargs)
  File "/deploy/app/iwoca_common/tests/test_logging.py", line 279, in fun_replace
    raise self.replace(fun_replacer.s())
  File "/usr/local/lib/python3.6/site-packages/celery/app/task.py", line 897, in replace
    return sig.apply().get()
  File "/usr/local/lib/python3.6/site-packages/celery/result.py", line 998, in get
    assert_will_not_block()
  File "/usr/local/lib/python3.6/site-packages/celery/result.py", line 37, in assert_will_not_block
    raise RuntimeError(E_WOULDBLOCK)
RuntimeError: Never call result.get() within a task!
```

I would like to add a test for this behaviour as well - is there any case you could direct me where the EAGER testing is done?


thedrow:
`t/unit/tasks/test_tasks.py` is the appropriate file.
Please add an integration test as well.

andreycizov:
I have added a unit test for retry that explicitly runs in eager mode.

> Please add an integration test as well.

There are multiple integration tests for `retry`. I would assume integration tests do not run with eager so they never matched the behaviour - would that imply I need to specifically run a task in eager mode within the integration tests?

thedrow:
> 
> 
> I have added a unit test for retry that explicitly runs in eager mode.
> 
> > Please add an integration test as well.
> 
> There are multiple integration tests for `retry`. I would assume integration tests do not run with eager so they never matched the behaviour - would that imply I need to specifically run a task in eager mode within the integration tests?

Yes, exactly. I want to see that everything runs as expected without mocks.

thedrow (dismissed):
So far, looks good to merge.

andreycizov:
I am struggling to find a way to set EAGER mode on a per-task basis, when launching a task. Is it reasonable to just mutate `celery.conf.task_always_eager = False` when I am launching a sub-task for the integration test?

Thanks for reviewing by the way.

thedrow:
> 
> 
> I am struggling to find a way to set EAGER mode on a per-task basis, when launching a task. Is it reasonable to just mutate `celery.conf.task_always_eager = False` when I am launching a sub-task for the integration test?
> 
> Thanks for reviewing by the way.

Since we use the session worker, it's not possible to change the worker's configuration mid-test run.
You should just `apply()` it and see if it's being replaced.

Also, could you please edit the PR description to describe what you're fixing instead of what is wrong?

thedrow:
Please mark this pull request as ready for review once you are done.

celery/app/task.py
```diff
@@ -919,6 +919,9 @@ def replace(self, sig):
 
         if self.request.is_eager:
             return sig.apply().get()
+            # task_result = sig.apply()
+            # with allow_join_result():
+            #     return task_result.get()
```
thedrow, line 924:
```suggestion
```

t/integration/test_canvas.py
```diff
@@ -803,6 +803,12 @@ def test_replaced_nested_chord(self, manager):
         res1 = c1()
         assert res1.get(timeout=TIMEOUT) == [29, 38]
 
+    @flaky
+    def test_replaced_eager(self):
+        # previously, eager tasks that used `Task.replace` would fail
+        assert add_replaced.apply(1, 1) == 2
+        raise AssertionError
```
thedrow, line 810: This test is meaningless since it always fails.

thedrow:
I think that at this point, it's hard to review this contribution.
@andreycizov Are you going to complete this PR?

celery/app/task.py
thedrow, line 924: This is probably what you meant.

```suggestion
            task_result = sig.apply()
            with allow_join_result():
                  return task_result.get()
```

thedrow (changes requested):

PR #6646, opened 2021-02-27
Simulate more exhaustive delivery info in apply()

Photonios:

## Description

When calling tasks directly, all options such as priority, exchange
and routing_key are accepted, but ignored.

All though these options take no effect when calling tasks
eagerly, it is useful to keep the data around. An application
might take action based on the priority, exchange or the
routing key.

It is especially useful when writing unit tests. This allows
tasks to be run eagerly, yet test that the application behaves
correctly when these options are passed.

### DIFF
--- a/celery/app/task.py
+++ b/celery/app/task.py
@@ -764,7 +764,12 @@ class Task:
             'callbacks': maybe_list(link),
             'errbacks': maybe_list(link_error),
             'headers': headers,
-            'delivery_info': {'is_eager': True},
+            'delivery_info': {
+                'is_eager': True,
+                'exchange': options.get('exchange'),
+                'routing_key': options.get('routing_key'),
+                'priority': options.get('priority'),
+            },
         }
         tb = None
         tracer = build_tracer(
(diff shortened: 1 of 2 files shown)


celery/app/task.py
```diff
@@ -764,7 +764,12 @@ def apply(self, args=None, kwargs=None,
             'callbacks': maybe_list(link),
             'errbacks': maybe_list(link_error),
             'headers': headers,
-            'delivery_info': {'is_eager': True},
+            'delivery_info': {
+                'is_eager': True,
+                'exchange': options.get('exchange'),
```
auvipy, line 769: what do these changes do? should we mention the changes in docs?

Photonios: They just make these options available inside the task when running a task eagerly:

```
@shared_task(bind=True)
def my_task(self):
     print(self.delivery_info.get('priority'))
     
my_task.apply_async(priority=2) # prints 2 in v5.0.0
my_task.apply(priority=2) # prints None v5.0.0
```

thedrow: I agree.

thedrow (approved):
LGTM.
This is well done. Thank you.

auvipy (approved):

PR #6649, opened 2021-02-28
Integration test & unit tests jobs separation

auvipy:


auvipy:
@xirdneh you can push improvement on this branch as use this as a base. I am checking how service containers going to work here for integration tests

auvipy:
pypy tests were failing before this PR.

auvipy:
redis connection is throwing errors. the service container might need some adjsutment

thedrow (changes requested):

auvipy:
@matusvalo please suggest modification too.

matusvalo:
Let me check CI in celery. It should be pretty straightforward to port kombu/py-amqp CI to celery also...

maybe-sybr:
Just freshened this on master. I think I'll spend a little time on it today to see if I can get the redis suite quite working as a first step and then the rabbitmq/rpc prior to merge

maybe-sybr:
The suite starts running now, but we have a busted test which fails the run after a few retries.

t/integration/test_inspect.py
```diff
@@ -117,7 +117,7 @@ def test_active(self, inspect):
                     'exchange': '',
                     'routing_key': 'celery',
                     'priority': 0,
-                    'redelivered': False
+                    'redelivered': ANY,
```
maybe-sybr, line 120: I'm not sure why these tests stopped passing. AFAICT the `redelivered` seems to be set to `None` implying that we "don't know" (or don't bother to default to `False`) if a task has been redelivered or not. These tests /probably/ (??) don't care about that so changing this to an `ANY` seems like it is okay.

I'd like for someone a bit more familiar with redelivery logic to think about this if possible. I just don't know what we should be expecting here and why it's not `False` like it must have been in the past.

maybe-sybr:
Not sure why that inspect test is failing. It passes on my box reliably

.github/workflows/python-package.yml
```diff
@@ -71,3 +75,85 @@ jobs:
...
+
+    runs-on: ubuntu-20.04
+    strategy:
+      fail-fast: false
+      matrix:
+        # XXX: RESTORE THIS MATRIX PRIOR TO MERGE
+        python-version: ['3.7', ] #'3.8', '3.9', 'pypy3']
+        # XXX: Does each matrix variation get their own copy of the service
+        # containers? Test cross-talk may be an issue...
+        toxenv: ['redis', ] #'rabbitmq']
```
maybe-sybr, line 99: This should probably actually be called `backend` since it's only the tail of the toxenv name.

t/integration/test_inspect.py
thedrow: `redelivered` is set when the broker redelivers the message (at least with AMQP brokers).
I honestly don't have an idea what has changed. Do you mind debugging this?

maybe-sybr: I'll see if I can find some time later this week

tarpas:
Hi. I'm curious what's the motivation for this change. What problem is it solving? Automated tests, their speed and UX for developer is an area of my (informal) research so I'm genuinely interested.

maybe-sybr:
> Hi. I'm curious what's the motivation for this change. What problem is it solving? Automated tests, their speed and UX for developer is an area of my (informal) research so I'm genuinely interested.

Our integration suits aren't being run as CI at the moment due to the slow migration we've had away from Travis which is where our CI jobs used to run. This change is mostly just intended to get them running again. There is an outstanding discussion around fast feedback and preferring to run the unit tests prior to the integration tests (as a prerequisite) which are intended to avoid running the more CI minute intensive suite(s) when the unit tests can show that something is clearly broken.

I think you'd need to ask a more specific question for any more detail.

tarpas:
That's good information! Do GitHub actions have some build minutes limits for OS projects? I wasn't able to google any mention of that. 

Another way to save CI minutes would be to use pytest -x (stop after first failure). Pro: saves probably even more minutes, Con: only indicates the first error/failure so the PR author might need to do more back and forth if not careful.

maybe-sybr:
Freshened this on current master, I assume that inspect test is probably still busted

maybe-sybr:
> That's good information! Do GitHub actions have some build minutes limits for OS projects? I wasn't able to google any mention of that.

I'm not sure tbh, I do know that minutes on Git*Lab* shared runners are limited per project/namespace so maybe GitHub is similar.

> Another way to save CI minutes would be to use pytest -x (stop after first failure). Pro: saves probably even more minutes, Con: only indicates the first error/failure so the PR author might need to do more back and forth if not careful.

We do use `-x` :) You might be able to see it in this diff, or have a look at the pytest commands we run via the `tox.ini` file. I find it more annoying than anything since I also use tox to test locally and find myself editing those lines to remove the `-x` pretty much any time I'm devving.

thedrow:
> 
> 
> > That's good information! Do GitHub actions have some build minutes limits for OS projects? I wasn't able to google any mention of that.
> 
> I'm not sure tbh, I do know that minutes on Git_Lab_ shared runners are limited per project/namespace so maybe GitHub is similar.
> 
> > Another way to save CI minutes would be to use pytest -x (stop after first failure). Pro: saves probably even more minutes, Con: only indicates the first error/failure so the PR author might need to do more back and forth if not careful.
> 
> We do use `-x` :) You might be able to see it in this diff, or have a look at the pytest commands we run via the `tox.ini` file. I find it more annoying than anything since I also use tox to test locally and find myself editing those lines to remove the `-x` pretty much any time I'm devving.

We need to enable `-x` only when the CI worker is building a pull request.

maybe-sybr:
Looks like the 3.7-redis suite passed last time I pushed. I've restored the unit tests (for sanity) and the integration matrix of python versions. I'm yet to re-enable the rabbitmq backend matrix element since I'm not sure if/how to avoid spinning up service containers which are for a backend not being tested in a particular job...

thedrow:
I rebased this PR.
We have to merge this soon so we can work on 5.2 safely.

thedrow:
What is wrong with the test suite?
Is it running locally just fine? @celery/core-developers

maybe-sybr:
```
     File "/home/runner/work/celery/celery/celery/app/trace.py", line 358, in build_tracer
      push_request = request_stack.push
  AttributeError: 'NoneType' object has no attribute 'push'
```

This seems pretty bad :grimacing:

auvipy:
> ```
>      File "/home/runner/work/celery/celery/celery/app/trace.py", line 358, in build_tracer
>       push_request = request_stack.push
>   AttributeError: 'NoneType' object has no attribute 'push'
> ```
> 
> This seems pretty bad grimacing

was the rabbitmq tests passing? I cant remember, I can only remember redis was failing. I will be doing some experiment in another PR to see if rabbit services & tests are running. I goose step at a time, lets see

PR #6654, opened 2021-03-02
Terminate tasks with late acknoledgement on connection loss

thedrow:

## Description

Tasks with late acknowledgement keep running after restart although the connection is lost and they cannot be acked anymore (to the best of my knowledge).

This results in log errors such as these:
```
[2021-03-02 18:29:58,355: CRITICAL/MainProcess] Couldn't ack 5, reason:RecoverableConnectionError(None, 'connection already closed', None, '')
  File "/home/thedrow/.virtualenvs/celery/lib/python3.9/site-packages/kombu/message.py", line 128, in ack_log_error
    self.ack(multiple=multiple)
  File "/home/thedrow/.virtualenvs/celery/lib/python3.9/site-packages/kombu/message.py", line 123, in ack
    self.channel.basic_ack(self.delivery_tag, multiple=multiple)
  File "/home/thedrow/.virtualenvs/celery/lib/python3.9/site-packages/amqp/channel.py", line 1391, in basic_ack
    return self.send_method(
  File "/home/thedrow/.virtualenvs/celery/lib/python3.9/site-packages/amqp/abstract_channel.py", line 54, in send_method
    raise RecoverableConnectionError('connection already closed')
amqp.exceptions.RecoverableConnectionError: connection already closed
[2021-03-02 18:29:58,356: CRITICAL/MainProcess] Couldn't ack 12, reason:RecoverableConnectionError(None, 'connection already closed', None, '')
```
If we cannot recover from this situation, we must terminate all tasks with late acknowledgement.
Any other suggestion, would be useful.

### DIFF
--- a/celery/app/defaults.py
+++ b/celery/app/defaults.py
@@ -290,6 +290,9 @@ NAMESPACES = Namespace(
         __old__=OLD_NS_WORKER,
         agent=Option(None, type='string'),
         autoscaler=Option('celery.worker.autoscale:Autoscaler'),
+        cancel_long_running_tasks_on_connection_loss=Option(
+            False, type='bool'
+        ),
         concurrency=Option(0, type='int'),
         consumer=Option('celery.worker.consumer:Consumer', type='string'),
         direct=Option(False, type='bool', old={'celery_worker_direct'}),
--- a/celery/worker/consumer/consumer.py
+++ b/celery/worker/consumer/consumer.py
@@ -345,7 +345,7 @@ class Consumer:
             if request.task.acks_late and not request.acknowledged:
                 warn(TERMINATING_TASK_ON_RESTART_AFTER_A_CONNECTION_LOSS,
                      request)
-                request.abort(self.pool)
+                request.cancel(self.pool)
 
     def register_with_event_loop(self, hub):
         self.blueprint.send_all(
--- a/celery/worker/consumer/consumer.py
+++ b/celery/worker/consumer/consumer.py
@@ -345,7 +345,7 @@ class Consumer:
             if request.task.acks_late and not request.acknowledged:
                 warn(TERMINATING_TASK_ON_RESTART_AFTER_A_CONNECTION_LOSS,
                      request)
-                request.terminate(self.pool)
+                request.abort(self.pool)
 
     def register_with_event_loop(self, hub):
         self.blueprint.send_all(
--- a/celery/worker/consumer/consumer.py
+++ b/celery/worker/consumer/consumer.py
@@ -29,8 +29,8 @@ from celery.utils.objects import Bunch
 from celery.utils.text import truncate
 from celery.utils.time import humanize_seconds, rate
 from celery.worker import loops
-from celery.worker.state import (maybe_shutdown, reserved_requests,
-                                 task_reserved)
+from celery.worker.state import (active_requests, maybe_shutdown,
+                                 reserved_requests, task_reserved)
 
 __all__ = ('Consumer', 'Evloop', 'dump_body')
 
@@ -104,6 +104,11 @@ MESSAGE_REPORT = """\
 body: {0}
 {{content_type:{1} content_encoding:{2}
   delivery_info:{3} headers={4}}}
+"""
+
+TERMINATING_TASK_ON_RESTART_AFTER_A_CONNECTION_LOSS = """\
+Task %s cannot be acknowledged after a connection loss since late acknowledgement is enabled for it.
+Terminating it instead.
 """
 
 
@@ -257,7 +262,7 @@ class Consumer:
     def _update_qos_eventually(self, index):
         return (self.qos.decrement_eventually if index < 0
                 else self.qos.increment_eventually)(
-                    abs(index) * self.prefetch_multiplier)
+            abs(index) * self.prefetch_multiplier)
 
     def _limit_move_to_pool(self, request):
         task_reserved(request)
@@ -335,6 +340,12 @@ class Consumer:
             self.connection.collect()
         except Exception:  # pylint: disable=broad-except
             pass
+
+        for request in tuple(active_requests):
+            if request.task.acks_late and not request.acknowledged:
+                warn(TERMINATING_TASK_ON_RESTART_AFTER_A_CONNECTION_LOSS,
+                     request)
+                request.terminate(self.pool)
 
     def register_with_event_loop(self, hub):
         self.blueprint.send_all(
@@ -487,7 +498,8 @@ class Consumer:
         signals.task_rejected.send(sender=self, message=message, exc=None)
 
     def on_unknown_task(self, body, message, exc):
-        error(UNKNOWN_TASK_ERROR, exc, dump_body(message, body), exc_info=True)
+        error(UNKNOWN_TASK_ERROR, exc, dump_body(message, body),
+              exc_info=True)
         try:
             id_, name = message.headers['id'], message.headers['task']
             root_id = message.headers.get('root_id')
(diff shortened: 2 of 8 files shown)


xirdneh:
@thedrow Can we add a test for this?

thedrow:
I'm not sure how since you need to cause a disconnect from the broker.
There are other problems with this PR. I'll update you soon.

xirdneh:
@thedrow That's what I was thinking. I was trying to figure out a way to mark a connection as disconnected so we can fake this case but I don't know if it's worth it.

matusvalo (approved):
LGTM

thedrow:
@matusvalo Actually, I was waiting on you to review this PR throughoutly.
Is there a way to revive the channel?

The problem is that by the time we restart the connection, the message's channel has its connection set to `None`.

> 
> 
> _Note_: Before submitting this pull request, please review our [contributing guidelines](https://docs.celeryproject.org/en/master/contributing.html).
> ## Description
> 
> Tasks with late acknowledgement keep running after restart although the connection is lost and they cannot be acked anymore (to the best of my knowledge).

> ```
> [2021-03-02 18:29:58,355: CRITICAL/MainProcess] Couldn't ack 5, reason:RecoverableConnectionError(None, 'connection already closed', None, '')
> Traceback (most recent call last):
>   File "/home/thedrow/.virtualenvs/celery/lib/python3.9/site-packages/kombu/message.py", line 128, in ack_log_error
>     self.ack(multiple=multiple)
>   File "/home/thedrow/.virtualenvs/celery/lib/python3.9/site-packages/kombu/message.py", line 123, in ack
>     self.channel.basic_ack(self.delivery_tag, multiple=multiple)
>   File "/home/thedrow/.virtualenvs/celery/lib/python3.9/site-packages/amqp/channel.py", line 1391, in basic_ack
>     return self.send_method(
>   File "/home/thedrow/.virtualenvs/celery/lib/python3.9/site-packages/amqp/abstract_channel.py", line 54, in send_method
>     raise RecoverableConnectionError('connection already closed')
> amqp.exceptions.RecoverableConnectionError: connection already closed
> [2021-03-02 18:29:58,356: CRITICAL/MainProcess] Couldn't ack 12, reason:RecoverableConnectionError(None, 'connection already closed', None, '')
> ```

Is there an alternative here?
I'd rather ack the message since the task is done.

thedrow:
@xirdneh @matusvalo There's also a request from a client of mine to terminate these tasks without revoking them.
Maybe we should introduce a setting for it? Is it always better to revoke these tasks?

matusvalo:
@thedrow can you show also the celery part of the stack? Kombu is able to recover automatically connection/channel during failed connection - see [1] - it also supports refreshing the channel if needed. 

Regarding your second question about revoking tasks, In this PR when you terminate the request, the task is revoked? (I don't know deep all details about Celery so maybe stupid question).

[1] https://docs.celeryproject.org/projects/kombu/en/master/userguide/failover.html#operation-failover

matusvalo:
My humble opinion: In distributed systems you can guarantee only on of the following two possibilities:
1. you can guarantee that message is processed at most once (this is early acknowledge)
2. you can guarantee that message is processed at least once (this is late acknowledge)

So it means that when user is using late acknowledge, he must be aware (and must design solution in that way) that data can be processed multiple times (they are idempotent). So  processing task multiple times is fine in this case and user must be take that in account.

thedrow:
> @thedrow can you show also the celery part of the stack? Kombu is able to recover automatically connection/channel during failed connection - see [1] - it also supports refreshing the channel if needed. 
> 

That's the only traceback we get. 
The message object has a channel with its connection set to None. 

> Regarding your second question about revoking tasks, In this PR when you terminate the request, the task is revoked? (I don't know deep all details about Celery so maybe stupid question).
> 
> [1] https://docs.celeryproject.org/projects/kombu/en/master/userguide/failover.html#operation-failover
> 
> 

Yes. A client of mine thinks it's the wrong behavior. According to them, we should terminate the task's without revoking.

thedrow:
Yes we shouldn't revoke as it will prevent the task from running again.
I feel like we need to introduce a new event.
I also hope that marking the task as retry is good enough and it won't have any side effects. I'm testing this now.

What I'm also working on now is to mark the tasks that were already done as successful, if that's possible.
If we can do that and check if the task was done, we can ack it when we receive it again.
@celery/core-developers Does that sound feasible?

ahopkins (dismissed):

auvipy (commented):
is it possible to add a test for the changes?

thedrow:
> is it possible to add a test for the changes?

I already replied to @xirdneh.

>  
> I'm not sure how since you need to cause a disconnect from the broker.

We can add a unit test but I'm not sure what it will cover exactly.

thedrow:
I still need to introduce a feature flag and write some tests.

celery/app/defaults.py
```diff
@@ -290,6 +290,9 @@ def __repr__(self):
         __old__=OLD_NS_WORKER,
         agent=Option(None, type='string'),
         autoscaler=Option('celery.worker.autoscale:Autoscaler'),
+        cancel_long_running_tasks_on_connection_loss=Option(
```
auvipy, line 293: any shorter name in mind?

maybe-sybr: @thedrow and I spoke about this a bit out of band and it sounded to me like being more explicit about what the option does is probably best. I could only imaging this being called something like `connection_loss_strategy` and having an enum of different behaviours (which would be like `DEFAULT_STRATEGY`, `IGNORE_EVENT` and `CANCEL_TASKS` or something like that).

I think since there are only two strategies right now, this long and more explicit name is probably okay.

celery/worker/consumer/consumer.py
```diff
@@ -336,6 +351,15 @@ def on_connection_error_after_connected(self, exc):
         except Exception:  # pylint: disable=broad-except
             pass
 
+        if self.app.conf.worker_cancel_long_running_tasks_on_connection_loss:
+            for request in tuple(active_requests):
+                if request.task.acks_late and not request.acknowledged:
+                    warn(TERMINATING_TASK_ON_RESTART_AFTER_A_CONNECTION_LOSS,
+                         request)
```
maybe-sybr, line 358: Should this be using `request.task.id` to ensure we only stringify the ID? Happy for this to remain as is or change as appropriate.

t/unit/worker/test_request.py
```diff
@@ -449,6 +453,23 @@ def test_terminate__task_started(self):
...
+        job._apply_result = Mock(name='_apply_result')
+        with self.assert_signal_called(
+                task_retry, sender=job.task, request=job._context,
+                einfo=None):
+            job.time_start = monotonic()
+            job.worker_pid = 314
+            job.cancel(pool, signal='TERM')
+            job._apply_result().terminate.assert_called_with(signum)
+
+            job._apply_result = Mock(name='_apply_result2')
```
maybe-sybr, line 469: Does this need an `assert_called_once_with()` as well?

maybe-sybr:
Overall this lgtm. I'm happy to approve the structure of this an allow @thedrow to address or resolve my minor comments as he sees fit.

maybe-sybr (approved):
Overall this lgtm. I'm happy to approve the structure of this an allow @thedrow to address or resolve my minor comments as he sees fit.

Edit: Oops, forgot to do this as a review rather than comments!

celery/worker/consumer/consumer.py
thedrow: No, we need to stringify the request as intended.

t/unit/worker/test_request.py
thedrow: The original test didn't include it. Feel free to do so, if you wish this to be fixed.
We can open an issue about it but I think it's rather minor.

PR #6664, opened 2021-03-08
add 'second' attribute in crontab class

f5reestyle:

## Description

I add 'second' attribute in the 'crontab' class. There were many questions in stackoverflow asking 'how to manipulate the crontab in seconds?'. I think celery crontab would be more useful to have 'second' attribute like other alternatives (ex.APScheduler)

 1. add 'second' argument in '__init__'method
 2. add 'second' part in return value in '__reduce__' and '__eq__' methods
  3. add some logic for calculating remaining seconds in 'remaining_delta' method
  
'second' attribute works like 'minute' attribute, I guess. The fixed class works well in my beat-scheduling cases, but I have not tested for all cases.

Thank you.

### DIFF
--- a/celery/schedules.py
+++ b/celery/schedules.py
@@ -314,89 +314,23 @@ class crontab_parser:
 class crontab(BaseSchedule):
     """Crontab schedule.
 
-    A Crontab can be used as the ``run_every`` value of a
-    periodic task entry to add :manpage:`crontab(5)`-like scheduling.
+    1. add 'second' argument in '__init__'method
+    2. add 'second' part in return value in '__reduce__' and '__eq__' methods
+    3. add some logic for remaining seconds in 'remaining_delta' method
 
-    Like a :manpage:`cron(5)`-job, you can specify units of time of when
-    you'd like the task to execute.  It's a reasonably complete
-    implementation of :command:`cron`'s features, so it should provide a fair
-    degree of scheduling needs.
-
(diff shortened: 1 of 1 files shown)


celery/schedules.py
```diff
@@ -314,89 +314,23 @@ def _expand_number(self, s):
...
-
-    It's important to realize that any day on which execution should
-    occur must be represented by entries in all three of the day and
-    month attributes.  For example, if ``day_of_week`` is 0 and
-    ``day_of_month`` is every seventh day, only months that begin
-    on Sunday and are also in the ``month_of_year`` attribute will have
-    execution events.  Or, ``day_of_week`` is 1 and ``day_of_month``
-    is '1-7,15-21' means every first and third Monday of every month
-    present in ``month_of_year``.
+    1. add 'second' argument in '__init__'method
```
thedrow, line 317: Why did you erase the docstring?

celery/schedules.py
```diff
@@ -627,13 +570,14 @@ def is_due(self, last_run_at):
         return schedstate(due, rem)
 
     def __eq__(self, other):
-        if isinstance(other, crontab):
+        if isinstance(other, customedCrontab):
```
thedrow, line 573: This is probably the wrong type name.

thedrow (changes requested):
This PR is still not ready for review.

PR #6671, opened 2021-03-09
Prefer more specific glob routing patterns

### LINKED ISSUE #6670: Less specific task routing is preferred if it comes first
# Checklist

  master branch.

#### Related Issues

- #5646 

#### Possible Duplicates

- None

# Brief Summary

The routes in `MapRoute` take precedence by their order of insertion.
This is by design but as far as I can tell undocumented.
However it doesn't make sense that the route `'proj.tasks.foo*'` doesn't always take precedence over 'proj.tasks.*' as `'proj.tasks.foo*'` is more specfic.

# Design

None

## Proposed Behavior

A more specific glob route should take precedence over a less specific one.

This may introduce a breaking change for those who specified their task routes incorrectly (as the less specific routing would apply).
However, I don't believe it will affect most of our users since they must have adjusted the order so that the more specific route would apply, found another workaround or given up.

## Proposed UI/UX

We're relaxing a requirement so therefore this will improve the UX as both:
```python
task_routes = ([
    ('web.tasks.notifications.emails.*', {'queue': 'emails'}),
    ('web.tasks.*', {'queue': 'web'}),
],)
```
and:

```python
task_routes = ([
    ('web.tasks.*', {'queue': 'web'}),
    ('web.tasks.notifications.emails.*', {'queue': 'emails'}),
],)
```
are equivalent.

## Diagrams

N/A

## Alternatives

None

thedrow:

## Description
The order of specification for task routes no longer matter as we now sort by pattern length.
If two overlapping patterns will match a route, we now prefer the longer one as it is more specific.

Fixes #6670.

### DIFF
--- a/celery/app/routes.py
+++ b/celery/app/routes.py
@@ -29,20 +29,19 @@ class MapRoute:
     def __init__(self, map):
         map = map.items() if isinstance(map, Mapping) else map
         self.map = {}
-        self.patterns = {}
+        patterns = {}
         for k, v in map:
             if isinstance(k, Pattern):
-                self.patterns[k] = v
+                patterns[k] = v
             elif '*' in k:
-                self.patterns[re.compile(fnmatch.translate(k))] = v
+                patterns[re.compile(fnmatch.translate(k))] = v
             else:
                 self.map[k] = v
 
         # We sort by the regex pattern's length since longer regex patterns
         # are likely to be more specific.
-        self.patterns = tuple(reversed(sorted(tuple(self.patterns.items()),
-                                              key=lambda pattern: len(
-                                                  pattern[0].pattern))))
+        self.patterns = tuple(reversed(sorted(patterns.items(),
+                                              key=lambda item: len(item[0].pattern))))
 
     def __call__(self, name, *args, **kwargs):
         try:
--- a/celery/app/routes.py
+++ b/celery/app/routes.py
@@ -4,7 +4,6 @@ Contains utilities for working with task routers, (:setting:`task_routes`).
 """
 import fnmatch
 import re
-from collections import OrderedDict
 from collections.abc import Mapping
 
 from kombu import Queue
@@ -23,19 +22,27 @@ except AttributeError:  # pragma: no cover
 __all__ = ('MapRoute', 'Router', 'prepare')
 
 
+GLOB_PATTERNS = ('*', '?', '[', ']', '!')
+
 class MapRoute:
     """Creates a router out of a :class:`dict`."""
 
     def __init__(self, map):
+        # map is either a mapping or a an iterable of tuples
         map = map.items() if isinstance(map, Mapping) else map
         self.map = {}
         patterns = {}
         for k, v in map:
             if isinstance(k, Pattern):
+                # This is already a regular expression so we simply store it.
                 patterns[k] = v
-            elif '*' in k:
+            elif any(glob_pattern in k for glob_pattern in GLOB_PATTERNS):
+                # This is a glob pattern so we
+                # must to translate it into a regular expression.
                 patterns[re.compile(fnmatch.translate(k))] = v
             else:
+                # This is a direct mapping between a task and a routing
+                # so we simply store it.
                 self.map[k] = v
 
         # We sort by the regex pattern's length since longer regex patterns
--- a/celery/app/routes.py
+++ b/celery/app/routes.py
@@ -21,8 +21,8 @@ except AttributeError:  # pragma: no cover
 
 __all__ = ('MapRoute', 'Router', 'prepare')
 
-
 GLOB_PATTERNS = ('*', '?', '[', ']', '!')
+
 
 class MapRoute:
     """Creates a router out of a :class:`dict`."""
@@ -56,12 +56,16 @@ class MapRoute:
         except KeyError:
             pass
         except ValueError:
+            # if self.map[name] is a string we consider it to be the queue's
+            # name.
             return {'queue': self.map[name]}
         for regex, route in self.patterns:
             if regex.match(name):
                 try:
                     return dict(route)
                 except ValueError:
+                    # if route is a string we consider it to be the queue's
+                    # name.
                     return {'queue': route}
 
 
--- a/celery/app/routes.py
+++ b/celery/app/routes.py
@@ -29,7 +29,7 @@ class MapRoute:
     def __init__(self, map):
         map = map.items() if isinstance(map, Mapping) else map
         self.map = {}
-        self.patterns = OrderedDict()
+        self.patterns = {}
         for k, v in map:
             if isinstance(k, Pattern):
                 self.patterns[k] = v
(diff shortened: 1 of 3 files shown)


celery/app/routes.py
```diff
@@ -38,14 +38,20 @@ def __init__(self, map):
             else:
                 self.map[k] = v
 
+        # We sort by the regex pattern's length since longer regex patterns
+        # are likely to be more specific.
+        self.patterns = tuple(reversed(sorted(tuple(self.patterns.items()),
```
ahopkins, line 43: Why the explicit cast from `.items()` to `tuple`?

```python
>>> sorted({1:1,3:3,2:2}.items())
[(1, 1), (2, 2), (3, 3)]
```

celery/app/routes.py
```diff
@@ -29,7 +29,7 @@ class MapRoute:
     def __init__(self, map):
         map = map.items() if isinstance(map, Mapping) else map
         self.map = {}
-        self.patterns = OrderedDict()
+        self.patterns = {}
```
ahopkins, line 32: Instead of defining and then redefining this variable, there is no need to assign it to `self` here. I realize there is no practical impact since there is no typing check. But, _if_ `mypy` were implemented someday, it would complain.

ahopkins (dismissed):

celery/app/routes.py
thedrow: Because sorted expects it. Maybe iter() would also be sufficient.

thedrow: It didn't. I was wrong and you're right.
Thanks.

celery/app/routes.py
```diff
@@ -22,34 +22,60 @@
...
             else:
+                # This is a direct mapping between a task and a routing
+                # so we simply store it.
                 self.map[k] = v
 
+        # We sort by the regex pattern's length since longer regex patterns
+        # are likely to be more specific.
+        self.patterns = tuple(reversed(sorted(patterns.items(),
+                                              key=lambda item: len(item[0].pattern))))
+
```
xirdneh, line 62: **[suggestion]**: Treat each type of pattern different.

After trying out some examples manually in `ipython` I suggest giving different weights to each of the type of patterns we can receive from the user.

We can receive either a regex, a glob pattern or a direct mapping.

We could keep the `OrderedDict` but insert direct mapping first, then glob patterns then regular expressions. Keeping the length sorting. This way we avoid any weird cases where a string is actually longer but is not more specific than a direct mapping.

Something like this:

```python
    mappings = []
    globs = []
    regexes = []

    for k, v in map:
        if isinstance(k, Patter):
            regexes.append((k, v))
        elif any(glob_pattern in k for glob_pattern in GLOB_PATTERNS):
            globs.append(re.compile(fnmatch.translate(k)),  v)
        else:
            mappings.append((k, v))

    globs = reversed(sorted(globs, key=lambda item: len(item[0].pattern)))
    mappings = reversed(sorted(mappings, key=lambda item: len(item[0].pattern)))

    self.patterns = OrderedDict(mappings + globs + regexes)
```

This way any direct mapping will always take precedence over everything else, then glob patterns then regexes. 
This is mainly because in my opinion `some.direct.mapping` is more specific than `some.direct.mapping.*`

Then, we only sort direct mappings and glob patterns by length because that does make sense but we leave regexes in the same order we receive them. That way we keep the same behavior for regexes that's used in most frameworks using regex to match stuff, the first one wins.

auvipy: if isinstance(k, Patter): should be, if isinstance(k, Pattern):

auvipy: and  self.patterns = OrderedDict() to  {}

thedrow: Why does the order of the regexes matter?

thedrow: Also, I'd prefer to have consistent behavior across all route types.

matusvalo (approved):
LGTM

celery/app/routes.py
xirdneh: The order of the regexes matter because that's usually how regexes are matched. This is mainly thinking about other python web frameworks that accept regexes as url routes. But, it might make more sense for urls and not here.

I'm OK with also sorting the regexes by length. But I do think that the precedence should be direct mappings > glob patterns > regexes.

It makes more sense from the user perspective that a direct mapping would work over a glob pattern or a regex since a direct mapping is not leaving room for any other value and is an exact string match.

celery/app/routes.py
```diff
@@ -24,10 +24,10 @@
 
 GLOB_PATTERNS = ('*', '?', '[', ']', '!')
 
-MAP_ROUTES_MUST_BE_A_DICTIONARY = """\
-Starting from Celery 5.1 the task_routes configuration must be a dictionary.
-Support for providing a list of router objects will be removed in 6.0.
-""".strip()
+MAP_ROUTES_MUST_BE_A_DICTIONARY = (
+    "Starting from Celery 5.1 the task_routes configuration must be a dictionary. "
+    "Support for providing a list of router objects will be removed in 6.0."
```
thedrow, line 29: @xirdneh there's no newline here.
Also, why is this better?

celery/app/routes.py
thedrow: In the current code (even after this change), we prefer direct mappings over patterns.
What I don't understand is why we should prefer glob patterns over other regexes.

xirdneh: I'd assume we want to give preference to glob patterns over regexes because glob patterns are more specific by definition.

For example `using.glob.patterns.*` is more specific than `using\.(glob|regexes)\.patterns\..*`.
This is a very specific example but since we cannot know how specific or general a regex is then I think it's safe to assume that a user would expect a string with a single glob pattern will have precedence than a regex with a bunch of rules.

And, you are correct I forgot about the code in `__call__` that checks for direct mappings before looping through the rest of the configured routing patterns.

celery/app/routes.py
xirdneh: Good point about the new line, it should be added there.

This change was marked as a *nit* because either way works and I'm OK with both versions.
But, I find using `strip()` at the end of a multi line string unnecessary.

celery/app/routes.py
thedrow: If it has a bunch of rules then it's longer by definition no?

xirdneh: Yes. The code in this PR would prefer the longer regex, right?
In this case, the longer regex is not more specific than the sorter pattern with a glob.

Or, what do you mean?

thedrow: Why is it not more specific?

maybe-sybr: I think we could arguably sort patterns by the number of groups in them, since that likely how most users would be using them. Aside from that, I think the behaviour as described in this comment thread seems sensible enough, and we should just aim to document it in a fairly obvious way so that if/when it surprises someone they can find out why it's not doing what they think it should.

PR #6674, opened 2021-03-11
Drop the lzma extra

thedrow:

## Description

The lzma backport is no longer needed since we don't support Python<3.3 anymore.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -27,7 +27,6 @@ EXTENSIONS = {
     'eventlet',
     'gevent',
     'librabbitmq',
-    'lzma',
     'memcache',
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -27,7 +27,6 @@
     'eventlet',
     'gevent',
     'librabbitmq',
-    'lzma',
```
auvipy, line 30: backports.lzma; & 'lzma' same?

thedrow: An extra has a name. It can install multiple packages.

auvipy (approved):

PR #6683, opened 2021-03-17
sqla 1.4 migration initialization

auvipy:
https://docs.sqlalchemy.org/en/14/changelog/migration_14.html#declarative-is-now-integrated-into-the-orm-with-new-features

### DIFF
--- a/celery/backends/database/session.py
+++ b/celery/backends/database/session.py
@@ -4,8 +4,7 @@ import time
 from kombu.utils.compat import register_after_fork
 from sqlalchemy import create_engine
 from sqlalchemy.exc import DatabaseError
-from sqlalchemy.ext.declarative import declarative_base
-from sqlalchemy.orm import sessionmaker
+from sqlalchemy.orm import sessionmaker, declarative_base
 from sqlalchemy.pool import NullPool
 
 from celery.utils.time import get_exponential_backoff_interval
--- a/examples/django/requirements.txt
+++ b/examples/django/requirements.txt
@@ -1,3 +1,3 @@
 django>=2.2.1
-sqlalchemy>=1.0.14
+sqlalchemy==1.4
 celery>=5.0.5
--- a/examples/django/requirements.txt
+++ b/examples/django/requirements.txt
@@ -1,3 +1,3 @@
 django>=2.2.1
-sqlalchemy==1.4
+sqlalchemy>=1.4.0
 celery>=5.0.5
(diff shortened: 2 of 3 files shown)


examples/django/requirements.txt
```diff
@@ -1,3 +1,3 @@
 django>=2.2.1
-sqlalchemy>=1.0.14
+sqlalchemy==1.4
```
thedrow, line 2: We can relax this requirement can't we?
```suggestion
sqlalchemy>=1.4.0
```

celery/backends/database/session.py
```diff
@@ -4,8 +4,7 @@
 from kombu.utils.compat import register_after_fork
 from sqlalchemy import create_engine
 from sqlalchemy.exc import DatabaseError
-from sqlalchemy.ext.declarative import declarative_base
-from sqlalchemy.orm import sessionmaker
+from sqlalchemy.orm import sessionmaker, declarative_base
```
thedrow, line 7: Why should we drop support for older versions?

thedrow (changes requested):

celery/backends/database/session.py
auvipy: declarative_base() - this is the same declarative base class that has been in use throughout the span of the declarative system, except it now references a registry object internally and is implemented by the registry.generate_base() method which can be invoked from a registry directly. The declarative_base() function creates this registry automatically so there is no impact on existing code. The sqlalchemy.ext.declarative.declarative_base name is still present, emitting a 2.0 deprecation warning when 2.0 deprecations mode is enabled.

auvipy: this is not a so big change from a sqla users perspective. as we are just updating the code path, nothing functional is changing with this and not in database level.

thedrow: I understand. But why should we drop support for older versions?

auvipy: sqla 1.4 is much slimmer in API & lay the groundwork for sqla 2.0 which will be python 3 only. if we are going to support multiple versions support, that won;t be hard as its just import and fallback to another version if import fail.

thedrow: I think we should support multiple versions until the next major version of Kombu.
It should be easy to do so.
Please add a TODO so we don't forget to remove the fallback once we drop support for sqlalchemy<1.4.

auvipy: ok

auvipy:
https://github.com/celery/kombu/pull/1318

thedrow:
I did it myself. I'm opening a PR right now.

PR #6733, opened 2021-04-19
fix: Chord counting of group children

### LINKED ISSUE #6721: 5.1.0b1 chord of chains dont work
# Checklist

- [ ] I have read the relevant section in the
  on reporting bugs.
- [ ] I have checked the [issues list](https://github.com/celery/celery/issues?q=is%3Aissue+label%3A%22Issue+Type%3A+Bug+Report%22+-label%3A%22Category%3A+Documentation%22)
- [ ] I have checked the [pull requests list](https://github.com/celery/celery/pulls?q=is%3Apr+label%3A%22PR+Type%3A+Bugfix%22+-label%3A%22Category%3A+Documentation%22)
  for existing proposed fixes.
- [ ] I have checked the [commit log](https://github.com/celery/celery/commits/master)
- [ ] I have included all related issues and possible duplicate issues

     version affected).
  to reproduce this bug.

  and/or implementation.
  result backend.
  broker and/or result backend.
- [x] I have tried reproducing the issue after downgrading

#### Related Issues

- None

#### Possible Duplicates

- None

## Environment & Settings

**Celery version**:  ```5.1.0b1 (singularity)```

<details>
<p>

```
software -> celery:5.1.0b1 (singularity) kombu:5.1.0b1 py:3.8.5
            billiard:3.6.4.0 py-amqp:5.0.6
            kernel version:5.8.0-48-generic imp:CPython
settings -> transport:amqp results:redis://localhost/

result_backend: 'redis://localhost/'
deprecated_settings: None
```

</p>
</details>

# Steps to Reproduce

## Required Dependencies

### Python Packages

<details>
<p>

```
```

</p>
</details>

### Other Dependencies

<details>
<p>
N/A
</p>
</details>

<details>
<p>

```python
import celery

app = celery.Celery(
    broker='amqp://localhost',
    backend='redis://localhost',
)


@app.task
def task_a(number):
    print(f'Task A: {number}')
    return number


@app.task
def task_b(number):
    print(f'Task B: {number}')
    return number


@app.task
def task_c(number):
    print(f'Task C: {number}')


if __name__ == '__main__':
    celery.chord(
      (celery.chain(task_a.s(i), task_b.s()))
      for i in range(3)
    )(task_c.s())  # yapf: disable

    app.start()
```

</p>
</details>

# Expected Behavior

Expected output:
```
[2021-04-11 16:44:51,364: WARNING/ForkPoolWorker-15] Task A: 0
[2021-04-11 16:44:51,364: WARNING/ForkPoolWorker-16] Task A: 1
[2021-04-11 16:44:51,364: WARNING/ForkPoolWorker-1] Task A: 2
[2021-04-11 16:44:51,387: WARNING/ForkPoolWorker-4] Task B: 2
[2021-04-11 16:44:51,387: WARNING/ForkPoolWorker-3] Task B: 0
[2021-04-11 16:44:51,387: WARNING/ForkPoolWorker-2] Task B: 1
[2021-04-11 16:44:51,413: WARNING/ForkPoolWorker-15] Task C: [0, 1, 2]
```

On `celery==5.0.5` it works fine

# Actual Behavior

```
[2021-04-11 16:33:24,583: WARNING/ForkPoolWorker-15] Task A: 0
[2021-04-11 16:33:24,583: WARNING/ForkPoolWorker-1] Task A: 2
[2021-04-11 16:33:24,583: WARNING/ForkPoolWorker-16] Task A: 1
[2021-04-11 16:33:24,608: WARNING/ForkPoolWorker-4] Task B: 2
[2021-04-11 16:33:24,608: WARNING/ForkPoolWorker-2] Task B: 0
[2021-04-11 16:33:24,608: WARNING/ForkPoolWorker-3] Task B: 1
```

maybe-sybr:
## Description

Fixes #6721

This PR fixes the faulty chord counting reported in #6721, as well as an edge case where the tail task of a chain being counted for chord size could be an empty group (or other zero sized thing).

### DIFF
--- a/celery/canvas.py
+++ b/celery/canvas.py
@@ -1424,8 +1424,15 @@ class _chord(Signature):
             subtasks = getattr(sig_obj.tasks, "tasks", sig_obj.tasks)
             return sum(cls._descend(task) for task in subtasks)
         elif isinstance(sig_obj, _chain):
-            # The last element in a chain counts toward this chord
-            return cls._descend(sig_obj.tasks[-1])
+            # The last non-empty element in a chain counts toward this chord
+            for child_sig in sig_obj.tasks[-1::-1]:
+                child_size = cls._descend(child_sig)
+                if child_size > 0:
+                    return child_size
+            else:
+                # We have to just hope this chain is part of some encapsulating
+                # signature which is valid and can fire the chord body
+                return 0
         elif isinstance(sig_obj, chord):
             # The child chord's body counts toward this chord
             return cls._descend(sig_obj.body)
--- a/celery/canvas.py
+++ b/celery/canvas.py
@@ -1415,20 +1415,20 @@ class _chord(Signature):
         )
 
     @classmethod
-    def __descend(cls, sig_obj):
+    def _descend(cls, sig_obj):
         # Sometimes serialized signatures might make their way here
         if not isinstance(sig_obj, Signature) and isinstance(sig_obj, dict):
             sig_obj = Signature.from_dict(sig_obj)
         if isinstance(sig_obj, group):
             # Each task in a group counts toward this chord
             subtasks = getattr(sig_obj.tasks, "tasks", sig_obj.tasks)
-            return sum(cls.__descend(task) for task in subtasks)
+            return sum(cls._descend(task) for task in subtasks)
         elif isinstance(sig_obj, _chain):
             # The last element in a chain counts toward this chord
-            return cls.__descend(sig_obj.tasks[-1])
+            return cls._descend(sig_obj.tasks[-1])
         elif isinstance(sig_obj, chord):
             # The child chord's body counts toward this chord
-            return cls.__descend(sig_obj.body)
+            return cls._descend(sig_obj.body)
         elif isinstance(sig_obj, Signature):
             # Each simple signature counts as 1 completion for this chord
             return 1
@@ -1437,7 +1437,7 @@ class _chord(Signature):
 
     def __length_hint__(self):
         tasks = getattr(self.tasks, "tasks", self.tasks)
-        return sum(self.__descend(task) for task in tasks)
+        return sum(self._descend(task) for task in tasks)
 
     def run(self, header, body, partial_args, app=None, interval=None,
             countdown=1, max_retries=None, eager=False,
--- a/celery/canvas.py
+++ b/celery/canvas.py
@@ -1296,8 +1296,8 @@ class group(Signature):
         return app if app is not None else current_app
 
 
-@Signature.register_type()
-class chord(Signature):
+@Signature.register_type(name="chord")
+class _chord(Signature):
     r"""Barrier synchronization primitive.
 
     A chord consists of a header and a body.
@@ -1535,6 +1535,11 @@ class chord(Signature):
 
     tasks = getitem_property('kwargs.header', 'Tasks in chord header.')
     body = getitem_property('kwargs.body', 'Body task of chord.')
+
+
+# Add a back-compat alias for the previous `chord` class name which conflicts
+# with keyword arguments elsewhere in this file
+chord = _chord
 
 
 def signature(varies, *args, **kwargs):
--- a/celery/canvas.py
+++ b/celery/canvas.py
@@ -1170,21 +1170,25 @@ class group(Signature):
             # we are able to tell when we are at the end by checking if
             # next_task is None.  This enables us to set the chord size
             # without burning through the entire generator.  See #3021.
+            chord_size = 0
             for task_index, (current_task, next_task) in enumerate(
                 lookahead(tasks)
             ):
+                # We expect that each task must be part of the same group which
+                # seems sensible enough. If that's somehow not the case we'll
+                # end up messing up chord counts and there are all sorts of
+                # awful race conditions to think about. We'll hope it's not!
                 sig, res, group_id = current_task
-                _chord = sig.options.get("chord") or chord
-                if _chord is not None and next_task is None:
-                    chord_size = task_index + 1
-                    if isinstance(sig, _chain):
-                        if sig.tasks[-1].subtask_type == 'chord':
-                            chord_size = sig.tasks[-1].__length_hint__()
-                        else:
-                            chord_size = task_index + len(sig.tasks[-1])
+                chord_obj = sig.options.get("chord") or chord
+                # We need to check the chord size of each contributing task so
+                # that when we get to the final one, we can correctly set the
+                # size in the backend and the chord can be sensible completed.
+                chord_size += _chord._descend(sig)
+                if chord_obj is not None and next_task is None:
+                    # Per above, sanity check that we only saw one group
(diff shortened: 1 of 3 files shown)


maybe-sybr:
I have to duck out for a bit, but I still need to run the MRTC from #6721 with this change and also add it to the integration test suite. FWICT this diff is enough to fix the issue, may need some massaging to be mergable.

auvipy (commented):
changes look good to me, can you also add an integration test for this?

celery/canvas.py
```diff
@@ -1170,21 +1170,29 @@ def _apply_tasks(self, tasks, producer=None, app=None, p=None,
...
-                            chord_size = sig.tasks[-1].__length_hint__()
-                        else:
-                            chord_size = task_index + len(sig.tasks[-1])
+                group_ids.add(group_id)
+                chord_obj = sig.options.get("chord") or chord
+                # We need to check the chord size of each contributing task so
+                # that when we get to the final one, we can correctly set the
+                # size in the backend and the chord can be sensible completed.
+                chord_size += _chord._descend(sig)
+                if chord_obj is not None and next_task is None:
```
thedrow, line 1190: If this is a sanity check, please add `if __debug__` so that the entire check would not be executed in production.

celery/canvas.py
```diff
@@ -1296,8 +1304,8 @@ def app(self):
         return app if app is not None else current_app
 
 
-@Signature.register_type()
-class chord(Signature):
+@Signature.register_type(name="chord")
+class _chord(Signature):
```
thedrow, line 1308: Why the name change?

celery/canvas.py
```diff
@@ -1415,20 +1423,27 @@ def apply(self, args=None, kwargs=None,
         )
 
     @classmethod
-    def __descend(cls, sig_obj):
+    def _descend(cls, sig_obj):
```
thedrow, line 1426: Why the name change?

celery/canvas.py
```diff
@@ -1170,21 +1170,29 @@ def _apply_tasks(self, tasks, producer=None, app=None, p=None,
             # we are able to tell when we are at the end by checking if
             # next_task is None.  This enables us to set the chord size
             # without burning through the entire generator.  See #3021.
+            group_ids = set()
```
thedrow, line 1173: If this set is used only for the sanity check we should avoid using it if `__debug__` is False.

thedrow (changes requested):
I've left some comments and questions.

celery/canvas.py
maybe-sybr: The method gets mangled with the old name and we need to call it from the `group` class now rather than just `chord`.

celery/canvas.py
maybe-sybr: This is so we can touch the `chord` class (as `_chord`) from methods in `group` which shadow the class' name with their kwargs (e.g. `group._apply_tasks(..., chord=None, ...)`)

celery/canvas.py
maybe-sybr: This bit isn't but the assertion below is. I'll make sure that only happens `if __debug__` but build the `group_ids` set as usual and just not use it when `not __debug__`

maybe-sybr: This is done now. I'll resolve the other thread and leave this one for you.

t/integration/test_canvas.py
```diff
@@ -704,6 +704,109 @@ def test_nested_group_group(self, manager):
...
+        # underlying promise might not be getting fulfilled (ref #6734). Pick a
+        # short timeout since we don't want to block for ages and this is a
+        # fairly simple signature which should run pretty quickly.
+        expected_result = [[42]]
+        with pytest.raises(TimeoutError):
+            res.get(timeout=TIMEOUT / 10)
+        # Get the child `AsyncResult` manually so that we don't have to wait
+        # again for the `GroupResult`
+        assert res.children[0].get(timeout=TIMEOUT) == expected_result[0]
+        assert res.get(timeout=TIMEOUT) == expected_result
```
maybe-sybr, line 808: I toyed with making this test xfail with the reason being #6734, but I think it's important that we continue to test the current behaviour doesn't change, where the `GroupResult` can be resolved if all of its children are manually resolved first.

Is there a way for me to xfail a test based on seeing a `TimeoutError` but to fail it if some other error occurs (e.g. if I assert something and an `AssertionError` gets raised)? Something for me to just look up but if anyone knows and can save me from doing so, please do :)

thedrow: Mark it as `@flaky`.

maybe-sybr: Oh, it doesn't need to be flaky I think. The behaviour is guaranteed AFAICT, it would just be useful for me to do something like:
```python
@pytest.mark.xfail(if_raised=TimeoutError)
def the_test():
  with pytest.raises(TImeoutError) as expected_exc:
    broken_thing()
  assert fallback_thing() == "working"
  raise expected_exc
```

maybe-sybr:
pypy3 failure is that bool/event confusion again. Unrelated to this diff AFAICT

maybe-sybr:
@mjhoffman65 if you have time to look over this changeset wrt evolution from the changes we added in #6576 which were needed for executing tasks from generators, it'd be appreciated :)

celery/canvas.py
thedrow: asserts are already removed when `not __debug__`.

There's no reason to execute the `if chord_obj is not None and next_task is None` condition and there's no reason to declare or to add group ids to the set other than debugging purposes as far as I can see.

maybe-sybr: Oh, I see what you mean. We do need to execute that condition to check that we're actually at the final task in the iterable, because we don't set the actual chord size in the backend until we've exhausted the iterable created from `enumerate(lookahead(tasks))`, which is indicated by `next_task is None` and done when that conditional is truthy.

As a note, I have NFI if there's a race condition there with the chord being completed early because we don't set the final size until some later point and perhaps tasks which this code fires off start landing. The potential issue crossed my mind but I haven't had time to think about it much and this PR is mostly aimed to avoid the chord size mis-calculation.

I put the group IDs in a set mostly because it felt like a low cost sanity check. I'll probably just remove it rather than committing a bunch of `if __debug__` conditionals. Thoughts?

thedrow: It is a low-cost sanity check if there aren't many groups.
Since we can't really expect what our users will do with the code, we shouldn't assume how large a canvas is.

maybe-sybr: I'll just trim it out then, and push an amended changeset shortly. :)

maybe-sybr: This has been removed now

maybe-sybr:
Labelling this as a blocker since it's needed to fix some of the behaviours tested in #6746

maybe-sybr:
I think this should be good to go now, @thedrow

thedrow (approved):

Issue #6738, opened 2021-04-23
“Load task modules from all registered Django app configs”


aruseni:
Hey, I think this phrase should be changed.

https://docs.celeryproject.org/en/stable/django/first-steps-with-django.html

```
# Load task modules from all registered Django app configs.
app.autodiscover_tasks()
```

We actually load the “tasks” modules from the apps themselves.

In other words, app configs do not have “tasks” modules, apps do.

So it would probably be better to say precisely that:

_Load task modules from all registered Django apps._

auvipy:
let's change it. need not to open issue for this type of tiny changes in docs

aruseni:
Sure. Just wanted any of you guys to confirm. Thanks.

PR #6743, opened 2021-04-24
Minor changes to the django/proj/celery.py example

aruseni:
See #6738 for an explanation why it’s better to say “registered apps”.


auvipy (approved):
looks OK to me

PR #6745, opened 2021-04-26
Update periodic-tasks.rst : add "add" task in the example 

jenhaoyang:
add "add" task in the example


thedrow:
Can you please explain why do we need this change?

jenhaoyang:
Because the config below use "task":"tasks.add".
I think this will more smooth for reader to use this example.

thedrow (approved):

PR #6754, opened 2021-05-04
fix child task inherit parent priority

supplayer:
fix child tasks cannot inherit the priority of the parent task.

## Description
In addition to the other priority-related PR #5313. 
fix child tasks cannot inherit the priority of the parent task. 
It has been tested in production and seems to work as designed.

### DIFF
--- a/celery/app/base.py
+++ b/celery/app/base.py
@@ -731,8 +731,7 @@ class Celery:
                     parent_id = parent.request.id
 
                 if conf.task_inherit_parent_priority:
-                    options.setdefault('priority',
-                                       parent.request.delivery_info.get('priority'))
+                    options.update({'priority': parent.request.delivery_info.get('priority')})
 
         message = amqp.create_task_message(
             task_id, name, args, kwargs, countdown, eta, group_id, group_index,


auvipy (changes requested):
could you please add a unit test for the suggested change?

thedrow:
> 
> 
> could you please add a unit test for the suggested change?

Before that, let's find out why this is needed.

@supplayer What's wrong with setting the default? I don't really understand why you'd always want to override the priority.
The current code inherits the priority only if it was not set explicitly.

supplayer:
> could you please add a unit test for the suggested change?
I think we can reuse the unit test of #5313 because the new fix code was revised on Mr madprogrammer's code basis.
Do I need to write another unit test?

supplayer:
> > could you please add a unit test for the suggested change?
> 
> Before that, let's find out why this is needed.
> 
> @supplayer What's wrong with setting the default? I don't really understand why you'd always want to override the priority.
> The current code inherits the priority only if it was not set explicitly.

I set task_inherit_parent_priority=True, task_default_priority=3, task_queue_max_priority=10 in Celery.conf. when i set parent priority is 5, the child task no inherit parent priority is still 3. Or where i lost the setting.

supplayer:
> > could you please add a unit test for the suggested change?
> 
> Before that, let's find out why this is needed.
> 
> @supplayer What's wrong with setting the default? I don't really understand why you'd always want to override the priority.
> The current code inherits the priority only if it was not set explicitly.

```
# tasks.py
@app.task(queue='test.task1', name='celeryapp.task1')
def task1(msg):
    print(f"task1 priority: {task1.request.delivery_info.get('priority')}")
    task2.apply_async((msg,), priority=5)

@app.task(queue='test.task2', name='celeryapp.task2')
def task2(msg):
    print(f"task2 priority: {task2.request.delivery_info.get('priority')}")
    task3.delay(msg)

@app.task(queue='test.task3', name='celeryapp.task3')
def task3(msg):
    print(f"task3 priority: {task3.request.delivery_info.get('priority')}")
    print(msg)

# printout
[2021-05-05 15:19:31,624: WARNING/ForkPoolWorker-1] task1 priority: 3
[2021-05-05 15:19:31,665: WARNING/ForkPoolWorker-1] task2 priority: 5
[2021-05-05 15:19:31,700: WARNING/ForkPoolWorker-1] task3 priority: 3
[2021-05-05 15:19:31,700: WARNING/ForkPoolWorker-1] test
```

thedrow:
The setting doesn't work this way. It only applies for canvases as far as I recall and that was the original intention.

thedrow:
I'll write an integration test case for this.

supplayer:
> I'll write an integration test case for this.

Thank you very much

PR #6770, opened 2021-05-14
fix: Preserve call/errbacks of replaced tasks

### LINKED ISSUE #6441: Errback inherited from an encapsulating chain is dropped when replacing tasks
# Checklist

  on reporting bugs.
  for existing proposed fixes.

     version affected).
  to reproduce this bug.

  and/or implementation.
  result backend.
  broker and/or result backend.

#### Related Issues

- None

#### Possible Duplicates

- None

## Environment & Settings

**Celery version**:

<details>
<p>

```
```

</p>
</details>

# Steps to Reproduce

## Required Dependencies

### Python Packages

<details>
<p>

```
```

</p>
</details>

### Other Dependencies

<details>
<p>
N/A
</p>
</details>

<details>
<p>
Swap the commented line setting `c` for expected behaviour.

```python
import celery

app = celery.Celery(broker="redis://", backend="redis://")

@app.task
def nop(*_):
    pass

@app.task
def die(*_):
    raise RuntimeError

@app.task(bind=True)
def replace(self, with_):
    with_ = celery.Signature.from_dict(with_)
    raise self.replace(with_)

@app.task
def cb(*args):
    print("CALLBACK", *args)

#c = celery.chain(nop.s(), die.s())
c = celery.chain(nop.s(), replace.si(die.s()))
c.link_error(cb.s())
c.apply_async()
```

</p>
</details>

# Expected Behavior
`cb` should be called as a new-style errback because it accepts starargs

# Actual Behavior
`cb` is not called

maybe-sybr:
## Description

This change fixes the handling of callbacks and errbacks on tasks which get replaced to
ensure they get called as expected if the replacement signature succeeds or fails.

Fixes #6441

### DIFF
--- a/celery/app/task.py
+++ b/celery/app/task.py
@@ -6,9 +6,9 @@ from kombu import serialization
 from kombu.exceptions import OperationalError
 from kombu.utils.uuid import uuid
 
-from celery import current_app, group, states
+from celery import current_app, states
 from celery._state import _task_stack
-from celery.canvas import _chain, signature
+from celery.canvas import _chain, group, signature
 from celery.exceptions import (Ignore, ImproperlyConfigured,
                                MaxRetriesExceededError, Reject, Retry)
 from celery.local import class_property
@@ -893,39 +893,40 @@ class Task:
             raise ImproperlyConfigured(
                 "A signature replacing a task must not be part of a chord"
             )
+        if isinstance(sig, _chain) and not getattr(sig, "tasks", True):
+            raise ImproperlyConfigured("Cannot replace with an empty chain")
 
+        # Ensure callbacks or errbacks from the replaced signature are retained
         if isinstance(sig, group):
-            sig |= self.app.tasks['celery.accumulate'].s(index=0).set(
-                link=self.request.callbacks,
-                link_error=self.request.errbacks,
-            )
-        elif isinstance(sig, _chain):
-            if not sig.tasks:
-                raise ImproperlyConfigured(
-                    "Cannot replace with an empty chain"
-                )
-
-        if self.request.chain:
-            if "link" in sig.options:
-                final_task_links = sig.tasks[-1].options.setdefault("link", [])
-                final_task_links.extend(maybe_list(sig.options["link"]))
-            # Construct the new remainder of the task by chaining the signature
-            # we're being replaced by with signatures constructed from the
-            # chain elements in the current request.
-            for t in reversed(self.request.chain):
-                sig |= signature(t, app=self.app)
-
+            # Groups get uplifted to a chord so that we can link onto the body
+            sig |= self.app.tasks['celery.accumulate'].s(index=0)
+        for callback in maybe_list(self.request.callbacks) or []:
+            sig.link(callback)
+        for errback in maybe_list(self.request.errbacks) or []:
+            sig.link_error(errback)
+        # If the replacement signature is a chain, we need to push callbacks
+        # down to the final task so they run at the right time even if we
+        # proceed to link further tasks from the original request below
+        if isinstance(sig, _chain) and "link" in sig.options:
+            final_task_links = sig.tasks[-1].options.setdefault("link", [])
+            final_task_links.extend(maybe_list(sig.options["link"]))
+        # We need to freeze the replacement signature with the current task's
+        # ID to ensure that we don't disassociate it from the existing task IDs
+        # which would break previously constructed results objects.
+        sig.freeze(self.request.id)
+        # Ensure the important options from the original signature are retained
         sig.set(
             chord=chord,
             group_id=self.request.group,
             group_index=self.request.group_index,
             root_id=self.request.root_id,
         )
-        # We need to freeze the new signature with the current task's ID to
-        # ensure that we don't disassociate the new signature from the existing
-        # task IDs which would break previously constructed results objects.
-        sig.freeze(self.request.id)
-
+        # If the task being replaced is part of a chain, we need to re-create
+        # it with the replacement signature - these subsequent tasks will
+        # retain their original task IDs as well
+        for t in reversed(self.request.chain or []):
+            sig |= signature(t, app=self.app)
+        # Finally, either apply or delay the new signature!
         if self.request.is_eager:
             return sig.apply().get()
         else:
--- a/celery/app/task.py
+++ b/celery/app/task.py
@@ -906,11 +906,6 @@ class Task:
                 )
 
         if self.request.chain:
-            # We need to freeze the new signature with the current task's ID to
-            # ensure that we don't disassociate the new chain from the existing
-            # task IDs which would break previously constructed results
-            # objects.
-            sig.freeze(self.request.id)
             if "link" in sig.options:
                 final_task_links = sig.tasks[-1].options.setdefault("link", [])
                 final_task_links.extend(maybe_list(sig.options["link"]))
@@ -926,6 +921,9 @@ class Task:
             group_index=self.request.group_index,
             root_id=self.request.root_id,
         )
+        # We need to freeze the new signature with the current task's ID to
+        # ensure that we don't disassociate the new signature from the existing
+        # task IDs which would break previously constructed results objects.
         sig.freeze(self.request.id)
 
         if self.request.is_eager:
--- a/celery/canvas.py
+++ b/celery/canvas.py
@@ -642,7 +642,8 @@ class _chain(Signature):
 
     def run(self, args=None, kwargs=None, group_id=None, chord=None,
             task_id=None, link=None, link_error=None, publisher=None,
-            producer=None, root_id=None, parent_id=None, app=None, **options):
+            producer=None, root_id=None, parent_id=None, app=None,
+            group_index=None, **options):
         # pylint: disable=redefined-outer-name
         #   XXX chord is also a class in outer scope.
         args = args if args else ()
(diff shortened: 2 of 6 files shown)


maybe-sybr:
This is a bugfix but I think we can land it on master once we sort out 5.1 since it's been broken for a while. :) It'll improve my life but FWICT there are no issues from other users tracking this misbehaviour.

celery/app/task.py
```diff
@@ -906,11 +906,6 @@ def replace(self, sig):
                 )
 
         if self.request.chain:
-            # We need to freeze the new signature with the current task's ID to
-            # ensure that we don't disassociate the new chain from the existing
-            # task IDs which would break previously constructed results
-            # objects.
-            sig.freeze(self.request.id)
```
thedrow, line 913: This may break previous behavior.

We used to have sig.freeze() in #6189 (See https://github.com/celery/celery/pull/6189/files#diff-db604dd7cb51e386710260ff2eba378aac19ba11eec97904bbf097b68caeada6R905) and it didn't fix the problem.

celery/canvas.py
```diff
@@ -1072,11 +1072,8 @@ def skew(self, start=1.0, stop=None, step=1.0):
     def apply_async(self, args=None, kwargs=None, add_to_parent=True,
                     producer=None, link=None, link_error=None, **options):
         args = args if args else ()
-        if link is not None:
-            raise TypeError('Cannot add link to group: use a chord')
-        if link_error is not None:
-            raise TypeError(
-                'Cannot add link to group: do that on individual tasks')
```
thedrow, line 1079: Why change the exception message?

thedrow (changes requested):
I'll review the integration tests later on.
Please check the build log as we have a genuine test failure, I believe.

celery/app/task.py
maybe-sybr: I saw the `sig.freeze(self.request.id)` down on L929/932 which is why I removed this one assuming it was a duplicate.

I do see that was also the case when you added the line removed here, back in #6189. My comment at [0] suggested the extra `sig.freeze(self.request.id)` as a fix for the `test_chain_replaced_with_a_chain_and_an_error_callback()` test which was failing prior to it being added. I'll take some time to run through that PR and this one to see if this is truly a duplicate line or if there's something subtle (perhaps the ordering of the `sig.freeze` and the later `sig.set()` on L923?).

[0] https://github.com/celery/celery/pull/6189#pullrequestreview-511455254

celery/canvas.py
maybe-sybr: No real reason - I can change the text back. The actual change here is to falsey checks rather than `None` checks since linking an empty iterable of tasks does nothing anyway so we can probably allow it. That said, doing so could allow weird behaviours to go unnoticed or introduce "sometimes it breaks" conditions (when there is a link/link_error) which could be annoying.

Happy to pull this change out of the diff. It's incidental and was only added because I had an unconditional `self.request.callbacks or []` in `task.py` before I decided that wasn't very nice.

maybe-sybr:
`t/unit/tasks/test_tasks.py::test_tasks::test_replace_callback()` is failing because I'm linking the callbacks and errbacks of the replaced signature to a replacement group even though they should should only be linked to the accumulate task that we chain onto the end of the group signature. Since a callback/errback was being set for a group, when that group was eventually applied it would die due to the `TypeError()` we were commenting on above:

```
self = group([t.unit.tasks.test_tasks.return_True()]), args = (), kwargs = None, add_to_parent = True, producer = None
link = ['callbacks'], link_error = ['errbacks']
options = {'group_id': 'group', 'group_index': None, 'root_id': 'root_id', 'task_id': '739d4d46-a741-4db5-aa5b-eacdb75de6bf'}

    def apply_async(self, args=None, kwargs=None, add_to_parent=True,
                    producer=None, link=None, link_error=None, **options):
        args = args if args else ()
        if link is not None or link_error is not None:
>           raise TypeError('Cannot add linked tasks to groups: use a chord')
E           TypeError: Cannot add linked tasks to groups: use a chord

celery/canvas.py:1076: TypeError
```

I've changed the logic in `replace()` to only link call/errbacks to the replacement signature if it's not a group. I'll push a new diff once I've done some more sanity checking on the integration tests.

celery/canvas.py
thedrow: I'd rather not break our public API (and exceptions are public API).
If you'd like, we can raise something more specific that inherits from a TypeError.

maybe-sybr: Sure thing, I'll just strip it out since it's not a particularly useful change. Will resolve once I push a new diff.

maybe-sybr:
I've reworked the replacement logic entirely now. It could do with a re-review. AFAICT, removing the early `replacement_sig.freeze(self.request.id)` hasn't disrupted any of our unit or integration tests (against redis) so I'm reasonably confident that the diff is fine to merge.

celery/app/task.py
maybe-sybr: Per below, I've run through the behaviour of this a bit more and can't really explain why we had the early `sig.freeze()` call here. The test I mentioned in the comment above still runs and succeeds, so either it wasn't really necessary or whatever subtlety previously caused it to be required is now changed/gone. The new reworked diff continues to only `freeze()` the original (to be replaced) task ID into the new signature after everything else is done, and the canvas integration tests appear to function as they do on master (a couple of failures with the retries tests :/).

I plan to come back and take another look over this with fresh eyes to make sure I'm not missing anything, but for now I think it's OK.

maybe-sybr: Scratch that, I've spent some time dumping task IDs again and have come to a more sensible version of the same conclusion I reached last time. `freeze()`ing the signature before we chain any others onto it ensures that the signature we're going to use to replace the original one with will have its ID set correctly, which ensures that the entire chain of parent tasks can be evaluated by the original result object created when the tasks were first applied.

It actually seems to be the second `freeze()` which is redundant, since by that time (when there are tasks to chain onto the replacement signature) `sig` is a chain signature with a final task which has already retained its original task ID. However, that second `freeze()` was necessary because in situation where there were no other tasks to chain onto the replacement signature, the first `freeze()` would not have been called. So the first `freeze()` was critical or it wasn't called, and the second was a no-op or it was critical.

Despite this, the result code (at least for redis) was resilient enough that it was passing the tests anyway!

The new code structure for `replace()` ensures that we only call `freeze()` once and it happens before we chain anything onto the replacement signature. This should be a fine solution. However, I'd like to work out some integration tests which will capture any regressions here since it's difficult to reason about when things seem to work fine despite being subtly broken.

maybe-sybr: The new diff handles this properly now and there are some tests which catch regression if I commentout the single `sig.freeze()` which is done now. I'll leave this or you to resolve if you're happy with it.

maybe-sybr:
This now behaves as I expect and we have some new integration tests to catch regressions or misbehaviours. That actually exposed an issue with chord header result ordering which I've also fixed in the commit on the tip of this branch.

celery/app/task.py
```diff
@@ -893,41 +893,40 @@ def replace(self, sig):
...
-                final_task_links.extend(maybe_list(sig.options["link"]))
-            # Construct the new remainder of the task by chaining the signature
-            # we're being replaced by with signatures constructed from the
-            # chain elements in the current request.
-            for t in reversed(self.request.chain):
-                sig |= signature(t, app=self.app)
-
+            # Groups get uplifted to a chord so that we can link onto the body
+            sig |= self.app.tasks['celery.accumulate'].s(index=0)
+        for callback in self.request.callbacks or []:
```
maybe-sybr, line 903: I think I just hit a condition where this is a simple signature rather than a list when using an editable install on this branch in some of my own application code. This line and L905 might be suspect >:(

t/integration/test_canvas.py
```diff
@@ -1784,61 +2038,133 @@ def test_errback_called_by_chord_from_group_fail_multiple(
...
+            raise pytest.skip(e.args[0])
+
+        orig_sig = chord(
+            (identity.s(42), replace_with_chain.s(1337), ),
+            identity.s(),
+        )
+        res_obj = orig_sig.delay()
+        assert res_obj.get(timeout=TIMEOUT) == [42, 1337]
+
+    def test_chord_body_task_replaced_with_chain(self):
```
maybe-sybr, line 2117: Urgh, these need the `manager` fixture. ^c^v fail

celery/app/task.py
maybe-sybr: I'm making them read `maybe_list(self.request.{call,err}backs) or []` which will handle the case where either of those attrs are a single signature/dict, a list of such objects, or `None` (since `maybe_list(None)` returns `None`).

maybe-sybr:
That last fixup broke a unit test, it's fixed again now.

celery/app/task.py
```diff
@@ -893,41 +893,40 @@ def replace(self, sig):
             raise ImproperlyConfigured(
                 "A signature replacing a task must not be part of a chord"
             )
+        if isinstance(sig, _chain) and not getattr(sig, "tasks", True):
+            raise ImproperlyConfigured("Cannot replace with an empty chain")
```
thedrow, line 897: Didn't I write a test for this?

thedrow (dismissed):
LGTM.

This code is much cleaner than before.

maybe-sybr:
> LGTM.
> 
> This code is much cleaner than before.

Great thanks :) I've just marked this for 5.2 so we can come back and merge it once we drop 5.1

maybe-sybr:
I'll mark this as needs rebase to remind me to rebase on top of 5.1 as prep for merge

auvipy (changes requested):
rebase please. as it's a bug fix we can land this on 5.1.x as well

auvipy (changes requested):
rebase please. as it's a bug fix we can land this on 5.1.x as well

maybe-sybr:
> rebase please. as it's a bug fix we can land this on 5.1.x as well

rebased on master

auvipy (approved):

PR #6779, opened 2021-05-23
Prepared Hacker News Post on Release Action

MicahLyle:
@thedrow and/or @auvipy 

`HN_USERNAME` and `HN_PASSWORD` should be provided in the secrets. I've tested this live and was able to verify that it works. However, if you'd like to do one test run first,  you can keep `HN_TEST_MODE` as `true`. Otherwise, if you're fully ready to go once the secrets are in, they feel free to set `HN_TEST_MODE` and maybe we can use this for the first time on the 5.1.0 release!

### DIFF
--- a/.github/workflows/post_release_to_hacker_news.yml
+++ b/.github/workflows/post_release_to_hacker_news.yml
@@ -1,0 +1,17 @@
+on:
+  release:
+    types: [released]
+
+jobs:
+  post_release_to_hacker_news:
+    runs-on: ubuntu-latest
+    name: Post Release to Hacker News
+    steps:
+      - name: Post the Release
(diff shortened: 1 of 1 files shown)


MicahLyle:
I'd be happy to give/transfer my repository to the Celery project.

And/or, I also could publish it as a public action in the GitHub Actions marketplace, however, it's kind of customized for Celery's use case currently, and I'd rather work on other Celery-related stuff right now than make the action more flexible.

I'm also less experienced with licensing right now so I made my repository an MIT license, but if you need it to be anything else just let me know!

thedrow:
I think we can merge this after 5.1 is released since that's going to happen now. :)

auvipy (approved):

thedrow:
I'll need to create an API key for it.

Issue #4834, opened 2018-06-21
Celery 4.2.0 - 5.6.0: Exception in group nested in double chain is raising internal error


kstachniuk:
## Checklist
celery -A proj report

software -> celery:4.2.0 (windowlicker) kombu:4.2.0 py:3.5.2
            billiard:3.5.0.3 py-amqp:2.3.2 
platform -> system:Linux arch:64bit, ELF imp:CPython
loader   -> celery.loaders.app.AppLoader
settings -> transport:amqp results:redis:///

## Steps to reproduce
Create a group with failing task. Group should be nested in chain, which will be in another chain.

```
@shared_task
def my_failing_task(raise_error):
    if raise_error:
        raise Exception("Raising exception")
    return "OK my_failing_task"

@shared_task
def my_task_result(p1):
    logger.info("my_task_result: {}".format(p1))
    return "OK my_task_result"

@shared_task
def on_error_task(*args, **kwargs):
    logger.info("on_error_task: {}, {}".format(args, kwargs))
    return "OK on_error_task"

def create_nested_chain_with_group():
    group_of_tasks = group(my_failing_task.si(raise_error=False), my_failing_task.si(raise_error=True))
    return chain(group_of_tasks, my_task_result.s())

def run():
    second_chain = chain(create_nested_chain_with_group(),
                         my_task_result.s())
    second_chain.link_error(on_error_task.s("Error caught by second_chain"))
    second_chain()
```

This as expected will raise Exception('Raising exception',)

## Expected behavior
`` on_error_task `` should be run and log which task has failed

## Actual behavior
`` on_error_task `` is not runned and whole task crashes with below exception:

```
/libs/celery/app/trace.py:561: RuntimeWarning: Exception raised outside body: TypeError('sequence item 1: expected a bytes-like object, NoneType found',):
Traceback (most recent call last):
  File "/libs/celery/backends/redis.py", line 368, in on_chord_part_return
    callback.delay([unpack(tup, decode) for tup in resl])
  File "/libs/celery/backends/redis.py", line 368, in <listcomp>
    callback.delay([unpack(tup, decode) for tup in resl])
  File "/libs/celery/backends/redis.py", line 323, in _unpack_chord_result
    raise ChordError('Dependency {0} raised {1!r}'.format(tid, retval))
celery.exceptions.ChordError: Dependency 3b405d80-90df-4426-8782-c8002ba3000d raised Exception('Raising exception',)

During handling of the above exception, another exception occurred:

[...]

PR #6784, opened 2021-05-25
Make `ResultSet.on_ready` promise hold a weakref to self

thedrow:

## Description

This avoids a memory leak since the promise creates a reference cycle to `self`.
See https://github.com/celery/celery/issues/4843#issuecomment-847634277.

### DIFF
--- a/celery/result.py
+++ b/celery/result.py
@@ -2,6 +2,7 @@
 
 import datetime
 import time
+from weakref import proxy
 from collections import deque
 from contextlib import contextmanager
 
@@ -535,7 +536,7 @@ class ResultSet(ResultBase):
     def __init__(self, results, app=None, ready_barrier=None, **kwargs):
         self._app = app
         self.results = results
-        self.on_ready = promise(args=(self,))
+        self.on_ready = promise(args=(proxy(self),))
         self._on_full = ready_barrier or barrier(results)
         if self._on_full:
             self._on_full.then(promise(self._on_ready, weak=True))


maybe-sybr:
This seems like it would help when result sets get thrown on the floor but not GCed, but we'll need to make some corresponding change to call the ref to actually get `self` back when `on_ready` is fulfilled prior to the provided callback getting called. Maybe we could use a caching wrapping helper so that we can avoid wrapping the same callback(s) over and over in larger scale environments? IDK how necessary that'd be, maybe wrapping every callback we get passed is OK

thedrow:
Who appends callbacks to `on_ready` anyway?

maybe-sybr:
> Who appends callbacks to `on_ready` anyway?

The change you mentioned on #4834 (6c78582) suggests that `group().delay().then()` or something like that was broken and someone noticed. Whether that someone was @ask or a user would be interesting to know but I suspect it might require some archaeology or be lost to time.

xirdneh:
Do we need to keep a reference to `self`?
Wouldn't it be safer to create a new object in `_on_ready` when it's called based on the same results?
I mean, if this is called async why do we need to keep the same object?

```python
def _on_ready(self):
    if self.backend.is_async:
        ResultSet(self.results, self.app, self.ready_barrier)

def then(self, callback, on_error=None, weak=False):
    return promise(
        ResultSet(self.results, self.app, self.ready_barrier)
    ).then(callback, on_error)
```
*note*: the previous code is only a proof of concept, I'm not sure it works.

And, why does `def _on_ready` doesn't return anything? Shouldn't it return the resolved promise

thedrow:
> 
> 
> Do we need to keep a reference to `self`?
> Wouldn't it be safer to create a new object in `_on_ready` when it's called based on the same results?
> I mean, if this is called async why do we need to keep the same object?
> 
> ```python
> def _on_ready(self):
>     if self.backend.is_async:
>         ResultSet(self.results, self.app, self.ready_barrier)
> 
> def then(self, callback, on_error=None, weak=False):
>     return promise(
>         ResultSet(self.results, self.app, self.ready_barrier)
>     ).then(callback, on_error)
> ```
> 
> _note_: the previous code is only a proof of concept, I'm not sure it works.
> 
> And, why does `def _on_ready` doesn't return anything? Shouldn't it return the resolved promise

These are very good questions.
I don't have any answers though and because of that, I'd rather be conservative.

thedrow:
> 
> 
> This seems like it would help when result sets get thrown on the floor but not GCed, but we'll need to make some corresponding change to call the ref to actually get `self` back when `on_ready` is fulfilled prior to the provided callback getting called. Maybe we could use a caching wrapping helper so that we can avoid wrapping the same callback(s) over and over in larger scale environments? IDK how necessary that'd be, maybe wrapping every callback we get passed is OK

@maybe-sybr I'm trying this with `weakref.proxy` now. Let's see what happens.

thedrow:
Codecov is wrong and everything is fine :/

maybe-sybr (approved):
Oh, I had no idea `weakref.proxy` existed. That's really useful. LGTM!

Issue #6786, opened 2021-05-26
regen() duplicates any non-list and non-tuple sequence


Suor:
# Checklist

- [x] I have verified that the issue exists against the `master` branch of Celery.
- [ ] This has already been asked to the [discussion group](https://groups.google.com/forum/#!forum/celery-users) first.
- [ ] I have read the relevant section in the
  [contribution guide](http://docs.celeryproject.org/en/latest/contributing.html#other-bugs)
  on reporting bugs.
- [x] I have checked the [issues list](https://github.com/celery/celery/issues?q=is%3Aissue+label%3A%22Issue+Type%3A+Bug+Report%22+-label%3A%22Category%3A+Documentation%22)
  for similar or identical bug reports.
- [x] I have checked the [pull requests list](https://github.com/celery/celery/pulls?q=is%3Apr+label%3A%22PR+Type%3A+Bugfix%22+-label%3A%22Category%3A+Documentation%22)
  for existing proposed fixes.
- [x] I have checked the [commit log](https://github.com/celery/celery/commits/master)
  to find out if the bug was already fixed in the master branch.
- [x] I have included all related issues and possible duplicate issues
  in this issue (If there are none, check this box anyway).

## Mandatory Debugging Information

- [ ] I have included the output of ``celery -A proj report`` in the issue.
    (if you are not able to do this, then at least specify the Celery
     version affected).
- [x] I have verified that the issue exists against the `master` branch of Celery.
- [x] I have included the contents of ``pip freeze`` in the issue.
- [x] I have included all the versions of all the external dependencies required
  to reproduce this bug.

## Optional Debugging Information

- [x] I have tried reproducing the issue on more than one Python version
  and/or implementation.
- [ ] I have tried reproducing the issue on more than one message broker and/or
  result backend.
- [ ] I have tried reproducing the issue on more than one version of the message
  broker and/or result backend.
- [ ] I have tried reproducing the issue on more than one operating system.
- [ ] I have tried reproducing the issue on more than one workers pool.
- [ ] I have tried reproducing the issue with autoscaling, retries,
  ETA/Countdown & rate limits disabled.
- [x] I have tried reproducing the issue after downgrading
  and/or upgrading Celery and its dependencies.

#### Related Issues

- None

#### Possible Duplicates

- None

## Environment & Settings

**Celery version**:
5.1.0 (sun-harmonics)

<details>
<summary><b><code>celery report</code> Output:</b></summary>
<p>

```
```

</p>
</details>

[...]

PR #6789, opened 2021-05-28
Ensure regen utility class gets marked as done when concretised

maybe-sybr:
## Description

Ensure `regen()` instances get marked as done when concretised via the
`UserList.data` property implementation. We also improve laziness of
`regen.__repr__()` so it reflects the fact that it's lazy.

Fixes: #6786

### DIFF
--- a/celery/utils/functional.py
+++ b/celery/utils/functional.py
@@ -249,6 +249,13 @@ class _regen(UserList, list):
             pass
         return self.__consumed
 
+    def __repr__(self):
+        return "<{}: [{}{}]>".format(
+            self.__class__.__name__,
+            ", ".join(repr(e) for e in self.__consumed),
+            "..." if not self.__done else "",
+        )
+
 
 def _argsfromspec(spec, replace_defaults=True):
     if spec.defaults:
--- a/celery/utils/functional.py
+++ b/celery/utils/functional.py
@@ -241,12 +241,9 @@ class _regen(UserList, list):
 
     @property
     def data(self):
-        try:
-            if not self.__done:
-                self.__consumed.extend(list(self.__it))
+        if not self.__done:
+            self.__consumed.extend(self.__it)
             self.__done = True
-        except StopIteration:
-            pass
         return self.__consumed
 
     def __repr__(self):
--- a/celery/utils/functional.py
+++ b/celery/utils/functional.py
@@ -242,7 +242,9 @@ class _regen(UserList, list):
     @property
     def data(self):
         try:
-            self.__consumed.extend(list(self.__it))
+            if not self.__done:
+                self.__consumed.extend(list(self.__it))
+            self.__done = True
         except StopIteration:
             pass
         return self.__consumed
--- a/t/unit/utils/test_functional.py
+++ b/t/unit/utils/test_functional.py
@@ -170,6 +170,42 @@ class test_regen:
         g = regen(die())
         assert "..." in repr(g)
 
+    def test_partial_reconcretisation(self):
+        class WeirdIterator():
+            def __init__(self, iter_):
+                self.iter_ = iter_
+                self._errored = False
+
+            def __iter__(self):
+                yield from self.iter_
+                if not self._errored:
+                    try:
+                        # This should stop the regen instance from marking
+                        # itself as being done
+                        raise AssertionError("Iterator errored")
+                    finally:
+                        self._errored = True
+
+        original_list = list(range(42))
+        g = regen(WeirdIterator(original_list))
+        iter_g = iter(g)
+        for e in original_list:
+            assert e == next(iter_g)
+        with pytest.raises(AssertionError, match="Iterator errored"):
+            next(iter_g)
+        # The following checks are for the known "misbehaviour"
+        assert getattr(g, "_regen__done") is False
+        # If the `regen()` instance doesn't think it's done then it'll dupe the
+        # elements from the underlying iterator if it can be re-used
+        iter_g = iter(g)
+        for e in original_list * 2:
+            assert next(iter_g) == e
+        with pytest.raises(StopIteration):
+            next(iter_g)
+        assert getattr(g, "_regen__done") is True
+        # Finally we xfail this test to keep track of it
+        raise pytest.xfail(reason="#6794")
+
 
 class test_head_from_fun:
 
--- a/t/unit/utils/test_functional.py
+++ b/t/unit/utils/test_functional.py
@@ -161,6 +161,15 @@ class test_regen:
         assert g == original_list
         assert g == original_list
 
+    def test_repr(self):
+        def die():
+            raise AssertionError("Generator died")
+            yield None
+
+        # Confirm that `regen()` instances are not concretised when represented
+        g = regen(die())
+        assert "..." in repr(g)
+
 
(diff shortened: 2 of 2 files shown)


t/unit/utils/test_functional.py
```diff
@@ -150,6 +152,56 @@ def build_generator():
...
+    def test_repr(self):
+        def die():
+            raise AssertionError("Generator died")
+            yield None
+
+        # Confirm that `regen()` instances are not concretised when represented
+        g = regen(die())
+        assert "..." in repr(g)
+
+    @pytest.mark.xfail(raises=StopIteration, reason="XXX NEEDS AN ISSUE")
```
maybe-sybr, line 173: This test is for an edge case where a repeatable iterator could raise an exception at some point before completion which would result in `regen()` stashing the elements it saw but never marking itself as done. Then the next time we yield elements from an iterator derived from the `regen()` instance, we'd yield the stashed elements, then repeat them from the underlying iterator if it's capable of giving them to us again.

IDK if this is something we can even bother handling. If the underlying iterator explodes, we could choose to mark the regen instance as `__done` and just give up on attempting to get more elements? If we want to go down that route, I can roll a change into this PR since it'd be pretty simple and still thematically relevant. If we think we need to do something more involved or optimistic, I'll make a follow up issue for discussion.

celery/utils/functional.py
```diff
@@ -242,11 +242,20 @@ def __bool__(self):
     @property
     def data(self):
         try:
-            self.__consumed.extend(list(self.__it))
+            if not self.__done:
+                self.__consumed.extend(list(self.__it))
+            self.__done = True
```
Suor, line 247: Might be movedunder `if`
```suggestion
                self.__done = True
```

maybe-sybr:
> Why do we have `except StopIteration` here at all? `list()` should handle that.

Good question. That was added in 99827730 and looks like it could have been deliberate, but I have no idea why that might be the case. AFAICT you'd have to be raising `StopIteration` from inside a generator (which is wrong) to need something like that :/ A quick sanity check of `list(iter(range(5)))` works without raising and, as expected, spits out `StopIteration` when I `next()` the iterator manually. Not sure what the point is there.

Removing the `try: except StopIteration:` doesn't seem to break any unit or integration tests (against redis), so I'll amend the diff and re-push shortly.

celery/utils/functional.py
maybe-sybr: The newly pushed diff makes this change due to the dedent of the rest of the try body

t/unit/utils/test_functional.py
```diff
@@ -150,6 +152,56 @@ def build_generator():
...
+                self._errored = False
+
+            def __iter__(self):
+                yield from self.iter_
+                if not self._errored:
+                    try:
+                        # This should stop the regen instance from marking
+                        # itself as being done
+                        #raise AssertionError("Iterator errored")
+                        pass
```
maybe-sybr, line 187: whoops, this needs to be uncommented

maybe-sybr: done

t/unit/utils/test_functional.py
maybe-sybr: I've made #6794 to track the discussion on this point. Could just be a thing we need to document, happy to have this PR merged without addressing that issue yet.

celery/utils/functional.py
```diff
@@ -241,12 +241,18 @@ def __bool__(self):
 
     @property
     def data(self):
-        try:
+        if not self.__done:
             self.__consumed.extend(list(self.__it))
```
Suor, line 245: An intermediate `list()` should not be needed. I.e. `self.__consumed.extend(self.__it)` should do.

Suor:
> AFAICT you'd have to be raising StopIteration from inside a generator (which is wrong) to need something like that

Since Python 3.7 raising `StopIteration` inside generator results in a `RuntimeError`.

thedrow:
> 
> 
> > AFAICT you'd have to be raising StopIteration from inside a generator (which is wrong) to need something like that
> 
> Since Python 3.7 raising `StopIteration` inside generator results in a `RuntimeError`.

This is probably an oversight we made during the Python 3 only migration in 5.0.
Thanks for noticing 😄

celery/utils/functional.py
maybe-sybr: changed

maybe-sybr:
I just pushed a change which removed what I thought was a pointless `iter(self)` but I now realise that it was to ensure that the `__done` attribute gets set when the underlying iterator is completed without having to do hard stuff in the `__getitem__()` method. I've stripped that commit out, but I still think it's wasteful to spend time iterating on the (potentially many) initial elements of ourself just to get some subsequent element from the underlying iterator.

Suor:
> I just pushed a change which removed what I thought was a pointless iter(self) but I now realise that it was to ensure that the __done attribute gets set when the underlying iterator is completed without having to do hard stuff in the __getitem__() method. I've stripped that commit out, but I still think it's wasteful to spend time iterating on the (potentially many) initial elements of ourself just to get some subsequent element from the underlying iterator.

I wonder how much use that `.__getitem__()` call gets. At least for `.map()` and friends we simply serialize, i.e. convert to list the thing eventually, we only need `regen()` to not consume immediately, since a signature might never get sent.

maybe-sybr:
> I wonder how much use that `.__getitem__()` call gets. At least for `.map()` and friends we simply serialize, i.e. convert to list the thing eventually, we only need `regen()` to not consume immediately, since a signature might never get sent.

I'd guess the only time it really gets called is to access the 0th or `-1`th item which are used for various things in the canvas code. Otherwise I'd expect it to be mostly simple `for x in some_regen_thing:` usage.

t/unit/utils/test_functional.py
```diff
@@ -150,6 +152,60 @@ def build_generator():
...
+        # If the `regen()` instance doesn't think it's done then it'll dupe the
+        # elements from the underlying iterator if it can be re-used
+        iter_g = iter(g)
+        for e in original_list * 2:
+            assert next(iter_g) == e
+        with pytest.raises(StopIteration):
+            next(iter_g)
+        assert getattr(g, "_regen__done") is True
+        # Finally we xfail this test to keep track of it
+        raise pytest.xfail(reason="#6794")
```
thedrow, line 207: Why do we raise this instead of marking it as xfail?
xfailed tests still run.

maybe-sybr: The test is verifying the incorrect behaviour in full before xfailing itself, rather than relying on some assertion failing partway through and marking it as an xfailure. I chose to do it that way because it's not clear to me how we should be handling situations like this so I can't write the "correct" behaviour at the moment. I'm sort of expecting the test to start hard failing if/when we work out how to treat something like this, or alternatively we simply leave it as an expected edge case and remove the `raise xfail()` on this line.

auvipy (approved):
let's merge it after some 5.1.x point releases

thedrow:
> 
> 
> let's merge it after some 5.1.x point releases

I think we can include it in 5.1.1 unless there's a special reason not to do so that I missed.

auvipy:
OK. merged as bugfix

PR #6799, opened 2021-06-03
improv: Make regen consume items more reliably

maybe-sybr:
## Description

This is a follow up to #6789 with some more significant changes to the `regen`
implementation to:
  * avoid re-iterating previously consumed elements in `__getitem__()`
  * mark the instance as done when a single lookahead stops iterating
  ** this fixes an edge case where a repeating underlying iterator could be duplicated if regen is not fully iterated
  ** element duplication from repeatable underlying iterables is not generally fixed since we still store the iterable not its iterator (IIUC)

### DIFF
--- a/celery/utils/functional.py
+++ b/celery/utils/functional.py
@@ -195,7 +195,6 @@ class _regen(UserList, list):
         # UserList creates a new list and sets .data, so we don't
         # want to call init here.
         self.__it = it
-        self.__index = 0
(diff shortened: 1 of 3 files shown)


maybe-sybr:
This is based on the diff currently up at #6789 - the substantive diff is just 08066a1

auvipy (approved):

PR #6805, opened 2021-06-07
fix: Typo in Tasks

danmichaelo:
Just noticed that it should be `super(MyRequest,`, not `super(Request,`.

While at it, I also added a `from celery import Task` to make the example self-contained.

### DIFF
--- a/docs/userguide/tasks.rst
+++ b/docs/userguide/tasks.rst
@@ -1605,6 +1605,7 @@ limits, and other failures.
 .. code-block:: python
 
    import logging
+   from celery import Task
    from celery.worker.request import Request
 
    logger = logging.getLogger('my.package')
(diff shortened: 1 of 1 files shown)


docs/userguide/tasks.rst
```diff
@@ -1621,7 +1622,7 @@ limits, and other failures.
               )
 
        def on_failure(self, exc_info, send_failed_event=True, return_ok=False):
-           super(Request, self).on_failure(
+           super(MyRequest, self).on_failure(
```
auvipy, line 1625: what about just super(). ?

thedrow: That's better!

thedrow, line 1625:
```suggestion
           super().on_failure(
```

PR #6806, opened 2021-06-08
test: Fix unexpected failure from bad mocking

maybe-sybr:
## Description

This test would attempt to mock the `request_stack` of a task so as to
confirm that it could confirm that the request object pushed onto it
contained simulated delivery information as expected. However, it did
not wrap the original call target which led to an unfortunate
interaction with the worker optimisations in `app/trace.py` which would
not find the request on the stack and therefore not end up calling the
task's `run()` method.

The worker optimisations can be enabled as a side effect of other tests
like `test_regression_worker_startup_info()` in the mongo and cache
backend suites. This led to a situation where the test changed in the
diff would fail if those tests happened to run before it! Luckily, the
side effect of the worker optimizations being enabled are not what cause
the unrelated failure, the test in this diff was a just a bit unaware of
the consequences of its mocking.

### DIFF
--- a/t/unit/tasks/test_tasks.py
+++ b/t/unit/tasks/test_tasks.py
@@ -1289,17 +1289,20 @@ class test_apply_task(TasksCase):
             f.get()
 
     def test_apply_simulates_delivery_info(self):
-        self.task_check_request_context.request_stack.push = Mock()
+        task_to_apply = self.task_check_request_context
+        with patch.object(
+            task_to_apply.request_stack, "push",
+            wraps=task_to_apply.request_stack.push,
+        ) as mock_push:
+            task_to_apply.apply(
+                priority=4,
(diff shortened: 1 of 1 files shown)


maybe-sybr:
To replicate the failure on master, you should be able to run:
```
$ tox -e 3.9-unit -- -k '(test_regression_worker_startup_info) or test_apply_simulates_delivery_info'
```
picking whatever toxenv fits your local python obviously :)

auvipy (approved):

auvipy:
can some more coverage be increased?

maybe-sybr:
Not sure why the coverage has changed, especially in the couchdb source. IMO it's unrelated to this PR and maybe just codecov getting a bit confused. Hard to be sure with everything apparently being broken because of Fastly right now :/

PR #6807, opened 2021-06-08
Add Python 3.10 support

thedrow:

## Description

We will support Python 3.10 starting from Celery 5.2.
We should start testing with it now to figure out what kind of problems we may have.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -17,7 +17,7 @@ python =
     3.7: 3.7-unit
     3.8: 3.8-unit
     3.9: 3.9-unit
-    3.10-beta.3: 3.10-unit
+    3.10: 3.10-unit
     pypy3: pypy3-unit
 
 [testenv]
--- a/tox.ini
+++ b/tox.ini
@@ -31,9 +31,8 @@ deps=
     -r{toxinidir}/requirements/test.txt
     -r{toxinidir}/requirements/pkgutils.txt
 
-    3.6,3.7,3.8,3.9: -r{toxinidir}/requirements/test-ci-default.txt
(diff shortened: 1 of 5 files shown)


auvipy (approved):
looks good, merge when 3.10 stable released

thedrow:
> 
> 
> looks good, merge when 3.10 stable released

This won't fail the build anyway starting from 7ff872b.
Is there any problem merging this just so we'll know what's the status of our Python 3.10 support?

maybe-sybr (approved):
lgtm, I reckon we merge and make a follow up issue to bump along the release cycle defined in PEP-0619
> https://www.python.org/dev/peps/pep-0619/

Keep in mind, @thedrow, that the release schedule for Python 3.10 will define our release schedule for 5.2 if we want to release with official support. Given our vaguely accelerating release cycle, I'd say we could alter that expectation slightly to keep the door open for a celery 5.2 before October with wording in the changelog about provisional 3.10 support which we could turn official in a 5.2.x release when the Python 3.10 rolls around. Depends on your views on the 5.2 release cycle timeline and what we want to land in it.

If @auvipy doesn't merge this in the next 8ish hours, I'll do it at the end of my work day.

tox.ini
```diff
@@ -1,7 +1,9 @@
 [tox]
+requires =
+    tox-gh-actions
```
maybe-sybr, line 3: Do we need this here since we install it in the actions yaml? No big deal, we can merge without an answer on this, it should be fine

thedrow: It's seems like a good idea to specify the deps for tox to work.

graingert: it's best to specify tox requires here

maybe-sybr: :+1:

PR #6814, opened 2021-06-16
fix: Calling of errbacks when chords fail

maybe-sybr:
## Description

We had a special case for calling errbacks when a chord failed which
assumed they were old style. This change ensures that we call the proper
errback dispatch method which understands new and old style errbacks,
and adds test to confirm that things behave as one might expect now.

### DIFF
--- a/celery/canvas.py
+++ b/celery/canvas.py
@@ -1123,18 +1123,17 @@ class group(Signature):
             task.set_immutable(immutable)
 
     def link(self, sig):
-        # Simply link to first task
+        # Simply link to first task. Doing this is slightly misleading because
+        # the callback may be executed before all children in the group are
+        # completed and also if any children other than the first one fail.
+        #
+        # The callback signature is cloned and made immutable since it the
+        # first task isn't actually capable of passing the return values of its
+        # siblings to the callback task.
         sig = sig.clone().set(immutable=True)
         return self.tasks[0].link(sig)
 
     def link_error(self, sig):
-        try:
-            sig = sig.clone().set(immutable=True)
-        except AttributeError:
-            # See issue #5265.  I don't use isinstance because current tests
-            # pass a Mock object as argument.
-            sig['immutable'] = True
-            sig = Signature.from_dict(sig)
         # Any child task might error so we need to ensure that they are all
         # capable of calling the linked error signature. This opens the
         # possibility that the task is called more than once but that's better
--- a/t/integration/test_canvas.py
+++ b/t/integration/test_canvas.py
@@ -3,7 +3,7 @@ import re
 import tempfile
 import uuid
 from datetime import datetime, timedelta
-from time import sleep
+from time import monotonic, sleep
 
 import pytest
 import pytest_subtests  # noqa: F401
@@ -18,8 +18,8 @@ from .conftest import (TEST_BACKEND, get_active_redis_channels,
                        get_redis_connection)
 from .tasks import (ExpectedException, add, add_chord_to_chord, add_replaced,
                     add_to_all, add_to_all_to_chord, build_chain_inside_task,
-                    chord_error, collect_ids, delayed_sum,
-                    delayed_sum_with_soft_guard, fail, fail_replaced,
+                    collect_ids, delayed_sum, delayed_sum_with_soft_guard,
+                    errback_new_style, errback_old_style, fail, fail_replaced,
                     identity, ids, print_unicode, raise_error, redis_count,
                     redis_echo, replace_with_chain,
                     replace_with_chain_which_raises, replace_with_empty_chain,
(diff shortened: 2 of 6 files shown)


t/integration/test_canvas.py
```diff
@@ -2020,7 +2167,45 @@ def test_errback_called_by_chord_from_group(self, manager, subtests):
...
+        child_sig = group(identity.si(42), fail_sig)
+
+        chord_sig = chord((child_sig, ), identity.s())
+        chord_sig.link_error(errback)
+        expected_redis_key = chord_sig.body.freeze().id
+        redis_connection.delete(expected_redis_key)
+        with subtests.test(msg="Error propagates from header group"):
+            res = chord_sig.delay()
+            with pytest.raises(ExpectedException):
+                res.get(timeout=TIMEOUT)
```
maybe-sybr, line 2192: This and the `res.get()` in the subtest below seem to sometimes time out and I can't work out why. It seems to only really occur when I run the full suite (redis backend) or run `tox -e ... -- -k test_canvas` to select the canvas integration tests.

t/unit/backends/test_base.py
```diff
@@ -534,17 +534,23 @@ def test_mark_as_revoked__chord(self):
...
 
     def test_chord_error_from_stack_raises(self):
+        class ExpectedException(Exception):
+            pass
+
         b = BaseBackend(app=self.app)
-        exc = KeyError()
         callback = Mock(name='callback')
         callback.options = {'link_error': []}
+        callback.keys.return_value = []
```
maybe-sybr, line 543: This mock doesn't work on pypy3 :(
```
___________ test_BaseBackend_dict.test_chord_error_from_stack_raises ___________

self = <t.unit.backends.test_base.test_BaseBackend_dict object at 0x000000000aad6e20>

    def test_chord_error_from_stack_raises(self):
        class ExpectedException(Exception):
            pass
    
        b = BaseBackend(app=self.app)
        callback = Mock(name='callback')
        callback.options = {'link_error': []}
        callback.keys.return_value = []
        task = self.app.tasks[callback.task] = Mock()
        b.fail_from_current_stack = Mock()
        group = self.patching('celery.group')
        with patch.object(
            b, "_call_task_errbacks", side_effect=ExpectedException()
        ) as mock_call_errbacks:
>           b.chord_error_from_stack(callback, exc=ValueError())

t/unit/backends/test_base.py:550: 

self = <celery.backends.base.BaseBackend object at 0x000000000a6552f0>
callback = <Mock name='callback' id='174412584'>, exc = ValueError()

    def chord_error_from_stack(self, callback, exc=None):
        app = self.app
        try:
            backend = app._tasks[callback.task].backend
        except KeyError:
            backend = self
        # We have to make a fake request since either the callback failed or
        # we're pretending it did since we don't have information about the
        # chord part(s) which failed. This request is constructed as a best
        # effort for new style errbacks and may be slightly misleading about
        # what really went wrong, but at least we call them!
        fake_request = Context({
            "id": callback.options.get("task_id"),
            "errbacks": callback.options.get("link_error", []),
            "delivery_info": dict(),
>           **callback
        })
E       TypeError: 'Mock' object is not a mapping
```

maybe-sybr: `MagicMock`s appear to work okay for both Python3 and pypy3

celery/canvas.py
```diff
@@ -1123,18 +1123,17 @@ def set_immutable(self, immutable):
...
+        # completed and also if any children other than the first one fail.
+        #
+        # The callback signature is cloned and made immutable since it the
+        # first task isn't actually capable of passing the return values of its
+        # siblings to the callback task.
         sig = sig.clone().set(immutable=True)
         return self.tasks[0].link(sig)
 
     def link_error(self, sig):
-        try:
```
auvipy, line 1131: curious, removing this part won't have any negative effect?

maybe-sybr: From what I can tell, no. I believe it was only cloned so that the `.set(immutable=True)` wouldn't alter the signature passed in by the user, and IMO making the errback immutable is probably the wrong thing to do. I explained (as I understand) the `.clone().set(immutable=True)` in the `link()` method above and since that explanation doesn't apply to the errbacks, I figured it made sense to remove it here.

t/integration/test_canvas.py
```diff
@@ -1497,11 +1498,14 @@ def test_chord_on_error(self, manager):
...
 
-        # Run the chord and wait for the error callback to finish.
+        # Run the chord and wait for the error callback to finish. Note that
+        # this only works for old style callbacks since they get dispatched to
+        # run async while new style errbacks are called synchronously so that
+        # they can be passed the request object for the failing task.
         c1 = chord(
             header=[add.s(1, 2), add.s(3, 4), fail.s()],
             body=print_unicode.s('This should not be called').on_error(
-                chord_error.s()),
```
maybe-sybr, line 1504: Oh, the `chord_error` test task is unused now I think. I can probably remove it prior to merge, just forgot to do so when I pushed this diff.

auvipy: ok

maybe-sybr: done

auvipy (approved):

PR #6818, opened 2021-06-19
Add message properties to app.tasks.Context (#6777)

### LINKED ISSUE #6777: Make AMQP message properties available in task
# Checklist

- [x] I have checked the [issues list](https://github.com/celery/celery/issues?utf8=%E2%9C%93&q=is%3Aissue+label%3A%22Issue+Type%3A+Feature+Request%22+)
- [x] I have checked the [pull requests list](https://github.com/celery/celery/pulls?utf8=%E2%9C%93&q=is%3Apr+label%3A%22PR+Type%3A+Feature%22+)
  master branch.

#### Related Issues

- None

#### Possible Duplicates

- None

# Brief Summary
I would like to be able to access arbitrary AMQP message properties (e.g. `user-id`) within my tasks.
# Design

None

## Proposed Behavior
If the message received from the transport has a `properties` attribute, a reference to the value should be available in the `app.task.Context` instance available as `self.request` in a bound task. If there is no `properties` attribute, the new `Context` attribute may be set to `None`.

## Proposed UI/UX
The new attribute in `app.task.Context` might be named `properties`.

## Diagrams

N/A

## Alternatives

Application-defined headers are already available to the task, but some functionality may be provided through broker interactions which cannot be replicated using application-defined headers (e.g. RabbitMQ's validated User-ID).
Another option would be to expose each of the remaining AMQP 0-9-1 message properties on the `app.task.Context` object similar to `reply_to`. This would have the disadvantage of introducing broker-specific logic.

InvalidInterrupt:
## Description

Fixes #6777 by exposing message properties in `app.task.Context` objects.

Includes a basic integration test. When I tried to run the full test suite, I encountered many failures on `master`, so I'm not too confident this hasn't broken anything. That being said, it is a fairly simple change.

### DIFF
--- a/celery/worker/request.py
+++ b/celery/worker/request.py
@@ -93,7 +93,8 @@ class Request:
                  maybe_make_aware=maybe_make_aware,
                  maybe_iso8601=maybe_iso8601, **opts):
         self._message = message
-        self._request_dict = message.headers if headers is None else headers
+        self._request_dict = (message.headers.copy() if headers is None
+                              else headers.copy())
         self._body = message.body if body is None else body
         self._app = app
         self._utc = utc
@@ -157,6 +158,7 @@ class Request:
             'redelivered': delivery_info.get('redelivered'),
         }
         self._request_dict.update({
+            'properties': properties,
             'reply_to': properties.get('reply_to'),
             'correlation_id': properties.get('correlation_id'),
             'hostname': self._hostname,
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -281,4 +281,5 @@ Frazer McLean, 2020/09/29
 Henrik Bruåsdal, 2020/11/29
 Tom Wojcik, 2021/01/24
 Ruaridh Williamson, 2021/03/09
+Garry Lawrence, 2021/06/19
 Patrick Zhang, 2017/08/19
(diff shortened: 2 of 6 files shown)


auvipy (approved):

celery/worker/request.py
```diff
@@ -93,7 +93,8 @@ def __init__(self, message, on_ack=noop,
                  maybe_make_aware=maybe_make_aware,
                  maybe_iso8601=maybe_iso8601, **opts):
         self._message = message
-        self._request_dict = message.headers if headers is None else headers
+        self._request_dict = (message.headers.copy() if headers is None
```
thedrow, line 96: Why are we copying the headers?

thedrow (commented):
This looks good but I'd like you to explain why are you copying the headers now?

celery/worker/request.py
InvalidInterrupt: `message.headers` is included in `message.properties` as `application_headers`; copying the headers here avoids creating a self-referential object. There's no reason (that I can recall, at least) that it absolutely cannot be self-referential, but it did lead to serialization errors in my test case. I think this is the least surprising way to handle this for end-users.

auvipy: it would great if you can also add unit test for the changes

PR #6823, opened 2021-06-27
The Consul backend must correctly associate requests and responses

### LINKED ISSUE #5605: Race condition in Consul result backend when saving a result
I'm encountering an [issue](https://groups.google.com/forum/#!topic/consul-tool/U1GXnuNmMN8) using the Consul result backend which I suspect is NOT a Celery issue as such but rather [an issue in the lack of thread safety in the library the backend uses](https://github.com/cablehead/python-consul/issues/144).

The symptoms are consistent with confusion between the sorts of responses that Consul gives to the library when creating saved results, and expiring those saved results. So, if Celery both writes and expires saved results using threads and, the lack of thread-safety in the python-consul library could produce the results described.

Can somebody confirm if Celery access to a results backend involves multiple threads?

ShaheedHaque:
## Description

As per #5605, the Consul backend does not cleanly associate responses
from Consul with the outbound Celery request that caused it. This leaves
it prone to mistaking the (final) response from an operation N as the
response to an (early) part of operation N + 1.

This changes fix that by using a separate connection for each request.
That of course has the downside of (a) being relatively expensive and (b)
increasing the rate of connection requests into Consul:

- The former is annoying, but at least the backend works reliably.

- The latter can cause Consul to reject excessive connection attempts, but
  if it does, at least it returns a clear indication of this (IIRC, it responds
  with an HTTP 429 "too many connections" indication).

  Additionally, this issue can be ameliorated by enabling retries in the
  python-consul2 (which I believe should be turned on regards less to handle
  transient network issues). This is fixed by poppyred/python-consul2#31 (which
  is yet to be merged).

Note that we have never seen (b) outside a test specifically trying to hammer
the system, but we see (a) all the time in our normal system tests.

## Tests

The problem only shows up under load, and so a reliable unit test is not easy to engineer. However, as per the discussion in #5605, our system level tests fail without this fix, and pass reliably with the fix.

### DIFF
--- a/celery/backends/consul.py
+++ b/celery/backends/consul.py
@@ -31,7 +31,6 @@ class ConsulBackend(KeyValueStoreBackend):
 
     supports_autoexpire = True
 
-    client = None
     consistency = 'consistent'
     path = None
 
@@ -40,15 +39,33 @@ class ConsulBackend(KeyValueStoreBackend):
 
         if self.consul is None:
             raise ImproperlyConfigured(CONSUL_MISSING)
-
+        #
+        # By default, for correctness, we use a client connection per
+        # operation. If set, self.one_client will be used for all operations.
+        # This provides for the original behaviour to be selected, and is
+        # also convenient for mocking in the unit tests.
+        #
+        self.one_client = None
         self._init_from_params(**parse_url(self.url))
 
     def _init_from_params(self, hostname, port, virtual_host, **params):
         logger.debug('Setting on Consul client to connect to %s:%d',
                      hostname, port)
         self.path = virtual_host
-        self.client = consul.Consul(host=hostname, port=port,
-                                    consistency=self.consistency)
+        self.hostname = hostname
+        self.port = port
+        #
+        # Optionally, allow a single client connection to be used to reduce
+        # the connection load on Consul by adding a "one_client=1" parameter
+        # to the URL.
+        #
+        if params.get('one_client', None):
+            self.one_client = self.client()
+
+    def client(self):
+        return self.one_client or consul.Consul(host=self.hostname,
+                                                port=self.port,
+                                                consistency=self.consistency)
 
     def _key_to_consul_key(self, key):
         key = bytes_to_str(key)
@@ -58,7 +75,7 @@ class ConsulBackend(KeyValueStoreBackend):
         key = self._key_to_consul_key(key)
         logger.debug('Trying to fetch key %s from Consul', key)
         try:
-            _, data = self.client.kv.get(key)
+            _, data = self.client().kv.get(key)
             return data['Value']
         except TypeError:
             pass
@@ -84,17 +101,16 @@ class ConsulBackend(KeyValueStoreBackend):
 
         logger.debug('Trying to create Consul session %s with TTL %d',
                      session_name, self.expires)
-        session_id = self.client.session.create(name=session_name,
-                                                behavior='delete',
-                                                ttl=self.expires)
+        client = self.client()
+        session_id = client.session.create(name=session_name,
+                                           behavior='delete',
+                                           ttl=self.expires)
         logger.debug('Created Consul session %s', session_id)
 
         logger.debug('Writing key %s to Consul', key)
-        return self.client.kv.put(key=key,
-                                  value=value,
-                                  acquire=session_id)
+        return client.kv.put(key=key, value=value, acquire=session_id)
 
     def delete(self, key):
         key = self._key_to_consul_key(key)
         logger.debug('Removing key %s from Consul', key)
-        return self.client.kv.delete(key)
+        return self.client().kv.delete(key)
--- a/docs/userguide/configuration.rst
+++ b/docs/userguide/configuration.rst
@@ -2016,14 +2016,52 @@ without any further configuration. For larger clusters you could use NFS,
 Consul K/V store backend settings
 ---------------------------------
 
-The Consul backend can be configured using a URL, for example:
+.. note::
+
+    The Consul backend requires the :pypi:`python-consul2` library:
+
+    To install this package use :command:`pip`:
+
+    .. code-block:: console
+
+        $ pip install python-consul2
+
+The Consul backend can be configured using a URL, for example::
 
     CELERY_RESULT_BACKEND = 'consul://localhost:8500/'
 
-The backend will storage results in the K/V store of Consul
-as individual keys.
+or::
 
-The backend supports auto expire of results using TTLs in Consul.
+    result_backend = 'consul://localhost:8500/'
+
+The backend will store results in the K/V store of Consul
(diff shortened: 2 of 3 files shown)


maybe-sybr:
Thanks for putting this PR up, @ShaheedHaque . Given that this introduces a new failure mode which might disrupt some people who haven't noticed (or don't care about) the current response confusion issue, could we perhaps make this an opt-out behaviour (ie. we turn it on for people) and add something to the changelog and consul backend doco about how to turn it off if people start experiencing 429s?

ShaheedHaque:
> Thanks for putting this PR up, @ShaheedHaque . Given that this introduces a new failure mode which might disrupt some people who haven't noticed (or don't care about) the current response confusion issue, could we perhaps make this an opt-out behaviour (ie. we turn it on for everyone by default) and add something to the changelog and consul backend doco about how to turn it off if people start experiencing 429s?

I can see the point, and will cook up a change accordingly.

auvipy (changes requested):
besides, tests needs update as well

ShaheedHaque:
I finally have the unit tests working (with some pre-existing errors), and can see that between the design change suggested and the difficulty of using the current "Mock"ing approach, I have to rethink this.

ShaheedHaque:
> besides, tests needs update as well

Indeed. I had trouble getting the unit tests to pass - and there are still several errors unrelated to my changes even in just t/unit/backends.

I have a new set of diffs with the suggested opt-out and test changes which I am testing locally. When done, I propose to squash them onto this branch (since the old approach is not usable) for feedback on the approach b4 attempting to update the changelog/docs. I hope that makes sense.

maybe-sybr:
Sounds great, @ShaheedHaque - thanks! Let us know if there are any specific pain points you're hitting wrt the testing as one of us may know how to work around it :)

ShaheedHaque:
> Sounds great, @ShaheedHaque - thanks! Let us know if there are any specific pain points you're hitting wrt the testing as one of us may know how to work around it :)

FWIW, here is what I saw:

- 4 t/unit tests fail:
    ```
    t/unit/backends/test_base.py::test_BaseBackend_dict::test_chord_error_from_stack_raises FAILED
    t/unit/tasks/test_canvas.py::test_group::test_link_error FAILED
    t/unit/utils/test_timer2.py::test_Timer::test_on_tick FAILED
    t/unit/utils/test_timer2.py::test_Timer::test_gc_race_lost FAILED
    ```
- t/unit tests seems to be bailing out like this:
    ```
    t/unit/worker/test_autoscale.py::test_Autoscaler::test_run 'Scaler' crashed: AttributeError("'Scaler' object has no attribute '_bgThread__is_shutdown'")
      File "/usr/local/lib/python3.8/dist-packages/celery/utils/threads.py", line 67, in run
        body()
      File "/main/srhaque/kdedev/celery/t/unit/worker/test_autoscale.py", line 128, in body
       getattr(self, "_bgThread__is_shutdown").set()
   AttributeError: 'Scaler' object has no attribute '_bgThread__is_shutdown'
    ```
- There are other failures outside t/unit.

I'm proceeding on the basis that there are known issues (or - maybe - don't happen in CI).

ShaheedHaque:
If the approach looks OK, I'll write something for the changelog and docs.

maybe-sybr:
> AttributeError: 'Scaler' object has no attribute '_bgThread__is_shutdown'
Interesting, that seems related to the changes taken to avoid shadowing `threading.Thread` attributes merged recently in #6417 . I'll have a closer look at that and see if I can repro.

celery/backends/consul.py
```diff
@@ -40,15 +39,33 @@ def __init__(self, *args, **kwargs):
...
+        self.port = port
+        #
+        # Optionally, allow a single client connection to be used to reduce
+        # the connection load on Consul by adding a "one_client=1" parameter
+        # to the URL.
+        #
+        if params.get('one_client', None):
+            self.one_client = self.client()
+
+    def client(self):
```
maybe-sybr, line 65: This could probably just be a `@property`, right?

maybe-sybr (commented):
I kind of like this approach because it's sneaky and doesn't require a large chunk of new code and config handling. Although it's a bit sketchy to be adding application level parameters to a transport URI which would be invalid if passed to anything other then celery.

Not going to +1 or -1 for the moment, will leave for someone like @auvipy or @thedrow in the event that they would prefer for it to live somewhere else.

celery/backends/consul.py
ShaheedHaque: It was originally a property, but that's the bit I had trouble mocking. I tried using both Mock and PropertyMock, but couldn't make the ".kv.get.return_value" work. I tried nested Mock instances, but eventually decided to just use what was right in front of me.

maybe-sybr: Ah, interesting - that's annoying. Thanks for the clarification

ShaheedHaque:
Just wondering if folk are waiting on me to do something that I missed? (I did try to write a new test for the set() operation to increase coverage, but could not work out how to make that work).

maybe-sybr:
> Just wondering if folk are waiting on me to do something that I missed? (I did try to write a new test for the set() operation to increase coverage, but could not work out how to make that work).

Not me. Waiting from some input from a second reviewer on my end

ShaheedHaque:
> ~(I did try to write a new test for the set() operation to increase coverage, but could not work out how to make that work).~

Added extra tests to bump code coverage.

ShaheedHaque:
Gentle reminder...I'm still waiting for an indication that this design approach is acceptable before I draft the doc changes.

maybe-sybr:
> Gentle reminder...I'm still waiting for an indication that this design approach is acceptable before I draft the doc changes.

Given that nobody else has objected, I'm happy for this to land as is. Since @thedrow and @auvipy are currently working on the 5.2 release, I think we should aim to do the documentation for this opt-in behaviour and land it in that minor release window.

I spent a little time this morning trying to trick consul into giving me a desync like what you've been observing but I just can't seem to manage it with the std backend. The mystery of #5605 continues :sob:

auvipy:
does  the patch solve your problem?

ShaheedHaque:
> does the patch solve your problem?

Yes. Without it, our system test suite typically gets 5-20 failures per run due to this issue. With the patch, no failures.

ShaheedHaque:
This PR now includes documentation updates for your kind review (and has been rebased to current master).

maybe-sybr (approved):
Looks like `t/unit/backends/test_asynchronous.py::test_EventletDrainer::test_drain_checks_on_interval ` times out sometimes on the Windows platforms. Not your fault as it appears to happen on master also.

LGTM

auvipy (approved):

maybe-sybr:
Thanks, @auvipy :)

Issue #6846, opened 2021-07-08
Chord across multiple apps using signatures breaks on task not registered


ghoulmaster:
# Checklist

- [x] I have verified that the issue exists against the `master` branch of Celery.
- [x] This has already been asked to the [discussion group](https://groups.google.com/forum/#!forum/celery-users) first.
- [x] I have read the relevant section in the
  [contribution guide](http://docs.celeryproject.org/en/latest/contributing.html#other-bugs)
  on reporting bugs.
- [x] I have checked the [issues list](https://github.com/celery/celery/issues?q=is%3Aissue+label%3A%22Issue+Type%3A+Bug+Report%22+-label%3A%22Category%3A+Documentation%22)
  for similar or identical bug reports.
- [x] I have checked the [pull requests list](https://github.com/celery/celery/pulls?q=is%3Apr+label%3A%22PR+Type%3A+Bugfix%22+-label%3A%22Category%3A+Documentation%22)
  for existing proposed fixes.
- [x] I have checked the [commit log](https://github.com/celery/celery/commits/master)
  to find out if the bug was already fixed in the master branch.
- [x] I have included all related issues and possible duplicate issues
  in this issue (If there are none, check this box anyway).

## Mandatory Debugging Information

- [x] I have included the output of ``celery -A proj report`` in the issue.
    (if you are not able to do this, then at least specify the Celery
     version affected).
- [x] I have verified that the issue exists against the `master` branch of Celery.
- [x] I have included the contents of ``pip freeze`` in the issue.
- [x] I have included all the versions of all the external dependencies required
  to reproduce this bug.

## Optional Debugging Information

- [x] I have tried reproducing the issue on more than one Python version
  and/or implementation.
- [x] I have tried reproducing the issue on more than one message broker and/or
  result backend.
- [ ] I have tried reproducing the issue on more than one version of the message
  broker and/or result backend.
- [x] I have tried reproducing the issue on more than one operating system.
- [ ] I have tried reproducing the issue on more than one workers pool.
- [ ] I have tried reproducing the issue with autoscaling, retries,
  ETA/Countdown & rate limits disabled.
- [x] I have tried reproducing the issue after downgrading
  and/or upgrading Celery and its dependencies.

#### Related Issues

- https://groups.google.com/g/celery-users/c/3Z07zPfUZQ0/m/VpOCC2PGDQAJ
- (semi-related but different) https://github.com/celery/celery/issues/6690

#### Possible Duplicates

- None

## Environment & Settings

[...]

PR #6847, opened 2021-07-08
Chords, get body_type independently to handle cases where body.type does not exist ...

ghoulmaster:
… due to tasks being created via Signatures

## Description

When celery Chords are created by Signatures where the workflow is Task_1+Group+Task_2, when the Task_1 gets executed and moves to start the Chain, if the callback Chord Task_2 does not belong to the same queue as Task_1, body.type does not exist and ends up throwing error as the Task_2 does not belong to Task_1's task list and we end up with a NotRegistered error.

Fixes https://github.com/celery/celery/issues/6846

### DIFF
--- a/celery/backends/base.py
+++ b/celery/backends/base.py
@@ -644,7 +644,11 @@ class Backend:
     def fallback_chord_unlock(self, header_result, body, countdown=1,
                               **kwargs):
         kwargs['result'] = [r.as_tuple() for r in header_result]
-        body_type = body.get('type', None)
+        try:
+            body_type = getattr(body, 'type', None)
+        except NotRegistered:
+            body_type = None
+
         queue = body.options.get('queue', getattr(body_type, 'queue', None))
         priority = body.options.get('priority', getattr(body_type, 'priority', 0))
         self.app.tasks['celery.chord_unlock'].apply_async(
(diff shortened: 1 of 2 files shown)


celery/backends/base.py
```diff
@@ -644,8 +644,9 @@ def set_chord_size(self, group_id, chord_size):
     def fallback_chord_unlock(self, header_result, body, countdown=1,
                               **kwargs):
         kwargs['result'] = [r.as_tuple() for r in header_result]
-        queue = body.options.get('queue', getattr(body.type, 'queue', None))
-        priority = body.options.get('priority', getattr(body.type, 'priority', 0))
+        body_type = body.get('type', None)
```
auvipy, line 647: this change need to be unit test covered

auvipy (changes requested):

celery/backends/base.py
ghoulmaster: Will add unit tests and fix current failing one.

Oddly enough the auto unit tests here fail, yet running on windows 10 via cli the same test passes, even odder if I debug through the test it fails.

I will attempt to resolve all this tonight.

ghoulmaster: @auvipy Unit tests added to existing test_chord_unlock_queue test. If you want a separate test for this I can split it out. Just let me know!

auvipy (approved):

PR #6860, opened 2021-07-19
Update configuration.rst

przlada:
Update default worker_task_log_format value


auvipy (approved):

thedrow:
What is the motivation behind this patch?

przlada:
The motivation behind this patch is to update docs to reflect the actual default configuration

thedrow:
> 
> 
> The motivation behind this patch is to update docs to reflect the actual default configuration

Thanks. I just want to keep every commit documented for future's sake.
You never know when you'll need it.

PR #6869, opened 2021-07-22
Remove celery.task references in modules, docs

### LINKED ISSUE #6859: Missing celery.task is not innocuous, breaks pydoc
In #6558 the lack of `celery.task` was closed as "not a bug", but because `task` is still in the `__all__` for the `celery` <s>module</s> package (in `celery/__init__.py`), among other things the missing module breaks `pydoc`:

```bash
$ python3 -m venv .
$ . ./bin/activate
$ python3 -m pip install celery
[...]
$ python3 -m pydoc celery
No module named 'celery.task'

$
```

Other introspection tools, not to mention freezing/packaging tools, will likely also be affected.

# Checklist

- [X] I have read the relevant section in the
  on reporting bugs.
- [X] I have checked the [issues list](https://github.com/celery/celery/issues?q=is%3Aissue+label%3A%22Issue+Type%3A+Bug+Report%22+-label%3A%22Category%3A+Documentation%22)
- [X] I have checked the [pull requests list](https://github.com/celery/celery/pulls?q=is%3Apr+label%3A%22PR+Type%3A+Bugfix%22+-label%3A%22Category%3A+Documentation%22)
  for existing proposed fixes.
- [X] I have checked the [commit log](https://github.com/celery/celery/commits/master)
- [X] I have included all related issues and possible duplicate issues

     version affected).
- [X] I have included the contents of ``pip freeze`` in the issue.
- [X] I have included all the versions of all the external dependencies required
  to reproduce this bug.

  and/or implementation.
  result backend.
  broker and/or result backend.

#### Related Issues

- #6558

#### Possible Duplicates

- None

## Environment & Settings

**Celery version**: 5.1.2 (sun-harmonics)

<details>
<p>
```sh
$ celery report
software -> celery:5.1.2 (sun-harmonics) kombu:5.0.2 py:3.9.6
            billiard:3.6.4.0 py-amqp:5.0.6
            kernel version:5.12.15-200.fc33.x86_64 imp:CPython

deprecated_settings: None
```
</p>
</details>

# Steps to Reproduce

## Required Dependencies

### Python Packages

<details>
<p>

```
amqp==5.0.6
billiard==3.6.4.0
celery==5.1.2
click==7.1.2
click-didyoumean==0.0.3
click-plugins==1.1.1
click-repl==0.2.0
kombu==5.1.0
prompt-toolkit==3.0.19
pytz==2021.1
six==1.16.0
vine==5.0.0
wcwidth==0.2.5
```

</p>
</details>

### Other Dependencies

<details>
<p>
N/A
</p>
</details>

```bash
$ python3 -m venv .
$ . ./bin/activate
$ python3 -m pip install celery
[...]
$ python3 -m pydoc celery
No module named 'celery.task'

$
```

# Expected Behavior

Module documentation is displayed

# Actual Behavior

ferdnyc:

## Description

This PR attempts to clean up lingering references to `celery.task` in both the Celery module code and accompanying documentation (excepting previous release notes and etc.), primarily to prevent failures when attempting to display in-module documentation using `pydoc`.

There are a few references in the documentation to modules previously found below `celery.task` (`celery.task.ping`, for example), for which I couldn't find adequate replacements. I've tagged them in this PR by adding `XXX:` comments, to direct attention to those sections of (example) code.

Fixes #6859

### DIFF
--- a/docs/internals/protocol.rst
+++ b/docs/internals/protocol.rst
@@ -292,8 +292,6 @@ format:
 
 .. code-block:: javascript
 
-    // XXX: No longer applicable?
-
     {"id": "4cc7438e-afd4-4f8f-a2f3-f46567e7ca77",
      "task": "celery.task.PingTask",
      "args": [],
--- a/docs/internals/protocol.rst
+++ b/docs/internals/protocol.rst
@@ -292,6 +292,8 @@ format:
 
 .. code-block:: javascript
 
+    // XXX: No longer applicable?
+
     {"id": "4cc7438e-afd4-4f8f-a2f3-f46567e7ca77",
      "task": "celery.task.PingTask",
      "args": [],
--- a/docs/userguide/application.rst
+++ b/docs/userguide/application.rst
@@ -360,28 +360,21 @@ Finalizing the object will:
 .. topic:: The "default app"
 
     Celery didn't always have applications, it used to be that
-    there was only a module-based API, and for backwards compatibility
-    the old API is still there until the release of Celery 5.0.
+    there was only a module-based API. A compatibility API was
+    available at the old location until the release of Celery 5.0,
+    but has been removed.
 
     Celery always creates a special app - the "default app",
     and this is used if no custom application has been instantiated.
 
-    The :mod:`celery.task` module is there to accommodate the old API,
-    and shouldn't be used if you use a custom app. You should
-    always use the methods on the app instance, not the module based API.
-
-    For example, the old Task base class enables many compatibility
-    features where some may be incompatible with newer features, such
-    as task methods:
+    The :mod:`celery.task` module is no longer available. Use the
+    methods on the app instance, not the module based API:
 
     .. code-block:: python
 
         from celery.task import Task   # << OLD Task base class.
 
         from celery import Task        # << NEW base class.
-
-    The new base class is recommended even if you use the old
-    module-based API.
 
 
 Breaking the chain
@@ -456,7 +449,7 @@ chain breaks:
 
     .. code-block:: python
 
-        from celery.task import Task
+        from celery import Task
         from celery.registry import tasks
 
         class Hello(Task):
@@ -475,7 +468,7 @@ chain breaks:
 
     .. code-block:: python
 
-        from celery.task import task
+        from celery import task
 
         @task(queue='hipri')
         def hello(to):
@@ -513,7 +506,7 @@ class: :class:`celery.Task`.
 
     If you override the task's ``__call__`` method, then it's very important
     that you also call ``self.run`` to execute the body of the task.  Do not
-    call ``super().__call__``.  The ``__call__`` method of the neutral base 
+    call ``super().__call__``.  The ``__call__`` method of the neutral base
     class :class:`celery.Task` is only present for reference.  For optimization,
     this has been unrolled into ``celery.app.trace.build_tracer.trace_task``
     which calls ``run`` directly on the custom task class if no ``__call__``
--- a/docs/userguide/application.rst
+++ b/docs/userguide/application.rst
@@ -468,16 +468,16 @@ chain breaks:
 
     .. code-block:: python
 
-        from celery import task
+        from celery import app
 
-        @task(queue='hipri')
+        @app.task(queue='hipri')
         def hello(to):
             return 'hello {0}'.format(to)
 
 Abstract Tasks
 ==============
 
-All tasks created using the :meth:`~@task` decorator
+All tasks created using the :meth:`@task` decorator
 will inherit from the application's base :attr:`~@Task` class.
 
 You can specify a different base class using the ``base`` argument:
--- a/docs/userguide/periodic-tasks.rst
+++ b/docs/userguide/periodic-tasks.rst
@@ -106,19 +106,19 @@ beat schedule list.
     @app.task
     def test(arg):
         print(arg)
-        
+
     @app.task
     def add(x, y):
         z = x + y
-        print(z) 
+        print(z)
 
 
 
 Setting these up from within the :data:`~@on_after_configure` handler means
-that we'll not evaluate the app at module level when using ``test.s()``. Note that 
+that we'll not evaluate the app at module level when using ``test.s()``. Note that
 :data:`~@on_after_configure` is sent after the app is set up, so tasks outside the
-module where the app is declared (e.g. in a `tasks.py` file located by 
-:meth:`celery.Celery.autodiscover_tasks`) must use a later signal, such as 
+module where the app is declared (e.g. in a `tasks.py` file located by
+:meth:`celery.Celery.autodiscover_tasks`) must use a later signal, such as
 :data:`~@on_after_finalize`.
 
 The :meth:`~@add_periodic_task` function will add the entry to the
@@ -192,7 +192,7 @@ Available Fields
     Execution options (:class:`dict`).
 
     This can be any argument supported by
-    :meth:`~celery.task.base.Task.apply_async` --
+    :meth:`~celery.app.task.Task.apply_async` --
     `exchange`, `routing_key`, `expires`, and so on.
 
 * `relative`
--- a/celery/__init__.py
+++ b/celery/__init__.py
@@ -27,7 +27,7 @@ __keywords__ = 'task job queue distributed messaging actor'
 # -eof meta-
 
 __all__ = (
-    'Celery', 'bugreport', 'shared_task', 'task', 'Task',
+    'Celery', 'bugreport', 'shared_task', 'Task',
     'current_app', 'current_task', 'maybe_signature',
     'chain', 'chord', 'chunks', 'group', 'signature',
     'xmap', 'xstarmap', 'uuid',
@@ -161,7 +161,6 @@ old_module, new_module = local.recreate_module(  # pragma: no cover
         ],
         'celery.utils': ['uuid'],
     },
-    direct={'task': 'celery.task'},
     __package__='celery', __file__=__file__,
     __path__=__path__, __doc__=__doc__, __version__=__version__,
     __author__=__author__, __contact__=__contact__,
--- a/celery/app/control.py
+++ b/celery/app/control.py
@@ -536,7 +536,7 @@ class Control:
             task_name (str): Name of task to change rate limit for.
             rate_limit (int, str): The rate limit as tasks per second,
                 or a rate limit string (`'100/m'`, etc.
-                see :attr:`celery.task.base.Task.rate_limit` for
+                see :attr:`celery.app.task.Task.rate_limit` for
                 more information).
 
         See Also:
--- a/celery/app/registry.py
+++ b/celery/app/registry.py
@@ -36,7 +36,7 @@ class TaskRegistry(dict):
 
         Arguments:
             name (str): name of the task to unregister, or a
-                :class:`celery.task.base.Task` with a valid `name` attribute.
+                :class:`celery.app.task.Task` with a valid `name` attribute.
 
         Raises:
             celery.exceptions.NotRegistered: if the task is not registered.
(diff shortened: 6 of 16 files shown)


docs/userguide/application.rst
```diff
@@ -475,7 +468,7 @@ chain breaks:
 
     .. code-block:: python
 
-        from celery.task import task
+        from celery import task
```
ferdnyc, line 471: Whoops, that's wrong. Will fix.

docs/userguide/periodic-tasks.rst
```diff
@@ -106,19 +106,19 @@ beat schedule list.
...
 
 
 Setting these up from within the :data:`~@on_after_configure` handler means
-that we'll not evaluate the app at module level when using ``test.s()``. Note that 
+that we'll not evaluate the app at module level when using ``test.s()``. Note that
 :data:`~@on_after_configure` is sent after the app is set up, so tasks outside the
-module where the app is declared (e.g. in a `tasks.py` file located by 
-:meth:`celery.Celery.autodiscover_tasks`) must use a later signal, such as 
+module where the app is declared (e.g. in a `tasks.py` file located by
+:meth:`celery.Celery.autodiscover_tasks`) must use a later signal, such as
```
ferdnyc, line 121: Sorry, my editor automatically removes trailing whitespace on save.

docs/internals/guide.rst
```diff
@@ -186,6 +186,7 @@ Here's an example using Celery in single-mode:
 
 .. code-block:: python
 
+    # XXX: No longer valid?
```
thedrow, line 189: Yes this is no longer valid.

docs/internals/protocol.rst
```diff
@@ -292,6 +292,8 @@ format:
 
 .. code-block:: javascript
 
+    // XXX: No longer applicable?
```
thedrow, line 295: The response is still the same but the ping task is elsewhere.

docs/userguide/application.rst
```diff
@@ -475,16 +468,16 @@ chain breaks:
...
-        @task(queue='hipri')
+        @app.task(queue='hipri')
         def hello(to):
             return 'hello {0}'.format(to)
 
 Abstract Tasks
 ==============
 
-All tasks created using the :meth:`~@task` decorator
+All tasks created using the :meth:`~@app.task` decorator
```
thedrow, line 480: Does this render properly?

thedrow (changes requested):
This looks great overall. Thank you for helping us making the documentation more accurate. :heart:

docs/userguide/application.rst
f0xxx1: It does render properly, just tested it.

thedrow:
Could you please address the rest of the review's comments so we can merge this?
Thanks!

thedrow:
@ferdnyc Ping?

ferdnyc:
@thedrow Whoops! Thanks, missed the prev notifications somehow. I'll take a look now, sorry for the delay.

docs/internals/guide.rst
ferdnyc: That's what I figured. Is the "here's the same thing..." version still valid? Should I just rip out the first version and make that second sample the one-and-only?

docs/internals/protocol.rst
ferdnyc: Not sure I understand... is the response _exactly_ the same, or is it a same-except-the-`"task"`-value-changes sort of sameness? If the former I'll just delete my comment, if the latter... perhaps someone can supply an updated JSON response snippet?

docs/userguide/application.rst
ferdnyc: Yeah, I stole that from elsewhere in the docs. :grin:

docs/internals/guide.rst
ferdnyc: (I'll proceed on the assumption that both answers are "yes", and add a commit as described. It's easy enough to drop or tweak further if need be.)

ferdnyc: Hrm. Actually, reading the intro to this section, the whole point is to illustrate the difference between (as the heading _says_, point-blank): "Applications vs. 'single mode'".So it needs two examples, if illustrating that dichotomy is still desirable at all. If not, I can just pull the entire section out.

If it _is_ worth illustrating, it'll need a new first (single-mode) example. Can the current Tasks API still be used single-mode style; is it just a matter of updating the imports?

auvipy (changes requested):
can you please check the readthedocs failures?

ferdnyc:
@auvipy Yeah, I started looking at those yesterday, and ended up down a rabbit hole of Sphinx intricacies. I hate their documentation, and never seem to get the information I actually need out of it.

I'm not at all sure, now, that the decorator references in the documentation are being handled correctly. Not merely the ones I changed, but all of them. In fact, I have concerns about a lot of the cross-references in the docs, and feel like I may have opened a can of worms by touching them. If I'm right, the issues go deeper than just my changes.

It seems that in the celery docs, references can sometimes be preceded by an `@` character, which has some special meaning. Most of these work, even though I can find nothing in the Sphinx docs that supports such a usage.

However, sometimes it breaks. See, for example, the `add_to_chord` docstring:
https://github.com/celery/celery/blob/ebeb4a4607d83cb5668fad5aaac5d5d8f2fb05b4/celery/app/task.py#L936-L947

That type [actually comes out as a **mailto:** link in the rendered documentation](https://docs.celeryproject.org/en/master/reference/celery.app.task.html#celery.app.task.Task.add_to_chord) because of the `@` sign, instead of linking to `celery.Signature` as presumably intended. Changing it to `~celery.Signature` causes it to display as `Signature` and link to the `celery.Signature` class docs in `reference/celery.html`, which seems to be the desired outcome. So I'm completely confused by all these `@` signs.

Meanwhile, coming back to the decorators, Sphinx actually does support both [`py:decorator`](https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#directive-py-decorator) and [`py:decoratormethod`](https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#directive-py-decoratormethod) domains, for writing decorator documentation. I... can't tell, honestly, if those are actually being used for the decorator documentation, because that's generated by autodoc. Autodoc _does_ (as of version 2.0) have an [`autodecorator`](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#directive-autodecorator) directive which I assume makes the answer to that question "yes", but also apparently comes with some caveats:
> #### Note
>
> If you document decorated functions or methods, keep in mind that autodoc retrieves its docstrings by importing the module and inspecting the `__doc__` attribute of the given function or method. That means that if a decorator replaces the decorated function with another, it must copy the original `__doc__` to the new function.

And, as a commenter on [this stackoverflow answer specifically related to celery](https://stackoverflow.com/a/15693082/200794) mentions, documenting the way it's suggested in that answer can cause warnings about duplicate descriptions. The celery docs currently abound with such warnings. A build of all docs with a clean output directory results in 26 instances of 'WARNING: duplicate object description' being emitted.

Even if decorators are documented as decorators, they're still referenced with the `:func:` or `:meth:` inline roles. The Sphinx docs even [take the time to explicitly point that out](https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#directive-py-decorator) (see the bottom of the entry). But they never address the elephant in the room, which is that a <code>:meth:&grave;some.py.function&grave;</code> reference will be formatted as `some.py.function()` and a <code>:meth:&grave;~other.Class.method&grave;</code> reference will format as `method()`. It would be _extremely useful_ to have a <code>:decor:&grave;~some.decorator&grave;</code> role, which would then format in the output as `@decorator`. But maybe that's automatic when the decorators are explicitly documented as decorators. I have no idea, because as  mentioned the Sphinx docs are turrible.

Putting _allllll_ that aside, the crashes in building the docs reference a function, `maybe_resolve_abbreviations`, that I can't seem to find anywhere:
```
Extension error (celerydocs):
Handler <function maybe_resolve_abbreviations at 0x7fe13797a8b0> for event 'missing-reference' threw an exception (exception: 'Unknown abbreviation: @app.task (meth)')
make: *** [Makefile:55: html] Error 2
```

So, the intuition that <code>:meth:&grave;~@app.task&grave;</code> might not format right was spot-on. If  replace all of those references (including the preexisting ones I didn't touch) with something else, the crashes go away and the docs build. The actual effects, however, vary:
* If I use <code>:meth:&grave;app.task&grave;`, the output is **app.task()** and doesn't link anywhere.
* If I use <code>:meth:&grave;~celery.app.task&grave;` or <code>:func:&grave;~celery.app.task&grave;` the output is **task()** and it links to the `celery.app.task` module documentation, which is presumably not the correct location.
* If I use <code>:func:&@task&grave;</code>, the output is **app.task()** (for reasons I cannot comprehend), and it links to the decorator documentation in `celery.Celery.task` (also for reasons I cannot comprehend).

So, presumably the best course of action would be to use <code>:func:&@task&grave;</code> when referencing the `@app.task` decorator, and <code>:func:&~@task&grave;</code> when referencing the `@task` decorator, which will name them both accordingly (except for the `()`-instead-of-`@` formatting issue). They'll both point to the same documentation, which... I _think_... is... right? Honestly at this point I'm completely confused and feeling like I didn't sign up for this.

ferdnyc:
I've updated the changes to use <code>:meth:&grave;@task&grave:</code> anywhere **app.task()** should be displayed as a link to the decorator documentation, even though that feels way too magical for my taste. I've also added a commit that fixes those mailto: links from `~@Signature`. I expect the docs will build correctly now; we shall see.

docs/internals/guide.rst
ferdnyc: @thedrow / @auvipy Any thoughts on this? I believe it's the last outstanding issue. For context, this is the documentation section that reads:

> #### Applications vs. “single mode”
> In the beginning Celery was developed for Django, simply because this enabled us get the project started quickly, while also having a large potential user base.
> 
> In Django there’s a global settings object, so multiple Django projects can’t co-exist in the same process space, this later posed a problem for using Celery with frameworks that don’t have this limitation.
> 
> Therefore the app concept was introduced. When using apps you use ‘celery’ objects instead of importing things from Celery sub-modules, this (unfortunately) also means that Celery essentially has two API’s.
> 
> Here’s an example using Celery in single-mode:
> 
> ```
> from celery import task
> from celery.task.control import inspect
> 
> from .models import CeleryStats
> 
> @task
> def write_stats_to_db():
>     stats = inspect().stats(timeout=1)
>     for node_name, reply in stats:
>         CeleryStats.objects.update_stat(node_name, stats)
> 
> ```
> and here’s the same using Celery app objects:
> 
> ```
> from .celery import celery
> from .models import CeleryStats
> 
> @app.task
> def write_stats_to_db():
>     stats = celery.control.inspect().stats(timeout=1)
>     for node_name, reply in stats:
>         CeleryStats.objects.update_stat(node_name, stats)
> ```
> 
> In the example above the actual application instance is imported from a module in the project, this module could look something like this:
> 
> ```
> from celery import Celery
> 
> app = Celery(broker='amqp://')
> 
> ```

The issue being that the first example is no longer valid code, since the removal of `celery.task`. I am not qualified to write a replacement snippet (or even to determine whether it's still necessary to have one), but I'll be happy to paste one in to the docs if someone can supply it.

auvipy (approved):

docs/internals/guide.rst
ferdnyc: I removed the added comment I'd placed here, as well. Either the first example will get updated, or it won't; since it's only a piece of example code in the documentation, it doesn't affect anything _operationally_ regardless, so this conversation can serve as indication that it needs to be addressed. (Unless someone feels like migrating it to a new Issue.)

auvipy (approved):

Issue #6888, opened 2021-08-01
chord_unlock isn't using the same queue as chord body task


MelnykR:
# Checklist

- [x] I have verified that the issue exists against the `master` branch of Celery.
- [ ] This has already been asked to the [discussion group](https://groups.google.com/forum/#!forum/celery-users) first.
- [x] I have read the relevant section in the
  [contribution guide](http://docs.celeryproject.org/en/latest/contributing.html#other-bugs)
  on reporting bugs.
- [x] I have checked the [issues list](https://github.com/celery/celery/issues?q=is%3Aissue+label%3A%22Issue+Type%3A+Bug+Report%22+-label%3A%22Category%3A+Documentation%22)
  for similar or identical bug reports.
- [x] I have checked the [pull requests list](https://github.com/celery/celery/pulls?q=is%3Apr+label%3A%22PR+Type%3A+Bugfix%22+-label%3A%22Category%3A+Documentation%22)
  for existing proposed fixes.
- [x] I have checked the [commit log](https://github.com/celery/celery/commits/master)
  to find out if the bug was already fixed in the master branch.
- [x] I have included all related issues and possible duplicate issues
  in this issue (If there are none, check this box anyway).

## Mandatory Debugging Information

- [x] I have included the output of ``celery -A proj report`` in the issue.
    (if you are not able to do this, then at least specify the Celery
     version affected).
- [x] I have verified that the issue exists against the `master` branch of Celery.
- [ ] I have included the contents of ``pip freeze`` in the issue.
- [ ] I have included all the versions of all the external dependencies required
  to reproduce this bug.

## Optional Debugging Information

- [x] I have tried reproducing the issue on more than one Python version
  and/or implementation.
- [x] I have tried reproducing the issue on more than one message broker and/or
  result backend.
- [x] I have tried reproducing the issue on more than one version of the message
  broker and/or result backend.
- [x] I have tried reproducing the issue on more than one operating system.
- [x] I have tried reproducing the issue on more than one workers pool.
- [ ] I have tried reproducing the issue with autoscaling, retries,
  ETA/Countdown & rate limits disabled.
- [x] I have tried reproducing the issue after downgrading
  and/or upgrading Celery and its dependencies.

#### Related Issues

- None

#### Possible Duplicates

https://github.com/celery/celery/issues/4337 -- closed duplicate, partially fixed with https://github.com/celery/celery/pull/4448 for manual routing (explicitly passing queue name in signature parameters).

## Environment & Settings

[...]

PR #6896, opened 2021-08-05
Route chord_unlock task to the same queue as chord body

MelnykR:

## Description

Fixes https://github.com/celery/celery/issues/6888

### DIFF
--- a/celery/backends/base.py
+++ b/celery/backends/base.py
@@ -650,6 +650,12 @@ class Backend:
             body_type = None
 
         queue = body.options.get('queue', getattr(body_type, 'queue', None))
+
+        if queue is None:
+            # fallback to default routing if queue name was not
+            # explicitly passed to body callback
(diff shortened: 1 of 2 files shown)


thedrow (changes requested):
This was just the fix I had in mind so well done!
Could you please include a unit test in this patch to verify that this works as expected?

thedrow:
The relevant test fails as the expected outcome has changed.
We need to adjust it before we merge this PR.

MelnykR:
> This was just the fix I had in mind so well done!

Glad to hear that! That was the intention of creating issue and PR -- get feedback whether I am on the right path.


> The relevant test fails as the expected outcome has changed.
We need to adjust it before we merge this PR.

I guess I fixed existing tests. But I think new test should be also added. I will try to work on this on the weekend.

Thanks for the feedback!

auvipy (approved):

auvipy:
thanks!

Issue #6865, opened 2021-07-20
'False' will be ignored in Celery kwargs config


gundam78:
# Checklist

- [x] I have verified that the issue exists against the `master` branch of Celery.
- [x] This has already been asked to the [discussion group](https://groups.google.com/forum/#!forum/celery-users) first.
- [x] I have read the relevant section in the
  [contribution guide](http://docs.celeryproject.org/en/latest/contributing.html#other-bugs)
  on reporting bugs.
- [x] I have checked the [issues list](https://github.com/celery/celery/issues?q=is%3Aissue+label%3A%22Issue+Type%3A+Bug+Report%22+-label%3A%22Category%3A+Documentation%22)
  for similar or identical bug reports.
- [x] I have checked the [pull requests list](https://github.com/celery/celery/pulls?q=is%3Apr+label%3A%22PR+Type%3A+Bugfix%22+-label%3A%22Category%3A+Documentation%22)
  for existing proposed fixes.
- [x] I have checked the [commit log](https://github.com/celery/celery/commits/master)
  to find out if the bug was already fixed in the master branch.
- [x] I have included all related issues and possible duplicate issues
  in this issue (If there are none, check this box anyway).

## Mandatory Debugging Information

- [x] I have included the output of ``celery -A proj report`` in the issue.
    (if you are not able to do this, then at least specify the Celery
     version affected).
- [x] I have verified that the issue exists against the `master` branch of Celery.
- [x] I have included the contents of ``pip freeze`` in the issue.
- [x] I have included all the versions of all the external dependencies required
  to reproduce this bug.

## Optional Debugging Information

- [ ] I have tried reproducing the issue on more than one Python version
  and/or implementation.
- [ ] I have tried reproducing the issue on more than one message broker and/or
  result backend.
- [ ] I have tried reproducing the issue on more than one version of the message
  broker and/or result backend.
- [ ] I have tried reproducing the issue on more than one operating system.
- [ ] I have tried reproducing the issue on more than one workers pool.
- [ ] I have tried reproducing the issue with autoscaling, retries,
  ETA/Countdown & rate limits disabled.
- [ ] I have tried reproducing the issue after downgrading
  and/or upgrading Celery and its dependencies.

#### Related Issues

- None

#### Possible Duplicates

- None

## Environment & Settings

**Celery version**: 5.1.2

<details>
<summary><b><code>celery report</code> Output:</b></summary>
<p>

[...]

PR #6897, opened 2021-08-05
Allow using non-true values in app kwargs

alesolda:
## Description

Trying to instantiate Celery app with *non-true kwargs* will not work
for those configs which have *True* as default, for example,
this will not have effect:

```
>>> app = Celery(task_create_missing_queues=False)
>>> app.conf['task_create_missing_queues']
True
```

This fix simply changes the filtering which from now on will discard
*None* values only.

Fixes: #6865

### DIFF
--- a/celery/app/base.py
+++ b/celery/app/base.py
@@ -323,7 +323,7 @@ class Celery:
         """Optional callback called at init."""
 
     def __autoset(self, key, value):
-        if value:
+        if value is not None:
             self._preconf[key] = value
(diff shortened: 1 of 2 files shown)


auvipy (commented):
is the windows build failure related?

alesolda:
Hi @auvipy !! Hope you are well !! I triggered the pipeline again and nows *Celery / build (3.6, windows-2019)?* succeeded :+1: :smile:

auvipy (approved):

PR #6914, opened 2021-08-16
use same serializer as task seriazlier

ericman93:

## Description

Make `asynpool` to use the same serializer as `task_seriazlier`
If one of the arguments isn't pickleable, and I'm using a different serializer that I registered (dill for example)
The task is being serialized great, but the `asynpool` isn't working since it is using pickle hardcodded

### DIFF
--- a/celery/concurrency/asynpool.py
+++ b/celery/concurrency/asynpool.py
@@ -30,11 +30,12 @@ from billiard.compat import buf_t, isblocking, setblocking
 from billiard.pool import ACK, NACK, RUN, TERMINATE, WorkersJoined
 from billiard.queues import _SimpleQueue
 from kombu.asynchronous import ERR, WRITE
-from kombu.serialization import pickle as _pickle
+from kombu.serialization import pickle as _pickle, registry as serializer_registry
 from kombu.utils.eventio import SELECT_BAD_FD
 from kombu.utils.functional import fxrange
 from vine import promise
 
+from celery import current_app
 from celery.utils.functional import noop
 from celery.utils.log import get_logger
 from celery.worker import state as worker_state
@@ -508,7 +509,7 @@ class AsynPool(_pool.Pool):
         self.handle_result_event = self._result_handler.handle_event
         self._create_timelimit_handlers(hub)
         self._create_process_handlers(hub)
-        self._create_write_handlers(hub)
+        self._create_write_handlers(hub, dumps=serializer_registry._encoders[current_app.conf.task_serializer][-1])
 
         # Add handler for when a process exits (calls maintain_pool)
         [self._track_child_process(w, hub) for w in self._pool]


celery/concurrency/asynpool.py
```diff
@@ -508,7 +509,8 @@ def register_with_event_loop(self, hub):
         self.handle_result_event = self._result_handler.handle_event
         self._create_timelimit_handlers(hub)
         self._create_process_handlers(hub)
-        self._create_write_handlers(hub)
+        self._create_write_handlers(hub,
```
auvipy, line 512: please add test and docs for this change

auvipy (changes requested):

thedrow (changes requested):
We definitely want to have a different configuration setting for this functionality.
I'd rather not break compatibility even if the risk is minimal since I don't know what would happen if we defaulted to json serialization when transmitting messages to our workers.

thedrow (changes requested):
I have some minor comments we need to address before merging this PR.

celery/concurrency/asynpool.py
ericman93: there are no tests for asynpool 
I ran all tests on debug and it got there only in 2 tests that were tested worker controller
what do you suggest?

docs/userguide/configuration.rst
```diff
@@ -3152,6 +3152,15 @@ Name of the pool class used by the worker.
...
 .. setting:: worker_pool_restarts
 
+``worker_pool_serializer``
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+Default: pickle.
+
+If pool is ``"prefork"`` (``celery.concurrency.prefork:TaskPool``) what seralizer to use
+
+.. setting:: worker_pool_serializer
```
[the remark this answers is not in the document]
ericman93: there was another line that needed to be changed, fixed that

docs/userguide/configuration.rst
```diff
@@ -3152,6 +3152,15 @@ Name of the pool class used by the worker.
 
 .. setting:: worker_pool_restarts
 
+``worker_pool_serializer``
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+Default: pickle.
+
+If pool is ``"prefork"`` (``celery.concurrency.prefork:TaskPool``) what seralizer to use
```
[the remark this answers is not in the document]
ericman93: I updated it to have some extra fields
what do you think? I will glad to have some ideas

celery/concurrency/asynpool.py
ericman93: this is why `serializer_registry._encoders['pickle'][-1](obj, protocol=protocol)` passed
it fails when trying running it locally because it doesn't have protocol param 
I'm fixing it, but `asynpool` isn't being testing as part of the uts

celery/app/defaults.py
```diff
@@ -314,6 +314,7 @@ def __repr__(self):
         max_tasks_per_child=Option(type='int'),
         pool=Option(DEFAULT_POOL),
         pool_putlocks=Option(True, type='bool'),
+        pool_serializer=Option('pickle', type='string'),
```
auvipy, line 317: why not json?

auvipy (changes requested):
and we also need the missing tests

PR #6916, opened 2021-08-17
Give indication whether the task is replacing another

thedrow:

## Description

We now increase the `replaced_task_nesting` option each time we replace a task.

### DIFF
--- a/celery/app/task.py
+++ b/celery/app/task.py
@@ -88,6 +88,7 @@ class Context:
     properties = None
     retries = 0
     reply_to = None
+    replaced_task_nesting = 0
     root_id = None
     shadow = None
     taskset = None   # compat alias to group
@@ -128,6 +129,7 @@ class Context:
             'headers': self.headers,
             'retries': self.retries,
             'reply_to': self.reply_to,
+            'replaced_task_nesting': self.replaced_task_nesting,
             'origin': self.origin,
         }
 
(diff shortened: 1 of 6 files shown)


celery/app/task.py
```diff
@@ -88,6 +88,7 @@ class Context:
     properties = None
     retries = 0
     reply_to = None
+    is_replacing = False
```
thedrow, line 91: Is there a better name for this flag?

thedrow:
This still needs documentation but it is ready for review.

t/unit/tasks/test_canvas.py
```diff
@@ -91,7 +91,7 @@ def test_reduce(self):
         assert fun(*args) == x
 
     def test_replace(self):
-        x = Signature('TASK', ('A'), {})
+        x = Signature('TASK', ('A',), {})
```
vlad-outscraper, line 94: why?

vlad-outscraper (dismissed):

auvipy (commented):
I would also like to see integration test for this if possible

t/unit/tasks/test_canvas.py
thedrow: Because the args argument should be a tuple with one item, not a string.
I can extract this change to a different PR as I added this by mistake.

thedrow:
I've found out that `task.replace` isn't documented anywhere in the user guide. I've filed #6967 about it.
I don't have the time to write an integration test and in any case, the integration suite is currently not running and is failing to run. I've filed #6968 so we don't forget to address this later.

PR #6919, opened 2021-08-19
Add cleanup implementation to filesystem backend

eprigorodov:
## Description

Filesystem back-end stores task results in files and currently has no `cleanup()` method. Without other ways of maintenance (e.g. `logrotate` or `systemd-tmpfiles`) that will eventually overflow file system. This PR adds `cleanup()` implementation, which we use in our production.

### DIFF
--- a/celery/backends/filesystem.py
+++ b/celery/backends/filesystem.py
@@ -1,6 +1,7 @@
 """File-system result store backend."""
 import locale
 import os
+from datetime import datetime
 
 from kombu.utils.encoding import ensure_bytes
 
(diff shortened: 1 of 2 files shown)


auvipy (changes requested):
if you can  try to improve the code coverage that would be really great

eprigorodov:
Hi @auvipy, thank you for fast response and explanation. The pipelines in Celery project are great, but I need some advise on how to fix the coverage.

1. The `test_cleanup()` is marked to be skipped on Windows. That is done because test makes a small sleep to divide expired and non-expired sets of task result flies. To make that work reliably on Windows + FAT the sleep time (according to the [docs](https://docs.python.org/3/library/os.html#os.stat_result.st_ctime_ns)) should be longer than two seconds, which is not a good thing for testing, I believe.

If no unit testing is run on Windows + FAT then the test can be enabled on Windows as is. Otherwise the only options are either add longer delay or ignore decreasing coverage. Please advise.

2. To fix new coverage finding I would need to add a third commit. Is that fine or should all commits better be squashed into one?

Thank you

auvipy:
we can sqash commits while merge. i am ok with the current coverage. can you just check on pypy failures?

eprigorodov:
From what I can understand:

```
  collecting ... ERROR: InvocationError for command 'D:\a\celery\celery\.tox\pypy3-unit\Scripts\pytest.EXE' --maxfail=10 -v --cov=celery --cov-report=xml --cov-report term (exited with code 3221225477)
```

exit code 3221225477 means segfault. And here is an issue that looks quite relevant: https://github.com/pytest-dev/pytest/issues/6419.

I'll try to rewrite with `os.listdir()` and `os.stat()`.

auvipy (approved):

eprigorodov:
hi @auvipy, thank you for reviewing and approval.

One last question: we have to maintain a number of boxes with Debian 9 -> Python 3.5 -> Celery 4.4. Can I somehow make a PR for backporting this feature to branch "4.x"?

auvipy:
4.x is EOL. you should use pyenv to upgrade your python and should not rely on systems python.

PR #6923, opened 2021-08-22
Add before_start hook (fixes #4110)

### LINKED ISSUE #4110: Question: any chance to have a on_started handler in the Task class?
It would be very handy to have an on_started call back in the Task class. In addition to after_return, on_failure, on_success and on_retry.

Thoughts?

Krismix1:
## Description
Add `before_start` handler to the `Task` class.

Fixes #4110.

### DIFF
--- a/celery/app/task.py
+++ b/celery/app/task.py
@@ -972,6 +972,18 @@ class Task:
         self.backend.store_result(
             task_id, meta, state, request=self.request, **kwargs)
 
+    def before_start(self, task_id, args, kwargs):
+        """Handler called before the task starts.
+
+        Arguments:
+            task_id (str): Unique id of the task to execute.
+            args (Tuple): Original arguments for the retried task.
+            kwargs (Dict): Original keyword arguments for the retried task.
+
+        Returns:
+            None: The return value of this handler is ignored.
+        """
+
     def on_success(self, retval, task_id, args, kwargs):
         """Success handler.
 
--- a/docs/userguide/tasks.rst
+++ b/docs/userguide/tasks.rst
@@ -1440,6 +1440,16 @@ The default value is the class provided by Celery: ``'celery.app.task:Task'``.
 Handlers
 --------
 
+.. method:: before_start(self, task_id, args, kwargs)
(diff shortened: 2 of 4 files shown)


celery/app/task.py
```diff
@@ -972,6 +972,18 @@ def update_state(self, task_id=None, state=None, meta=None, **kwargs):
         self.backend.store_result(
             task_id, meta, state, request=self.request, **kwargs)
 
+    def before_start(self, task_id, args, kwargs):
+        """Handler called before the task starts.
+
+        Arguments:
+            task_id (str): Unique id of the task to execute.
+            args (Tuple): Original arguments for the retried task.
+            kwargs (Dict): Original keyword arguments for the retried task.
```
thedrow, line 981: The documentation is wrong here.
This is called for every task, not just retried tasks.

thedrow (changes requested):
I have just one comment that we need to address and we can merge this.

docs/userguide/tasks.rst
```diff
@@ -1440,6 +1440,16 @@ The default value is the class provided by Celery: ``'celery.app.task:Task'``.
 Handlers
 --------
 
+.. method:: before_start(self, task_id, args, kwargs)
+
+    Run by the worker before the task starts executing.
```
auvipy, line 1445: you can add added in version 5.2 notation

celery/app/task.py
Krismix1: Updated.

docs/userguide/tasks.rst
Krismix1: Added the directive in the current file.
I have noticed that the notation is sometimes added in the docstring of the source code. Should I add it there as well or is it not needed?

thedrow: It'd be nice to have both, yes.

Krismix1: Added in docstring as well.

auvipy:
could you plz check the read the docs failure?

PR #6949, opened 2021-09-07
update setup with python 3.7 as minimum

auvipy:
i thought it was needed for 5.2 stable


thedrow:
I was about to do it myself. Thanks!

PR #6957, opened 2021-09-16
amqp: send expiration field to broker if requested by user

nicolaerosia:


auvipy (changes requested):
it would be great if you could adjust unit tests as well. and build failures. thanks for the patch

nicolaerosia:
@auvipy sure - I was waiting for bots feedback :)

I see that we can set expires as an absolute time, do you have a suggestion here? Should I try to detect if expires is datetime then subtract if from now() and get the seconds?
https://github.com/celery/celery/pull/6957/checks?check_run_id=3619410907#step:8:3084
```
              # With ETA.
              presult2 = self.mytask.apply_async(
                  kwargs={'name': 'George Costanza'},
                  eta=self.now() + timedelta(days=1),
  >               expires=self.now() + timedelta(days=2),
              )
```
this fails because I'm assuming expires is seconds - e.g. after how many seconds it shoud expire.
https://github.com/celery/celery/pull/6957/checks?check_run_id=3619410907#step:8:3163

nicolaerosia:
@auvipy I could do
```
        if expires is not None:
            if isinstance(expires, datetime.datetime):
                expires_s = (expires - self.now()).total_seconds()
            else:
                expires_s = expires

            options["expiration"] = expires_s
```
however what happens if expires_s is negative? I think I should just restrict it to be positive, since according to https://www.rabbitmq.com/ttl.html , the TTL must always be >= 0
```
The value of the TTL argument or policy must be a non-negative integer (0 <= n), describing the TTL period in milliseconds. Thus a value of 1000 means that a message added to the queue will live in the queue for 1 second or until it is delivered to a consumer. The argument can be of AMQP 0-9-1 type short-short-int, short-int, long-int, or long-long-int.
```

auvipy:
please improve the code coverage

nicolaerosia:
@auvipy how can I re-run codecov?

auvipy:
this should regen now

auvipy (approved):

nicolaerosia:
@auvipy thanks!

Issue #6810, opened 2021-06-11
Executing a chord in a task of a beat schedule leads to recursion exhaustion while dumping.


rainbowtux:
# Checklist

- [X] I have verified that the issue exists against the `master` branch of Celery.
- [X] This has already been asked to the [discussion group](https://groups.google.com/forum/#!forum/celery-users) first.
- [X] I have read the relevant section in the
  [contribution guide](http://docs.celeryproject.org/en/latest/contributing.html#other-bugs)
  on reporting bugs.
- [X] I have checked the [issues list](https://github.com/celery/celery/issues?q=is%3Aissue+label%3A%22Issue+Type%3A+Bug+Report%22+-label%3A%22Category%3A+Documentation%22)
  for similar or identical bug reports.
- [X] I have checked the [pull requests list](https://github.com/celery/celery/pulls?q=is%3Apr+label%3A%22PR+Type%3A+Bugfix%22+-label%3A%22Category%3A+Documentation%22)
  for existing proposed fixes.
- [ ] I have checked the [commit log](https://github.com/celery/celery/commits/master)
  to find out if the bug was already fixed in the master branch.
- [X] I have included all related issues and possible duplicate issues
  in this issue (If there are none, check this box anyway).

## Mandatory Debugging Information

- [X] I have included the output of ``celery -A proj report`` in the issue.
    (if you are not able to do this, then at least specify the Celery
     version affected).
- [X] I have verified that the issue exists against the `master` branch of Celery.
- [X] I have included the contents of ``pip freeze`` in the issue.
- [X] I have included all the versions of all the external dependencies required
  to reproduce this bug.

## Optional Debugging Information

- [ ] I have tried reproducing the issue on more than one Python version
  and/or implementation.
- [ ] I have tried reproducing the issue on more than one message broker and/or
  result backend.
- [ ] I have tried reproducing the issue on more than one version of the message
  broker and/or result backend.
- [ ] I have tried reproducing the issue on more than one operating system.
- [ ] I have tried reproducing the issue on more than one workers pool.
- [ ] I have tried reproducing the issue with autoscaling, retries,
  ETA/Countdown & rate limits disabled.
- [ ] I have tried reproducing the issue after downgrading
  and/or upgrading Celery and its dependencies.

## Environment & Settings
**Celery version**:

<details>
<summary><b><code>celery report</code> Output:</b>
</summary>
<p>
software -> celery:5.1.0 (sun-harmonics) kombu:5.1.0 py:3.9.2
[...]

PR #6959, opened 2021-09-16
canvas: fix kwargs argument to prevent recursion (#6810)

rainbowtux:
## Description

When using a beat schedule running a task taking a chord as argument, celery crashed with max recursion depth reached. After debugging, kwargs were discovered to be nested every time \_\_init\_\_ is called. This PR fixes that.

### DIFF
--- a/celery/canvas.py
+++ b/celery/canvas.py
@@ -1352,10 +1352,10 @@ class _chord(Signature):
     def __init__(self, header, body=None, task='celery.chord',
                  args=None, kwargs=None, app=None, **options):
         args = args if args else ()
-        kwargs = kwargs if kwargs else {}
+        kwargs = kwargs if kwargs else {'kwargs': {}}
         Signature.__init__(
             self, task, args,
-            {'kwargs': kwargs, 'header': _maybe_group(header, app),
+            {**kwargs, 'header': _maybe_group(header, app),
              'body': maybe_signature(body, app=app)}, app=app, **options
         )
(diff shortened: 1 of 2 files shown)


auvipy (changes requested):
could you please add test coverage for the change?

rainbowtux:
> could you please add test coverage for the change?

@auvipy what exactly do you expect from me? Only 2 lines were changed and no new paths of execution were created. None of the existing tests breaks. I could try writing a specific test for this with my still limited knowledge of celery internals, but could you please have a look and see if my reasoning regarding the kwargs is correct? Otherwise I might be writing a test for something wrong.

As I see it now, before my changes, the kwargs argument is added in a dict with 'kwargs': kwargs and that dict is assigned to the kwargs parameter of Signature.\_\_init\_\_, which results in the nesting and recursion and I took that away. It also explains why the problem does not always shows: only repeatedly serializing/deserializing the chord causes the \_\_init\_\_ function to be called over and over with deeper going kwargs every time. I just would like to know your opinion with regard to potential issues somewhere else. Both existing unit tests nor my personal real-life testing did uncover any issues.

Thank you for your time!

rainbowtux:
> could you please add test coverage for the change?

I came up with a test... it seems to prove the point: running on the master branch the test fails, on this branch it works. It verifies whether or not the kwargs get nested.

auvipy (approved):

PR #6965, opened 2021-09-20
Warn when expiration date is in the past

thedrow:

## Description

This is a followup on #6957.
When we do non-obvious things we should let the user know, loudly.
The least surprising default would be to ignore the expiration date, I think.
But since we did it this way, we should warn the users that what is going to happen is that we set the expiration date to 0 and what it means.

@nicolaerosia Would you mind reviewing this as well, please?

### DIFF
--- a/celery/app/base.py
+++ b/celery/app/base.py
@@ -739,6 +739,17 @@ class Celery:
                 expires_s = expires
 
             if expires_s < 0:
+                logger.warning(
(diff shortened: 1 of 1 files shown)


auvipy (approved):

nicolaerosia:
thanks for doing this, I was not aware of this practice. LGTM

Issue #6910, opened 2021-08-12
Events can only be enabled via CLI, `worker_send_task_events` is ignored


AlexHill:
Template filled out below but this is pretty clear.

This line populates Worker.task_events using `either`:
https://github.com/celery/celery/blob/04771d65597f62ccf2f9d901c0d1f7c1d0f24d42/celery/worker/worker.py#L377

However, as of version 5, the `task_events` arg is `False` here when the argument is omitted from the command line, not `None`. So `either`, which looks for the first non-`None` value, always returns `False` and ignores the value from `app.conf`.

Works as expected in 4.4.7.

# Checklist

- [x] I have verified that the issue exists against the `master` branch of Celery.
- [ ] This has already been asked to the [discussion group](https://groups.google.com/forum/#!forum/celery-users) first.
- [x] I have read the relevant section in the
  [contribution guide](http://docs.celeryproject.org/en/latest/contributing.html#other-bugs)
  on reporting bugs.
- [x] I have checked the [issues list](https://github.com/celery/celery/issues?q=is%3Aissue+label%3A%22Issue+Type%3A+Bug+Report%22+-label%3A%22Category%3A+Documentation%22)
  for similar or identical bug reports.
- [x] I have checked the [pull requests list](https://github.com/celery/celery/pulls?q=is%3Apr+label%3A%22PR+Type%3A+Bugfix%22+-label%3A%22Category%3A+Documentation%22)
  for existing proposed fixes.
- [x] I have checked the [commit log](https://github.com/celery/celery/commits/master)
  to find out if the bug was already fixed in the master branch.
- [x] I have included all related issues and possible duplicate issues
  in this issue (If there are none, check this box anyway).

## Mandatory Debugging Information

- [x] I have included the output of ``celery -A proj report`` in the issue.
    (if you are not able to do this, then at least specify the Celery
     version affected).
- [x] I have verified that the issue exists against the `master` branch of Celery.
- [x] I have included the contents of ``pip freeze`` in the issue.
- [x] I have included all the versions of all the external dependencies required
  to reproduce this bug.

## Optional Debugging Information

- [ ] I have tried reproducing the issue on more than one Python version
  and/or implementation.
- [ ] I have tried reproducing the issue on more than one message broker and/or
  result backend.
- [ ] I have tried reproducing the issue on more than one version of the message
  broker and/or result backend.
- [ ] I have tried reproducing the issue on more than one operating system.
- [ ] I have tried reproducing the issue on more than one workers pool.
[...]

PR #6853, opened 2021-07-12
Fix setting worker concurrency option after signal

alesolda:
### Current behaviour

1. `click.option` decorator for `--concurrency` option is executed, its callback returns `0` when evaluating https://github.com/celery/celery/blob/5fd182417d9a6cb1b5aebe29916814d7a725e62a/celery/bin/worker.py#L188 with the values `None or 0`; this default `0` comes from: https://github.com/celery/celery/blob/5fd182417d9a6cb1b5aebe29916814d7a725e62a/celery/app/defaults.py#L297

2. Celery `user_preload_options` signal is processed, then    `app.conf.worker_concurrency` value is correctly updated through `Settings.update`.

3. Celery `worker.worker.WorkController.setup_defaults` kicks off and `concurrency` attribute is resolved with https://github.com/celery/celery/blob/5fd182417d9a6cb1b5aebe29916814d7a725e62a/celery/worker/worker.py#L376

4. `either` method chains calls to `first` function with `None` as predicate (returns the first item that's not `None`): https://github.com/celery/celery/blob/5fd182417d9a6cb1b5aebe29916814d7a725e62a/celery/app/base.py#L925 in our case `first(None, defaults)` (defaults is `(0,)`) will take precedence and `0` will be returned, whatever value is in `app.conf.worker_concurrency`.

### Proposed solution

This fix changes `worker_concurrency` default from `0` to `None` allowing `either` method to correctly resolve in favor of `app.conf.worker_concurrency` value.

The final value used as _concurrency_ is resolved in https://github.com/celery/celery/blob/5fd182417d9a6cb1b5aebe29916814d7a725e62a/celery/worker/worker.py#L109 thus having `None` as default value for `self.concurrency` doesn't break things.

Fixes: #6836

PR #6966, opened 2021-09-20
Allow to enable Events with app.conf mechanism

alesolda:
# Allow to enable Events with app.conf mechanism

### Description

`--task-events` is defined as a [Click Boolean Flag](https://click.palletsprojects.com/en/8.0.x/options/#boolean-flags), without an *off-switch* and `False` as the implicit default value, so when this parameter is omitted in CLI invocation, *Click* will set it to `False`. Because the aforementioned, *Events* only can be enabled via CLI (values in `app.conf.worker_send_task_events` will be ignored).

### Current behaviour

1. `click.option` decorator for `--task-events` sets `task_events` flag to `False`
https://github.com/celery/celery/blob/e2e3e95bf8ac9f85e1ee91753602c47bac878380/celery/bin/worker.py#L205

2. `either` function resolves to the first non-None value (in our case `False` from `task_events` argument ) ignoring values from `app.conf`
https://github.com/celery/celery/blob/e2e3e95bf8ac9f85e1ee91753602c47bac878380/celery/worker/worker.py#L377

### Proposed solution

This fix changes `--task-events` default value from *implicit `False`* to *explicit `None`*, allowing `either` method to correctly resolve in favor of `app.conf.worker_send_task_events` value when set.

Fixes: #6910

### DIFF
--- a/celery/bin/worker.py
+++ b/celery/bin/worker.py
@@ -206,6 +206,7 @@ def detach(path, argv, logfile=None, pidfile=None, uid=None,
               '--task-events',
               '--events',
               is_flag=True,
+              default=None,
               cls=CeleryOption,
               help_group="Pool Options",
               help="Send task-related events that can be captured by monitors"


alesolda:
Hi @auvipy !! Hope you are well :smiley:. According to https://github.com/celery/celery/pull/6853#discussion_r668484926 there aren't unittests to update/add related to Celery's CLI.

auvipy:
yeah you are right

AlexHill:
Author of the linked issue here. Thanks!

PR #6975, opened 2021-09-29
Prevent worker to send expired revoked items upon hello command

olii:
This PR solves the issue when a newly spawned worker receives a lot of old and expired revocations.
There is no reason to send old revocations that are already expired.

In bigger systems that rely on revocations and use a much larger `LimitedSet` this is a problem as the hello response may be hundreds of megabytes large. Receiving such a large message from other workers takes a lot of time and also stresses RabbitMQ (the broker we use). This is also a way to reclaim some memory back in case there are no longer active revocations.

Let me know what do you think about it.

### DIFF
--- a/celery/worker/control.py
+++ b/celery/worker/control.py
@@ -310,6 +310,8 @@ def hello(state, from_node, revoked=None, **kwargs):
         logger.info('sync with %s', from_node)
         if revoked:
             worker_state.revoked.update(revoked)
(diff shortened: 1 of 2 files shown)


auvipy (changes requested):
pleasecheck test failures

olii:
Alerts are not caused by my PR. They are already in `master`.

auvipy (approved):

PR #6980, opened 2021-10-03
Fire task failure signal on final reject

thedrow:

## Description

When a task is rejected but not re-queued, it is marked as failed.
However, the `task_failure` signal isn't sent in this case, although it should have been fired.

This PR fixes that and slightly improves our test quality.

### DIFF
--- a/celery/worker/request.py
+++ b/celery/worker/request.py
@@ -579,6 +579,12 @@ class Request:
                 store_result=self.store_errors,
             )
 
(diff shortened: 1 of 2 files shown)


auvipy (approved):

auvipy:
can u check the lint error?

auvipy:
should we consider this for 5.2?

thedrow:
> 
> 
> should we consider this for 5.2?

Yes, it is a very small change.
I just need to fix the lint errors.

PR #6988, opened 2021-10-07
Change unix socket protocol for Redis from `socket` to `redis+socket`

erictapen:
Otherwise this throws an error:

```
$ CELERY_BROKER_URL=socket::///run/redis/redis.sock celery -A myapp worker
...
KeyError: 'No such transport: socket'
```

So I guess just specifying `socket` should not be recommended in the documentation? Didn't run the test though.


thedrow (approved):

auvipy (changes requested):
not sure why CI is failing

thedrow (changes requested):
The relevant test is failing because of this change.
Please take a look.

PR #6595, opened 2021-01-13
Add new setting: 'task_args_repr_function'.

arnau126:
## Description
Add new setting: `task_args_repr_function` (perhaps there is a more appropriate name).
This setting allows you to choose which function do you want to use to generate `argsrepr` and `kwargsrepr` (the string representation of the task args and kwargs used for logging purposes).

Default is `"celery.utils.saferepr.saferepr"` which is the currently used function.

## Examples
```python3
def task_args_repr_function(o, **kwargs):
    return json.dumps(o)
```

so it will be easier to parse them later on in some backends like the one of `django-celery-results`.

or

```python3
def task_args_repr_function(o, **kwargs):
    if isinstance(o, dict):
        o = o.copy()
        if 'card_number' in o:
            o['card_number'] = '***'
        if 'phone_number' in o:
            o['phone_number'] = o['phone_number'][:3]
        if 'email' in o:
            o.pop('email')
    return repr(o)
```

so we don't have to specify `kwargsrepr` param at each task call.

## Tests
Tests not done yet. Please let me know if you agree with this approach, and I will add some tests.

PR #6989, opened 2021-10-07
Repr function2

AidaPaul:
## Description
This is followup on https://github.com/celery/celery/pull/6595 in an effort to finish it up. I've taken notes from the existing PR comments and added the few missing, and next on my board is to add sufficient test coverage. But in the meanwhile I decided to open this one up in case there is more feedback so it can be tackled promptly.

### DIFF
--- a/docs/userguide/tasks.rst
+++ b/docs/userguide/tasks.rst
@@ -629,7 +629,7 @@ arguments:
     ...     kwargsrepr=repr({'card': '**** **** **** 5678'})
     ... ).delay()
 
-.. versionadded:: 5.1
+.. versionadded:: 5.3
 
 You can also define a custom function to generate the representation of the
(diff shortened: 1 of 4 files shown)


auvipy:
thanks for taking this!

auvipy (changes requested):
you will also need to  add tests for the code changes

docs/userguide/tasks.rst
```diff
@@ -629,6 +629,24 @@ arguments:
     ...     kwargsrepr=repr({'card': '**** **** **** 5678'})
     ... ).delay()
 
+.. versionadded:: 5.1
```
auvipy, line 632:
```suggestion
.. versionadded:: 5.3
```

auvipy (changes requested):

PR #6995, opened 2021-10-16
start with switching to debian slim image

auvipy:
started so that i dont forget about it. lots of works needed to be done. will verify them locally and push incrementally.


auvipy:
updating and building successful with new OS and images & python versions releases

auvipy (commented):
can be merged as is

PR #7082, opened 2021-11-13
not needed anymoreas python 2.7 is not suported

auvipy:


auvipy:
mmm need to revisit the tests

thedrow:
If we remove the requirement, I think we'll have to remove the extra entirely which will break people's production systems.
Let's defer this.

auvipy:
ok

PR #7105, opened 2021-11-18
tried slots on some Celery class attributes to experiment

auvipy:
This is an experimental approach to try with slots in celery class attributes and see how they work. related works https://github.com/celery/py-amqp/pull/368 https://github.com/celery/kombu/pull/1439


auvipy:
humm so it seems the dependencies need slots as well

PR #7125, opened 2021-11-26
Add API to schedules to get next run time

### LINKED ISSUE #6974: Celery Beat: pass trigger time to the task (for bind=True)
# Checklist

  master branch.

#### Related Issues

- None

#### Possible Duplicates

- None

# Brief Summary
Period task is a great addition to Celery. In some use cases you need to know when it was triggered (execution time can not be used to infer it, because of delay,etc).

[A workaround](https://stackoverflow.com/questions/40000663/get-celery-beat-trigger-time-on-task) I saw was using `expires` to calculate it, but would be great if supported natively.

# Design

None

## Proposed Behavior

Setting `eta` as the trigger time in task could do the job. This needs to be done by beat scheduler, while dispatching the task.

I am not well versed with Celery internals, this could break certain things.

## Diagrams

N/A

## Alternatives

None

matthewpull:
This would fix #6974 by

1. introducing a new `next_scheduled_run` API to `BaseSchedule` subclasses (i.e. all concrete schedule classes); and
2. using this API in the beat process to pass the trigger time to the task through the `eta` parameter.

While some users have raised potential drawbacks of (effectively) overloading the `eta` parameter in this way (https://github.com/celery/celery/issues/6974#issuecomment-956323126), I would argue that it is the right approach to provide this new functionality.

Firstly, there is the problem that the `eta` parameter has a side effect - it prevents a task being run before the given datetime. However, this should be no problem for us as (assuming correctness of the implementation) the beat process should not even schedule the task before the eta has occurred, so the eta should always be in the past.

Second, users may already be using the `eta` parameter somehow. While I would think this unlikely, as I cannot see a way in which it could be set per instance of a schedule at runtime given the current codebase - and without this ability it would be functionally redundant - it is technically possible I think. Therefore, my solution only sets the `eta` if it is not already set, maintaining backwards compatibility just in case.

I am open to hearing other suggestions about how to achieve this functionality (all I really want is a reliable and precise solution to #6974, I don't mind how), but in my opinion this is a sensible approach to at least consider.

### DIFF
--- a/celery/beat.py
+++ b/celery/beat.py
@@ -400,6 +400,8 @@ class Scheduler:
         try:
             entry_args = _evaluate_entry_args(entry.args)
             entry_kwargs = _evaluate_entry_kwargs(entry.kwargs)
+            if entry.schedule and "eta" not in entry_kwargs:
+                entry_kwargs["eta"] = entry.schedule.next_scheduled_run(entry.last_run_at)
             if task:
                 return task.apply_async(entry_args, entry_kwargs,
                                         producer=producer,
--- a/celery/schedules.py
+++ b/celery/schedules.py
@@ -12,7 +12,8 @@ from kombu.utils.objects import cached_property
 from . import current_app
 from .utils.collections import AttributeDict
 from .utils.time import (ffwd, humanize_seconds, localize, maybe_make_aware,
-                         maybe_timedelta, remaining, timezone, weekday)
+                         maybe_timedelta, remaining, timezone, weekday,
+                         delta_resolution)
 
 __all__ = (
     'ParseException', 'schedule', 'crontab', 'crontab_parser',
@@ -65,6 +66,9 @@ class BaseSchedule:
 
     def now(self):
         return (self.nowfun or self.app.now)()
+
+    def next_scheduled_run(self, last_run_at, now=None):
+        raise NotImplementedError()
 
     def remaining_estimate(self, last_run_at):
         raise NotImplementedError()
@@ -121,11 +125,23 @@ class schedule(BaseSchedule):
         self.relative = relative
         super().__init__(nowfun=nowfun, app=app)
 
+    def next_scheduled_run(self, last_run_at, now=None):
+        now = now or self.maybe_make_aware(self.now())
+        ends_in = self.run_every
+        start = self.maybe_make_aware(last_run_at)
+        if str(start.tzinfo) == str(now.tzinfo) and now.utcoffset() != start.utcoffset():
+            # DST started/ended
+            start = start.replace(tzinfo=now.tzinfo)
+        end_date = start + ends_in
+        if self.relative:
+            end_date = delta_resolution(end_date, ends_in).replace(
+                microsecond=0)
+        return end_date
+
     def remaining_estimate(self, last_run_at):
-        return remaining(
-            self.maybe_make_aware(last_run_at), self.run_every,
-            self.maybe_make_aware(self.now()), self.relative,
-        )
+        now = self.maybe_make_aware(self.now())
+        next = self.next_scheduled_run(last_run_at, now=now)
+        return next - now
 
     def is_due(self, last_run_at):
         """Return tuple of ``(is_due, next_time_to_check)``.
@@ -767,15 +783,7 @@ class solar(BaseSchedule):
             self.event, self.lat, self.lon,
         )
 
-    def remaining_estimate(self, last_run_at):
-        """Return estimate of next time to run.
-
-        Returns:
-            ~datetime.timedelta: when the periodic task should
-                run next, or if it shouldn't run today (e.g., the sun does
-                not rise today), returns the time when the next check
-                should take place.
-        """
+    def next_scheduled_run(self, last_run_at, now=None):
         last_run_at = self.maybe_make_aware(last_run_at)
         last_run_at_utc = localize(last_run_at, timezone.utc)
         self.cal.date = last_run_at_utc
@@ -797,10 +805,20 @@ class solar(BaseSchedule):
                 self.cal.next_antitransit(self.ephem.Sun()) +
                 timedelta(minutes=1)
             )
-        next = self.maybe_make_aware(next_utc.datetime())
+        return self.maybe_make_aware(next_utc.datetime())
+
+    def remaining_estimate(self, last_run_at):
+        """Return estimate of next time to run.
+
+        Returns:
+            ~datetime.timedelta: when the periodic task should
+                run next, or if it shouldn't run today (e.g., the sun does
+                not rise today), returns the time when the next check
+                should take place.
+        """
         now = self.maybe_make_aware(self.now())
-        delta = next - now
-        return delta
+        next = self.next_scheduled_run(last_run_at)
+        return next - now
 
     def is_due(self, last_run_at):
         """Return tuple of ``(is_due, next_time_to_run)``.


matthewpull:
This is a draft PR as the implementation is not finished yet - I still need to add an implementation for crontab schedule.

matthewpull:
So this is my first time properly looking at the Celery codebase and there are a couple of bits I don't really understand. The most important is line 195 of `utils/time.py` (surrounding function given below).

https://github.com/celery/celery/blob/fad54a99db1aafba505d26c79d7d9368bbeaa4df/celery/utils/time.py#L174-L204

I am probably just missing something, but from my reading of the code I think that line 195 is setting the time zone of `start` to be the time zone of `now` ... which I can kinda understand why, except that on line 193 we know that the `tzinfo` is the same, no?

Also, if both `start` and `now` have accurate time zones, shouldn't Python be able to calculate the timedelta between them accurately, even if DST starts / ends during this time?

Indeed, we don't even actually calculate the timedelta between `start` and `now` directly, but rather after we add `ends_in` to `start` - but this only occurs after line 195, and there appears to be no allowance for a DST threshold in this case. The adding of `ends_in` could easily have caused `end_date` to cross in to / out of DST, so if it is not a problem here then why account for it on line 195?

To be clear, I'm not trying to find fault but am genuinely curious, particularly as this area of the code concerns this PR.

auvipy:
> So this is my first time properly looking at the Celery codebase and there are a couple of bits I don't really understand. The most important is line 195 of `utils/time.py` (surrounding function given below).
> 
> https://github.com/celery/celery/blob/fad54a99db1aafba505d26c79d7d9368bbeaa4df/celery/utils/time.py#L174-L204
> 
> I am probably just missing something, but from my reading of the code I think that line 195 is setting the time zone of `start` to be the time zone of `now` ... which I can kinda understand why, except that on line 193 we know that the `tzinfo` is the same, no?
> 
> Also, if both `start` and `now` have accurate time zones, shouldn't Python be able to calculate the timedelta between them accurately, even if DST starts / ends during this time?
> 
> Indeed, we don't even actually calculate the timedelta between `start` and `now` directly, but rather after we add `ends_in` to `start` - but this only occurs after line 195, and there appears to be no allowance for a DST threshold in this case. The adding of `ends_in` could easily have caused `end_date` to cross in to / out of DST, so if it is not a problem here then why account for it on line 195?
> 
> To be clear, I'm not trying to find fault but am genuinely curious, particularly as this area of the code concerns this PR.

we won't mind if you can find any fault in the code! we are open to suggestions to improve. can you check the related existing tests for the code?

matthewpull:
> This pull request **introduces 3 alerts** when merging [6428c14](https://github.com/celery/celery/commit/6428c1407ff4c0f393370135003c6a1985114646) into [fad54a9](https://github.com/celery/celery/commit/fad54a99db1aafba505d26c79d7d9368bbeaa4df) - [view on LGTM.com](https://lgtm.com/projects/g/celery/celery/rev/pr-7ccb25aa6196f0c1adbd47ca696e20c5b10b867f)
> 
> **new alerts:**
> 
>     * 2 for Wrong name for an argument in a call
> 
>     * 1 for Clear-text logging of sensitive information

Having checked the [LGTM.com link](https://lgtm.com/projects/g/celery/celery/rev/pr-7ccb25aa6196f0c1adbd47ca696e20c5b10b867f), the 3 alerts occur in 2 files unchanged by this PR.

Further, 2 of the alerts are "_Keyword argument 'threads' is not a supported parameter name of function oid_from_". Having checked `oid_from`'s definition in `kombu/common.py`, I can confirm that `threads` __is__ a keyword parameter of that function.

Unless anyone objects, I'm going to ignore the LGTM.com bot's analysis here.

auvipy:
lgtm alerts are not major issue here. was talking about unit tests

matthewpull:
> lgtm alerts are not major issue here. was talking about unit tests

Sorry for the confusion, I did understand what you meant earlier but was just trying to address the low-hanging fruit - cross that minor issue off the list.

> we won't mind if you can find any fault in the code! we are open to suggestions to improve. can you check the related existing tests for the code?

Running the test bench against current celery master (i.e. without any changes in this PR), 3 tests do indeed fail - specifically ones testing DST boundaries, so it is clear that there is something going on here.

I'm still trying to figure out exactly _how_ line 195 addresses it - as I need to make sure that any changes in this PR continue to correct for DST - but I am now sure that it does.

matthewpull:
> I am probably just missing something, but from my reading of the code I think that line 195 is setting the time zone of `start` to be the time zone of `now` ... which I can kinda understand why, except that on line 193 we know that the `tzinfo` is the same, no?

No. As I should have spotted earlier, comparing `str(x) == str(y)` will not necessarily be the same as comparing `x == y`, and indeed in this case it is not. The first part of the if statement is checking if the time zones are the same, _not including DST_. 

> Also, if both `start` and `now` have accurate time zones, shouldn't Python be able to calculate the timedelta between them accurately, even if DST starts / ends during this time?

It can, but this is not what we actually want. The three failing tests all relate to using ffwd (either directly or as part of a crontab schedule) - not the case of wanting to run a task every x hours. For example, if we have a crontab of `0 8 * * *`, executing every 4am local time, we actually would need a delta of 23 or 25 hours when entering or leaving DST (respectively). In all cases it appears the ffwd is relative to the current time, not the start time, hence the necessary conversion.

> Indeed, we don't even actually calculate the timedelta between `start` and `now` directly, but rather after we add `ends_in` to `start` - but this only occurs after line 195, and there appears to be no allowance for a DST threshold in this case. The adding of `ends_in` could easily have caused `end_date` to cross in to / out of DST, so if it is not a problem here then why account for it on line 195?

I'm actually still not 100% sure that this couldn't be an issue. However, I've spent a while trying to find an edge case without success, so I'd say maybe 98%. If it is indeed not an issue I think its because it doesn't matter if `end_date` crossed a DST boundary (i.e. if `start` and `now` are both not in DST but `end_date` was in the future and did cross into DST), as when we calculate `end_date - now` we _are_ actually calculating a time delta. Python does indeed calculate this correctly across different time zones and/or DST boundaries.

> To be clear, I'm not trying to find fault but am genuinely curious, particularly as this area of the code concerns this PR.

To help achieve correctness, I will change the existing `remaining` and `remaining_estimate` methods to use the new ends date system (well, more like extracting out the ends date functionality, as both already calculate it but then discard it). This means that existing test cases will also pass if and only if the new APIs are correct with respect to DST.

matthewpull:
I think the implementation is pretty much there now. Tests are present for the new / changed methods in `time.py`, but I still need to add them for `beat.py` and `schedules.py`.

Given that the existing `schedules.py` tests pass, I am pretty confident of the implementation there, so I think that objective (1) of the PR is now met.

auvipy:
so it is ready for review?

matthewpull:
Sorry, this one sort of fell through the cracks the last couple of weeks while I've been working on other projects. I've just added a couple of test cases for `beat.py`, so now I think it's all there except for documentation (which I can look at if/when the code passes review).

I'd be very happy to hear your (or anyone else's) thoughts on this PR

auvipy:
I started the CI & will thoroughly review it tomorrow. thanks for your work.

auvipy (changes requested):
Looks like you need to pull from master

matthewpull:
I have rerun the test suite with the latest master and I do now get a test failing, but its in the mongodb.py code and the error is `"The DNS query name does not exist: _mongodb._tcp.dns-seedlist-host.example.com"`, which I'm going to assume is unrelated to my PR.

auvipy:
yeah thats un related. @naomielst can you please investigate "The DNS query name does not exist: _mongodb._tcp.dns-seedlist-host.example.com" failure? as you recently handled this.

auvipy:
i can see the following 
  FAILED t/unit/app/test_beat.py::test_Scheduler::test_apply_async_sets_eta - A...
  FAILED t/unit/app/test_beat.py::test_Scheduler::test_apply_async_keeps_existing_eta

matthewpull:
Those are the new ones I made today, to cover the changes in `beat.py`. Just ran the tests again locally and they're still working, and they appear to have passed on at least some of the CI checks - could you share more info on what error / failing assertion you're getting?

matthewpull:
Also which version of Python are you running? I tested on 3.8 but will run on 3.7 and let you know.

auvipy:
check the CI failures of python 3.7 please

matthewpull:
Okay, so something really weird is going on and I'm not quite sure what. It seems that when running on 3.7, it's using celery the package PyPi, as opposed to the local directory.

I'm not exactly sure how to fix it, or if its just a local issue with my setup, but I'm pretty sure it's just not running my code at all. I added some `raise RuntimeError`'s at the start of key functions and pytest on 3.7 appears to plough straight through, with pytest on 3.8 immediately failing (due to the added errors). Same directory on my machine, just different terminals with different venvs.

matthewpull:
I am now pretty sure that it is indeed an error with my setup, but I think it's unrelated to the issue causing those 2 tests to fail - the commit I just pushed fixes that problem on my machine at least; looks like `Mock`'s behave differently in 3.8.

auvipy:
we will drop python37 next year with celery v6

matthewpull:
If this latest commit does indeed fix it, then the problem was only with the test cases; my actual implementation appears to work fine in 3.7, I was just using the Mock class in a way that wasn't compatible with 3.7.

auvipy:
we will merge this after a bug fix release on 5.2.x and before 5.3 start rolling. you can add docs in the mean time when you are free.

matthewpull:
Sounds good, will try to get that done this week 👍

auvipy (changes requested):
please fix the merge conflicts and add the missing docs

thedrow:
@matthewpull In the future, please avoid from contributing from your fork's master branch.
Doing so prevents us from editing this PR ourselves.
I'll review this PR in the upcoming weeks.

PR #7135, opened 2021-12-01
Split Signature.__or__ into subclasses' __or__

treszkai:
## Description
When task chaining with the bar operator (`X | Y | Z`) was introduced in 2016, the following comment was put in `Signature.__or__`:

> These could be implemented in each individual class, I'm sure, but for now we have this.

I saw this as an invitation to move the quite complex branching into `_chain.__or__`, `group.__or__` and `_chord.__or__`.

The commits of micro-refactoring steps are left as is for easier review.

### DIFF
--- a/celery/canvas.py
+++ b/celery/canvas.py
@@ -396,15 +396,15 @@ class Signature(dict):
     def __or__(self, other):
         # These could be implemented in each individual class,
         # I'm sure, but for now we have this.
-        if isinstance(other, _chain):
-            # task | chain -> chain
-            return _chain(seq_concat_seq(
-                (self,), other.unchain_tasks()), app=self._app)
-        elif isinstance(self, chord) and not isinstance(other, group):
+        if isinstance(self, chord) and not isinstance(other, (group, _chain)):
             # chord | task ->  attach to body
             sig = self.clone()
             sig.body = sig.body | other
             return sig
+        elif isinstance(other, _chain):
+            # task | chain -> chain
+            return _chain(seq_concat_seq(
+                (self,), other.unchain_tasks()), app=self._app)
         elif isinstance(other, group):
             # unroll group with one member
             other = maybe_unroll_group(other)
--- a/celery/canvas.py
+++ b/celery/canvas.py
@@ -394,8 +394,6 @@ class Signature(dict):
         )))
 
     def __or__(self, other):
-        # These could be implemented in each individual class,
-        # I'm sure, but for now we have this.
         if isinstance(other, _chain):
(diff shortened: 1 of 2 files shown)


thedrow:
OMG I've been meaning to refractor this for ages. 😳
Thank you! Will review soon.

thedrow:
It looks like the PR is perfect and ready for merge but our integration test suite is currently broken.
@treszkai Did you try running it locally? Also, could you please rebase so that the build will be green?

#7149 should resolve all our problems with the integration suite.

@celery/core-developers Should we merge this now and fix whatever problems which arise (if any) after the integration tests run again?

treszkai:
@thedrow I rebased now on `master`.

I tried running [the dockerized `make test`](https://docs.celeryproject.org/en/master/contributing.html#developing-and-testing-with-docker), which exits immediately with errors (`rm: missing operand` and `make: python: No such file or directory`).

Running [the tests locally](https://docs.celeryproject.org/en/master/contributing.html#running-the-unit-test-suite) also resulted in immediate errors (`'flaky' not found in `markers` configuration option` in the integration tests and `ModuleNotFoundError: No module named 'dns'` in the unit tests), after installing packages from both requirements files.

The canvas-related tests passed last time in the GitHub workflows.

auvipy:
The integration test is not active in CI yet. I returned home today and will try to check the relevant issues

treszkai:
@auvipy The decrease in code coverage is due to missing both a `(chord | group) == chain` and `(chord | chain) == chain` in `_chord.__or__` [(L1387)](https://github.com/celery/celery/pull/7135/files#diff-3a80ff45da16a11b96e26a63973d7d490187a68ddc1949e2dfd7fd090b208841R1387)

L414 in the original was _covered_ (but not tested) as `Signature | group` (in `test_group_to_chord`, `test_chord_size_nested_group_chain_group_mid_single`, and similar).

I now added tests for these two cases and rebased on latest `master`. You can verify that only this changed since @thedrow 's review with `diff <(git -P diff 6fb4f9f6 e538d07c) <(git -P diff cbdebaec 505505881)`.

auvipy:
I think we can include this in a point release, if not that have to hold it as there is a 5.2.2 release pending tomorrow. any aditional docs change is not needed right?

auvipy (approved):
looks good. just want to wait some more time before merge

treszkai:
Thanks! No docs change necessary, as the code is logically equivalent to the previous version.

auvipy:
then we can include this to next point release! thanks a lot for patience!

PR #7181, opened 2021-12-20
Terminate celery entirely on connection error.

thedrow:

## Description
Celery did not provide a way to perform a warm shutdown whenever a connection error occurs.
Production environments are complex and some jobs are long running. It is unacceptable to some of our users to prefetch any new tasks while others are running and `worker_cancel_long_running_tasks_on_connection_loss` is still a compromise to them.

`worker_terminate_on_connection_error`, when set to True will shut down Celery entirely and leave the supervisor the responsibility to revive the Celery process

### DIFF
--- a/celery/app/defaults.py
+++ b/celery/app/defaults.py
@@ -330,6 +330,10 @@ NAMESPACES = Namespace(
         ),
         state_db=Option(),
         task_log_format=Option(DEFAULT_TASK_LOG_FMT),
+        terminate_on_connection_error=Option(
+            False,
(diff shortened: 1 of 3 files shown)


naomielst:
To do: 

- Add a CLI option 
- Add test coverage

thedrow:
We shouldn't implement this. Setting `[broker_connection_retry](https://docs.celeryproject.org/en/stable/userguide/configuration.html#broker-connection-retry)` to False and the worker would shut down.

PR #7183, opened 2021-12-21
Header Stamping

naomielst:

## Description
Implement group stamping - when a group ID is frozen, its ID is stamped on all tasks in the group. 
A task's group stamps are ordered according to the order of freezing.


.github/workflows/python-package.yml
```diff
@@ -1,74 +0,0 @@
-# This workflow will install Python dependencies, run tests and lint with a variety of Python versions
```
auvipy, line 1: I am not sure why the whole CI is deleted?

auvipy (changes requested):
can you please check the build failures?

auvipy:
btw, for future contribution you might follow https://opensource.guide/how-to-contribute/ , and please fork celery and create a branch there and make a PR. that will be better for you IMHO.

.github/workflows/python-package.yml
thedrow: It's just the comment. Not the entire file.

PR #7196, opened 2022-01-02
update docs

auvipy:
https://github.com/celery/celery/issues/7182

https://github.com/celery/celery/blob/master/celery/app/task.py#L778


auvipy (commented):
will be merging as is

PR #7216, opened 2022-01-11
[WIP] - intial tox-docker config for dockerised integration tests

auvipy:
A different approach to  https://github.com/celery/celery/pull/6649 to run integration tests in github action using tox-docker. going to add github action part as well and tweak both. to be  able to make the integrations tests running

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,6 +1,7 @@
 [tox]
 requires =
     tox-gh-actions
+    tox-docker>=3.0
 envlist =
     {3.7,3.8,3.9,3.10,pypy3}-unit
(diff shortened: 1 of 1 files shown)


tox.ini
```diff
@@ -82,6 +84,33 @@ basepython =
 usedevelop = True
 
 
+docker =
+    integration-rabbitmq: rabbitmq
+    integration-redis: redis
+
+dockerenv =
+    PYAMQP_INTEGRATION_INSTANCE=1
```
auvipy, line 92: not sure if this line is needed for some adjustment needed

thedrow:
Why do we need this PR if we're working on a new pytest-celery plugin?

auvipy:
both are different thing

PR #7217, opened 2022-01-11
Switch from pkg_resources importlib-metadata

sim1234:
Based on https://github.com/celery/kombu/pull/1071

Rationale:
- Use of pkg_resources is discouraged in favor of importlib.resources, importlib.metadata, and their backports (resources, metadata). Please consider using those libraries instead of pkg_resources.
- setuptools package is considered to be unsafe in a requirements file

### DIFF
--- a/requirements/default.txt
+++ b/requirements/default.txt
@@ -6,4 +6,4 @@ click>=8.0.3,<9.0
 click-didyoumean>=0.0.3
 click-repl>=0.2.0
 click-plugins>=1.1.1
-setuptools>=59.1.1,<59.7.0
+importlib-metadata>=0.18; python_version<"3.8"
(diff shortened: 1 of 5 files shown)


requirements/default.txt
```diff
@@ -6,4 +6,4 @@ click>=8.0.3,<9.0
 click-didyoumean>=0.0.3
 click-repl>=0.2.0
 click-plugins>=1.1.1
-setuptools>=59.1.1,<59.7.0
```
auvipy, line 9: why setuptools is removed?

sim1234: Because it's no longer required to run the lib. Usage of pkg_resources (that is provided by setuptools) was replaced with usage of importlib.metadata.

thedrow:
I'd like to test this first before we merge.

auvipy:
want to close this in favor of https://github.com/celery/celery/pull/7218

PR #7215, opened 2022-01-10
fix an error which is introduced by pr7040. 

uuip:
## Description

Nothing to subscribe so should not  assign to _pubsub ,  then  drain_events  just sleep.    The error   introduced by pr7040 :  `RuntimeError: pubsub connection not set: did you forget to call subscribe() or psubscribe()?`

[discussion](https://github.com/celery/celery/discussions/7212)

PR #7040, opened 2021-11-03
Prevent from subscribing to emtpy channels

Androidown:
*Note*: Before submitting this pull request, please review our [contributing
guidelines](https://docs.celeryproject.org/en/master/contributing.html).

## Description

If an error occurs while calling `self._pubsub.get_message`, method `_reconnect_pubsub` will be called.
after the following loop is completed:
```
  for meta in metas:
      self.on_state_change(self._decode_result(meta), None)
```
all subscribed tasks might be discarded and the set `self.subscribed_to` will be empty.  Thus a call to `self._pubsub.subscribe(*self.subscribed_to)` will leads the following error:

> redis.exceptions.ResponseError: wrong number of arguments for 'subscribe' command

Maybe related issues:
https://github.com/redis/redis-py/issues/1351
https://github.com/celery/celery/issues/6993

PR #7220, opened 2022-01-12
call ping to set connection for avoiding error (take place of #7215)

uuip:
fix an error which is introduced by [pr7040](https://github.com/celery/celery/pull/7040).

## Description
when subscribed_to is empty , call ping to set connection attr .  
After this operation,  `drain_events` exec self._pubsub.get_message is ok,    
and redis parse_response no error .


auvipy (approved):

auvipy:
looks good to me

PR #7224, opened 2022-01-13
WIP: Include `dont_autoretry_for` option in tasks.

mvaled:
## Description

Add an option `dont_autoretry_for` to exclude some types of exceptions from retrying. 

This allows to opt-out from exceptions which you know are not worthy retrying, but still might be caught by `autoretry_for`.

### DIFF
--- a/docs/userguide/tasks.rst
+++ b/docs/userguide/tasks.rst
@@ -785,6 +785,15 @@ You can also set `autoretry_for`, `max_retries`, `retry_backoff`, `retry_backoff
     and the actual delay value will be a random number between zero and that
     maximum. By default, this option is set to ``True``.
 
+.. versionadded:: 5.2.4
+
+.. attribute:: Task.dont_autoretry_for
+
+    A list/tuple of exception classes.  These exceptions won't be autoretried.
+	This allows to exclude some exceptions that match `autoretry_for
+	<Task.autoretry_for>`:attr: but for which you don't want a retry.
+
+
 .. _task-options:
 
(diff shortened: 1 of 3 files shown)


docs/userguide/tasks.rst
```diff
@@ -785,6 +785,15 @@ You can also set `autoretry_for`, `max_retries`, `retry_backoff`, `retry_backoff
     and the actual delay value will be a random number between zero and that
     maximum. By default, this option is set to ``True``.
 
+.. versionadded:: 5.2.4
```
auvipy, line 788: i think this is a new feaure and should be in 5.3.0 :)

auvipy (changes requested):
thanks for taking the stub! any related issues you could imazine?

mvaled:
> thanks for taking the stub! any related issues you could imazine?

I've just started to work on this.  So far, I don't foresee any issues.  It's just a simple opt-out option which could avoid having to program the [retry feature](https://github.com/merchise-autrement/odoo/blob/merchise-develop-14.0/odoo/jobs.py#L1066) by yourself.

Of course, more *specialized* situations (like inspecting `error.pgcode` to know if we should retry or not the task) should still be left to the app.

I'll keep you posted about the progress in this.

It's been a long time since my last contribution to celery ;) so I'm not up to speed yet.

docs/userguide/tasks.rst
mvaled: I will include this in my next push.

PR #6985, opened 2021-10-06
Fix redis connections leak

D0d0:
Tries to fix https://github.com/celery/celery/issues/6819

_Note_: Before submitting this pull request, please review our [contributing
guidelines](https://docs.celeryproject.org/en/master/contributing.html).

## Description
Tries to fix redis connections leak with caching backend https://github.com/celery/celery/issues/6819. Also should fix https://github.com/andymccurdy/redis-py/issues/1495#

PR #7232, opened 2022-01-18
Fix subscribed_to maybe empty

uuip:
## Description
According to these pull requests:  https://github.com/celery/celery/pull/7040 and https://github.com/celery/celery/pull/7220, I improved the behavior when subscribed_to has been  empty, no additional requests.

### DIFF
--- a/celery/backends/redis.py
+++ b/celery/backends/redis.py
@@ -118,9 +118,8 @@ class ResultConsumer(BaseResultConsumer):
             self._pubsub.connection = self._pubsub.connection_pool.get_connection(
                 'pubsub', self._pubsub.shard_hint
             )
-            # even if nothing to subscribe, we should not lost the callback after being connected.
-            # maybe on_connect will be changed in the future. at present, this callback
-            # re-subscribe to any channels previously subscribed to.
+            # even if there is nothing to subscribe, we should not lose the callback after connecting.
+            # The on_connect callback will re-subscribe to any channels we previously subscribed to.
             self._pubsub.connection.register_connect_callback(self._pubsub.on_connect)
 
     @contextmanager
--- a/celery/backends/redis.py
+++ b/celery/backends/redis.py
@@ -118,6 +118,9 @@ class ResultConsumer(BaseResultConsumer):
             self._pubsub.connection = self._pubsub.connection_pool.get_connection(
                 'pubsub', self._pubsub.shard_hint
             )
+            # even if nothing to subscribe, we should not lost the callback after being connected.
+            # maybe on_connect will be changed in the future. at present, this callback
+            # re-subscribe to any channels previously subscribed to.
             self._pubsub.connection.register_connect_callback(self._pubsub.on_connect)
 
     @contextmanager
--- a/celery/backends/redis.py
+++ b/celery/backends/redis.py
@@ -103,17 +103,22 @@ class ResultConsumer(BaseResultConsumer):
         self.backend.client.connection_pool.reset()
         # task state might have changed when the connection was down so we
         # retrieve meta for all subscribed tasks before going into pubsub mode
-        metas = self.backend.client.mget(self.subscribed_to)
-        metas = [meta for meta in metas if meta]
-        for meta in metas:
-            self.on_state_change(self._decode_result(meta), None)
+        if self.subscribed_to:
+            metas = self.backend.client.mget(self.subscribed_to)
+            metas = [meta for meta in metas if meta]
+            for meta in metas:
+                self.on_state_change(self._decode_result(meta), None)
         self._pubsub = self.backend.client.pubsub(
             ignore_subscribe_messages=True,
         )
+        # subscribed_to maybe empty after on_state_change
         if self.subscribed_to:
             self._pubsub.subscribe(*self.subscribed_to)
         else:
-            self._pubsub.ping()
+            self._pubsub.connection = self._pubsub.connection_pool.get_connection(
+                'pubsub', self._pubsub.shard_hint
+            )
+            self._pubsub.connection.register_connect_callback(self._pubsub.on_connect)
 
     @contextmanager
     def reconnect_on_error(self):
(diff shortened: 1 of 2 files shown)


auvipy (changes requested):
could you please add additional unit tests for the changes proposed?

celery/backends/redis.py
```diff
@@ -103,17 +103,22 @@ def _reconnect_pubsub(self):
...
+            for meta in metas:
+                self.on_state_change(self._decode_result(meta), None)
         self._pubsub = self.backend.client.pubsub(
             ignore_subscribe_messages=True,
         )
+        # subscribed_to maybe empty after on_state_change
         if self.subscribed_to:
             self._pubsub.subscribe(*self.subscribed_to)
         else:
-            self._pubsub.ping()
```
auvipy, line 116: what are the limitations of self._pubsub.ping() ?

uuip: pubsub will receive an empty reply b'$0\r\n\r\n\0x00...', this has no effect on celery's functionality, but causes extra connections and memory usage.  The reason is not to send a ping, but when redis using hiredis to receive data, each connection has a 64 bytes bytearray. Without sending ping, there are no these empty replies. </br>
 This problem was found when analyzing the high memory after running high-frequency tasks for a long time----there are many bytearray(b'$0\r\n\r\n\0x00...') takes up some memory.  
The unexpected growth in memory is related to redis-py, not celery.
redis-py, I also made some changes [redis-py 3.5.3 less memory usage](https://github.com/uuip/redis-py/tree/3.5)

auvipy: OK can you add additional unit tests for the changes?

thedrow: A unit test that verifies this behavior would be nice to have.

celery/backends/redis.py
```diff
@@ -103,17 +103,22 @@ def _reconnect_pubsub(self):
...
+            self._pubsub.connection = self._pubsub.connection_pool.get_connection(
+                'pubsub', self._pubsub.shard_hint
+            )
+            self._pubsub.connection.register_connect_callback(self._pubsub.on_connect)
```
thedrow, line 121: Why was this line added? Can you please add a comment describing what this does?

thedrow (changes requested):
Overall this looks to be the correct fix. I do have a question and we do need a unit test to merge this.

uuip:
@thedrow test_asynchronous.py reduced the coverage, but this pr didn't modify that file.

auvipy (approved):

auvipy:
> @thedrow test_asynchronous.py reduced the coverage, but this pr didn't modify that file.

no worries the PR coverage looks good.

uuip:
There is another problem than I haven't submitted.
```python
    def _reconnect_pubsub(self):
        self._pubsub = None
```
```python
    def _reconnect_pubsub(self):
        # release old connection before assign to new connection
        self._pubsub.reset()
```
I'm not sure yet.

auvipy:
> There is another problem than I haven't submitted.
> 
> ```python
>     def _reconnect_pubsub(self):
>         self._pubsub = None
> ```
> 
> ```python
>     def _reconnect_pubsub(self):
>         # release old connection before assign to new connection
>         self._pubsub.reset()
> ```
> 
> I'm not sure yet.

can you please elaborate more?

uuip:
Until 6 hours ago, I completed a test which running 12 hours.   
Logically, the app is very simple, repeating tasks every 10 seconds. 
At the first 5 minute，there were less than 50 connections，at the ending of test, there were actually more than 1200 connections, and their status were not time_wait, but  ESTAB, I suspect that those connections were not closed normally by celery or redis-py. 

> There is another problem than I haven't submitted.
> 
> ```python
>     def _reconnect_pubsub(self):
>         self._pubsub = None
> ```
> 
> ```python
>     def _reconnect_pubsub(self):
>         # release old connection before assign to new connection
>         self._pubsub.reset()
> ```
> 
> I'm not sure yet.

auvipy:
@pawl can you share your thought on last comment of @uuip please?

uuip:
Shall we focus this pr `Fix subscribed_to maybe empty`?  As for the number of connections, I will analyze it again, an issue or pr in future. Thanks for your kind help.

auvipy:
> Shall we focus this pr `Fix subscribed_to maybe empty`? As for the number of connections, I will analyze it again, an issue or pr in future. Thanks for your kind help.

you can also search for related issues & create a new discussion first to have more consensus

uuip:
@thedrow here is a review assigned to you, if you have time.

auvipy (changes requested):
you will need to adjust tests with the codes as well

auvipy:
Also please do more thorough check of codes and tests before another round of review

uuip:
@auvipy I have removed redundant code, please recheck.

auvipy:
ok, i will re check tomorrow. in the mean time can you  please recheck that you are done with the issue you are trying to resolve?

uuip:
When using redis as backend, gevent or eventlet as concurrency, celery app hold too many redis connections. There are some discussion on https://github.com/celery/celery/pull/6985. 
redis-py always use independent connection for every command, one backend instance is enough.
I fix this scene via a workaround but missing spaces around operator -- this cause previous checking error. This is a workaround, not common solution. After removed it, now all checks have passed.

celery/backends/redis.py
```diff
@@ -103,17 +103,25 @@ def _reconnect_pubsub(self):
...
         if self.subscribed_to:
             self._pubsub.subscribe(*self.subscribed_to)
         else:
-            self._pubsub.ping()
+            self._pubsub.connection = self._pubsub.connection_pool.get_connection(
+                'pubsub', self._pubsub.shard_hint
+            )
+            # even if nothing to subscribe, we should not lost the callback after being connected.
+            # maybe on_connect will be changed in the future. at present, this callback
+            # re-subscribe to any channels previously subscribed to.
```
thedrow, line 123:
```suggestion
            # even if there is nothing to subscribe, we should not lose the callback after connecting.
            # maybe on_connect will be changed in the future. at present, this callback
            # will re-subscribe to any channels we previously subscribed to.
```

thedrow (changes requested):
Just a minor fix and we're there.

auvipy (approved):

auvipy:
thanks both

PR #7234, opened 2022-01-18
Clarify relation between visibility timeout & predefined queues in SQS

browniebroke:
## Description

Clarify documentation about SQS. 

With SQS, the `visibility_timeout` setting is only used when Celery creates a queue in AWS. It is therefore not used when `predefined_queues` is also configured. I had no idea these 2 settings were related like this, hopefully this change will clarify this point in the docs.

The relevant code is [in Kombu](https://github.com/celery/kombu/blob/master/kombu/transport/SQS.py#L364).


auvipy (approved):
looks good

PR #7242, opened 2022-01-20
Ensure all our dependencies are properly maintained

thedrow:

## Description

I'd like to try deadpendency to see if it will help us maintain the project better by getting alerts on unmaintained dependencies which we should replace.


thedrow:
We need to exclude a few packages before merging this.

Issue #6072, opened 2020-05-04
Calling get() on a synchronous chain result triggers E_WOULDBLOCK


neuroid:
# Checklist

- [x] I have verified that the issue exists against the `master` branch of Celery.
- [ ] This has already been asked to the [discussion group](https://groups.google.com/forum/#!forum/celery-users) first.
- [x] I have read the relevant section in the
  [contribution guide](http://docs.celeryproject.org/en/latest/contributing.html#other-bugs)
  on reporting bugs.
- [x] I have checked the [issues list](https://github.com/celery/celery/issues?q=is%3Aissue+label%3A%22Issue+Type%3A+Bug+Report%22+-label%3A%22Category%3A+Documentation%22)
  for similar or identical bug reports.
- [x] I have checked the [pull requests list](https://github.com/celery/celery/pulls?q=is%3Apr+label%3A%22PR+Type%3A+Bugfix%22+-label%3A%22Category%3A+Documentation%22)
  for existing proposed fixes.
- [x] I have checked the [commit log](https://github.com/celery/celery/commits/master)
  to find out if the bug was already fixed in the master branch.
- [x] I have included all related issues and possible duplicate issues
  in this issue (If there are none, check this box anyway).

## Mandatory Debugging Information

- [x] I have included the output of ``celery -A proj report`` in the issue.
    (if you are not able to do this, then at least specify the Celery
     version affected).
- [x] I have verified that the issue exists against the `master` branch of Celery.
- [x] I have included the contents of ``pip freeze`` in the issue.
- [x] I have included all the versions of all the external dependencies required
  to reproduce this bug.

## Optional Debugging Information

- [ ] I have tried reproducing the issue on more than one Python version
  and/or implementation.
- [ ] I have tried reproducing the issue on more than one message broker and/or
  result backend.
- [ ] I have tried reproducing the issue on more than one version of the message
  broker and/or result backend.
- [ ] I have tried reproducing the issue on more than one operating system.
- [ ] I have tried reproducing the issue on more than one workers pool.
- [ ] I have tried reproducing the issue with autoscaling, retries,
  ETA/Countdown & rate limits disabled.
- [ ] I have tried reproducing the issue after downgrading
  and/or upgrading Celery and its dependencies.

#### Related Issues

- https://github.com/celery/celery/issues/5330

#### Possible Duplicates

- https://github.com/celery/celery/issues/5330

## Environment & Settings

**Celery version**:

<details>
[...]

PR #7253, opened 2022-01-23
EagerResult shouldn't care if sync subtasks are enabled

treszkai:
## Description

`EagerResult.get()` shouldn't check if it is executed in an async context (Unlike `AsyncResult.get()`.)

The version on `master` prevents a chain from being executed eagerly because `_chain.apply()` runs `result.get()` on the result of each task in the chain. As this result is an `EagerResult`, this `get()` operation fails.

I don't see the purpose of `EagerResult` ever disabling synchronous subtasks, but maybe I'm missing something.

## Alternative

If that conditional check in `EagerResult.get()` is correct, then `result.get()` could be called with `disable_sync_subtasks=False` in `_chain.apply()`.


auvipy:
@thedrow need your insight here

treszkai:
@thedrow Do you see a need to ever prevent calling `.get()` on an EagerResult in an async task?

celery/result.py
```diff
@@ -993,11 +993,7 @@ def __copy__(self):
     def ready(self):
         return True
 
-    def get(self, timeout=None, propagate=True,
-            disable_sync_subtasks=True, **kwargs):
```
auvipy, line 997: this change seems to be backward incompatible

treszkai: Do you mean some users would expect the assertion, or use `disable_sync_subtasks` as a positional argument? Otherwise, as a keyword argument it's consumed in `kwargs`.

I could leave the parameter and remove the conditional assertion.

thedrow (changes requested):
I haven't read through the code yet but the fundamental problem is that some tasks can run eagerly by forcing them specifically to run eagerly and others can be delayed. If the configuration alone controlled whether a task is applied or applied asynchronously, we could do something like what you're suggesting.

celery/result.py
treszkai: @auvipy I pushed a version that leaves the parameter list unchanged, now fully backwards compatible.

I don't understand @thedrow 's response. `EagerResult` is about a result where the task is already executed and finished, disabling sync subtasks makes no sense to me once you already have the result.

thedrow: But what if that eager result depends on `AsyncResult`s?

treszkai: `EagerResult.get` is so short that I can quote it in full:

```python
def get(self, propagate=True, disable_sync_subtasks=True):
    # << there is currently a check here about disable_sync_subtasks that I'd like to remove
    if self.successful():  
        return self.result  
    elif self.state in states.PROPAGATE_STATES:  
        if propagate:  
            raise self.result if isinstance(  
                self.result, Exception) else Exception(self.result)  
        return self.result
```

This accesses the following attributes of `self` :
 - `def successful(self): return self.state == states.SUCCESS`
 - `EagerResult.state`: a property, returns `self._state`
 - `self.result`: returns `self._result`

None of these depend on other tasks. Therefore, `EagerResult.get` always returns a value or raises an exception in a handful of instructions, and there's no need to disable sync subtasks _inside it_. I'm _not_ talking about `task.apply`.

@thedrow Your response is appreciated, our application uses a not-nice workaround now.

thedrow: The problem arises from mixing sync and async tasks which is possible and a slight change in the could make an EagerResult an AsyncResult.
Since they have the same interface it's not easy to tell until you run the code. This PR can create confusing behavior.
The way to wait for a task to complete and execute another task is using a `chain`.

auvipy (changes requested):
do you have bandwidth to revisit this? making the CI green should be the first priority

treszkai:
@auvipy I'll fix it on Monday, 16 March.

treszkai:
@auvipy The CI is green now. (Merging main into my branch caused `isort` to fail.) My changes are still relevant.

auvipy:
hey @ChickenBenny can you review this

ChickenBenny:
I'll take a look over the weekend.

celery/result.py
```diff
@@ -1038,10 +1039,22 @@ def ready(self):
...
+            propagate (bool): Re-raise exception if the task failed.
+            disable_sync_subtask: No effect.
+
+        Raises:
+            Exception: If the task raised an exception and `propagate` is True.
+        """
+        if disable_sync_subtasks is not None:
+            warnings.warn(
+                'Passing `disable_sync_subtasks` to EagerResult.get has no effect.',
+                CDeprecationWarning, stacklevel=2)
```
ChickenBenny: `stacklevel=2` points to the direct caller of `EagerResult.get()`, which is correct in the straightforward case. However, this class has a `wait = get` alias just below. If a user calls `result.wait(disable_sync_subtasks=True)`, the alias doesn't add a stack frame (it's the same function object), so `stacklevel=2` still works. But if any subclass or wrapper adds a layer, the warning would point to the wrong location.

Could we double-check that `stacklevel=2` is correct for all call paths, or at least add a comment explaining why 2 is the right value here?

celery/result.py
```diff
@@ -1038,10 +1039,22 @@ def ready(self):
...
-        if disable_sync_subtasks:
-            assert_will_not_block()
+            disable_sync_subtasks=None,         # deprecated
+            **kwargs):
+        """Return the result of an eagerly executed task.
 
+        Arguments:
+            timeout: No effect. This method never waits.
+            propagate (bool): Re-raise exception if the task failed.
+            disable_sync_subtask: No effect.
```
ChickenBenny: Typo: `disable_sync_subtask` should be `disable_sync_subtasks`

celery/result.py
ChickenBenny: It appears that unit testing does not cover this section.

treszkai: The fact that unit testing doesn’t cover the deprecation warning speaks to the argument not being used by Celery itself (for this method). I now added a test to check the deprecation (and the return value).

treszkai: Searching for `get\(.*(\n.*){0,3}disable_sync` returns the only relevant result `ResultSet.join`. I couldn’t figure out what method returns a `ResultSet`, but as `ResultSet.join` also has the `if disable_sync_subtasks: assert_will_not_block()` line, it will not affect the subsequent `result.get(...)` if `result` is an `EagerResult`.

Searching for `\bget\(.*(\n.*){0,3}\*\*kwargs` returns :
- `AsyncResult.collect` as vaguely relevant (but not really: it’s only async results)
- `ManagerMixin.join`: will call `ResultSet.get`, which is already cleared

treszkai: (Not sure if you'd still prefer a comment justifying `stacklevel=2`. IMO lines without comments have been reviewed and are expected to work, that's why I didn't add one.)

ChickenBenny: The `stacklevel=2` analysis seems incomplete for the indirect call path.

When `disable_sync_subtasks=False` is passed to `ResultSet.join`, it skips the `assert_will_not_block()` guard (which only fires on `True`) and propagates the value down to each `result.get()` call ([result.py line 801](https://github.com/celery/celery/blob/main/celery/result.py#L801)):

```python
# ResultSet.join
if disable_sync_subtasks:      # False → skipped
    assert_will_not_block()
...
for result in self.results:
    value = result.get(
        ...,
        disable_sync_subtasks=disable_sync_subtasks,  # False propagated here
    )
```

If `result` is an `EagerResult`, the deprecation warning fires — but `stacklevel=2` points to the `result.get(...)` line inside `ResultSet.join`, not to the user's call site:

```
user_code: result_set.join(disable_sync_subtasks=False)   # ← user wants to see this
  └─ ResultSet.join
       └─ result.get(disable_sync_subtasks=False)          # ← stacklevel=2 points here ❌
            └─ EagerResult.get
                 └─ warnings.warn(..., stacklevel=2)
```

This is a real (if rare) case: `disable_sync_subtasks=False` is a deliberate opt-out, so it can appear in production code. Could we either:
- Adjust `stacklevel` dynamically (e.g., check `inspect.stack()` depth), or
- Add a comment acknowledging this known limitation with `stacklevel=2`?

celery/result.py
treszkai: Correct, fixed.

celery/result.py
treszkai: @ChickenBenny Very true.

Stack inspection turned out fairly concise, so now implemented and tested.

As such, @auvipy's (and Copilot's) fix of warning only when `disable_sync_subtasks is False` turned out unnecessary.

treszkai (commented):
@ChickenBenny Thanks for the thorough review. Fixed according to your requests.

celery/result.py
treszkai: The test suite keeps failing on Windows; I’ll see tomorrow what the issue is, as the logs don’t even state the failing test case.

treszkai: (The failing test was in the logs of course, I just couldn't find it.)

`t/unit/tasks/test_canvas.py::test_chain::test_chord_to_chain` keeps failing on the server but not locally, probably because I only run the failing test case. On the server, [calling `inspect.stack()` fails](https://github.com/celery/celery/actions/runs/23959998652/job/69887317403?pr=7253#step:6:33788) because somewhere there is a mocked module -- I expect this is because another test case doesn't mock properly.

At this point I'd like to call it a day, I've been at it for hours and am making progress, will come back to it on Monday.

treszkai: I'd also say that stack inspection in `EagerResult.get` is too heavy a thing to execute. I'd rather just drop the whole warning.

But the test case failing is rather curious. I bisected the minimum failing test command to `pytest  .\t\unit\fixups\test_django.py::test_DjangoWorkerFixup::test_validate_models  .\t\unit\tasks\test_canvas.py::test_chain::test_chord_to_chain`

I'll keep grinding on this now, to see what's wrong.

treszkai: @ChickenBenny Accordingly, I removed the deprecation warning altogether. It technically wasn't a deprecation anyhow, as EagerResult subclassed AsyncResult, so the keyword argument is there to stay.

I'll open another issue about the test failings (which are now irrelevant again).

treszkai: @auvipy @ChickenBenny I investigated the `inspect.stack()` issue and proposed a fix in #10256. We could technically do stack inspection for this feature but _at most_ with `inspect.stack(context=0)` (with source lines omitted), though even that is fairly expensive for such a simple method (`EagerResult.get`). Leaving the default `context=1` would add _another_ order of magnitude execution time.

treszkai:
@auvipy Rebased and squashed commits.

auvipy:
is this PR related to this? https://github.com/celery/celery/pull/8070

treszkai:
@auvipy Yes, #8070 is the very same issue, and same fix but for the docstring and test changes (which that issue doesn't yet deal with).

auvipy (commented):
@Nusnus need your final call here

treszkai:
@auvipy @Nusnus This PR fixes https://github.com/celery/celery/issues/6072

PR #7257, opened 2022-01-24
Remove __ne__ methods

atombrella:
I'm not really sure what the sentiments are for this type of change. Python's documentation was slightly unclear to me, and I wrote some sample code to clarify how it works.

https://gist.github.com/atombrella/c16693f564b4bbf375f70e5d56fe30ad

I think this `__ne__` method can be removed as well:

```
def with_unique_field(attr):

        def __ne__(this, other):
            res = this.__eq__(other)
            return True if res is NotImplemented else not res
        cls.__ne__ = __ne__
```

## Description

These are already defined as the opposite of __eq__ in Python 3, and when __eq__
returns NotImplemented, Python by default will return True.

### DIFF
--- a/celery/beat.py
+++ b/celery/beat.py
@@ -194,14 +194,6 @@ class ScheduleEntry:
         """
         return self.editable_fields_equal(other)
 
-    def __ne__(self, other):
-        """Test schedule entries inequality.
-
-        Will only compare "editable" fields:
-        ``task``, ``schedule``, ``args``, ``kwargs``, ``options``.
-        """
-        return not self == other
-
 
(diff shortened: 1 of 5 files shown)


auvipy (changes requested):
please check the test failures as well

atombrella:
@auvipy I'm not quite sure my changes caused the test failures. They all appear to be related to codecov connectivity/token issues. If you can re-trigger the jobs, and they still fail, I'll have a closer look. The `tox` steps all pass, but `codecov` does not.

auvipy (approved):

PR #7263, opened 2022-01-25
Set max_line_length in .editorconfig to match flake8 config.

atombrella:
https://github.com/celery/celery/blob/master/setup.cfg#L9 I think these two values should be identical.


auvipy:
yeah make sense to me

auvipy (approved):

PR #7268, opened 2022-01-27
Silence backend warning when eager results are stored

kwikwag:
## Description

When task_always_eager is enabled, the backend issues a warning when trying to get task data, This cancels this warning in case task_store_eager_result is enabled as well.

This began with #2275, and closed in c71cd08 (later changed from an error to a warning), but v5.1 introduced `task_store_eager_result ` and this wasn't updated.

### DIFF
--- a/celery/backends/base.py
+++ b/celery/backends/base.py
@@ -569,9 +569,10 @@ class Backend:
             pass
 
     def _ensure_not_eager(self):
-        if self.app.conf.task_always_eager:
+        if self.app.conf.task_always_eager and not self.app.conf.task_store_eager_result:
             warnings.warn(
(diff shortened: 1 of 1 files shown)


celery/backends/base.py
```diff
@@ -569,9 +569,10 @@ def get_children(self, task_id):
             pass
 
     def _ensure_not_eager(self):
-        if self.app.conf.task_always_eager:
+        if self.app.conf.task_always_eager and not self.app.conf.task_store_eager_result:
```
auvipy, line 572: would you mind adding some unit tests for this small but important changes?

auvipy (changes requested):

auvipy (approved):

PR #7271, opened 2022-01-31
Update link to exponential backoff and jitter

DmytroLitvinov:
## Description

Update link to exponential backoff and jitter


auvipy (approved):
yeah its working

PR #6863, opened 2021-07-19
Discard jobs on flush if synack isn't enabled

thedrow:
*Note*: Before submitting this pull request, please review our [contributing
guidelines](https://docs.celeryproject.org/en/master/contributing.html).

## Description

Fixes #6855.

A connection loss flushes the asynpool (See https://github.com/celery/celery/blob/117cd9ca410e8879f71bd84be27b8e69e462c56a/celery/worker/consumer/consumer.py#L414).
This is expected as these jobs cannot be completed anymore.
However, jobs that have not been accepted yet (that is, they are not running yet) are canceled. This only works if the synack keyword argument is set to True.
In our case, it isn't and therefore the jobs remain in the pool's cache forever.

This is a memory leak which we have now resolved by discarding the job (which clears it from the cache) as they will never be cancelled.

PR #7273, opened 2022-02-01
Only clear the cache if there are no active writers.

naomielst:
In #6863 we discarded all jobs if synack isn't enabled for the pool.
This fixed a severe memory leak which occurs on connection restart.

Instead of going over each job and checking if we should discard it, we should clear the entire cache when there are no active writers.

If there are active writers, we should discard the jobs from the cache after we're done writing them since they also may remain on the cache forever.


auvipy (commented):
should we try to add some tests or increase some test coverage?

auvipy (approved):

PR #7279, opened 2022-02-03
Update example in docs

VojtechH:
## Description

I think the first example in the Connections was intended to look like this in order to correspond to the second example.

### DIFF
--- a/docs/userguide/calling.rst
+++ b/docs/userguide/calling.rst
@@ -673,13 +673,13 @@ publisher:
 
 .. code-block:: python
 
-
+    numbers = [(2, 2), (4, 4), (8, 8), (16, 16)]
     results = []
     with add.app.pool.acquire(block=True) as connection:
         with add.get_publisher(connection) as publisher:
             try:
-                for args in numbers:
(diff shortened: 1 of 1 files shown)


auvipy:
can you check and add the proposed changes in unit test please?

VojtechH:
@auvipy I wasn't aware that the docs are directly interconnected with unit tests and I wasn't able to find the specific test that would include this example. Could you please give me a hint about where should I look?

The change in the docs shouldn't affect any functionality of the library. It is only about improving the correctness and clarity of the example.

auvipy:
docs are not directly related to unit test, they represent codes functionality. the reason I asked for that is to improve test cases a bit better using more clear examples :)

VojtechH:
Thank you for the clarification. I'll try to look into it, but it may take some time. I'm currently at the stage of reading the docs to get familiar with Celery and my suggestion was more on a general Python level rather than specific to the library. I'll need to dig deeper to get familiar with the unit tests of Celery too.

thedrow (approved):

auvipy (approved):

Issue #7290, opened 2022-02-09
Celery beat sleeps 300 seconds sometimes even when it should run a task within a few seconds (e.g. 13 seconds)


rubgombar1:
# Checklist
- [x] I have verified that the issue exists against the `master` branch of Celery.
- [ ] This has already been asked to the [discussions forum](https://github.com/celery/celery/discussions) first.
- [x] I have read the relevant section in the
  [contribution guide](http://docs.celeryproject.org/en/latest/contributing.html#other-bugs)
  on reporting bugs.
- [x] I have checked the [issues list](https://github.com/celery/celery/issues?q=is%3Aissue+label%3A%22Issue+Type%3A+Bug+Report%22+-label%3A%22Category%3A+Documentation%22)
  for similar or identical bug reports.
- [x] I have checked the [pull requests list](https://github.com/celery/celery/pulls?q=is%3Apr+label%3A%22PR+Type%3A+Bugfix%22+-label%3A%22Category%3A+Documentation%22)
  for existing proposed fixes.
- [x] I have checked the [commit log](https://github.com/celery/celery/commits/master)
  to find out if the bug was already fixed in the master branch.
- [x] I have included all related issues and possible duplicate issues
  in this issue (If there are none, check this box anyway).

## Mandatory Debugging Information

- [x] I have included the output of ``celery -A proj report`` in the issue.
    (if you are not able to do this, then at least specify the Celery
     version affected).
- [x] I have verified that the issue exists against the `master` branch of Celery.
- [x] I have included the contents of ``pip freeze`` in the issue.
- [x] I have included all the versions of all the external dependencies required
  to reproduce this bug.

## Optional Debugging Information
- [ ] I have tried reproducing the issue on more than one Python version
  and/or implementation.
- [ ] I have tried reproducing the issue on more than one message broker and/or
  result backend.
- [ ] I have tried reproducing the issue on more than one version of the message
  broker and/or result backend.
- [ ] I have tried reproducing the issue on more than one operating system.
- [ ] I have tried reproducing the issue on more than one workers pool.
- [ ] I have tried reproducing the issue with autoscaling, retries,
  ETA/Countdown & rate limits disabled.
- [x] I have tried reproducing the issue after downgrading
  and/or upgrading Celery and its dependencies.

#### Related Issues

- None

#### Possible Duplicates

- None

## Environment & Settings
4.4.7 (cliffs)
**Celery version**:
<details>
<summary><b><code>celery report</code> Output:</b></summary>
<p>

[...]

PR #7291, opened 2022-02-09
Fix: Celery beat sleeps 300 seconds sometimes even when it should run a task within a few seconds (e.g. 13 seconds) #7290

rubgombar1:
Fixes the issue reported in #7290

### DIFF
--- a/celery/utils/functional.py
+++ b/celery/utils/functional.py
@@ -389,3 +389,7 @@ def seq_concat_seq(a, b):
     if not isinstance(b, prefer):
         b = prefer(b)
     return a + b
+
+
+def is_numeric_value(value):
+    return isinstance(value, (int, float)) and not isinstance(value, bool)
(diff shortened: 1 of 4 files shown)


celery/utils/functional.py
```diff
@@ -389,3 +389,7 @@ def seq_concat_seq(a, b):
     if not isinstance(b, prefer):
         b = prefer(b)
     return a + b
+
+
+def is_numeric_value(value):
```
auvipy, line 394: this function also need test coverage.

auvipy (changes requested):
hope this is backward compatible

rubgombar1:
> hope this is backward compatible

Hi @auvipy! 

Thank you so much for your feedback.

What do you mean by backwards compatible? With older versions of celery? I have added several tests for the is_numeric_value function in order to check with the current python versions.

celery/utils/functional.py
rubgombar1: Done! I added the test (with several test cases) in https://github.com/celery/celery/pull/7291/commits/0e8d9db9a1ecc9a3192b647b690fec7b482f778a

PR #7300, opened 2022-02-15
`broker_connection_retry` should no longer apply on startup

thedrow:

### DIFF
--- a/celery/worker/consumer/consumer.py
+++ b/celery/worker/consumer/consumer.py
@@ -333,10 +333,11 @@ class Consumer:
                                          else 'broker_connection_retry')
                 connection_retry = self.app.conf[connection_retry_type]
                 if not connection_retry:
-                    raise WorkerShutdown(
+                    crit(
                         f"Retrying to {'establish' if is_connection_loss_on_startup else 're-establish'} "
                         f"a connection to the message broker after a connection loss has "
-                        f"been disabled (app.conf.{connection_retry_type}=False). Shutting down...") from exc
+                        f"been disabled (app.conf.{connection_retry_type}=False). Shutting down...")
+                    raise WorkerShutdown(1) from exc
                 if isinstance(exc, OSError) and exc.errno == errno.EMFILE:
                     raise WorkerTerminate("Too many open files. Aborting...") from exc
                 maybe_shutdown()
--- a/celery/worker/consumer/consumer.py
+++ b/celery/worker/consumer/consumer.py
@@ -327,16 +327,14 @@ class Consumer:
                 # If we're not retrying connections, we need to properly shutdown or terminate
                 # the Celery main process instead of abruptly aborting the process without any cleanup.
                 is_connection_loss_on_startup = self.restart_count == 0
-                connection_retry_type = ('broker_connection_retry_on_startup'
-                                         if (is_connection_loss_on_startup
-                                             and self.app.conf.broker_connection_retry_on_startup is not None)
-                                         else 'broker_connection_retry')
+                connection_retry_type = self._get_connection_retry_type(is_connection_loss_on_startup)
                 connection_retry = self.app.conf[connection_retry_type]
                 if not connection_retry:
                     crit(
                         f"Retrying to {'establish' if is_connection_loss_on_startup else 're-establish'} "
                         f"a connection to the message broker after a connection loss has "
-                        f"been disabled (app.conf.{connection_retry_type}=False). Shutting down...")
+                        f"been disabled (app.conf.{connection_retry_type}=False). Shutting down..."
+                    )
                     raise WorkerShutdown(1) from exc
                 if isinstance(exc, OSError) and exc.errno == errno.EMFILE:
                     crit("Too many open files. Aborting...")
@@ -349,6 +347,12 @@ class Consumer:
                         self.on_connection_error_before_connected(exc)
                     self.on_close()
                     blueprint.restart(self)
+
+    def _get_connection_retry_type(self, is_connection_loss_on_startup):
+        return ('broker_connection_retry_on_startup'
+                if (is_connection_loss_on_startup
+                    and self.app.conf.broker_connection_retry_on_startup is not None)
+                else 'broker_connection_retry')
 
     def on_connection_error_before_connected(self, exc):
         error(CONNECTION_ERROR, self.conninfo.as_uri(), exc,
--- a/celery/worker/consumer/consumer.py
+++ b/celery/worker/consumer/consumer.py
@@ -465,20 +465,17 @@ class Consumer:
 
         # TODO: Rely only on broker_connection_retry_on_startup to determine whether connection retries are disabled.
         #       We will make the switch in Celery 6.0.
-        deprecated_no_retry_on_startup = (self.app.conf.broker_connection_retry_on_startup is None
-                                          and not self.app.conf.broker_connection_retry)
-        if (
-            deprecated_no_retry_on_startup
-            or not self.app.conf.broker_connection_retry_on_startup
-        ):
-            if deprecated_no_retry_on_startup:
-                warnings.warn(
-                    CPendingDeprecationWarning(
-                        "The broker_connection_retry configuration setting will no longer determine\n"
-                        "whether broker connection retries are made during startup in Celery 6.0 and above.\n"
-                        "If you wish to refrain from retrying connections on startup,\n"
-                        "you should set broker_connection_retry_on_startup to False instead.")
-                )
+
+        if self.app.conf.broker_connection_retry_on_startup is None:
+            warnings.warn(
+                CPendingDeprecationWarning(
+                    f"The broker_connection_retry configuration setting will no longer determine\n"
+                    f"whether broker connection retries are made during startup in Celery 6.0 and above.\n"
+                    f"If you wish to retain the existing behavior for retrying connections on startup,\n"
+                    f"you should set broker_connection_retry_on_startup to {self.app.conf.broker_connection_retry}.")
+            )
+
+        if not self.app.conf.broker_connection_retry and not self.app.conf.broker_connection_retry_on_startup:
             # Retry disabled, just call connect directly.
             conn.connect()
             return conn
--- a/celery/worker/consumer/consumer.py
+++ b/celery/worker/consumer/consumer.py
@@ -468,9 +468,9 @@ class Consumer:
             if deprecated_no_retry_on_startup:
                 warnings.warn(
                     CPendingDeprecationWarning(
-                        "The broker_connection_retry configuration setting will no longer determine"
+                        "The broker_connection_retry configuration setting will no longer determine\n"
                         "whether broker connection retries are made during startup in Celery 6.0 and above.\n"
-                        "If you wish to refrain from retrying connections on startup, "
+                        "If you wish to refrain from retrying connections on startup,\n"
                         "you should set broker_connection_retry_on_startup to False instead.")
                 )
             # Retry disabled, just call connect directly.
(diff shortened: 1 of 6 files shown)


celery/worker/consumer/consumer.py
```diff
@@ -323,12 +324,21 @@ def start(self):
...
+                connection_retry_type = ('broker_connection_retry_on_startup'
+                                         if (is_connection_loss_on_startup
+                                             and self.app.conf.broker_connection_retry_on_startup is not None)
+                                         else 'broker_connection_retry')
+                connection_retry = self.app.conf[connection_retry_type]
+                if not connection_retry:
+                    raise WorkerShutdown(
+                        f"Retrying to {'establish' if is_connection_loss_on_startup else 're-establish'} "
+                        f"a connection to the message broker after a connection loss has "
+                        f"been disabled (app.conf.{connection_retry_type}=False). Shutting down...") from exc
```
thedrow, line 339: @naomielst You should extract the message to a critical log entry (Just call [crit](https://github.com/celery/celery/blob/c889a5cb7606aec57708247aa05934d0b2dab4fc/celery/worker/consumer/consumer.py#L42) with the message as an argument)

```suggestion
                    crit(f"Retrying to {'establish' if is_connection_loss_on_startup else 're-establish'} "
                        f"a connection to the message broker after a connection loss has "
                        f"been disabled (app.conf.{connection_retry_type}=False). Shutting down...")
                    raise WorkerShutdown(1) from exc
```

The same goes for the other shutdown messages. Celery doesn't print these messages although it should since these exceptions should behave just like `sys.exit()` would.

celery/worker/consumer/consumer.py
```diff
@@ -442,10 +452,28 @@ def _error_handler(exc, interval, next_step=CONNECTION_RETRY_STEP):
...
+        # TODO: Rely only on broker_connection_retry_on_startup to determine whether connection retries are disabled.
+        #       We will make the switch in Celery 6.0.
+        deprecated_no_retry_on_startup = (self.app.conf.broker_connection_retry_on_startup is None
+                                          and not self.app.conf.broker_connection_retry)
+        if (
+            deprecated_no_retry_on_startup
+            or not self.app.conf.broker_connection_retry_on_startup
+        ):
+            if deprecated_no_retry_on_startup:
+                warnings.warn(
```
thedrow, line 469: @naomielst Please write a test which checks that we are warning about the pending deprecation here.

celery/worker/consumer/consumer.py
naomielst: Done.

celery/worker/control.py
```diff
@@ -20,6 +20,7 @@
 
 DEFAULT_TASK_INFO_ITEMS = ('exchange', 'routing_key', 'rate_limit')
 logger = get_logger(__name__)
+crit = logger.critical
```
thedrow, line 23: Please revert. This is no longer needed.

celery/worker/control.py
```diff
@@ -515,7 +516,8 @@ def autoscale(state, max=None, min=None):
 def shutdown(state, msg='Got shutdown from remote', **kwargs):
     """Shutdown worker(s)."""
     logger.warning(msg)
-    raise WorkerShutdown(msg)
+    logger.critical(msg)
```
thedrow, line 519: We already provide the message as a warning here. Don't change existing behavior.

celery/worker/state.py
```diff
@@ -86,7 +90,8 @@ def maybe_shutdown():
     if should_terminate is not None and should_terminate is not False:
         raise WorkerTerminate(should_terminate)
     elif should_stop is not None and should_stop is not False:
-        raise WorkerShutdown(should_stop)
+        crit(should_stop)
+        raise WorkerShutdown(1)
```
thedrow, line 94: Please revert. should_stop is a boolean.

celery/worker/state.py
```diff
@@ -17,6 +17,10 @@
 from celery import __version__
 from celery.exceptions import WorkerShutdown, WorkerTerminate
 from celery.utils.collections import LimitedSet
+from celery.utils.log import get_logger
+
+logger = get_logger(__name__)
+crit = logger.critical
```
thedrow, line 23: Please revert.

celery/worker/consumer/consumer.py
```diff
@@ -323,12 +324,22 @@ def start(self):
...
+                connection_retry = self.app.conf[connection_retry_type]
+                if not connection_retry:
+                    crit(
+                        f"Retrying to {'establish' if is_connection_loss_on_startup else 're-establish'} "
+                        f"a connection to the message broker after a connection loss has "
+                        f"been disabled (app.conf.{connection_retry_type}=False). Shutting down...")
+                    raise WorkerShutdown(1) from exc
                 if isinstance(exc, OSError) and exc.errno == errno.EMFILE:
-                    raise  # Too many open files
+                    raise WorkerTerminate("Too many open files. Aborting...") from exc
```
thedrow, line 342: @naomielst What about here?

naomielst: You're right, added.

celery/worker/consumer/consumer.py
naomielst: Done

t/unit/worker/test_consumer.py
```diff
@@ -210,18 +214,37 @@ def bp_start(*args, **kwargs):
...
     def _closer(self, c):
         def se(*args, **kwargs):
             c.blueprint.state = CLOSE
         return se
 
+    def test_blueprint_restart_when_state_not_in_stop_conditions(self):
+        c = self.get_consumer()
+
+        # ensure that WorkerShutdown is not raised
+        c.app.conf['broker_connection_retry'] = True
```
thedrow, line 234: `broker_connection_retry` should be either True or False. In either of these cases, no warnings should be emitted.

t/unit/worker/test_consumer.py
```diff
@@ -306,6 +329,52 @@ def test_cancel_long_running_tasks_on_connection_loss__warning(self):
         with pytest.deprecated_call(match=CANCEL_TASKS_BY_DEFAULT):
             c.on_connection_error_after_connected(Mock())
 
+    def test_ensure_connected_warns_if_deprecated_no_retry_on_startup(self):
+        c = self.get_consumer()
+        c.app.conf.broker_connection_retry_on_startup = None
+        c.app.conf.broker_connection_retry = False
```
thedrow, line 335: What should happen if `broker_connection_retry` is True and `broker_connection_retry_on_startup` is still None?

t/unit/worker/test_consumer.py
```diff
@@ -210,18 +214,37 @@ def bp_start(*args, **kwargs):
...
+        c = self.get_consumer()
+
+        # ensure that WorkerShutdown is not raised
+        c.app.conf['broker_connection_retry'] = True
+        c.app.conf['broker_connection_retry_on_startup'] = True
+        c.restart_count = -1
+
+        # ensure that blueprint state is not in stop conditions
+        c.blueprint.state = bootsteps.RUN
+        c.blueprint.start.side_effect = ConnectionError()
```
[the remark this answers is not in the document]
naomielst: In this test I'm recreating a flow that gets me to [`blueprint.restart` ](https://github.com/celery/celery/blob/fac135ebb8367104c8507b77c389bd3e5d25b8a4/celery/worker/consumer/consumer.py#L349). That is why I mock an error which isn't caught and thrown as a WorkerShutdown or as a WorkerTerminate. 

Testing `ensure_connected()` is an entirely different test. What did I miss here?

t/unit/worker/test_consumer.py
naomielst: In that case, `deprecated_no_retry_on_startup = False`, so the warning is not triggered.

t/unit/worker/test_consumer.py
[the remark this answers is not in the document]
thedrow: You're missing that this is a different code path and that the `ConnectionError` is raised in https://github.com/celery/celery/pull/7300/files#diff-15905920605b6295205bfe3661dec9557f310e0cad1d14d17fe060e895323cd0R483 when we don't restart the connection to the broker on startup. Therefore, you're testing the wrong unit.

t/unit/worker/test_consumer.py
thedrow: That sounds like a bug to me since we want `broker_connection_retry_on_startup` to indicate if we retry connection losses on startup.
What do you think?

thedrow: Maybe we should output a different deprecation warning message instead of not printing anything at all.
We need to let the users know they should set `broker_connection_retry_on_startup` to `True` so they'll have a safe migration path to the next major version.

naomielst: This is a very good idea. I will add the logic and test coverage.

naomielst: Added

t/unit/worker/test_consumer.py
```diff
@@ -306,6 +322,62 @@ def test_cancel_long_running_tasks_on_connection_loss__warning(self):
...
+            c.ensure_connected(Mock())
+
+    @pytest.mark.parametrize("broker_connection_retry_on_startup", [None, False])
+    def test_ensure_connected_calls_connect_when_retry_is_disabled(self, broker_connection_retry_on_startup):
+        c = self.get_consumer()
+        c.app.conf.broker_connection_retry = False
+        c.app.conf.broker_connection_retry_on_startup = broker_connection_retry_on_startup
+        conn = Mock()
+        c.ensure_connected(conn)
+        conn.connect.assert_called_with()
```
thedrow, line 341: This could be one tests with two subtests.

thedrow, line 341: We don't care that `c.ensure_connected()` calls `conn.connect()`. We care about what happens when `conn.connect()` raises a `ConnectionError`
Do we get a retry or is the `ConnectionError` raised?
What's the expected outcome here?

thedrow: In case you're wondering why, they share the same setup phase.

naomielst: Gotcha. The expected outcome is that the error is raised - since retry is set to false. I updated the test accordingly.

celery/worker/consumer/consumer.py
```diff
@@ -323,12 +324,21 @@ def start(self):
             try:
                 blueprint.start(self)
             except self.connection_errors as exc:
-                # If we're not retrying connections, no need to catch
-                # connection errors
-                if not self.app.conf.broker_connection_retry:
-                    raise
+                # If we're not retrying connections, we need to properly shutdown or terminate
+                # the Celery main process instead of abruptly aborting the process without any cleanup.
+                is_connection_loss_on_startup = self.restart_count == 0
```
eo2020, line 329: `is_connection_loss_on_startup` sounds like it's a boolean.  Please initialize with `False` instead of 0.

t/unit/worker/test_consumer.py
```diff
@@ -306,6 +322,63 @@ def test_cancel_long_running_tasks_on_connection_loss__warning(self):
...
+        c.app.conf['broker_connection_retry'] = False
+        c.app.conf['broker_connection_retry_on_startup'] = broker_connection_retry_on_startup
+        c.blueprint.start.side_effect = ConnectionError()
+
+        with subtests.test("Consumer raises WorkerShutdown on connection restart"):
+            with pytest.raises(WorkerShutdown):
+                c.start()
+
+        record = caplog.records[0]
+        with subtests.test("Critical error log message is outputted to the screen"):
```
eo2020, line 374: `outputted` -> `output`

celery/worker/consumer/consumer.py
thedrow: `is_connection_loss_on_startup` is being initialized with a boolean. Read 329 again.

t/unit/worker/test_consumer.py
thedrow: Seems like we can go [either way](https://pasttenses.com/output-past-tense).

PR #7337, opened 2022-03-06
added pypy-3.9 to github actions

auvipy:


thedrow (changes requested):
Good idea but we should compile the wheels for pypy3.9 in our wheelhouse.

PR #7373, opened 2022-03-23
Update the package links.

### LINKED ISSUE #7372: Update the links in PyPI to the new site.
The [PyPI page](https://pypi.org/project/celery/) stills points to the old URLs which confuses users because those domains no longer exist and in some cases the SSL certificate is invalid.

mvaled:
Fixes #7372.


mvaled:
Hi @auvipy, I think this is an easy one :)

auvipy (approved):

PR #7325, opened 2022-02-28
Add typing to Celery

atombrella:
This is a simple bootstrap of the process, adding some types to a
few selected functions, based on comment annotations. MyPy is chosen as the
default static analyzer for the types.

It's an attempt to solve #7258 Note that the work is quite demanding.

Issue #7258, opened 2022-01-24
Use typing annotations in Celery (PEP 484)?


atombrella:
Mainly intended as a question. There are a lot of type annotations as comment (compatible with Python 2). Some of them are outdated, as new arguments have since then been added to the functions. Are there plans to add type annotations to Celery, and enforce them as well? It's my experience that it increases readability of the code (and also IDE support), and may also find some bugs. I do know that not everyone is a big fan of this new addition in Python.

I couldn't find any issue relating this; searching for `typing` and `mypy` didn't yield results (only lists with packages from `pip freeze`).

https://github.com/ilevkivskyi/com2ann is a neat tool for doing the hard work.
https://www.python.org/dev/peps/pep-0484/

This is one example of a class with both type comments. The type for `_evict` does not reflect the current arguments.
```
class Evictable:
    """Mixin for classes supporting the ``evict`` method."""

    Empty = Empty

    def evict(self):
        # type: () -> None
        """Force evict until maxsize is enforced."""
        self._evict(range=count)

    def _evict(self, limit=100, range=range):
        # type: (int) -> None
        try:
            [self._evict1() for _ in range(limit)]
        except IndexError:
            pass

    def _evict1(self):
        # type: () -> None
        if self._evictcount <= self.maxsize:
            raise IndexError()
        try:
            self._pop_to_evict()
        except self.Empty:
            raise IndexError()
```

auvipy:
we would prefer it in future version may be v7, but int the mean time gradual typing would be OK. since we are 3.7+ now and in v6 we will be 3.8+ only

auvipy:
I would like the approach of https://github.com/sqlalchemy/sqlalchemy2-stubs first

Kludex:
My suggestion is to work in a progressive way, like we did on uvicorn: https://github.com/encode/uvicorn/blob/master/setup.cfg

In other words, create a PR per file and add to the list of files covered by mypy.

@auvipy opinions?

auvipy:
I am positive

PR #7383, opened 2022-03-29
Add `mypy` to the pipeline

Kludex:
- Replaces #7325
- Related to #7258

Thanks @atombrella for the initial work!

This PR adds `mypy` to `tox` and to the pipeline. A couple of differences between this PR and the previous one:
- Apply mypy per file. Meaning that we can do progressive work just adding a new file to the `pyproject.toml`.
- Use `pyproject.toml` instead of `mypy.ini`.
- `disallow_untyped_defs` is `True` now.

This PR is intended to be minimal, so we can start working _per files basis_. I've reused the previous branch, so there are also some small type annotation fixes, which are not required for this PR.


povilasb:
Nice, type hints are much appreciated when navigating the code base.
Thank you!

Kludex:
Yep. As soon as this gets in, I'll organize a similar issue as I did on `uvicorn`: https://github.com/encode/uvicorn/issues/998

Then we can start working on it :)

celery/contrib/testing/worker.py
```diff
@@ -90,16 +90,15 @@ def start_worker(
...
+                         pool: str = 'solo',
                          loglevel=WORKER_LOGLEVEL,
-                         logfile=None,
-                         WorkController=TestWorkController,
-                         perform_ping_check=True,
-                         shutdown_timeout=10.0,
-                         **kwargs):
-    # type: (Celery, int, str, Union[str, int], str, Any, **Any) -> Iterable
+                         logfile: Union[str, int] = None,
+                         WorkController: Any = TestWorkController,
```
auvipy, line 98: shouldn't we try not to use Any?

auvipy (commented):
good job

celery/contrib/testing/worker.py
Kludex: Yes. I didn't remove the annotations added from the previous PR. Let me check them.

Kludex:
All the annotations besides `__main__.py` are not needed for this PR. If you want, I can remove the other ones.

celery/utils/saferepr.py
```diff
@@ -15,7 +15,7 @@
 from itertools import chain
 from numbers import Number
 from pprint import _recursion
-from typing import Any, AnyStr, Callable, Dict, Iterator, List, Sequence, Set, Tuple
+from typing import Any, AnyStr, Callable, Dict, Iterator, List, Optional, Sequence, Set, Tuple
```
Kludex, line 18:
```suggestion
from typing import Any, AnyStr, Callable, Dict, Iterator, List, Sequence, Set, Tuple
```

Kludex:
I've just found out about this: https://github.com/encode/uvicorn/pull/1437#issuecomment-1086574265

Might be useful for celery as well.

auvipy (approved):

Kludex:
Thanks @auvipy ! :)

I'm going to create an issue so we can organize the work towards a fully annotated celery! 🙌

auvipy:
we can just modify this https://github.com/celery/celery/issues/7258

Kludex:
@auvipy I didn't modify it because I'd like to have rights to edit. See the created issue: #7394

PR #7385, opened 2022-03-29
Ensure expiration is of float type when migrating tasks

damjankuznar:
## Description

This fixes a subtle bug when migrating tasks with `expiration` set from Redis to RabbitMQ broker. The issue is that message properties contain expiration value as a `str` and not a `float` which is expected by the `kombu.messaging.Producer.publish`. This causes a bug where expiration value is multiplied by `1000` (to convert to miliseconds) and this means a string is multiplied 1000x times, e.g.:
```python
expiration = "3000"

# from kombu/messaging.py:163
if expiration is not None:
    properties['expiration'] = str(int(expiration * 1000))

assert len(expiration) == 4000
assert expiration.startswith("3000300030003000")
```


auvipy (changes requested):
can you also add basic unit test for the change please?

auvipy (approved):

PR #7390, opened 2022-03-31
Celery set prefetch count on restart

dobosevych:

### DIFF
--- a/celery/worker/consumer/consumer.py
+++ b/celery/worker/consumer/consumer.py
@@ -622,9 +622,9 @@ class Consumer:
                 try:
                     ack_log_error_promise = promise(call_soon, (message.ack_log_error,),
                                                     on_error=self._restore_prefetch_count_after_connection_restart)
-                    reject_log_error_promise = promise(call_soon,
-                                                       (message.reject_log_error,),
-                                                       on_error=self._restore_prefetch_count_after_connection_restart)
+                    reject_log_error_promise = \
+                        promise(call_soon, (message.reject_log_error,),
+                                on_error=self._restore_prefetch_count_after_connection_restart)
 
                     if (
                         not self._maximum_prefetch_restored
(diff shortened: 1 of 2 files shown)


celery/worker/consumer/consumer.py
```diff
@@ -649,7 +649,7 @@ def on_task_received(message):
 
         return on_task_received
 
-    def _restore_prefetch_count_after_connection_restart(self, p):
+    def _restore_prefetch_count_after_connection_restart(self, p, *args):
```
dobosevych, line 652: @thedrow based on my test, I need to add *args as an argument
```
>>> import vine
>>> from vine.promises import promise
def raise_exception():
    raise KeyError('Foo')

>>> p = promise(raise_exception, on_error=print)
>>> p()
Foo
```

celery/worker/consumer/consumer.py
```diff
@@ -622,9 +622,9 @@ def on_task_received(message):
                 try:
                     ack_log_error_promise = promise(call_soon, (message.ack_log_error,),
                                                     on_error=self._restore_prefetch_count_after_connection_restart)
-                    reject_log_error_promise = promise(call_soon,
-                                                       (message.reject_log_error,),
-                                                       on_error=self._restore_prefetch_count_after_connection_restart)
+                    reject_log_error_promise = \
```
thedrow, line 625: Is there something we can do with this code formatting? It's not readable.

thedrow (changes requested):

celery/worker/consumer/consumer.py
dobosevych: agree, I will improve that
whether there are plans to use something like black?

dobosevych: improved formating

auvipy (approved):

Issue #7394, opened 2022-04-02
Full Type Annotaed Celery


Kludex:
# Fully Type Annotated Celery

As per #7383, we added `mypy` to the pipeline, so we could start working on annotating this project in a way that work gets splitted.

Below, you can see a table of all celery files. If you want to contribute to our goal:
1. Choose a file with ❌ .
2. Comment here mentioning the PR, so I (or someone else with rights) can update the table.

To get more context about how this idea started refer to #7258.

# Annotate Celery
|  Status  |                Filename                 |  PR   |
|  :----:  |  -------------------------------------  |  ---  |
|   :x:    |  **\_\_init\_\_.py**                    |       |
|   ✅     |  **\_\_main\_\_.py**                    |   #7383    |
|   :x:    |  **\_state.py**                         |       |
|   :x:    |  **app/\_\_init\_\_.py**                |       |
|   :x:    |  **app/amqp.py**                        |       |
|   :x:    |  **app/annotations.py**                 |       |
|   :x:    |  **app/autoretry.py**                   |       |
|   :x:    |  **app/backends.py**                    |       |
|   :x:    |  **app/base.py**                        |       |
|   :x:    |  **app/builtins.py**                    |       |
|   :x:    |  **app/control.py**                     |       |
|   :x:    |  **app/defaults.py**                    |       |
|   :x:    |  **app/events.py**                      |       |
|   :x:    |  **app/log.py**                         |       |
|   :x:    |  **app/registry.py**                    |       |
|   :x:    |  **app/routes.py**                      |       |
|   :x:    |  **app/task.py**                        |       |
|   :x:    |  **app/trace.py**                       |       |
|   :x:    |  **app/utils.py**                       |       |
|   :x:    |  **apps/\_\_init\_\_.py**               |       |
|   :x:    |  **apps/beat.py**                       |       |
|   :x:    |  **apps/multi.py**                      |       |
|   :x:    |  **apps/worker.py**                     |       |
|   :x:    |  **backends/\_\_init\_\_.py**           |       |
|   :x:    |  **backends/arangodb.py**               |       |
|   :x:    |  **backends/asynchronous.py**           |       |
|   :x:    |  **backends/azureblockblob.py**         |       |
|   :x:    |  **backends/base.py**                   |       |
|   :x:    |  **backends/cache.py**                  |       |
|   :x:    |  **backends/cassandra.py**              |       |
|   :x:    |  **backends/consul.py**                 |       |
[...]

PR #7396, opened 2022-04-02
Annotate `celery/exceptions.py`

Kludex:
- Related to #7394

### DIFF
--- a/celery/exceptions.py
+++ b/celery/exceptions.py
@@ -51,11 +51,17 @@ Error Hierarchy
         - :exc:`~celery.exceptions.WorkerShutdown`
 """
 
+from datetime import datetime
 import numbers
+from types import TracebackType
+from typing import Any, Optional, Tuple, Type, TypeVar, TYPE_CHECKING, Union
 
 from billiard.exceptions import SoftTimeLimitExceeded, Terminated, TimeLimitExceeded, WorkerLostError
 from click import ClickException
 from kombu.exceptions import OperationalError
+
+if TYPE_CHECKING:
+    from celery.canvas import Signature
 
 __all__ = (
     'reraise',
@@ -101,7 +107,10 @@ Task of kind {0} never registered, please make sure it's imported.\
 """
 
 
-def reraise(tp, value, tb=None):
+_T = TypeVar("_T", bound="Exception")
+
+
+def reraise(tp: Type[_T], value: _T, tb: Optional[TracebackType] = None) -> _T:
     """Reraise exception."""
     if value.__traceback__ is not tb:
         raise value.with_traceback(tb)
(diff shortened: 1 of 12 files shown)


auvipy (changes requested):
please rebase & would be great if some unit test coverage can be increased

Kludex:
I don't want to add tests on those PRs, as I'm not including any functionally I don't think it makes sense (or is fair).

That being said... I'm willing to create a plan to increase the coverage of the whole project to 100%, analogous to the plan created to add annotations (progressive, and by file(s)). I did https://github.com/encode/uvicorn/issues/102 for uvicorn.

If you agree, I can create a similar issue, and setup for it as well (so the coverage doesn't drop, only goes up).

GautamDhiman (approved):
LGTM

celery/exceptions.py
```diff
@@ -144,17 +153,18 @@ class Retry(TaskPredicate):
...
     #: Exception (if any) that caused the retry to happen.
-    exc = None
+    exc: Optional[Exception] = None
+    excs: Optional[str] = None
 
-    #: Time of retry (ETA), either :class:`numbers.Real` or
-    #: :class:`~datetime.datetime`.
-    when = None
+    #: Time of retry (ETA)
+    when: Optional[Union[datetime, numbers.Number]] = None
```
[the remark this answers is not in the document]
Nusnus: I think you can leave it like that, but keep the original doc with the :class:`numbers.Real` doc, being a more specialized doc compared to the "general" annotation as `numbers.Number`.

What do you think?

Nusnus (approved):
Left one comment but not something critical to change, food for thought.
Besides, LGTM too.

celery/exceptions.py
```diff
@@ -144,17 +153,18 @@ class Retry(TaskPredicate):
...
 
     #: Exception (if any) that caused the retry to happen.
-    exc = None
+    exc: Optional[Exception] = None
+    excs: Optional[str] = None
 
-    #: Time of retry (ETA), either :class:`numbers.Real` or
-    #: :class:`~datetime.datetime`.
-    when = None
+    #: Time of retry (ETA)
```
Kludex, line 162:
```suggestion
    #: Time of retry (ETA), either :class:`numbers.Number` or
    #: :class:`~datetime.datetime`.
```
@Nusnus  I didn't get what you said. Is this enough?

auvipy: :class:numbers.Real is what he suggested as per my understanding

Kludex: So... A mismatch between doc and code? 🧐

Nusnus: TBH you're right. My original thought was that the doc would mention Real and the annotation would mention Number, being _upper in hierarchy_, but that is wrong as you are noting @Kludex  :)

PR #7398, opened 2022-04-02
Annotate `celery/security/certificate.py`

Kludex:
- Related to #7394

I've disabled `warn_return_any`, as it was not intended to be included in the configuration. Also, there are some issues with it: https://github.com/python/mypy/issues/5697. In this file, there were some comparison (`__eq__` and `__gt__`) that caused the issue.

### DIFF
--- a/celery/security/certificate.py
+++ b/celery/security/certificate.py
@@ -4,7 +4,7 @@ from __future__ import annotations
 import datetime
 import glob
 import os
-from typing import TYPE_CHECKING, Dict, Iterator, Union
+from typing import TYPE_CHECKING, Iterator
 
 from cryptography.hazmat.backends import default_backend
 from cryptography.hazmat.primitives.asymmetric import padding, rsa
@@ -70,7 +70,7 @@ class Certificate:
                 mgf=padding.MGF1(digest),
                 salt_length=padding.PSS.MAX_LENGTH)
 
-            self.get_pubkey().verify(signature, ensure_bytes(data), padd, digest)
+            self.get_pubkey().verify(signature, ensure_bytes(data), pad, digest)
 
 
 class CertStore:
--- a/celery/security/certificate.py
+++ b/celery/security/certificate.py
@@ -2,6 +2,7 @@
 import datetime
 import glob
 import os
+from typing import TYPE_CHECKING, Dict, Iterator, Union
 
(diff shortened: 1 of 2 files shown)


auvipy (changes requested):
please rebase & would be great if some unit test coverage can be increased

celery/security/certificate.py
```diff
@@ -2,6 +2,7 @@
 import datetime
```
thedrow, line 2:
```suggestion
from __future__ import annotations

import datetime
```

thedrow (changes requested):
Since we support Python 3.7, all modules with annotations should import annotations from `__future__`. See [StackOverflow](https://stackoverflow.com/a/61544901/85140) for why.

CC @auvipy

celery/security/certificate.py
Kludex: That's not quite what is written on that answer... That's only about postponed evaluation of annotations.

So instead of:
```python
    def verify(self, data: bytes, signature: bytes, digest: Union["HashAlgorithm", "Prehashed"]) -> None:
```
I would write:
```python
    def verify(self, data: bytes, signature: bytes, digest: Union[HashAlgorithm, Prehashed]) -> None:
```

thedrow: What I meant of course is that this import would not be necessary if we only supported 3.8+.

Kludex: I see the benefits now: https://lukasz.langa.pl/61df599c-d9d8-4938-868b-36b67fdb4448/

Nusnus:
Hey @Kludex , 
We are trying to increase the paste towards releasing Celery 5.3 (Currently `5.3.0b2`) and this PR looks very good and if I am not mistaken there isn't much work to get it finished.

Would you mind pushing this forward?
We'd be happy to have it for 5.3.
I can assist with quick response/review/merge.

Thanks!

Kludex:
What is missing? Rebase?

auvipy:
a re base should be OK

Nusnus:
> What is missing? Rebase?

I believe that's all my friend

Kludex:
Done :+1:

auvipy (approved):

Nusnus:
Thanks for closing this! 💯

PR #7399, opened 2022-04-03
Annotate `celery/fixups`

Kludex:
- Related to #7394

### DIFF
--- a/celery/fixups/django.py
+++ b/celery/fixups/django.py
@@ -4,12 +4,26 @@ import sys
 import warnings
 from datetime import datetime
 from importlib import import_module
+from typing import IO, TYPE_CHECKING, Any, List, Optional, cast
 
 from kombu.utils.imports import symbol_by_name
 from kombu.utils.objects import cached_property
 
 from celery import _state, signals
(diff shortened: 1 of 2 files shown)


celery/fixups/django.py
```diff
@@ -43,18 +57,19 @@ def fixup(app, env='DJANGO_SETTINGS_MODULE'):
         else:
             _verify_django_version(django)
             return DjangoFixup(app).install()
+    return None
```
Kludex, line 60: > Be consistent in return statements. Either all return statements in a function should return an expression, or none of them should. If any return statement returns an expression, any return statements where no value is returned should explicitly state this as return

https://peps.python.org/pep-0008/#programming-recommendations

It's either this or `# type: ignore[return]`. I'm not attached to any option.

auvipy (changes requested):

auvipy:
OK let's go with some of them

Kludex:
Thanks 👍 

I'll try to spend more time on each `Any` to see if I can do better on the next ones. 👍

PR #7401, opened 2022-04-03
Annotate `celery/security`

Kludex:
- Related to #7394

I had to remove `utils.py` and `serialization.py` from the `mypy` files, as there are some `mypy` errors that are sensitive i.e. I'd need to implement further changes (which can be considered breaking changes). I don't want to implement breaking changes on annotation PRs, so I'll open separate ones later on. If you want to know which are the sensitive ones I'm mentioning, you just need to add those two files on the `pyproject.toml`, and you'll understand the changes that are needed.

### DIFF
--- a/celery/security/__init__.py
+++ b/celery/security/__init__.py
@@ -1,10 +1,17 @@
 """Message Signing Serializer."""
+from typing import TYPE_CHECKING, Iterable, Literal, Optional
+
 from kombu.serialization import disable_insecure_serializers as _disable_insecure_serializers
 from kombu.serialization import registry
 
 from celery.exceptions import ImproperlyConfigured
 
 from .serialization import register_auth  # : need cryptography first
+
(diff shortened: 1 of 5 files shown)


celery/security/__init__.py
```diff
@@ -1,11 +1,25 @@
...
+    import sys
+
+    from celery.app.base import Celery, Settings
+
+    if sys.version_info < (3, 8):
+        from typing_extensions import Literal
+    else:
+        from typing import Literal
+
+    _Serializer = Literal["json", "msgpack", "yaml", "pickle"]
```
Kludex, line 21: Here is kind of tricky, because the user can also add custom serializers. Meaning that the annotation should be `str` instead. I'm open to change it.

Also, not related, but I have two `_Serializer` in this PR, on different files. I accept suggestion on where to put them, if duplicated is not wanted.

PR #7405, opened 2022-04-03
Expose more debugging information when receiving unkown tasks

thedrow:

## Description

Since the message might have been delivered to the wrong worker due to a
routing error,
we need to emit the headers and delivery_info when logging the error as
well as the message's body.


auvipy:
same PR twice?

PR #7406, opened 2022-04-03
load_extension_class_names - correct module_name

DamianZaremba:
95015a changed over to using importlib rather than pkg_resources,
unfortunately the object is not exactly the same.

Attempting to start up a celery instance with `django-celery-results`
installed results in an exception during `load_extension_class_names`;

```

  File "/usr/lib/python3.10/site-packages/celery/worker/worker.py", line 203, in start
    self.blueprint.start(self)
  File "/usr/lib/python3.10/site-packages/celery/bootsteps.py", line 112, in start
    self.on_start()
  File "/usr/lib/python3.10/site-packages/celery/apps/worker.py", line 136, in on_start
    self.emit_banner()
  File "/usr/lib/python3.10/site-packages/celery/apps/worker.py", line 170, in emit_banner
    ' \n', self.startup_info(artlines=not use_image))),
  File "/usr/lib/python3.10/site-packages/celery/apps/worker.py", line 232, in startup_info
    results=self.app.backend.as_uri(),
  File "/usr/lib/python3.10/site-packages/celery/app/base.py", line 1252, in backend
    self._local.backend = new_backend = self._get_backend()
  File "/usr/lib/python3.10/site-packages/celery/app/base.py", line 955, in _get_backend
    backend, url = backends.by_url(
  File "/usr/lib/python3.10/site-packages/celery/app/backends.py", line 69, in by_url
    return by_name(backend, loader), url
  File "/usr/lib/python3.10/site-packages/celery/app/backends.py", line 47, in by_name
    aliases.update(load_extension_class_names(extension_namespace))
  File "/usr/lib/python3.10/site-packages/celery/utils/imports.py", line 146, in load_extension_class_names
    yield ep.name, ':'.join([ep.module_name, ep.attrs[0]])
AttributeError: 'EntryPoint' object has no attribute 'module_name'
```

Move over to using the direct value should resolve this issue;

```
>>> from pkg_resources import iter_entry_points
>>> list(iter_entry_points('celery.result_backends'))[0].__dict__
{'name': 'django-cache', 'module_name': 'django_celery_results.backends', 'attrs': ('CacheBackend',), 'extras': (), 'dist': django-celery-results 2.3.0 (/usr/lib/python3.10/site-packages)}
```
vs
```
>>> from importlib.metadata import entry_points
>>> entry_points().get('celery.result_backends')[0]
EntryPoint(name='django-cache', value='django_celery_results.backends:CacheBackend', group='celery.result_backends')
```

### DIFF
--- a/celery/utils/imports.py
+++ b/celery/utils/imports.py
@@ -143,7 +143,7 @@ def gen_task_name(app, name, module_name):
 
 def load_extension_class_names(namespace):
     for ep in entry_points().get(namespace, []):
-        yield ep.name, ':'.join([ep.module_name, ep.attrs[0]])
+        yield ep.name, ep.value
 
 
 def load_extension_classes(namespace):


auvipy (changes requested):
can you please add unit test for the change please?

jensenbox:
I would consider this a high priority fix. 5.2.5 is essentially a broken release for me when using django with the django-db backend.

thedrow:
> I would consider this a high priority fix. 5.2.5 is essentially a broken release for me when using django with the django-db backend.

We will yank it after releasing 5.2.6.

DamianZaremba:
> can you please add unit test for the change please?

I can have a look at this later, but I'm not sure of the best way now... I would normall use pkg_resource to injection a mock endpoint, but the intent of the original change was to move away from that.

Accepting an implementation as a parameter for this function seems overkill, it would likely double the size of the code contained within it.

Do you have a suggestion for how to go about it in this code base?

thedrow:
Well I think that we need some sort of an integration test to verify it.
We can use pytest-virtualenv to install the checked out celery commit and django-celery-results and verify that the call to `load_extension_class_names` outputs to stdout what we expect.
I've done something similar in a fixture for our new pytest plugin that we're working on so you have an example to work with.
See https://github.com/celery/pytest-celery/blob/52e5d59f3f9931af219d5fc9876a9c18dc6f21f3/tests/integration/conftest.py#L30-L38

However, I do think we need to:
- [ ] Create a new PR and merge this contribution into the 5.2
- [ ] Release Celery 5.2.6
- [ ] Yank 5.2.5 from PyPi.

thedrow:
Done. This is already released.

To merge this into master, we need to write an integration test. @DamianZaremba What do you think about my plan to create it?

PR #7437, opened 2022-04-06
Update documentation on docs/userguide/application.rst file

krauss:
## Description

The documentation on [application](docs/userguide/application.rst) needs to be updated.
The call to `app.worker_main()` below no longer works without the `argv` argument

```python
from celery import Celery
app = Celery()

@app.task
def add(x, y): return x + y

if __name__ == '__main__':
    app.worker_main()  # <-- It's missing the 'argv' argument
```

So it should look like this:
```python
app.worker_main(argv=['worker', '--loglevel=INFO', '-Q celery', '--autoscale=10,4']) 
```

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,9 +1,9 @@
 [bumpversion]
-current_version = 5.2.6
+current_version = 5.2.5
 commit = True
 tag = True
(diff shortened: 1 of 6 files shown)


auvipy (changes requested):
please send exact commit from a new branch to master. your changes are not reflecting in the pr

krauss:
Hey @auvipy, I've created another one (#7438), which is the correct one. This one can be deleted.

PR #7441, opened 2022-04-07
revert celery#5941 so note below makes sense again

tisdall:
## Description

#5941 removed `raise` in the docstring example for `.retry()` but didn't edit the "Note" directly below that explained why the `raise` was there and that it was optional.   This reverts that change so that the docstring makes sense again.  This also [follows the convention that seems to be used several times in the docs](https://docs.celeryq.dev/en/stable/userguide/tasks.html).


auvipy:
@[AlexRiina](https://github.com/AlexRiina) can you verify

auvipy (approved):

PR #7447, opened 2022-04-10
Ensure a proper error message is raised when id for key is empty

thedrow:

## Description

I've encountered cases where a worker running the chord's body has crashed which rises an exception since
the group_id is None for some reason.
This PR adds a nicer error message so we'll see the problem more clearly later.


auvipy (changes requested):
can you please fix the build failures?

auvipy (approved):

PR #7476, opened 2022-04-21
Use inspect.getgeneratorstate in asynpool.gen_not_started

colesbury:
This improves compatibility with the nogil Python fork, which does not
have the gi_frame attribute on generators.

### DIFF
--- a/celery/concurrency/asynpool.py
+++ b/celery/concurrency/asynpool.py
@@ -14,6 +14,7 @@ This code deals with three major challenges:
 """
 import errno
 import gc
+import inspect
 import os
 import select
 import time
@@ -89,8 +90,7 @@ Ack = namedtuple('Ack', ('id', 'fd', 'payload'))
 
 def gen_not_started(gen):
     """Return true if generator is not started."""
-    # gi_frame is None when generator stopped.
-    return gen.gi_frame and gen.gi_frame.f_lasti == -1
+    return inspect.getgeneratorstate(gen) == "GEN_CREATED"
 
 
 def _get_job_writer(job):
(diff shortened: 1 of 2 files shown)


celery/concurrency/asynpool.py
```diff
@@ -89,8 +90,7 @@ def unpack_from(fmt, iobuf, unpack=unpack):  # noqa
 
 def gen_not_started(gen):
     """Return true if generator is not started."""
-    # gi_frame is None when generator stopped.
-    return gen.gi_frame and gen.gi_frame.f_lasti == -1
+    return inspect.getgeneratorstate(gen) == "GEN_CREATED"
```
auvipy, line 93: is this change backward compatible?

auvipy (changes requested):

celery/concurrency/asynpool.py
colesbury: Yes - `inspect.getgeneratorstate` is available since Python 3.2 and this function returns `True` in the same cases as the previous implementation (when the generator is created, but not yet started).

There is some testing of this in `test_AsynPool::test_gen_not_started`. I also did a bit more [manual testing](https://gist.github.com/c13541de1cd0426111ab8bcf58b0aaa5) with Python 3.7, 3.10, and PyPy 7.3.9.

auvipy: is it possible to add some additonal testing for the change?

colesbury: Sure. I added assertions for a few additional cases:

- while the generator is running
- after the generator exited due to an exception

That's all the cases I could think of.

auvipy (approved):

PR #7497, opened 2022-04-28
pyzmq>=22.3.0

auvipy:


dobosevych:
Rabbitmq failed, as the task timeout
I will take a look for optimization that would allow to skip tests faster
The test_mutable_errback_called_by_chord_from_group_fail_multiple[errback_old_style] failed, as the performance between delay and get is to good. I didn't investigate in detail, but it seems that adding 1s sleep will resolve an issue.

PR #7543, opened 2022-05-31
[Documentation] Fix order of arguments for clarity

EricAtORS:
## Description

I'm a little new to celery, but if I understand callbacks correctly, there's a typo in the documentation:
the order of the arguments applied will be `8+4` because the `8` comes first in the `signature` and then the `4` gets applied second.


thedrow:
This is incorrect. Links are callbacks that get called *after* the task is executed.

thedrow (approved):

PR #7575, opened 2022-06-16
MAINT: use f-string format

cyai:

## Description
Updated to f-string format for a clear and better understanding of code.


auvipy (changes requested):
thanks for the patch, can u also check the lint error plz?

PR #7587, opened 2022-06-19
Minor refactors, found by static analysis

gabrielsoldani:

## Description

This patch contains some simple refactors and cleanups found using SonarQube Community Edition and LGTM.

### DIFF
--- a/celery/app/trace.py
+++ b/celery/app/trace.py
@@ -369,7 +369,7 @@ def build_tracer(name, task, loader=None, hostname=None, store_errors=True,
     from celery import canvas
     signature = canvas.maybe_signature  # maybe_ does not clone if already
 
-    def on_error(request, exc, uuid, state=FAILURE, call_errbacks=True):
+    def on_error(request, exc, state=FAILURE, call_errbacks=True):
         if propagate:
             raise
         I = Info(state, exc)
@@ -459,10 +459,10 @@ def build_tracer(name, task, loader=None, hostname=None, store_errors=True,
                     traceback_clear(exc)
                 except Retry as exc:
                     I, R, state, retval = on_error(
-                        task_request, exc, uuid, RETRY, call_errbacks=False)
+                        task_request, exc, RETRY, call_errbacks=False)
                     traceback_clear(exc)
                 except Exception as exc:
-                    I, R, state, retval = on_error(task_request, exc, uuid)
+                    I, R, state, retval = on_error(task_request, exc)
                     traceback_clear(exc)
                 except BaseException:
                     raise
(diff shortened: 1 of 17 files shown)


celery/local.py
```diff
@@ -148,12 +148,6 @@ def __setitem__(self, key, value):
     def __delitem__(self, key):
         del self._get_current_object()[key]
 
-    def __setslice__(self, i, j, seq):
```
auvipy, line 151: can you please verify this removals won't create any regression in the code?

auvipy (changes requested):
good work btw

celery/backends/filesystem.py
```diff
@@ -53,10 +53,9 @@ def __init__(self, url=None, open=open, unlink=os.unlink, sep=os.sep,
         # Lets verify that we've everything setup right
         self._do_directory_test(b'.fs-backend-' + uuid().encode(encoding))
 
-    def __reduce__(self, args=(), kwargs={}):
-        kwargs.update(
-            dict(url=self.url))
-        return super().__reduce__(args, kwargs)
+    def __reduce__(self, args=(), kwargs=None):
+        kwargs = {} if not kwargs else kwargs
+        return super().__reduce__(args, dict(kwargs, url=self.url))
```
gabrielsoldani, line 58:
```suggestion
        return super().__reduce__(args, {**kwargs, 'url': self.url})
```

With this change it works. Is this preferable or should I go with the first option? I usually prefer avoiding a string literal if possible.

celery/local.py
gabrielsoldani: These were [deprecated back in the Python 2 days](https://docs.python.org/2.7/reference/datamodel.html#additional-methods-for-emulation-of-sequence-types) and were removed in Python 3, which [passes `slice` objects to the `__getitem__`, `__setitem__` and `__delitem__` methods instead](https://docs.python.org/3/reference/datamodel.html?highlight=__getitem__#object.__getitem__).

I can confirm they are not being used at the moment:

```
>>> from celery.local import Proxy
>>> a = list(range(10))
>>> proxy = Proxy(lambda: a)
>>> print(proxy)
[0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
>>> print(proxy[2:])
[2, 3, 4, 5, 6, 7, 8, 9]
>>> proxy[2:] = [0]
>>> print(proxy)
[0, 1, 0]
>>> del proxy[2:]
>>> print(proxy)
[0, 1]
```

celery/app/trace.py
```diff
@@ -369,7 +369,7 @@ def build_tracer(name, task, loader=None, hostname=None, store_errors=True,
     from celery import canvas
     signature = canvas.maybe_signature  # maybe_ does not clone if already
 
-    def on_error(request, exc, uuid, state=FAILURE, call_errbacks=True):
```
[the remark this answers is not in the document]
gabrielsoldani: Yes, static analysis showed the parameter was unused and this function is in the local scope of the `build_tracer` function, so the function signature isn't part of a public API.

It also passes the relevant integration tests.

gabrielsoldani:
@auvipy I've pushed the changes from #7588 to this branch as you requested and addressed your comments and suggestions :)

auvipy (approved):

PR #7593, opened 2022-06-22
Improve workflow primitive subclassing

EricAtORS:
## Description

In my application, I am subclassing the primitive `group` in order to override the `apply_async` function.
However, my actual workflow is something like:
```
group( ... ) | merge_results.s(...)
```
which ends up creating a chord.
The problem is when the chord calls `group.clone`, the new group is of type `celery.canvas.group` instead of `mymodule.my.group`, even though my type is registered correctly.

The issue arises from the call to `from_dict` which always creates a new group of type `clery.canvas.group` and ignoring the `cls` argument. 

This PR seeks to change this by letting the `from_dict` method of group to create the correct `group` type.
 Looking at the other primitives, similar issues could arise in chunks and chain, so those were fixed as well.

### DIFF
--- a/celery/canvas.py
+++ b/celery/canvas.py
@@ -897,7 +897,10 @@ class chain(_chain):
                 tasks = tasks[0] if len(tasks) == 1 else tasks
                 # if is_list(tasks) and len(tasks) == 1:
                 #     return super(chain, cls).__new__(cls, tasks, **kwargs)
-                return reduce(operator.or_, tasks, cls())
+                new_instance = reduce(operator.or_, tasks, _chain())
+                if cls != chain and isinstance(new_instance, _chain) and not isinstance(new_instance, cls):
+                    return super().__new__(cls, new_instance.tasks, **kwargs)
+                return new_instance
         return super().__new__(cls, *tasks, **kwargs)
 
 
--- a/celery/canvas.py
+++ b/celery/canvas.py
@@ -558,7 +558,7 @@ class _chain(Signature):
             if isinstance(tasks, tuple):  # aaaargh
                 tasks = d['kwargs']['tasks'] = list(tasks)
             tasks = [maybe_signature(task, app=app) for task in tasks]
-        return _chain(tasks, app=app, **d['options'])
+        return cls(tasks, app=app, **d['options'])
 
     def __init__(self, *tasks, **options):
         tasks = (regen(tasks[0]) if len(tasks) == 1 and is_list(tasks[0])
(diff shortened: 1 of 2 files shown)


auvipy (changes requested):
i would like to see unit & integration test for this change

EricAtORS:
@auvipy, sorry, I don't use github very much. How do I mark it as needs revisions again?

EricAtORS:
@auvipy 
I don't quite understand why the changes are causing the tests to fail. The tests are passing on my machine locally.
I have a hard time how test system works as I wasn't able to build the test environment in docker using tox.
Do you think you can point me in the right direction for what I should look for?

In any case,I'm not sure if I agree with adding integration tests.
There are 2 main issues, but I think it's beyond the scope of this PR.

1 - At the very least, if I subclass `group` or `chord`, then they should always return the correct types when I use them. So I 've added some unit tests that demonstrates this. But, beyond that, I'm not sure what integration tests will show more than now.
2 - At the moment, there's not a real way to change the primitives being without overriding the implementation. For example, for `group | task => chord`, in my usage, I have to override `__or__` of my group subclass otherwise, it would have just returned the default celery `chord` instead of the `chord_subclass` that I would like to use.

auvipy (changes requested):
  FAILED t/unit/tasks/test_canvas.py::test_chain::test_chain_of_chain_with_a_single_task
  FAILED t/unit/tasks/test_canvas.py::test_chain::test_accepts_generator_argument
  = 2 failed, 2663 passed, 19 skipped, 3 xfailed, 556 warnings, 54 subtests passed in 113.24s (0:01:53) =

EricAtORS:
@auvipy Sorry for all the "noise" that you may have gotten. I messed up my git merge and had to clean it up and then the coverage went down because I hadn't merged in the latest revisions that improved the coverage by 2%.

auvipy:
> 2 - At the moment, there's not a real way to change the primitives being without overriding the implementation. For example, for `group | task => chord`, in my usage, I have to override `__or__` of my group subclass otherwise, it would have just returned the default celery `chord` instead of the `chord_subclass` that I would like to use.

The PR looks good to me, but I would love to know more on this part from you

auvipy:
> I can't wait for the 5.3 release.

this will be on master and soon in 5.3a1

thedrow (approved):
Makes sense. LGTM!

auvipy (approved):

PR #7596, opened 2022-06-28
added reuse_task_id to replace()

anilaratna2:

## Description

`self.replace(sig_chain)` reuses task id 123 of the current task with taskname 1_2_3 and sets the task id of the last task in sig_chain say taskname 4_5_6 to be that task_id 123. Hence a task 1_2_3 runs with task id 123, then after some time task 4_5_6 with task id 123 again. This causes problems in task monitoring 
- worker logs show the same task id repeated
- storing task id + taskname in some db shows that 2 tasks ran with same id

I understand task id being reused in self.replace() was done to preserve AsyncResult.get() behavior. But we don't use AsyncResult.get(). We store results per task id using db backend.

Hence, this PR to allow users that dont use AsyncResult.get() to use self.replace without reusing task id . 
I believe nothing in Celery breaks if we dont reuse task id and dont use AsyncResult.get()

### DIFF
--- a/celery/app/task.py
+++ b/celery/app/task.py
@@ -886,7 +886,7 @@ class Task:
                 type_,
                 uuid=req.id, retry=retry, retry_policy=retry_policy, **fields)
 
-    def replace(self, sig):
+    def replace(self, sig, reuse_task_id=True):
         """Replace this task, with a new task inheriting the task id.
 
         Execution of the host task ends immediately and no subsequent statements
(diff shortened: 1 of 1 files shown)


auvipy (changes requested):
this will need unit test & if possible some integration tests. but unit test is a must

thedrow:
Why do you need this?
You can already tell that the task was replaced using the `replaced_task_nesting` header even if it has the same id.

As far as I recall, the problem isn't just with waiting for `AsyncResult`s but with chords, callbacks, and errbacks as well.

anilaratna2:
Yes, in chain(a, groupX(b, replaced_d), e) when replaced_d (replaces with another task) task's ID needs to be reused otherwise the groupX wont complete

PR #7597, opened 2022-06-28
async chords should pass it's kwargs to the group/body

EricAtORS:
## Description

I am trying to override the `apply_async` methods of `group` and `chord`. In my testing, when I did a `chord_instance.apply_async( args, kwargs)`, my kwargs were not sent. Looking at the code, if the chord was run in an eager mode, the kwargs were sent correctly.

I've created a fix and added 2 new test cases to cover this scenario.

### DIFF
--- a/celery/canvas.py
+++ b/celery/canvas.py
@@ -1705,7 +1705,7 @@ class _chord(Signature):
             task_id = option_task_id
 
         # chord([A, B, ...], C)
-        return self.run(tasks, body, args, task_id=task_id, **merged_options)
+        return self.run(tasks, body, args, task_id=task_id, kwargs=kwargs, **merged_options)
 
     def apply(self, args=None, kwargs=None,
               propagate=True, body=None, **options):
@@ -1755,7 +1755,7 @@ class _chord(Signature):
 
     def run(self, header, body, partial_args, app=None, interval=None,
             countdown=1, max_retries=None, eager=False,
-            task_id=None, **options):
+            task_id=None, kwargs=None, **options):
         app = app or self._get_app(body)
         group_id = header.options.get('task_id') or uuid()
         root_id = body.options.get('root_id')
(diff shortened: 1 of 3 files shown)


celery/canvas.py
```diff
@@ -1487,14 +1487,16 @@ def __length_hint__(self):
 
     def run(self, header, body, partial_args, app=None, interval=None,
             countdown=1, max_retries=None, eager=False,
-            task_id=None, **options):
+            task_id=None, partial_kwargs=None, **options):
```
auvipy, line 1490: what is the purpose of partial_kwqrs?

auvipy (changes requested):
also please check the integration test failure   INFO     celery.worker.strategy:strategy.py:161 Task celery.accumulate[f00dac24-d8a5-48e0-a5a3-adba7c40885f] received
  FAILED t/integration/test_canvas.py::test_chord::test_group_args_and_kwargs
  !!!!!!!!!!!!!!!!!!!!!!!!!! stopping after 1 failures !!!!!!!!!!!!!!!!!!!!!!!!!!!
  = 1 failed

celery/canvas.py
EricAtORS: The short answer is I think it will be ok to rename it to `kwargs`. I initially called it`partial_kwargs` to keep to the same naming converting with `partial_args` as I didn't fully grasp why `partial_args` was called `partial_args` and not just `args`, but I think I understand it now. It makes sense that `partial_args` is called partial because the results from the tasks will be the last arg.

If I missed your point and the real question is why do we even need the kwargs argument, it's because this is the async equivalent of [this](https://github.com/celery/celery/blob/59263b0409e3f02dc16ca8a3bd1e42b5a3eba36d/celery/canvas.py#L1435): when we eagerly apply the chord, the `apply` gets  both the `args` and `kwargs` while the true async call does not receive the kwargs. It is needed to fix the omission in the api. 

At the moment, I just pushed an updated commit that addresses the tests failing with the redis backend. If that's ok, then I'll update this PR with renaming `partial_kwargs` to `kwargs`

auvipy:
that is probably a network issue.

auvipy (commented):
since this is an enhancement, should we add some documentation?

auvipy (approved):

PR #7602, opened 2022-07-02
Reference Gopher Celery

marselester:
## Description

Please consider adding https://github.com/marselester/gopher-celery to a list of protocol implementations.


auvipy:
it is still in early stage

PR #7608, opened 2022-07-05
beat: Suppress banner output with the quiet option

### LINKED ISSUE #5836: celery worker has --quiet to suppress banner output but celery beat does not
Sorry for leaving out the issue template but I believe this is fairly trivial and straight-forward.

In master, there is code to enable the suppressing of printing the banner when running `celery worker` with `--quiet`:
https://github.com/celery/celery/blob/9773eba837982c84380c93bd3788470273e7674d/celery/apps/worker.py#L138-L139

This conditional is not present in the code that runs `celery beat`:
https://github.com/celery/celery/blob/9773eba837982c84380c93bd3788470273e7674d/celery/apps/beat.py#L77-L78
https://github.com/celery/celery/blob/9773eba837982c84380c93bd3788470273e7674d/celery/apps/beat.py#L100

This causes a few issues for us because we expect all our services to only emit JSON.

gabrielsoldani:

## Description

This adds missing support for the quiet command line option
(`--quiet` or `-q`) for the celery beat command, which suppresses
banner and version information output.

Fixes #5836.

### DIFF
--- a/celery/apps/beat.py
+++ b/celery/apps/beat.py
@@ -44,7 +44,8 @@ class Beat:
                  scheduler=None,
                  scheduler_cls=None,  # XXX use scheduler
                  redirect_stdouts=None,
-                 redirect_stdouts_level=None, **kwargs):
+                 redirect_stdouts_level=None,
+                 quiet=False, **kwargs):
         self.app = app = app or self.app
         either = self.app.either
         self.loglevel = loglevel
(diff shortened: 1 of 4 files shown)


auvipy (changes requested):
you need to add unit test using pytest-click, there is some test already, which you can check for ref. also documentation might be needed

auvipy:
https://github.com/celery/celery/blob/master/t/unit/bin/test_worker.py

thedrow (approved):

auvipy (approved):
thanks for the tests

PR #7612, opened 2022-07-06
Use spawn pool on POSIX platforms

### LINKED ISSUE #6036: Allow celery to spawn rather than fork
# Checklist

* [x] I have checked the [issues list](https://github.com/celery/celery/issues?utf8=%E2%9C%93&q=is%3Aissue+label%3A%22Issue+Type%3A+Feature+Request%22+)
* [x] I have checked the [pull requests list](https://github.com/celery/celery/pulls?utf8=%E2%9C%93&q=is%3Apr+label%3A%22PR+Type%3A+Feature%22+)
  master branch.

#### Related Issues
* None

#### Possible Duplicates
Similar to issue #5005 and issue #5021

# Brief Summary
When launching child processes, celery uses the fork method. Sometimes it is preferable to be able to spawn rather than fork, for instance:  https://bugs.python.org/issue33725. Celery 3.x had this ability with the `CELERYD_FORCE_EXECV` option, however it was apparently removed in version 4.0 for some reason ( https://docs.celeryproject.org/en/stable/history/whatsnew-4.0.html). Please re-introduce this feature.

# Design

None

## Proposed Behavior

This feature would behave exactly as it (presumably) did in version 3.x: using spawn to launch child workers rather than fork. It is likely this is already the behavior when run on windows anyway, since at least in the python multiprocessing module fork is not available on windows.

## Proposed UI/UX

Same as in version 3.x: provide a CELERYD_FORCE_EXECV option (although I guess it would be lowercase now)

## Diagrams

N/A

## Alternatives

None

itamarst:

## Description

Fixes #6036 

Rather than making the pool strategy configurable, this always uses `spawn`, for the following reason:

* `fork()` without `execve()` is fundamentally broken. See the many many complaints in #6036, and also https://pythonspeed.com/articles/python-multiprocessing/. I have helped multiple people fix mysterious problems caused by `fork()` without `execve()` as well in non-Celery contexts; it's just not possible to work correctly.
* `fork()` without `execve()` has no benefits! It won't save much memory: shared libraries and the Python executable will already be shared by the OS, and Python objects trigger copy-on-write anytime e.g. a refcount changes. It will speed up startup, very slightly. Neither is particularly useful.
* ~`spawn` is already what Windows uses, so it's known to work.~ Turns out Windows always uses threads (https://github.com/celery/celery/blob/master/celery/worker/components.py#L140).

So instead of an extra knob to disable a broken-by-default feature, this just uses a correct-by-default and consistent-across-platforms default, with no knob to turn it off. (The blocking issue on Windows is not related to spawn AFAICT, and Windows always spawn).

### DIFF
--- a/celery/concurrency/asynpool.py
+++ b/celery/concurrency/asynpool.py
@@ -28,6 +28,7 @@ from weakref import WeakValueDictionary, ref
 
 from billiard import pool as _pool
 from billiard.compat import isblocking, setblocking
+from billiard.context import SpawnContext
 from billiard.pool import ACK, NACK, RUN, TERMINATE, WorkersJoined
 from billiard.queues import _SimpleQueue
 from kombu.asynchronous import ERR, WRITE
@@ -460,7 +461,7 @@ class AsynPool(_pool.Pool):
 
         self.write_stats = Counter()
 
-        super().__init__(processes, *args, **kwargs)
+        super().__init__(processes, context=SpawnContext(), *args, **kwargs)
 
         for proc in self._pool:
             # create initial mappings, these will be updated
--- a/celery/concurrency/prefork.py
+++ b/celery/concurrency/prefork.py
@@ -1,4 +1,4 @@
-"""Prefork execution pool.
+"""Subprocess execution pool.
 
 Pool implementation using :mod:`multiprocessing`.
 """
@@ -6,8 +6,8 @@ import os
 
 from billiard import forking_enable
 from billiard.common import REMAP_SIGTERM, TERM_SIGNAME
+from billiard.context import SpawnProcess
 from billiard.pool import CLOSE, RUN
-from billiard.pool import Pool as BlockingPool
 
 from celery import platforms, signals
 from celery._state import _set_task_join_will_block, set_default_app
@@ -93,7 +93,10 @@ class TaskPool(BasePool):
     """Multiprocessing Pool implementation."""
 
     Pool = AsynPool
-    BlockingPool = BlockingPool
+
+    @staticmethod
+    def BlockingPool(*args, **kwargs):
+        return SpawnProcess().Pool(*args, **kwargs)
 
     uses_semaphore = True
     write_stats = None
--- a/celery/concurrency/prefork.py
+++ b/celery/concurrency/prefork.py
@@ -2,7 +2,6 @@
 
 Pool implementation using :mod:`multiprocessing`.
 """
-from functools import partial
 import os
 
 from billiard import forking_enable
--- a/celery/concurrency/prefork.py
+++ b/celery/concurrency/prefork.py
@@ -4,7 +4,6 @@ Pool implementation using :mod:`multiprocessing`.
 """
 import os
 
-from billiard import forking_enable
 from billiard.common import REMAP_SIGTERM, TERM_SIGNAME
 from billiard.pool import CLOSE, RUN
 from billiard.pool import Pool as BlockingPool
@@ -99,7 +98,6 @@ class TaskPool(BasePool):
     write_stats = None
 
     def on_start(self):
-        forking_enable(self.forking_enable)
         Pool = (self.BlockingPool if self.options.get('threads', True)
                 else self.Pool)
         proc_alive_timeout = (
--- a/celery/concurrency/prefork.py
+++ b/celery/concurrency/prefork.py
@@ -2,12 +2,13 @@
 
 Pool implementation using :mod:`multiprocessing`.
 """
+from functools import partial
 import os
 
 from billiard import forking_enable
 from billiard.common import REMAP_SIGTERM, TERM_SIGNAME
-from billiard.context import SpawnProcess
 from billiard.pool import CLOSE, RUN
+from billiard.pool import Pool as BlockingPool
 
 from celery import platforms, signals
 from celery._state import _set_task_join_will_block, set_default_app
@@ -93,10 +94,7 @@ class TaskPool(BasePool):
     """Multiprocessing Pool implementation."""
 
     Pool = AsynPool
-
-    @staticmethod
-    def BlockingPool(*args, **kwargs):
-        return SpawnProcess().Pool(*args, **kwargs)
+    BlockingPool = BlockingPool
 
     uses_semaphore = True
     write_stats = None
--- a/celery/worker/components.py
+++ b/celery/worker/components.py
@@ -160,7 +160,6 @@ class Pool(bootsteps.StartStopStep):
             threads=threaded,
             max_restarts=max_restarts,
             allow_restart=allow_restart,
-            forking_enable=True,
             semaphore=semaphore,
             sched_strategy=self.optimization,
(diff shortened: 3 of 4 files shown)


auvipy (changes requested):
thanks for tackling this, would you please add some tests for this?

itamarst:
Will do!

itamarst:
OK, I've updated with tests, and making it actually work :grin: (all existing tests used mock, to my surprise).

auvipy (changes requested):
one failure
  =========================== short test summary info ===========================
  FAILED t/unit/concurrency/test_asynpool.py::test_AsynPool::test_subprocesses_are_spawned
  = 1 failed,

auvipy:
also check the integration tests failures please

itamarst:
Will do.

itamarst:
Oops apparently I ran integration tests in previous testing but not at the right time after I got the right code path sorted. And indeed, it does not work. And since Windows was not actually using spawn, that validation of spawn working is not available... so perhaps it degraded since it was last supported.

In order for `spawn` to fail where `fork` succeeds, something must be wrong somewhere. Potential sites include:

1. Code running in the child worker relies on class or module-level state set in the parent process _before_ it was forked.
2. There's a bug in Billiard's `spawn` codepaths. Seems less likely?
3. Something involving file descriptor inheritance? And Python sets FD_CLOEXEC or Windows equivalent so that `execve()` (but not `fork()`!) closes file descriptors: https://peps.python.org/pep-0446/

Anything else?

itamarst:
Interestingly the first task (`ping` typically in tests) does succeed...

itamarst:
So I noticed billiard used to have a C extension that was supposedly required for execve mode to work. Is the extension module no longer necessary, or was it just not maintained, or something else? It's disabled on Python 3: https://github.com/celery/billiard/blob/master/setup.py#L232

itamarst:
(here's the reference to it being necessary for execv: https://github.com/celery/billiard/blob/master/setup.py#L33)

auvipy:
also upgrading/syncing the billiard codes is a long overdue

itamarst:
Here we can see some logs where the first task (`ping`) succeeds. But then there's a kombu error; the second task does not succeed:

```
INFO     celery.worker.consumer.connection:connection.py:22 Connected to amqp://guest:**@127.0.0.1:5672//
INFO     celery.worker.strategy:strategy.py:161 Task celery.ping[83f98219-6663-458c-bd01-437091802b4b] received
[2022-07-10 13:57:23,212: INFO/SpawnPoolWorker-2] child process 834713 calling self.run()
[2022-07-10 13:57:23,283: INFO/SpawnPoolWorker-2] Task celery.ping[83f98219-6663-458c-bd01-437091802b4b] succeeded in 0.025449884007684886s: 'pong'
ERROR    kombu.pidbox:pidbox.py:106 pidbox command error: AttributeError("'NoneType' object has no attribute 'groups'")
  File "/home/itamarst/devel/celery/.tox/3.7-integration-rabbitmq/lib/python3.7/site-packages/kombu/pidbox.py", line 102, in dispatch
    reply = handle(method, arguments)
  File "/home/itamarst/devel/celery/.tox/3.7-integration-rabbitmq/lib/python3.7/site-packages/kombu/pidbox.py", line 124, in handle_cast
    return self.handle(method, arguments)
  File "/home/itamarst/devel/celery/.tox/3.7-integration-rabbitmq/lib/python3.7/site-packages/kombu/pidbox.py", line 118, in handle
    return self.handlers[method](self.state, **arguments)
  File "/home/itamarst/devel/celery/celery/worker/control.py", line 275, in enable_events
    if dispatcher.groups and 'task' not in dispatcher.groups:
AttributeError: 'NoneType' object has no attribute 'groups'
```

Any idea what might cause that, and any sense if it's related?

thedrow:
> So I noticed billiard used to have a C extension that was supposedly required for execve mode to work. Is the extension module no longer necessary, or was it just not maintained, or something else? It's disabled on Python 3: https://github.com/celery/billiard/blob/master/setup.py#L232

Billiard itself needs a maintainer who knows what he's doing. The C extension, like every other component in billiard receives very little attention.

xboy2k11:
@itamarst Good day, sir.  We have had this same issue with celery in several different applications over the years. I last solved it with the spawn fix, but that's no longer an option since airflow needs a newer version of celery.  We would be happy to help test this at scale once you're confident the PR is complete.

itamarst:
Unfortunately I don't have time to work on this at the moment, so if someone else wants to take it up let me know.

ShaheedHaque:
> also upgrading/syncing the billiard codes is a long overdue

Is there some information - or documentation - on what Billiard actually *does*? What was/is the rationale for replacing multiprocessing?

sabetAI:
@itamarst I get a weird crash, and the following warning when trying to run celery with your PR:

/home//miniconda3/lib/python3.9/multiprocessing/resource_tracker.py:216: UserWarning: resource_tracker: There appear to be 1 leaked semaphore objects to clean up at shutdown

Any ideas why?

itamarst:
@sabetAI I never got this to working state, sorry... Will close to prevent confusion. Still would be nice if someone could figure out how to do this.

sabetAI:
@itamarst no worries. Do you have a dockerfile I can use to reproduce?

ShaheedHaque:
FWIW, I would love to see this taken forward. It seems clear that this core fork-versus-spawn functionality of Celery is broken in multiple use cases and with a variety of hard-to-track down symptoms, but it also seems clear that no one person has the knowledge+willingness to drive this to conclusion. Might it be possible to draw together a small group to work on this in some collaborative fashion?

sabetAI:
@ShaheedHaque @itamarst yes! In particular, Celery multi-processing doesn't play well with CUDA. Know any fixes?

auvipy:
I am not sure celery multiprocessing was designed to work with CUDA. you should try Dask / Ray for that I think

adampl:
It does work with CUDA, you just have to make 100% sure that no CUDA code is executed by the root (parent) process :)

sabetAI:
@adampl interesting, can you share a link so I can read up further on that? Do I disable CUDA on parent by running torch code only if os.getpid != os.getppid()?

adampl:
I just import all the CUDA-related stuff inside the task functions. The trick with PIDs may work too. That way no CUDA context is ever initialized in the parent process. I also use the Celery Max Tasks Per Child=1 setting so that the child process is deleted when the task is finished, to ensure full cleanup.

veonua:
>  deleted when the task is finished, to ensure full cleanup.

some models are 10+ Gb, sometimes it can be very costly to do cleanup after every task

adampl:
> sometimes it can be very costly to do cleanup after every task

Yes, but unfortunately many libraries don't properly manage GPU resources, causing subsequent tasks to fail.

auvipy:
hey good peoples! we got a new approach here to fix it https://github.com/celery/celery/pull/9810 . please check it out

Issue #4006, opened 2017-05-01
Celery no longer honoring Django's `TIME_ZONE` setting


izquierdo:
http://docs.celeryproject.org/en/latest/userguide/periodic-tasks.html states:

> For Django users the time zone specified in the TIME_ZONE setting will be used

This used to be true (checked in Celery 3.1.25) but is no longer the case in Celery 4.

I can send a patch but would like to know if the expected solution is to add back support for this, or to update the documentation to no longer indicate that this setting is read.

## Checklist

- [X] I have included the output of ``celery -A proj report`` in the issue.
      (if you are not able to do this, then at least specify the Celery
       version affected).

Version 4.0.2 (latentcall)

- [X] I have verified that the issue exists against the `master` branch of Celery.

## Steps to reproduce

Run Celery from a Django project that has the `TIME_ZONE` setting set, but which does not specify `CELERY_TIMEZONE`.

## Expected behavior

The configured timezone (e.g. `app.conf.timezone` is the same as that in Django's `TIME_ZONE` setting).

## Actual behavior

The configured timezone is `None`.

RabidCicada:
I also ran into and noticed this.  I reference this behavior on clery/django-celery-beat#57.  I would like it fixed soas to make it easier to keep them in sync from the config file.

georgepsarakis:
@izquierdo @RabidCicada I believe you need to combine it with the `USE_TZ=True`  setting. For reference see also [here](https://github.com/celery/django-celery/blob/79d9689b62db3d54ebd0346e00287f91785f6355/djcelery/tests/test_snapshot.py#L128).

RabidCicada:
USE_TZ is set to True as well documented.  This behavior persists.

georgepsarakis:
I think you also need to set `CELERY_ENABLE_UTC=False`. Check the discussion in https://github.com/celery/celery/issues/3753 .

bhrutledge:
@auvipy Can you elaborate on why this was closed? Was it fixed? Or is this now expected behavior (counter to [the docs](http://docs.celeryproject.org/en/latest/userguide/periodic-tasks.html))?

kylebebak:
@georgepsarakis @auvipy 
It's obvious this should not have been closed. This is a bug that runs counter to the documentation.

I'm running `celery==4.1.0` and `Django==1.11.6`.

[...]

PR #7613, opened 2022-07-07
Fix honor Django's TIME_ZONE setting

vanschelven:
See #4006

It appears that `app.conf.timezone` is simply `None` when `TIME_ZONE` is set in the Django settings, but the celery `timezone` setting is not set. This contrasts the documentation, which says that 

> Celery recommends and is compatible with the new USE_TZ setting introduced in Django 1.4.
> For Django users the time zone specified in the TIME_ZONE setting will be used, or you can specify a custom time zone for Celery alone by using the timezone setting.

Searching the code (case insensitive) for time_zone reveals only 1 location, the one in this commit.

It appears that's the part that's broken. Looking at the actual contents of the configuration object, I'd say that the latter `time_zone` should simply be upper-cased.

### DIFF
--- a/celery/app/utils.py
+++ b/celery/app/utils.py
@@ -128,7 +128,7 @@ class Settings(ConfigurationView):
     @property
     def timezone(self):
         # this way we also support django's time zone.
-        return self.first('timezone', 'time_zone')
+        return self.first('timezone', 'TIME_ZONE')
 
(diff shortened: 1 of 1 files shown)


auvipy:
thanks, lets see the CI

auvipy (approved):

PR #7620, opened 2022-07-11
feat: add global keyprefix for backend result keys

kaustavb12:
## Description

While using celery with a shared redis instance, we can currently add custom global key prefixes to keys used in kombu. However, there is currently no way to configure such a global key prefix when the shared redis instance is used as a result backend. This make it impossbile for multple applications to use a shared redis instance as a result backend.

This PR adds the capabliity to configure a global key prefix through `result_backend_transport_options` for the keys used to store the results in backend .

#### Testing instructions:

1. Start redis (`redis-server`)
2. In a different window connect to redis using redis-cli
3. Execute `FLUSHALL`
4. Create a file `tasks1.py` and add the following task :
```
from celery import Celery

app = Celery("tasks", broker='redis://localhost:6379/0', backend='redis://localhost:6379/0')

app.conf.result_backend_transport_options = {'global_keyprefix': 'test_prefix_'}

@app.task
def add(x, y):
    return x+y
```
5. Start the celery worker server `celery -A tasks1 worker --loglevel=INFO`
6. In a different window call the task 
```
>>> from tasks1 import add
>>> add.delay(4,4)
```
7. From redis-cli, execute `KEYS *`
8. Verify there is a key like `test_prefix__celery-task-meta-<task_id>`
9. Execute `FLUSHALL`
10. Create a new file `tasks2.py` with the below content :
```
from celery import Celery

app = Celery("tasks", broker='redis://localhost:6379/0', backend='redis://localhost:6379/0')

@app.task
def add(x, y):
    return x+y
```
11. Repeat steps 5, 6 and 7, replacing `tasks1` with `tasks2`
12. Verify there is a key like `celery-task-meta-<task_id>`

### DIFF
--- a/celery/backends/base.py
+++ b/celery/backends/base.py
@@ -817,10 +817,24 @@ class BaseKeyValueStoreBackend(Backend):
     def __init__(self, *args, **kwargs):
         if hasattr(self.key_t, '__func__'):  # pragma: no cover
             self.key_t = self.key_t.__func__  # remove binding
-        self._encode_prefixes()
         super().__init__(*args, **kwargs)
+        self._add_global_keyprefix()
+        self._encode_prefixes()
         if self.implements_incr:
             self.apply_chord = self._apply_chord_incr
+
+    def _add_global_keyprefix(self):
+        """
+        This method prepends the global keyprefix to the existing keyprefixes.
+
+        This method checks if a global keyprefix is configured in `result_backend_transport_options` using the
+        `global_keyprefix` key. If so, then it is prepended to the task, group and chord key prefixes.
+        """
+        global_keyprefix = self.app.conf.get('result_backend_transport_options', {}).get("global_keyprefix", None)
+        if global_keyprefix:
+            self.task_keyprefix = f"{global_keyprefix}_{self.task_keyprefix}"
+            self.group_keyprefix = f"{global_keyprefix}_{self.group_keyprefix}"
+            self.chord_keyprefix = f"{global_keyprefix}_{self.chord_keyprefix}"
 
     def _encode_prefixes(self):
         self.task_keyprefix = self.key_t(self.task_keyprefix)
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -290,3 +290,4 @@ Gabor Boros, 2021/11/09
 Tizian Seehaus, 2022/02/09
 Oleh Romanovskyi, 2022/06/09
 JoonHwan Kim, 2022/08/01
+Kaustav Banerjee, 2022/11/10
--- a/docs/getting-started/backends-and-brokers/redis.rst
+++ b/docs/getting-started/backends-and-brokers/redis.rst
@@ -100,6 +100,24 @@ If you are using Sentinel, you should specify the master_name using the :setting
 
     app.conf.result_backend_transport_options = {'master_name': "mymaster"}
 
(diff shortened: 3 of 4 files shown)


auvipy (changes requested):
The timeout was not related to this patch, however need concrete reason to add this feature

kaustavb12:
Hey @auvipy, 

Thanks for checking.

> need concrete reason to add this feature

We can currently configure custom global key prefixes through kombu when using Redis as a broker. This is very useful, when using a shared redis instance, so that access to specific keys can be controlled through redis ACL.

However, currently there is no such mechanism to configure global key prefixes, when using Redis as a results backend. This makes it impossible for multiple applications to use a common shared redis instance as a result backend.

This PR adds the capabliity to configure a global key prefix through `result_backend_transport_options` in line with `broker_transport_options`

BITSOLVER:
> The timeout was not related to this patch, however need concrete reason to add this feature

This feature seems essential for sharing a redis instance with multiple django instances where an instance is needed per customer.  I could really do with this feature now.

auvipy:
@celery/core-developers what do you think about this?

gabor-boros:
@auvipy @celery/core-developers Do you have any updates on this?

auvipy:
did you try this?

gabor-boros:
> did you try this?

@auvipy you mean who? We are using this in production for months now IIRC. Cc: @kaustavb12

celery/backends/base.py
```diff
@@ -817,11 +817,25 @@ class BaseKeyValueStoreBackend(Backend):
...
         if hasattr(self.key_t, '__func__'):  # pragma: no cover
             self.key_t = self.key_t.__func__  # remove binding
-        self._encode_prefixes()
         super().__init__(*args, **kwargs)
+        self._add_global_keyprefix()
+        self._encode_prefixes()
         if self.implements_incr:
             self.apply_chord = self._apply_chord_incr
 
+    def _add_global_keyprefix(self):
```
auvipy, line 826: from my understanding, this feature needs to be documented properly

gabor-boros: @auvipy If you could elaborate on what you exactly mean here, that would help a lot on our side to adjust the PR as per your needs.

Anyways, the method already has docstrings and the feature is already documented based on my understanding: https://docs.celeryq.dev/en/master/search.html?q=keyprefix.

auvipy (changes requested):
please rebase first and add version added 5.3 notation

gabor-boros:
@auvipy the PR is rebased and doc strings added, but what do you exactly mean by annotating? Some guidance would help a lot to sort it out quickly.

auvipy:
thanks you. in case of any edge cases reported, might ping you on those issues reported

ivan-klass:
Thank you for the awesome contribution!
We've spent multiple hours pulling our hairs out figuring out why our shared instance in ElastiCache is not picking up the global prefix, while it should be there according to the latest code. 
And eventually we've noticed that was implemented so recently that it hasn't been released yet 🤣 

@auvipy @thedrow  Is there any chance we can get it before stable 5.3 version - i.e. maybe it worth backporting to 5.2.x branch? I'll be happy to help with PR, if needed. If it doesn't sound like a good idea, could you please publish it as something like `5.3.0-c1`?
Thank you!

auvipy:
this will be part of 5.3.0b2 in few weeks

auvipy:
hopefully in february

PR #7626, opened 2022-07-13
Don't warn about DEBUG=True for Django

vanschelven:
This warning used to be correct, but is no longer relevant since Django 1.8.

See https://github.com/django/django/commit/cfcca7ccce3dc527d16757ff6dc978e50c4a2e61
for the Django-side fix.


auvipy (changes requested):
the change seems legit to me, can you please cross check the redis integration build failures?

auvipy (approved):

Issue #6819, opened 2021-06-21
Redis result backend connections leak


ronlut:
# Checklist

* [x] I have verified that the issue exists against the `master` branch of Celery.
* [x] This has already been asked to the [discussion group](https://groups.google.com/forum/#!forum/celery-users) first.
* [x] I have read the relevant section in the
  [contribution guide](http://docs.celeryproject.org/en/latest/contributing.html#other-bugs)
  on reporting bugs.
* [x] I have checked the [issues list](https://github.com/celery/celery/issues?q=is%3Aissue+label%3A%22Issue+Type%3A+Bug+Report%22+-label%3A%22Category%3A+Documentation%22)
  for similar or identical bug reports.
* [x] I have checked the [pull requests list](https://github.com/celery/celery/pulls?q=is%3Apr+label%3A%22PR+Type%3A+Bugfix%22+-label%3A%22Category%3A+Documentation%22)
  for existing proposed fixes.
* [x] I have checked the [commit log](https://github.com/celery/celery/commits/master)
  to find out if the bug was already fixed in the master branch.
* [x] I have included all related issues and possible duplicate issues
  in this issue (If there are none, check this box anyway).

## Mandatory Debugging Information
* [x] I have included the output of ``celery -A proj report`` in the issue.
    (if you are not able to do this, then at least specify the Celery
     version affected).
* [x] I have verified that the issue exists against the `master` branch of Celery.
* [x] I have included the contents of ``pip freeze`` in the issue.
* [x] I have included all the versions of all the external dependencies required
  to reproduce this bug.

## Optional Debugging Information

* [ ] I have tried reproducing the issue on more than one Python version
  and/or implementation.
* [ ] I have tried reproducing the issue on more than one message broker and/or
  result backend.
* [ ] I have tried reproducing the issue on more than one version of the message
  broker and/or result backend.
* [ ] I have tried reproducing the issue on more than one operating system.
* [ ] I have tried reproducing the issue on more than one workers pool.
* [ ] I have tried reproducing the issue with autoscaling, retries,
  ETA/Countdown & rate limits disabled.
* [ ] I have tried reproducing the issue after downgrading
  and/or upgrading Celery and its dependencies.

#### Possible Duplicates
* None

## Environment & Settings

**Celery version**:

<details>
<summary><b><code>celery report</code> Output:</b></summary>
<p>

```
software -> celery:5.1.0 (sun-harmonics) kombu:5.1.0 py:3.8.8
[...]

PR #6416, opened 2020-10-15
Multithreaded backend

matusvalo:
## Description

This PR contains two changes:
1. Storing backend to separate thread local variable to avoid sharing underlying Connection object between threads. This means that now each thread has it's own Backend object.
2. Change `oid` (Universally unique identifier) to be thread-based. Moreover, the `oid` is now different for each thread. This is causing that each thread has it's own reply-to queue (Before all threads shares one reply-to queue which was causing missing responses)

This PR fixes #6414 and probably also issues linked to it.

PR #7673, opened 2022-08-02
fix: share redis connection as it's thread-safe

wochinge:
## Description
- attempt to fix https://github.com/celery/celery/issues/6819

## Notes
- Celery shared Redis connections between threads and not between geventlets before [this PR](https://github.com/celery/celery/pull/6416/files)
- The linked PR then changed it to have a connection per thread which probably ([due to gevents monkeypatching](https://www.gevent.org/api/gevent.monkey.html)) translates to one connection per geventlet
- The PR tries to revert the changes for redis as `redis-py` is thread-safe

## Open questions
- Is `threading.local` cleaning up properly? (especially after `gevent` monkeypatching)?

### DIFF
--- a/celery/app/base.py
+++ b/celery/app/base.py
@@ -955,6 +955,22 @@ class Celery:
         backend, url = backends.by_url(
             self.backend_cls or self.conf.result_backend,
             self.loader)
+
+        from celery.backends.redis import RedisBackend
+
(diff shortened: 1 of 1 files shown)


wochinge:
see comment https://github.com/celery/celery/issues/6819#issuecomment-1203827633 for more details

Issue #6963, opened 2021-09-18
Redis result consumer is not drained periodically (memory leak and possible connection reset)


olii:
# Checklist

- [ ] I have verified that the issue exists against the `master` branch of Celery.
- [ ] This has already been asked to the [discussion group](https://groups.google.com/forum/#!forum/celery-users) first.
- [ ] I have read the relevant section in the
  [contribution guide](http://docs.celeryproject.org/en/latest/contributing.html#other-bugs)
  on reporting bugs.
- [ ] I have checked the [issues list](https://github.com/celery/celery/issues?q=is%3Aissue+label%3A%22Issue+Type%3A+Bug+Report%22+-label%3A%22Category%3A+Documentation%22)
  for similar or identical bug reports.
- [ ] I have checked the [pull requests list](https://github.com/celery/celery/pulls?q=is%3Apr+label%3A%22PR+Type%3A+Bugfix%22+-label%3A%22Category%3A+Documentation%22)
  for existing proposed fixes.
- [ ] I have checked the [commit log](https://github.com/celery/celery/commits/master)
  to find out if the bug was already fixed in the master branch.
- [ ] I have included all related issues and possible duplicate issues
  in this issue (If there are none, check this box anyway).

## Mandatory Debugging Information

- [ ] I have included the output of ``celery -A proj report`` in the issue.
    (if you are not able to do this, then at least specify the Celery
     version affected).
- [ ] I have verified that the issue exists against the `master` branch of Celery.
- [ ] I have included the contents of ``pip freeze`` in the issue.
- [ ] I have included all the versions of all the external dependencies required
  to reproduce this bug.

## Optional Debugging Information

- [ ] I have tried reproducing the issue on more than one Python version
  and/or implementation.
- [ ] I have tried reproducing the issue on more than one message broker and/or
  result backend.
- [ ] I have tried reproducing the issue on more than one version of the message
  broker and/or result backend.
- [ ] I have tried reproducing the issue on more than one operating system.
- [ ] I have tried reproducing the issue on more than one workers pool.
- [ ] I have tried reproducing the issue with autoscaling, retries,
  ETA/Countdown & rate limits disabled.
- [ ] I have tried reproducing the issue after downgrading
  and/or upgrading Celery and its dependencies.

#### Related Issues

- None

#### Possible Duplicates

- None

## Environment & Settings

**Celery version**: 5.0+

<details>
<summary><b><code>celery report</code> Output:</b></summary>
<p>

```
```

</p>
</details>

[...]

PR #7685, opened 2022-08-05
Fix redis connection leak with `app.send_task`

wochinge:
## Description

This test shows that there is a Redis connection leak when using `app.send_task`. 
The connection leak doesn't happen if `task.delay()` is used.

Todos:
- [ ] find solution

### DIFF
--- a/celery/backends/base.py
+++ b/celery/backends/base.py
@@ -667,8 +667,17 @@ class Backend:
         """Backend cleanup."""
 
     def process_cleanup(self):
-        """Cleanup actions to do at the end of a task worker process."""
-
+        try:
+            from eventlet import corolocal
+            task_backend_cache = self.app.local
+            # Whenever an eventlet pool spawns a coroutine it has a different identifier and hence will acquire
+            # a new coroutine-local storage. To avoid that we leak connections we explicitly clean out the
+            # cache once the coroutine is done.
+            if isinstance(task_backend_cache, corolocal.local):
+                task_backend_cache.__dict__.clear()
+        except ImportError:
+            # Eventlet is not installed and hence not used
+            pass
     def on_task_call(self, producer, task_id):
         return {}
 
--- a/celery/backends/base.py
+++ b/celery/backends/base.py
@@ -101,6 +101,9 @@ class Backend:
     #: Set to true if the backend is persistent by default.
     persistent = True
 
+    #: If the backend is thread-safe
+    is_threadsafe = False
+
     retry_policy = {
         'max_retries': 20,
         'interval_start': 0,
--- a/t/integration/test_tasks.py
+++ b/t/integration/test_tasks.py
@@ -5,8 +5,9 @@ import pytest
 
 import celery
 from celery import group
+from celery.result import AsyncResult
 
-from .conftest import get_active_redis_channels
+from .conftest import get_active_redis_channels, get_redis_connection
 from .tasks import (ClassBasedAutoRetryTask, ExpectedException, add, add_ignore_result, add_not_typed, fail,
                     print_unicode, retry, retry_once, retry_once_headers, retry_once_priority, return_properties,
                     sleeping)
@@ -328,3 +329,16 @@ class test_task_redis_result_backend:
 
         new_channels = [channel for channel in get_active_redis_channels() if channel not in channels_before_test]
         assert new_channels == []
+
+    def test_redis_connection_cleanup_with_send_task(self, manager):
+        redis = get_redis_connection()
+        connections_before = len(redis.client_list())
+
+        result = manager.app.send_task('t.integration.tasks.add', args=(1, 2))
+        assert result.get(timeout=TIMEOUT) == 3
+
+        # Result should be there after a few seconds
+        sleep(5)
+
+        # TODO: This fails but shouldn't
+        assert len(redis.client_list()) == connections_before
--- a/t/integration/test_tasks.py
+++ b/t/integration/test_tasks.py
@@ -1,3 +1,4 @@
+import gc
 from datetime import datetime, timedelta
 from time import perf_counter, sleep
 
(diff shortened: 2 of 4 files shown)


t/integration/test_tasks.py
```diff
@@ -288,29 +289,46 @@ def test_properties(self, celery_session_worker):
...
+            f"celery-task-meta-{result.id}".encode()
         ]
         assert result.get(timeout=3) == 3
         assert get_active_redis_channels() == []
+
+    def test_redis_connection_cleanup_with_send_task(self, manager):
+        redis = get_redis_connection()
+        connections_before = len(redis.client_list())
+
+        result = manager.app.send_task('t.integration.tasks.add', args=(1, 2))
```
wochinge, line 327: if we use `add.delay()` then everything works

t/integration/test_tasks.py
```diff
@@ -288,29 +289,46 @@ def test_properties(self, celery_session_worker):
...
+    def test_asyncresult_get_cancels_subscription(self, manager):
         result = add.delay(1, 2)
         assert get_active_redis_channels() == [
-            f"celery-task-meta-{result.id}"
+            f"celery-task-meta-{result.id}".encode()
         ]
         assert result.get(timeout=3) == 3
         assert get_active_redis_channels() == []
+
+    def test_redis_connection_cleanup_with_send_task(self, manager):
```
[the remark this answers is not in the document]
wochinge: it might be related to https://github.com/celery/celery/issues/6963 .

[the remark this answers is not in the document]
wochinge: It definitely seems like that the connection always stays in the `pending_unsubscribe_channels` within the Redis client when using `app.send_task`. I tried with the solution from the linked issue but that doesn't resolve things.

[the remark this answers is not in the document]
wochinge: @auvipy can you please give me some pointers where the issue might be?

auvipy (changes requested):
can you please rebase this?

wochinge:
done @auvipy

t/integration/test_tasks.py
```diff
@@ -328,3 +329,16 @@ def test_asyncresult_get_cancels_subscription(self, manager):
 
         new_channels = [channel for channel in get_active_redis_channels() if channel not in channels_before_test]
         assert new_channels == []
+
+    def test_redis_connection_cleanup_with_send_task(self, manager):
```
thedrow, line 333: I suggest we mark the test as `xfail`ed and merge this PR.
There should be a matching issue for this PR so that we'll know this isn't yet fixed.

thedrow (changes requested):

t/integration/test_tasks.py
wochinge: @thedrow I don't think that merging this before fixing this is helpful. There are already multiple issues pointing towards a connection leak:

- https://github.com/celery/celery/issues/6963
- https://github.com/celery/celery/issues/6819

What I would I need is some pointers / help. Besides that I'd be more than happy to investigate and help with a fix.

auvipy: > @thedrow I don't think that merging this before fixing this is helpful. There are already multiple issues pointing towards a connection leak:
> 
>     * [Redis result consumer is not drained periodically (memory leak and possible connection reset) #6963](https://github.com/celery/celery/issues/6963)
> 
>     * [Redis result backend connections leak #6819](https://github.com/celery/celery/issues/6819)
> 
> 
> What I would I need is some pointers / help. Besides that I'd be more than happy to investigate and help with a fix.

that should be the spirit

wochinge:
@auvipy I spent some time debugging it and that's the situation:
- the `local` cache [here](https://github.com/celery/celery/blob/6c5d5755ab3da0c1ff9af31b284e2b4415a29592/celery/app/base.py#L233) doesn't work with `eventlet` as each greenthread has unique ID and hence its own `local()` instance. You can see this in the code snippet below as well in the comment [here](https://github.com/eventlet/eventlet/issues/741#issuecomment-1005739191):
  ```python
  import threading
  from time import sleep
  
  import eventlet
  eventlet.monkey_patch()
  
  
  def f(idx: int):
      t = threading.current_thread()
		  # We always have a different ID 
      print(threading.get_ident(), idx)
  
  pool = eventlet.GreenPool(size=1)
  
  threads = []
  for i in range(10):
      t = pool.spawn(f, i)
      threads.append(t)
  
  for t in threads:
      t.wait()
  ```
- this means that with `eventlets` we create one backend connection per task. These caches are not always cleaned up correctly. I've added a piece of code which cleans up the cache for each task _in case_ it's eventlet cache. Note that we can't test this as the integration tests run in `prefork` mode. If you want to reproduce it you need to:
  - Change `prefork` to `eventlet` [here](https://github.com/celery/celery/blob/1c4ff33bd22cf94e297bd6449a06b5a30c2c1fbc/t/integration/conftest.py#L52)
   - Monkeypatch the `eventlet` stuff [here](https://github.com/celery/celery/blob/feaad3f9fdf98d0453a07a68e307e48c6c3c2550/celery/__init__.py#L1) by adding these lines:
     ```python
     import eventlet
     eventlet.monkey_patch()
     ```
- The above stuff should help with leaking connections. The other question is whether we can fix the cache for `eventlet` pools. I don't have a good solution for this at the moment. I'll explore using `cached_property` for Redis.

wochinge:
Added a potential fix for the caching issues with eventlet. Any idea why the CI is failing? In my opinion a bit weird as it's failing in somewhat unrelated parts.

celery/backends/base.py
```diff
@@ -667,8 +670,17 @@ def cleanup(self):
...
+            from eventlet import corolocal
+            task_backend_cache = self.app.local
+            # Whenever an eventlet pool spawns a coroutine it has a different identifier and hence will acquire
+            # a new coroutine-local storage. To avoid that we leak connections we explicitly clean out the
+            # cache once the coroutine is done.
+            if isinstance(task_backend_cache, corolocal.local):
+                task_backend_cache.__dict__.clear()
+        except ImportError:
+            # Eventlet is not installed and hence not used
+            pass
```
thedrow, line 683: I'm not sure I am content with coupling eventlet or any other concurrency mode with the backend. Is there another way to resolve this issue?

auvipy:
@wochinge would you mind trying https://github.com/celery/celery/pull/8058?

wochinge:
We're unfortunately no longer running into this (we put `ignore_result` whereever possible), sorry 🤷🏻

hzc989:
> @wochinge would you mind trying #8058?

after upgrading to 5.3.1 , with setting `result_backend_thread_safe=true`, works for us.

p.s. we've try 4.4.7 & 5.2.7 , both failed with the same problem.

PR #7688, opened 2022-08-06
Replace all Lock to RLock(Fixes #7686)

### LINKED ISSUE #7686: Deadlock
# Disclaimer
I don't know much about celery yet.
So, I create a Issue, not a PR, to report this.

I seen Celery is uses lot `theating.Lock()`, and multiple calls `acquire()`(same as `with Lock():`) in same thread.
This is the point of deadlock problem.

I have sent a [PR](https://github.com/celery/celery/pull/7652) to similar problem, it was merged to master.
It looks work normal.
But actually, I'm not completely sure is this PR stable or not.
I'm not sure if replacing most `Lock()` with `RLock()` in Celery, will cause other bugs or not.

# Bug description
I don't know what wrong will be cause with this deadlock.
But it must be a deadlock.

In there, It got the Lock().
https://github.com/celery/celery/blob/bdbf6d6ae1aca9addd81800b5dd2e8c3477afb18/celery/events/state.py#L444
Then, It will got Lock() again.
https://github.com/celery/celery/blob/bdbf6d6ae1aca9addd81800b5dd2e8c3477afb18/celery/events/state.py#L452

954-Ivory:


954-Ivory:
Did I make some mistake in PR opration?

auvipy:
1st thing is you did it on your forks masters branch, which you should not. you should create a new branch first

954-Ivory:
Sry, has other mistack? I will try again later.

PR #7690, opened 2022-08-06
Replace all `Lock` to `RLock`(Fixes #7686)

### LINKED ISSUE #7686: Deadlock

954-Ivory:
## Description
Try to fixed most of deadlock.
Replace all `Lock` to `RLock`(Fixes #7686).

### DIFF
--- a/celery/events/dispatcher.py
+++ b/celery/events/dispatcher.py
@@ -68,7 +68,7 @@ class EventDispatcher:
         self.buffer_limit = buffer_limit
         self.on_send_buffered = on_send_buffered
         self._group_buffer = defaultdict(list)
-        self.mutex = threading.Lock()
+        self.mutex = threading.RLock()
         self.producer = None
         self._outbound_buffer = deque()
         self.serializer = serializer or self.app.conf.event_serializer
--- a/celery/events/dispatcher.py
+++ b/celery/events/dispatcher.py
@@ -220,8 +220,9 @@ class EventDispatcher:
         """Close the event dispatcher."""
         try:
             self.mutex.release()
-        finally:
-            self.producer = None
+        except RuntimeError:
+            pass
+        self.producer = None
 
     def _get_publisher(self):
         return self.producer
--- a/celery/events/dispatcher.py
+++ b/celery/events/dispatcher.py
@@ -218,8 +218,10 @@ class EventDispatcher:
 
     def close(self):
         """Close the event dispatcher."""
-        self.mutex.locked() and self.mutex.release()
-        self.producer = None
+        try:
+            self.mutex.release()
+        finally:
+            self.producer = None
 
     def _get_publisher(self):
         return self.producer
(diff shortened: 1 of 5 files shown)


954-Ivory:
I get the failing reason from CI log.
But I have some question:
1. How should I do now? Close this PR and send a new PR?
2. Can I Fork CI in my Repo? I want to checking again when i changed code.

954-Ivory:
I use pycharm to `Edit Commit Message`, the commit seens a little dirty.
But this PR seens work well now.

auvipy:
no issue, we will squash commit.

celery/events/dispatcher.py
```diff
@@ -218,7 +218,10 @@ def extend_buffer(self, other):
 
     def close(self):
         """Close the event dispatcher."""
-        self.mutex.locked() and self.mutex.release()
+        try:
+            self.mutex.release()
+        except RuntimeError:
+            pass
```
thedrow, line 224: This is too broad. We need to catch specific exceptions.

thedrow (changes requested):

celery/events/dispatcher.py
954-Ivory: This just catch the `RuntimeError` from `RLock.release()`.
But I think maybe CI exception cause by RLock didn't release.

thedrow:
I'm a little bit concerned that we're replacing all locks with reentrant locks without any evidence it will resolve the deadlock (although it makes some sense).
Perhaps you can use [pydld](https://github.com/mindstorm38/pydld) or a similar tool to verify this.

auvipy:
@954-Ivory can you revisit this please?

954-Ivory:
> @954-Ivory can you revisit this please?

**Maybe we should consider temporarily closing this PR.**

I checked the document about `Lock` and `RLock`, there has a little different:

#### `Lock.release()`
_This can be called from any thread, not only the thread which has acquired the lock._
https://docs.python.org/3/library/threading.html?highlight=rlock#threading.Lock.release

#### `RLock.release()`
_Only call this method when the calling thread owns the lock._
https://docs.python.org/3/library/threading.html?highlight=rlock#threading.RLock.release

**I'm not sure about that, does celery has any scene, about release a thread lock in the not lock owning thread.**

PR #7692, opened 2022-08-07
Added integration test_chord_header_id_duplicated_on_rabbitmq_msg_duplication()

Nusnus:
When a task that predates a chord in a chain was duplicated by Rabbitmq (for whatever reason), the chord header id was not duplicated. This caused the chord header to have a different id. This test ensures that the chord header's id preserves itself in face of such an edge case.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -83,7 +83,7 @@ jobs:
           fail-fast: false
           matrix:
               python-version: ['3.7', '3.8', '3.9', '3.10']
-              toxenv: ['redis', 'rabbitmq']
+              toxenv: ['redis', 'rabbitmq', 'rabbitmq_redis']
 
       services:
           redis:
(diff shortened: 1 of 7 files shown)


auvipy (changes requested):
thanks for the integration test. can you also try to fix the integration test failures in CI?

thedrow:
@auvipy Could you please review the pull request **and** the build log before you make any comments regarding the quality of the PR or the state of the build?

If you were to check the build output, you'd find that the test @Nusnus contributed is failing. The next logical question is why would he contribute a failing test. That's because we have a test case for something that's currently not working in Celery. The next logical question is why this test case doesn't work.
I appreciate your contributions, but at this point we might as well create a Github Action that reminds contributors of what needs to be done. If you don't know what to do, you can always ask.

auvipy:
@thedrow, I used to check codes and logs before commenting in most of the time. however, for this PR, reading the PR description, and seeing all the integration tests failing, I wasn't able to understand initially it was just a failing test :) 

I have asked the contributor if he is up for fixing the failing codes. So from my point of view, this is not a big issue. And for the new contributors action, that would be a good stuff to get rid of lots of manual verification.

thedrow (approved):

PR #7730, opened 2022-08-28
Fixed bug in chord stamping with another chord as a body + unit test

Nusnus:


t/unit/tasks/test_canvas.py
```diff
@@ -88,7 +88,13 @@ def replace_with_chain(self, x, y):
...
         @self.app.task(shared=False)
         def xprod(numbers):
-            return math.prod(numbers)
+            try:
+                return math.prod(numbers)
+            except AttributeError:
+                import operator
+                from functools import reduce
+
+                return reduce(operator.mul, numbers)
```
thedrow, line 97:
```suggestion
                #  TODO: Drop this backport once
                #              we drop support for Python 3.7
                import operator
                from functools import reduce

                return reduce(operator.mul, numbers)
```

thedrow (changes requested):
Just a minor comment and we're ready to go.

t/unit/tasks/test_canvas.py
Nusnus: Done @thedrow

auvipy (approved):

PR #7734, opened 2022-08-30
Use "describe_table" not "create_table" to check for existence of DynamoDB table

### LINKED ISSUE #6622: dynamoDB result backend tries to create tables while role does not permit
The following code

https://github.com/celery/celery/blob/d0f5300691ca594f2311daf542aa63367622c027/celery/backends/dynamodb.py#L192

tries to create or get the table. In case the table exists and the role executing that code on AWS does not have the `CreateTable` permission, the raised exception is not the one expected by that particular line. Yet the table exists.

To reproduce:
* create the result backend table and give the following permissions to the role executing celery on that table:
  
  ```
      "dynamodb:DescribeTable",
      "dynamodb:PutItem",
      "dynamodb:UpdateItem",
      "dynamodb:DeleteItem",
      "dynamodb:BatchWriteItem",
      "dynamodb:GetItem",
      "dynamodb:BatchGetItem",
      "dynamodb:Scan",
      "dynamodb:Query",
      "dynamodb:ConditionCheckItem"
```
* remove the permission CreateTable from the role executing the code

Possible solution:

* handle the check of the table existence with another boto3 call such as [`describe_table`](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/dynamodb.html#DynamoDB.Client.describe_table) and then fall-back into the table creation
* provide a configuration that, when indicated, assumes the existence of the table
* handle that exception differently

maxfirman:
Fixes #6622 by testing for existence of table using `describe_table` rather than `create_table`

### DIFF
--- a/celery/backends/dynamodb.py
+++ b/celery/backends/dynamodb.py
@@ -201,28 +201,25 @@ class DynamoDBBackend(KeyValueStoreBackend):
         """Create table if not exists, otherwise return the description."""
         table_schema = self._get_table_schema()
         try:
-            table_description = self._client.create_table(**table_schema)
-            logger.info(
-                'DynamoDB Table {} did not exist, creating.'.format(
-                    self.table_name
-                )
-            )
-            # In case we created the table, wait until it becomes available.
-            self._wait_for_table_status('ACTIVE')
-            logger.info(
-                'DynamoDB Table {} is now available.'.format(
-                    self.table_name
-                )
-            )
-            return table_description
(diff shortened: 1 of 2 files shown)


auvipy (commented):
I am not sure about the cod coverage, can we improve it or the codcov is generating wrong coverage?

maxfirman:
@auvipy apologies my commit with refactored tests didn't get pushed yesterday for some reason. I've just pushed it now.

auvipy (approved):

PR #7739, opened 2022-09-02
Fix beat Service __reduce__ error

AXHMFHNP:
Fix beat Service `__reduce__` error

### DIFF
--- a/t/unit/test_beat.py
+++ b/t/unit/test_beat.py
@@ -1,0 +1,21 @@
+import pickle
+from unittest import mock
+
+from celery.beat import Service
+
+
+class test_Service:
+
+    def test___reduce__(self):
+
+        s1 = Service(mock.sentinel.app,
+                     max_interval=mock.sentinel.max_interval,
+                     schedule_filename=mock.sentinel.schedule_filename,
+                     scheduler_cls=mock.sentinel.scheduler_cls)
+
+        s2 = pickle.loads(pickle.dumps(s1))
+
+        assert s1.app == s2.app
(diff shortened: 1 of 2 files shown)


auvipy (commented):
can you please tell what error you are fixing with this?

AXHMFHNP:
“TypeError: 'Celery' object is not callable” reported at celery beat service start.
```
        debug('beat: Ticking with max interval->%s',
              humanize_seconds(self.scheduler.max_interval))
```
self.scheduler.max_interval -> self.scheduler -> self.get_scheduler() 
```
        return symbol_by_name(self.scheduler_cls, aliases=aliases)(
```
 self.scheduler_cls becomes Celery object due to the inconsistency between the order of `__init__` parameters and the order returned by `__reduce__`

AXHMFHNP:
Done

t/unit/test_beat.py
```diff
@@ -0,0 +1,21 @@
...
+class test_Service:
+
+    def test___reduce__(self):
+
+        s1 = Service(mock.sentinel.app,
+                     max_interval=mock.sentinel.max_interval,
+                     schedule_filename=mock.sentinel.schedule_filename,
+                     scheduler_cls=mock.sentinel.scheduler_cls)
+
+        s2 = pickle.loads(pickle.dumps(s1))
```
auvipy, line 16: why specifically pickle? can we also try json here beside pickle?

auvipy (changes requested):

AXHMFHNP:
The problems that occurred with the old version are not reproduced in the new version.

PR #7744, opened 2022-09-05
Enhancements for task_allow_error_cb_on_chord_header tests and docs

Nusnus:
Added a test to check `task_allow_error_cb_on_chord_header` with an upgraded chord works the same as an explicit chord when the flag is enabled.
Also enhanced the documents about this use case.


thedrow (approved):
LGTM

PR #7782, opened 2022-10-03
pin importlib-metadata>=1.4.0,<5.0

neutronth:
## Description

* Since importlib-metadata has released version 5.0.0 [1] and breaks a backward compatible with throwing error as below,

  ``` ImportError: cannot import name 'Celery' from 'celery' (/usr/local/lib/python3.7/site-packages/celery/__init__.py) ```

---
[1] https://pypi.org/project/importlib-metadata/5.0.0


jaraco:
I'd recommend #7785 instead, as this approach adds tech debt.

PR #7789, opened 2022-10-05
Do not recommend using easy_install anymore

jugmac00:
It has been deprecated since 2019:
https://setuptools.pypa.io/en/latest/history.html#v42-0-0


jugmac00:
The failure is unrelated to this PR, see https://stackoverflow.com/questions/73929564/

PR #5759, opened 2019-10-02
Fix #5597: chain priority

ptitpoulpe:

PR #7793, opened 2022-10-06
Add test to check that apply_async priority propogates to all tasks in chain

dumbPy:
Refer https://github.com/celery/celery/pull/5759#issuecomment-1270413045 for how this fails.

## Description

When chaining tasks with .s and .si and adding priority while calling .apply_async on the chain, it should ideally propagate to all tasks of the chain.

Added test checks that all tasks of the chain have same priority as the one provided in apply_async


dumbPy:
@auvipy  The test I added doesn't work. Sorry. I couldn't figure out how to add a test to check what get's added to the task after apply_async in chain. Hence Closing

PR #7807, opened 2022-10-15
Remove `int` cast on propagate

Kludex:
I had to go back a bit to see the commit it was introduced... https://github.com/celery/celery/commit/d05b523ee95d85fdeb487a24c734a54a8cc6cdee.

### DIFF
--- a/celery/app/log.py
+++ b/celery/app/log.py
@@ -6,6 +6,8 @@ Sets up logging for the worker and other programs,
 redirects standard outs, colors log output, patches logging
 related compatibility fixes, and so on.
 """
+from __future__ import annotations
+
 import logging
 import os
 import sys
@@ -153,8 +155,8 @@ class Logging:
             if loglevel:
                 logger.setLevel(loglevel)
 
-    def setup_task_loggers(self, loglevel=None, logfile=None, format=None,
-                           colorize=None, propagate=False, **kwargs):
+    def setup_task_loggers(self, loglevel: int | None = None, logfile=None, format=None,
+                           colorize=None, propagate: bool = False, **kwargs):
         """Setup the task logger.
 
         If `logfile` is not specified, then `sys.stderr` is used.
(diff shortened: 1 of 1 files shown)


celery/app/log.py
```diff
@@ -6,6 +6,8 @@
 redirects standard outs, colors log output, patches logging
 related compatibility fixes, and so on.
 """
+from __future__ import annotations
+
```
Kludex, line 10: Feel free to remove this if wanted.
```suggestion
```

celery/app/log.py
```diff
@@ -153,8 +155,8 @@ def _configure_logger(self, logger, logfile, loglevel,
             if loglevel:
                 logger.setLevel(loglevel)
 
-    def setup_task_loggers(self, loglevel=None, logfile=None, format=None,
-                           colorize=None, propagate=False, **kwargs):
+    def setup_task_loggers(self, loglevel: int | None = None, logfile=None, format=None,
+                           colorize=None, propagate: bool = False, **kwargs):
```
Kludex, line 159: Since this is also not part of the change... Feel free to remove it. 👀 
```suggestion
    def setup_task_loggers(self, loglevel=None, logfile=None, format=None,
                           colorize=None, propagate=False, **kwargs):
```

auvipy:
are the failures related?

celery/app/log.py
```diff
@@ -171,8 +173,7 @@ def setup_task_loggers(self, loglevel=None, logfile=None, format=None,
             formatter=TaskFormatter, **kwargs
         )
         logger.setLevel(loglevel)
-        # this is an int for some reason, better to not question why.
-        logger.propagate = int(propagate)
+        logger.propagate = propagate
```
auvipy, line 176: can we do some additional test on this?

auvipy (changes requested):

celery/app/log.py
Kludex: Sure. I'll add them during the week.

Thanks @auvipy :pray:

Issue #477, opened 2011-09-09
celery/apps/worker calls sys.stdout.isatty() directly


ask:
It should use `celery.utils.isatty(fh)`

Ref http://groups.google.com/group/celery-users/browse_thread/thread/138c30649622cd8c?hl=en_US

PR #7814, opened 2022-10-16
Remove `isatty` wrapper function

Kludex:
This function was added 11 years ago to avoid an issue with Python 2.6.

References: 
- https://groups.google.com/g/celery-users/c/E4wwZJYizYw?hl=en_US
- https://github.com/celery/celery/issues/477

### DIFF
--- a/celery/app/log.py
+++ b/celery/app/log.py
@@ -18,7 +18,6 @@ from celery import signals
 from celery._state import get_current_task
 from celery.exceptions import CDeprecationWarning, CPendingDeprecationWarning
 from celery.local import class_property
-from celery.platforms import isatty
 from celery.utils.log import (ColorFormatter, LoggingProxy, get_logger, get_multiprocessing_logger, mlevel,
                               reset_multiprocessing_logger)
 from celery.utils.nodenames import node_format
@@ -204,7 +203,7 @@ class Logging:
         if colorize or colorize is None:
             # Only use color if there's no active log file
             # and stderr is an actual terminal.
-            return logfile is None and isatty(sys.stderr)
+            return logfile is None and sys.stderr.isatty()
         return colorize
 
     def colored(self, logfile=None, enabled=None):
--- a/celery/apps/worker.py
+++ b/celery/apps/worker.py
@@ -20,7 +20,7 @@ from kombu.utils.encoding import safe_str
 from celery import VERSION_BANNER, platforms, signals
 from celery.app import trace
 from celery.loaders.app import AppLoader
-from celery.platforms import EX_FAILURE, EX_OK, check_privileges, isatty
+from celery.platforms import EX_FAILURE, EX_OK, check_privileges
 from celery.utils import static, term
 from celery.utils.debug import cry
 from celery.utils.imports import qualname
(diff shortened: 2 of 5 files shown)


auvipy (approved):

grembo:
Ok, so I can confirm that celery-worker does not start without stdin being available, which used to work until this change. Not sure if it's a big deal, I do have one user though whose setup does not work because of it. This can be worked around outside of celery, but maybe it would make sense to make that part of the code more resilient again (read: not requiring standard input to determine if the terminal supports images).

Kludex:
@auvipy can you revert this? Sorry.

Kludex:
I think the reason why this change was introduced initially was for Python 2.6, but it worked for other systems.

grembo:
To test the issue from a shell (close stdin, start worker):

```
(exec 0<&-; /usr/local/bin/celery --app myapp worker)
```

auvipy:
I can try to revert this

Kludex:
Thanks.

grembo:
It seems a bit strange to me, that this check

```python
def supports_images():
    return sys.stdin.isatty() and ITERM_PROFILE
```

happens on stdin (maybe it's because the check should work if either stdout or stderr are redirected?).

So instead of the reverting the whole patch, only this specific place could be wrapped in a try/except block and return false in case of attribute error (using the logic of the previous function):

```python
def supports_images():
    try:
        return sys.stdin.isatty() and ITERM_PROFILE
    except AttributeError:
        pass
```

I'm not an expert on celery though, so just sharing my thoughts here.

Thank you for your quick reaction on my initial comment, much appreciated.

Nusnus:
@auvipy Can you please summarize what happened in terms of why this change was introduced, and why it was reverted (+minor modification)?
I’ll use it in the release change log to describe the change eventually.

Nusnus:
Just noticed this 😂
CleanShot 2025-02-13 at 11 53 17

Unfortunately we’re not there yet haha:
CleanShot 2025-02-13 at 11 53 41

grembo:
@nusnus What about just scrolling back on this very issue, e.g., https://github.com/celery/celery/pull/7814#issuecomment-2585771143

Nusnus:
@grembo 
> @Nusnus What about just scrolling back on this very issue, e.g., https://github.com/celery/celery/pull/7814#issuecomment-2585771143

I’d be happy _just_ to scroll up if it would efficiently cross-reference the description with the PR comments of the revert PR and explain concisely what the bottom line was.

Overall, when someone reviews the release change log, it is not expected that they will study all of the change items to understand what happened. Providing the community with a concise and clear description is more efficient and will avoid confusion.
Lastly, as our release is extremely late, I’d appreciate any help in completing the checklists for the release, which include clarifications on each change since v5.4.0.

Nusnus:
@grembo 
> Fix issue that was introduced by accident when removing a workaround, which stopped celery from starting when executed without stdin available.

That’s great actually! Thanks!
Can we TL;DR what was the issue?

grembo:
@Nusnus 
The issue was that the code was written a way that required _stdin_ to exist, even though this isn't always the case, e.g.,  when being started from a service manager.

grembo:
The comment https://github.com/celery/celery/pull/7814#issuecomment-2585829789 demonstrates this:

```
def supports_images():
    return sys.stdin.isatty() and ITERM_PROFILE
```

So as far as I am concerned, the release note line could be:

```
Fix an issue that was introduced by accident when removing a workaround for _sys.stdout.isatty_ on old versions of Python that was not required anymore. This stopped celery from starting when executed without stdin available, as _sys.stdin.isatty_ was used to determine if the terminal supports images without taking into the account the situation in which celery is started without stdin available as it, e.g., can happen when started by a service manager on boot.
```

Nusnus:
Hey @grembo - thanks a ton!
This is great++ !

Really appreciate it 🙏

PR #7827, opened 2022-10-18
Fix sphinx pipeline

Kludex:
Well... At least the exception changed :rofl:


Kludex:
- To be continued on #7833

PR #7832, opened 2022-10-18
Remove flakeplus package

Kludex:
This PR removes the old `flakeplus` package, and all the references to flake8.

`flake8` is already running on the `pre-commit`, and in the pipeline.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,6 @@ PYTEST=pytest
 GIT=git
 TOX=tox
 ICONV=iconv
-FLAKE8=flake8
 PYROMA=pyroma
 SPHINX2RST=sphinx2rst
 RST2HTML=rst2html.py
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -5,9 +5,7 @@ PYTEST=pytest
 GIT=git
 TOX=tox
 ICONV=iconv
-FLAKE8=flake8
```
auvipy, line 8: should we remove flake8 from make file?

Kludex: The pre-commit already has it, and we also run the pre-commit on the pipeline.

Unless... You use it? :eyes:

auvipy: makefile is for local use as far I can tell

Kludex: pre-commit as well:

`pre-commit run flake8 --all-files`

My point is just to not have multiple ways to do the same thing.

Issue #7836, opened 2022-10-19
CI: test_mutable_errback_called_by_chord_from_group_fail_multiple fails often


woutdenolf:
I see this test fail often in CI

```python
 =================================== FAILURES ===================================
  _ test_chord.test_mutable_errback_called_by_chord_from_group_fail_multiple[errback_old_style] [Error propagates from body group] _
  
  self = <celery.backends.redis.ResultConsumer object at 0x7f6132ec6710>
  result = <GroupResult: 8d8dfee3-a5fc-4c82-8a89-fc1d10beef7d [aef15b21-9fa3-4862-b37c-3d0c04fb1577, 0d02d76a-82de-4788-944f-fb3e...dc3, 68473988-4cef-4f6e-b556-df28d8b22560, 16ef8f73-7e94-4cb6-8d95-dc38d1b52416, d88e2c60-e2dc-420c-9e57-76e07f27f6aa]>
  timeout = 60, on_interval = None, on_message = None
  kwargs = {'interval': 0.5, 'no_ack': True}, prev_on_m = None, _ = None
  
      def _wait_for_pending(self, result,
                            timeout=None, on_interval=None, on_message=None,
                            **kwargs):
          self.on_wait_for_pending(result, timeout=timeout, **kwargs)
          prev_on_m, self.on_message = self.on_message, on_message
          try:
  >           for _ in self.drain_events_until(
                      result.on_ready, timeout=timeout,
                      on_interval=on_interval):
  
  celery/backends/asynchronous.py:287: 
  _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
  
  self = <celery.backends.asynchronous.Drainer object at 0x7f6132ec6560>
  p = <promise@0x7f6132f0bd90>, timeout = 60, interval = 1, on_interval = None
  wait = <bound method ResultConsumer.drain_events of <celery.backends.redis.ResultConsumer object at 0x7f6132ec6710>>
  
      def drain_events_until(self, p, timeout=None, interval=1, on_interval=None, wait=None):
          wait = wait or self.result_consumer.drain_events
          time_start = time.monotonic()
      
          while 1:
              # Total time spent may exceed a single call to wait()
              if timeout and time.monotonic() - time_start >= timeout:
  >               raise socket.timeout()
  E               TimeoutError
  
              redis_connection.delete(fail_sig_id)
[...]

PR #7837, opened 2022-10-19
Stabilized test_mutable_errback_called_by_chord_from_group_fail_multiple

Nusnus:
`test_mutable_errback_called_by_chord_from_group_fail_multiple` was failing randomly for some time now.
I suspect the arbitrary task count used in the test was too high so I lowered it from 24 to 8.
PR Opened due to Issue #7836

### DIFF
--- a/t/integration/test_canvas.py
+++ b/t/integration/test_canvas.py
@@ -2472,10 +2472,8 @@ class test_chord:
             await_redis_count(fail_task_count, redis_key=redis_key)
         redis_connection.delete(redis_key)
 
-    @pytest.mark.parametrize(
-        "errback_task", [errback_old_style, errback_new_style, ],
-    )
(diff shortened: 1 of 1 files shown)


Nusnus:
I am trying to play with more arbitrary values in the test. Let's see if it works.

Nusnus:
The test had two different cases implemented under one roof. I'm trying to separate it for a test per (logical) case, hoping it would reduce the effort of each test and hopefully will make it stable.

As of making this comment, it seems my fix works.
CleanShot 2022-10-21 at 02 29 57@2x

I will re-run all of the automatic tests a few times in a row, just for fun, to see if the fix is deterministic or if I just had "luck" this run.

Nusnus:
I want to believe 3 successful runs in a row is good enough, what do you say @auvipy ?
CleanShot 2022-10-21 at 03 52 42@2x

auvipy (approved):

PR #7852, opened 2022-10-25
even if it is cached no_ack should affect to the get function

nistick21:

## Description

I'm beginer of celery. I'm trying to ampq as broker and backend.

in tutorial, result is not removed by get() function

for example,

case1
```
result = add.delay(4, 4)
result.ready()
result.get()
```

case2
```
result = add.delay(4, 4)
result.get()
```

when I was trying to case 1,  not removed from result queue.
but case 2 does.

so I add 1 line in get function.

I'm  not good at programming, If you have a another solution. plz tell me.

thanks.

### DIFF
--- a/celery/result.py
+++ b/celery/result.py
@@ -215,7 +215,7 @@ class AsyncResult(ResultBase):
         if on_interval:
             _on_interval.then(on_interval)
 
-        if self._cache:
+        if self._cache and not no_ack:
             if propagate:
                 self.maybe_throw(callback=callback)
             return self.result


celery/result.py
```diff
@@ -215,7 +215,7 @@ def get(self, timeout=None, propagate=True, interval=0.5,
         if on_interval:
             _on_interval.then(on_interval)
 
-        if self._cache:
+        if self._cache and not no_ack:
```
auvipy, line 218: can you please add a unit test to verify this change?

auvipy (changes requested):

celery/result.py
nistick21: I have no idea about unit test, can you explain how to do this?

nistick21: To explain a little more in detail, if it becomes a cache through the ready function, no_ack is False and the consumer operates. After that, through the get function, the consumer should operate with no_ack set to True, but the cache value is returned and the function is terminated without operating the consumer with no_ack set to True.

I don't know how make exactly test code.  I'm not sure how to write code that correct tests this phenomenon.

auvipy:
@Nusnus feel free to give this a tryand let me know what do you think about this change

Nusnus:
From what I can see, the issue makes sense, but I'm not sure if we want this exact solution.
`ready()` indeed creates a cache which explains why `get(no_ack=True)`  would return from the cache.
That being said, changing `get()` implementation to skip the cache on `no_ack==True` may lead to confusion, as this is a specific-case issue and you don't want to tie the caching to the task acknowledgment just for that.
Second, this may break the API for cases where a cache exists, and you call `get()` (with default `no_ack==True`), when the expected behavior would be to get from the cache, but this change will break this (== cache is valid but not used due to acknowledgment configuration).
So the problem is that we want to separate the caching and the acknowledgment and this PR creates a dependency that breaks the API as well which is why I don't think it qualifies with the current implementation @auvipy 

@nistick21 In other words, your fix applies well to your case but will cause other people to have a different behavior than expected without an explicit change, which is prone to issues that are hard to find and fix, especially in production systems. So we have to be careful about what we change, even if all of the unit/integration tests of the CI are passing.

Lastly, if I assume correctly and the end results are to be able to change the state of the task and get the result without invoking the cache, we need an explicit and clear way of implementing this. An example off the top of my head may be adding a (default `False`) argument to `get()` to ignore cache, for example:
```python
def get(self, timeout=None, propagate=True, interval=0.5,
            no_ack=True, follow_parents=True, callback=None, on_message=None,
            on_interval=None, disable_sync_subtasks=True,
            EXCEPTION_STATES=states.EXCEPTION_STATES,
            PROPAGATE_STATES=states.PROPAGATE_STATES,
            ignore_cache=False):
...
if not ignore_cache and self._cache:
    if propagate:
    return self.result
```
Then you can have your example like this:
```python
result = add.delay(4, 4)
result.ready()
result.get(ignore_cache=True)
case2
```
```python
result = add.delay(4, 4)
result.get()
```

This will make the code much more readable, as the `result.get(ignore_cache=True)` will tell 100% what is happening and will remove confusion. In addition, it will not create a dependency between the task acknowledgment and the caching.

If this makes sense @nistick21 and you wish to take this further, LMK and I'll try to push you in the right direction. For now here's my two cents :)

P.S
This "proposed solution" is just to convey my point - there might be better implementations, but it requires more investigation.

Nusnus (changes requested):
Changing to Draft.

Issue #7581, opened 2022-06-17
Option to disable running Django system checks on startup


joekohlsdorf:
On startup, Celery runs the Django system checks, there is currently no option to disable this.
See: https://github.com/celery/celery/blob/master/celery/fixups/django.py#L136

In large projects, this can take a considerable amount of time and typically you don't want to do this in production.
If you run Django in a WSGI server you don't get system checks either.

I wanted to get the communities opinion to see if it would be worth submitting a PR which lets the user disable this, similar to the `--skip-checks` option which Django commands have.

auvipy:
i think that would be a good addition, but need to provide unit/ integration tests

PR #7859, opened 2022-10-28
Add --skip-checks flag to bypass django core checks

mudetz:
## Description

Resolves feature request #7581

### DIFF
--- a/celery/bin/celery.py
+++ b/celery/bin/celery.py
@@ -131,9 +131,15 @@ else:
               cls=CeleryOption,
               is_flag=True,
               help_group="Global Options")
+@click.option('--skip-checks',
+              envvar='SKIP_CHECKS',
+              cls=CeleryOption,
+              is_flag=True,
+              help_group="Global Options",
+              help="Skip Django core checks on startup.")
 @click.pass_context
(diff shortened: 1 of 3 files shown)


docs/whatsnew-5.2.rst
```diff
@@ -391,3 +391,9 @@ a TLS connection to ArangoDB result backend.
 
 If you'd like to do so, set the ``verify`` key in the
 :setting:`arangodb_backend_settings` dictionary to ``True``.
+
```
auvipy, line 394: this will be added to v5.3.0, so this part should be removed and find a location in CLI part of the doc

auvipy (changes requested):

auvipy (changes requested):
we should have this new command flag documented here
https://docs.celeryq.dev/en/stable/reference/cli.html or look for appropriate place to describe the new command with versionadded 5.3 notation

mudetz:
> we should have this new command flag documented here https://docs.celeryq.dev/en/stable/reference/cli.html or look for appropriate place to describe the new command with versionadded 5.3 notation

It is documented automatically in the CLI reference: https://celery--7859.org.readthedocs.build/en/7859/reference/cli.html

PR #7876, opened 2022-11-02
fix typos in optional tests

hsk17:
had a look at some of the optional tests, i.e. tests that are skipped if the corresponding python module is not installed, and found two small glitches:  in test_schedules, a spurious 0 was appended to the module name, and in test_cache, though the python package is named memcached, the to-be-imported module is memcache

### DIFF
--- a/t/unit/backends/test_cache.py
+++ b/t/unit/backends/test_cache.py
@@ -143,7 +143,7 @@ class test_CacheBackend:
         assert b.as_uri() == backend
 
     def test_regression_worker_startup_info(self):
-        pytest.importorskip('memcached')
+        pytest.importorskip('memcache')
         self.app.conf.result_backend = (
             'cache+memcached://127.0.0.1:11211;127.0.0.2:11211;127.0.0.3/'
         )
--- a/t/unit/app/test_schedules.py
+++ b/t/unit/app/test_schedules.py
@@ -26,7 +26,7 @@ def patch_crontab_nowfun(cls, retval):
 class test_solar:
 
     def setup(self):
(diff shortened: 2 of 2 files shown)


t/unit/backends/test_cache.py
```diff
@@ -143,7 +143,7 @@ def test_as_uri_multiple_servers(self):
         assert b.as_uri() == backend
 
     def test_regression_worker_startup_info(self):
-        pytest.importorskip('memcached')
+        pytest.importorskip('memcache')
```
auvipy, line 146:
```suggestion
        pytest.importorskip('memcached')
```

auvipy: isn't it memcached? https://memcached.org/

auvipy (changes requested):

hsk17:
the package from https://memcached.org/ provides the C-written memcached daemon and does AFAICS not provide a python client to access the daemon.  that is done by python package python-memcached, https://github.com/linsomniac/python-memcached/, which provides python module memcache.  of course, to make sense the python module needs the memcached daemon to be running.  a more recent python package to access memcached might be pymemcache, https://github.com/pinterest/pymemcache, which provides python module pymemcache, but i did not check this.

summary:  no, it should be memcache here IMHO :-)

t/unit/backends/test_cache.py
hsk17: no, the to-be-imported python module is named "memcache", see my comment below

auvipy (approved):

PR #7879, opened 2022-11-05
Sphinx extension: use inspect.signature to make it Python 3.11 compatible

mathiasertl:
`celery.contrib.sphinx` uses `inspect.formatargspec`, which has been removed in Python 3.11 in favor of `inspect.signature()`.

As a bonus, the new implementation also no longer renders non-existent `*args, **kwargs`. For example the signature of this task:

```python
@shared_task
def example() -> None:
    pass
```

renders as `(*args, **kwds) → [None]` in the old version, while (correctly) as `() -> None` in the new version.

### DIFF
--- a/celery/contrib/sphinx.py
+++ b/celery/contrib/sphinx.py
@@ -30,7 +30,7 @@ syntax.
 
 Use ``.. autotask::`` to alternatively manually document a task.
 """
-from inspect import formatargspec, getfullargspec
+from inspect import signature
 
 from sphinx.domains.python import PyFunction
 from sphinx.ext.autodoc import FunctionDocumenter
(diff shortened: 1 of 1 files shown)


celery/contrib/sphinx.py
```diff
@@ -30,7 +30,7 @@
 
 Use ``.. autotask::`` to alternatively manually document a task.
 """
-from inspect import formatargspec, getfullargspec
+from inspect import signature
```
auvipy, line 33: this is backward compatible with python 3.7?

auvipy (approved):

auvipy:
thanks a lot.

PR #7883, opened 2022-11-07
Fixed bug where chain flattening removed previous links

Nusnus:
When using a chain of chains, the flattened chain was missing the existing link/link_error from before.
A new unit test reproduced this bug to confirm the fix works.
This changes an assumption from #4240, where the flattened chain did not copy the link_error, and now it does, so the tests were adjusted accordingly.


Nusnus:
My fix created a worse bug. Closing the PR until I will have a stable implementation.
@auvipy FYI as I see you've started reviewing it :)

PR #7886, opened 2022-11-08
cryptography==38.0.3

auvipy:


auvipy:
just an approval needed before merge. that will improve score in openssf

PR #7891, opened 2022-11-09
Canvas.py doc enhancement

Nusnus:


Nusnus:
I'm creating separated PRs for new documentation to canvas.py for organizational purposes.
More will come soon..

celery/canvas.py
```diff
@@ -311,6 +311,12 @@ class Signature(dict):
 
     @classmethod
     def register_type(cls, name=None):
+        """Register a new type of signature.
```
thedrow, line 314: We should also write something about this in our user guide.

celery/canvas.py
```diff
@@ -319,6 +325,10 @@ def _inner(subclass):
 
     @classmethod
     def from_dict(cls, d, app=None):
```
thedrow, line 327: While you're documenting this methods, it would also be useful to annotate the types of the arguments and the return value.

celery/canvas.py
```diff
@@ -413,6 +423,24 @@ def apply_async(self, args=None, kwargs=None, route_name=None, **options):
         return _apply(args, kwargs, **options)
 
     def _merge(self, args=None, kwargs=None, options=None, force=False):
```
thedrow, line 425: Please annotate argument types and return types.

celery/canvas.py
```diff
@@ -685,6 +750,16 @@ def election(self):
             return type.AsyncResult(tid)
 
     def reprcall(self, *args, **kwargs):
```
thedrow, line 752:
```suggestion
    def reprcall(self, *args, **kwargs) -> str:
```

thedrow (changes requested):
Type annotations are missing. The documentation itself seems accurate.

thedrow (approved):
On second thought, we can address annotations in another PR.

PR #7901, opened 2022-11-11
Fix Issue #604 in django-celery-beat: Celery Beat Crashing at the End of Daylight Savings 

### LINKED ISSUE #604: Celery Beat Crashing at the End of Daylight Savings
### Summary:
* Celery Version: 4.4.7
* Celery-Beat Version: 2.2.0 

### Exact steps to reproduce the issue:
1. Set `CELERY_TIMEZONE` equal to `America/Los_Angeles`
2. Run the Celery beat such that an `IntervalSchedule` task runs between 1 AM and 2 PM on the day that daylight savings ends. In our case, this was 2022-11-06.
3. Celery beat will crash with the following traceback:

```
[INFO] [2022-11-06 01:18:52,155] [18813] [schedulers.schedule:356] - DatabaseScheduler: Schedule changed.
Process Beat:
  File "/path/to/virtualenv/lib/python3.7/site-packages/billiard/process.py", line 327, in _bootstrap
    self.run()
  File "/path/to/virtualenv/lib/python3.7/site-packages/celery/beat.py", line 707, in run
    self.service.start(embedded_process=True)
  File "/path/to/virtualenv/lib/python3.7/site-packages/celery/beat.py", line 631, in start
    interval = self.scheduler.tick()
  File "/path/to/virtualenv/lib/python3.7/site-packages/celery/beat.py", line 329, in tick
    self.populate_heap()
  File "/path/to/virtualenv/lib/python3.7/site-packages/celery/beat.py", line 303, in populate_heap
    is_due, next_call_delay = entry.is_due()
  File "/path/to/virtualenv/lib/python3.7/site-packages/django_celery_beat/schedulers.py", line 135, in is_due
    return self.schedule.is_due(last_run_at_in_tz)
  File "/path/to/virtualenv/lib/python3.7/site-packages/celery/schedules.py", line 171, in is_due
    rem_delta = self.remaining_estimate(last_run_at)
  File "/path/to/virtualenv/lib/python3.7/site-packages/celery/schedules.py", line 137, in remaining_estimate
    self.maybe_make_aware(self.now()), self.relative,
  File "/path/to/virtualenv/lib/python3.7/site-packages/celery/schedules.py", line 76, in now
    return (self.nowfun or self.app.now)()
  File "/path/to/virtualenv/lib/python3.7/site-packages/django_celery_beat/models.py", line 175, in <lambda>
    nowfun=lambda: make_aware(now())
  File "/path/to/virtualenv/lib/python3.7/site-packages/django_celery_beat/utils.py", line 27, in make_aware
    value = timezone.make_aware(value, timezone.get_default_timezone())
  File "/path/to/virtualenv/lib/python3.7/site-packages/django/utils/timezone.py", line 270, in make_aware
    return timezone.localize(value, is_dst=is_dst)
  File "/path/to/virtualenv/lib/python3.7/site-packages/pytz/tzinfo.py", line 363, in localize
    raise AmbiguousTimeError(dt)
pytz.exceptions.AmbiguousTimeError: 2022-11-06 01:18:52.418857
```

An easier way to reproduce this issue is by running the following script in the shell (`python manage.py shell`):
```
from datetime import datetime
from django_celery_beat.utils import make_aware

dt = datetime(2022, 11, 6, 1, 36, 0)
make_aware(dt)
```

### Detailed information
We were able to pinpoint the root cause after investigating the issue.
1. The `IntervalSchedule` will define the `nowfun` function as`make_aware` ([Source](https://github.com/celery/django-celery-beat/blob/master/django_celery_beat/models.py#L179))
2. The `now()` function inside `celery/schedules.py` will call the `nowfun`, subsequently `make_aware` ([Source](https://github.com/celery/celery/blob/master/celery/schedules.py#L128))
3. The `make_aware` function will call the `make_aware` function from `django/utils/timezone.py` **without** passing in `is_dst` as a parameter ([Source](https://github.com/celery/django-celery-beat/blob/master/django_celery_beat/models.py#L179))
4. The second `make_aware` function will call `localize` from `pytz/timeinfo.py` ([Source](https://github.com/django/django/blob/stable/3.1.x/django/utils/timezone.py#L234))
5. The `localize` will find two timezones associated with the `datetime` ([Source](https://github.com/stub42/pytz/blob/master/src/pytz/tzinfo.py#L321))
6. Since two timezones were found and `is_dst` is undefined, the function will throw an `AmbiguousTimeError` ([Source](https://github.com/stub42/pytz/blob/master/src/pytz/tzinfo.py#L362))

We have many Celery beat tasks running throughout the day. Because of this exception, we were not even able to start the Celery beat scheduler creating an outage for two hours. We had alerts to catch the issue as soon as it occurred, but there would have been a longer outage had we not caught the issue earlier.

### Possible Solutions
We were wondering if there is a way to configure the Celery beat such that the `IntervalSchedule` tasks will run despite the fact that the time may be in two timezones. A change like the following would prevent such problems from occurring:

```
value = timezone.make_aware(value, timezone.get_default_timezone(), is_dst=bool(time.localtime().tm_isdst))
```
[Line to change](https://github.com/celery/django-celery-beat/blob/master/django_celery_beat/utils.py#L27)

This might require some changes in the `remaining_estimate`. We can remove the following lines such that the time elapsed between the two timestamps is accurate:

```
    if str(start.tzinfo) == str(now.tzinfo) and now.utcoffset() != start.utcoffset():
        # DST started/ended
        start = start.replace(tzinfo=now.tzinfo)
```

[Line to change](https://github.com/celery/celery/blob/master/celery/utils/time.py#L193)

If `last_run_at` is 1:15 AM PDT and `now` is 1:30 AM PST, then this should return 1 hour 15 minutes instead of 15 minutes.

### Note
I am aware of https://github.com/celery/django-celery-beat/issues/285. However, this does not solve our issue. Our issue is that the Celery beat scheduler will crash and fail to send tasks for an entire two hours due to the behavior of `make_aware`.

polarmt:
## Dependency
https://github.com/celery/django-celery-beat/pull/605

**Note**: The two PRs are codependent and need to be merged together.

## Description
These changes are necessary to fix https://github.com/celery/django-celery-beat/issues/604 and is a dependency for https://github.com/celery/django-celery-beat/pull/605. Both issues have more detailed information about the problem and how testing was performed for the changes.


celery/utils/time.py
```diff
@@ -190,9 +190,6 @@ def remaining(start, ends_in, now=None, relative=False):
         ~datetime.timedelta: Remaining time.
     """
     now = now or datetime.utcnow()
-    if str(start.tzinfo) == str(now.tzinfo) and now.utcoffset() != start.utcoffset():
```
auvipy, line 193: since we are doing to remove this part, wouldn't it better to have some test to make sure this change won't create any regression?

auvipy (changes requested):

celery/utils/time.py
auvipy: all the unit tests are failing

polarmt: Looking into them

polarmt: Added unit tests.

auvipy:
can you rebase on top top of master again please?

auvipy:
I am not sure why builds are failing, can you check please? are the tests passing locally?

polarmt:
> I am not sure why builds are failing, can you check please? are the tests passing locally?

I just check. It should be good now.

polarmt:
I am a little confused as to what the behavior should be for Celery beat. Currently, the only unit test that is failing is from https://github.com/celery/celery/blob/master/t/unit/app/test_schedules.py#L441. This unit test was added to deal with https://github.com/celery/celery/issues/1604.

This is where the confusion lies. During the end of DST, my understanding is that we do want the tasks to run twice. For example, a Celery task runs at 1:30 PDT and 1:30 PST. From our business perspective, this makes the most sense. However, the problem in https://github.com/celery/celery/issues/1604 seems to be that they **do not** want to the tasks to run twice.

What should be the expected behavior? Should we add a configuration parameter to control this? If so, what is the default parameter?

auvipy:
I will come back to this soon

auvipy:
> I am a little confused as to what the behavior should be for Celery beat. Currently, the only unit test that is failing is from https://github.com/celery/celery/blob/master/t/unit/app/test_schedules.py#L441. This unit test was added to deal with #1604.
> 
> This is where the confusion lies. During the end of DST, my understanding is that we do want the tasks to run twice. For example, a Celery task runs at 1:30 PDT and 1:30 PST. From our business perspective, this makes the most sense. However, the problem in #1604 seems to be that they **do not** want to the tasks to run twice.
> 
> What should be the expected behavior? Should we add a configuration parameter to control this? If so, what is the default parameter?

can you please revisit this? I want to know more of your thoughts

polarmt:
For instance, let us assume that we have a task that we want to run every hour. An example of such important task is sending a timely email every hour at the 30 minute mark. Then, we would want this task to run both at 1:30 AM PDT AND 1:30 AM PST.

In our business, we send such timely emails for many of our features and having one hour of downtime is detrimental.

auvipy:
> For instance, let us assume that we have a task that we want to run every hour. An example of such important task is sending a timely email every hour at the 30 minute mark. Then, we would want this task to run both at 1:30 AM PDT AND 1:30 AM PST.
> 
> In our business, we send such timely emails for many of our features hourly at different minute marks and having one hour of downtime is detrimental.

I think running a task multiple times in different timezone should be allowed.

polarmt:
Thanks, @auvipy ! In that case, keep us posted on the progress of the review.

auvipy:
can't review it as the builds are failing and there are merge conflicts. we need those two resolved before moving further.

polarmt:
Can you run the builds again?

auvipy:
re started and builds are failing

auvipy:
can you restart this?

polarmt:
Yes. Yesterday, we discovered a potential bug with this change that was exposed by the latest DST transition. I will summarize it and fix it when I have time.

polarmt:
## Root Cause of Bug
In the existing logic of the Celery source code, it will convert the `start` timezone to the current timezone (`PDT` to `PST`) and then compute the difference between `start` and `now`. If we compare 1:15 PM PST and 1:15 PM PDT, the difference should be one hour apart. However, with the unnecessary conversion, the `remaining` function would return 0 minutes. To fix this issue, we skipped the conversion of `start` from `PDT` to `PST` before computing the expected end time.

This works for periodic tasks that run every X minutes, but the logic seems to be incorrect if the beat task uses a `crontab`. `time.remaining` takes in three main parameters: `now`, `ends_in`, and `start`. The `ends_in` represents when we expect the task to run next. This value can either be a `timedelta` object or an `ffwd` object.

An `ffwd` object does not represent the time difference between two times. Instead, "adding" an `ffwd` object will replace the current `datetime` object with the times specified in the `ffwd` object. For instance, if we add `ffwd(hour=7, minute=0, weekday=6)` to 1:15 AM on Saturday, the resulting time would become 7:00 AM on the following Sunday.

If the Saturday in the example was prior to the transition and the Sunday was afterwards, then we expect the `start` (1:15 AM) to be in PDT and `end_date` (7:00 AM) to be PST. Because we skip the conversion of `start` in the new logic, the `end_date` will be incorrectly in PDT even though we have transitioned to PST. For `ffwd`, the changes will not work.

## Symptoms of Bug
The first time that the Celery beat task will run will be an hour earlier than the scheduled time. Then, the Celery beat task will run again at the expected time. This causes the Celery beat task to run an additional time unnecessarily.

## Proposed Fix
We can check if `ends_in` is an `ffwd` object or not before deciding whether to convert `ends_in` or not:

```
   if str(start.tzinfo) == str(now.tzinfo) and now.utcoffset() != start.utcoffset():
        if now.utcoffset() > start.utcoffset() or isinstance(ends_in, ffwd):
            # DST started
            start = start.replace(tzinfo=now.tzinfo)
```

polarmt:
@auvipy
1. What do you think about the proposed fix to the problem?
2. Can we set up some time to discuss how to keep the CI stable? The problem is that every time I rebase, the main branch will be updated causing the CI to fail again.

auvipy (changes requested):
can you please rebase now? CI is stable now, with sometimes codcov network outage.

auvipy:
> @auvipy
> 
>     1. What do you think about the proposed fix to the problem?
> 
>     2. Can we set up some time to discuss how to keep the CI stable? The problem is that every time I rebase, the main branch will be updated causing the CI to fail again.

you can propose the changes with additional tests case for review

auvipy (changes requested):
  ________________________________ test_remaining ________________________________
  
      def test_remaining():
          # Relative
          remaining(datetime.utcnow(), timedelta(hours=1), relative=True)
      
          """
          The upcoming cases check whether the next run is calculated correctly
          """
          eastern_tz = ZoneInfo("US/Eastern")
          tokyo_tz = ZoneInfo("Asia/Tokyo")
      
          # Case 1: `start` in UTC and `now` in other timezone
          start = datetime.now(ZoneInfo("UTC"))
          now = datetime.now(eastern_tz)
          delta = timedelta(hours=1)
          assert str(start.tzinfo) == str(ZoneInfo("UTC"))
          assert str(now.tzinfo) == str(eastern_tz)
          rem_secs = remaining(start, delta, now).total_seconds()
          # assert remaining time is approximately equal to delta
          assert rem_secs == pytest.approx(delta.total_seconds(), abs=1)
      
          # Case 2: `start` and `now` in different timezones (other than UTC)
          start = datetime.now(eastern_tz)
          now = datetime.now(tokyo_tz)
          delta = timedelta(hours=1)
          assert str(start.tzinfo) == str(eastern_tz)
          assert str(now.tzinfo) == str(tokyo_tz)
          rem_secs = remaining(start, delta, now).total_seconds()
          assert rem_secs == pytest.approx(delta.total_seconds(), abs=1)
      
          """
          Case 3: DST check
          Suppose start (which is last_run_time) is in EST while next_run is in EDT,
          then check whether the `next_run` is actually the time specified in the
          start (i.e. there is not an hour diff due to DST).
          In 2019, DST starts on March 10
          """
  >       start = eastern_tz.localize(datetime(day=9, month=3, year=2019, hour=10, minute=0))         # EST
  E       AttributeError: 'zoneinfo.ZoneInfo' object has no attribute 'localize'
  
  t/unit/utils/test_time.py:161: AttributeError

polarmt:
I think I fixed the issue. We can see whether the CI passes.

celery/utils/time.py
```diff
@@ -221,9 +221,11 @@ def remaining(
     if str(
             start.tzinfo) == str(
             now.tzinfo) and now.utcoffset() != start.utcoffset():
-        # DST started/ended
-        start = start.replace(tzinfo=now.tzinfo)
+        if now.utcoffset() > start.utcoffset() or isinstance(ends_in, ffwd):
+            # DST started
+            start = start.replace(tzinfo=now.tzinfo, fold=now.fold)
     end_date = start + ends_in
+    print(start, end_date)
```
auvipy, line 228: we might not keep print in final version

auvipy (changes requested):

auvipy (changes requested):
The following test fails as per CI report. we need to fix it to make everything green again

  ________________________________ test_remaining ________________________________
  
      def test_remaining():
          # Relative
          remaining(datetime.utcnow(), timedelta(hours=1), relative=True)
      
          """
          The upcoming cases check whether the next run is calculated correctly
          """
          eastern_tz = ZoneInfo("US/Eastern")
          tokyo_tz = ZoneInfo("Asia/Tokyo")
          eastern_tz_pytz = pytz.timezone("US/Eastern")
          tokyo_tz_pytz = pytz.timezone("Asia/Tokyo")
      
          # Case 1: `start` in UTC and `now` in other timezone
          start = datetime.now(ZoneInfo("UTC"))
          now = datetime.now(eastern_tz)
          delta = timedelta(hours=1)
          assert str(start.tzinfo) == str(ZoneInfo("UTC"))
          assert str(now.tzinfo) == str(eastern_tz)
          rem_secs = remaining(start, delta, now).total_seconds()
          # assert remaining time is approximately equal to delta
          assert rem_secs == pytest.approx(delta.total_seconds(), abs=1)
      
          # Case 2: `start` and `now` in different timezones (other than UTC)
          start = datetime.now(eastern_tz)
          now = datetime.now(tokyo_tz)
          delta = timedelta(hours=1)
          assert str(start.tzinfo) == str(eastern_tz)
          assert str(now.tzinfo) == str(tokyo_tz)
          rem_secs = remaining(start, delta, now).total_seconds()
          assert rem_secs == pytest.approx(delta.total_seconds(), abs=1)
      
          """
          Case 3: DST check
          Suppose start (which is last_run_time) is in EST while next_run is in EDT,
          then check whether the `next_run` is actually the time specified in the
          start (i.e. there is not an hour diff due to DST).
          In 2019, DST starts on March 10
          """
          start = datetime(day=9, month=3, year=2019, hour=10, minute=0, tzinfo=eastern_tz)         # EST
          now = datetime(day=11, month=3, year=2019, hour=1, minute=0, tzinfo=eastern_tz)           # EDT
          delta = ffwd(hour=10, year=2019, microsecond=0, minute=0, second=0, day=11, weeks=0, month=3)
          # `next_actual_time` is the next time to run (derived from delta)
          next_actual_time = datetime(
              day=11, month=3, year=2019, hour=10, minute=0, tzinfo=eastern_tz)  # EDT
          assert start.tzname() == "EST"
          assert now.tzname() == "EDT"
          assert next_actual_time.tzname() == "EDT"
          rem_time = remaining(start, delta, now)
          next_run = now + rem_time
          assert next_run == next_actual_time
      
          """
          Case 4: DST check (ZoneInfo, timedelta)
          Suppose start (which is last_run_time) is in PDT while next_run is in PST,
          Check whether there is an hour added to the time between now and the `next_run`
          In 2022, DST ends on Nov 6
          """
          start = datetime(day=6, month=11, year=2022, hour=1, minute=15, tzinfo=eastern_tz, fold=0)
          now = datetime(day=6, month=11, year=2022, hour=1, minute=34, tzinfo=eastern_tz, fold=1)
          ends_in = timedelta(minutes=80)
          next_actual_time = datetime(day=6, month=11, year=2022, hour=1, minute=35, tzinfo=eastern_tz, fold=1)
          assert start.tzname() == "EDT"
          assert now.tzname() == "EST"
          assert next_actual_time.tzname() == "EST"
          rem_time = remaining(start, ends_in, now)
          print(start + ends_in - now)
          next_run = now + rem_time
  >       assert next_run == next_actual_time
  E       AssertionError: assert datetime.datetime(2022, 11, 6, 2, 35, tzinfo=ZoneInfo(key='US/Eastern')) == datetime.datetime(2022, 11, 6, 1, 35, tzinfo=ZoneInfo(key='US/Eastern'), fold=1)
  
  t/unit/utils/test_time.py:193: AssertionError

t/unit/utils/test_time.py
```diff
@@ -177,6 +174,81 @@ def test_remaining():
...
+    """
+    start = datetime(day=6, month=11, year=2022, hour=1, minute=15, tzinfo=eastern_tz, fold=0)
+    now = datetime(day=6, month=11, year=2022, hour=1, minute=34, tzinfo=eastern_tz, fold=1)
+    ends_in = timedelta(minutes=80)
+    next_actual_time = datetime(day=6, month=11, year=2022, hour=1, minute=35, tzinfo=eastern_tz, fold=1)
+    assert start.tzname() == "EDT"
+    assert now.tzname() == "EST"
+    assert next_actual_time.tzname() == "EST"
+    rem_time = remaining(start, ends_in, now)
+    print(start + ends_in - now)
```
auvipy, line 191: I guess we can remove the print() func which might be used for debug purpose here?

auvipy (changes requested):

t/unit/utils/test_time.py
polarmt: Removed the print() function

auvipy:
It seems unit tests are now passing. I restarted the failing pypy build as it seems to be a network issue

auvipy (commented):
you need to check/fix the lint issues as well. thanks for all your efforts here!! I will come back to this next morning to check if it is merge able with 5.3.5 release

auvipy:
Integration tests are now passing

auvipy:
t/unit/utils/test_time.py:136:5: F841 local variable 'tokyo_tz_pytz' is assigned to but never used
t/unit/utils/test_time.py:193:1: W293 blank line contains whitespace
t/unit/utils/test_time.py:223:118: E501 line too long (118 > 117 characters)
t/unit/utils/test_time.py:230:1: W293 blank line contains whitespace

auvipy:
Will try to push lint fixes

auvipy:
I might consider this for 5.3.6 if it do not break backward compat

auvipy:
@Nusnus can we consider this for a patch release?

Nusnus:
@auvipy 
> @Nusnus can we consider this for a patch release?

I haven't fully reviewed the entire PR, but from the original description:
> https://github.com/celery/django-celery-beat/pull/605
>
> Note: The two PRs are codependent and need to be merged together.

It looks like the other PR isn't ready, besides this PR having failures in CI.

That being said, if this work can be baked enough, I support merging on, but we'll have it for v5.4 if it will be ready by then.

auvipy:
Yeah it's better off for 5.4

celery/utils/time.py
```diff
@@ -218,12 +218,15 @@ def remaining(
         ~datetime.timedelta: Remaining time.
     """
     now = now or datetime.utcnow()
```
cclauss, line 220: https://docs.python.org/3/library/datetime.html#datetime.datetime.utcnow is deprecated in Python 3.12

t/unit/utils/test_time.py
```diff
@@ -131,6 +132,8 @@ def test_remaining():
     """
     eastern_tz = ZoneInfo("US/Eastern")
     tokyo_tz = ZoneInfo("Asia/Tokyo")
+    eastern_tz_pytz = pytz.timezone("US/Eastern")
```
auvipy, line 135: instead of pytz, shouldn't we use ZoneInfo?

auvipy (changes requested):

t/unit/utils/test_time.py
auvipy: https://stackoverflow.com/questions/70249473/convert-an-aware-datetimes-timezone-to-utc-with-zoneinfo-migration-away-from-p

cclauss: https://pypi.org/project/pytz
> [!NOTE]
> Projects using Python 3.9 or later should be using the support now included as part of the standard library, and third-party packages work with it, such as [tzdata](https://pypi.org/project/tzdata/). ___pytz offers no advantages___ beyond backwards compatibility with code written for earlier versions of Python.

PR #7916, opened 2022-11-21
Stamping a task left the task properties dirty

Nusnus:


celery/app/base.py
```diff
@@ -778,6 +778,10 @@ def send_task(self, name, args=None, kwargs=None, countdown=None,
             **options
         )
 
+        stamped_headers = options.pop('stamped_headers', [])
```
auvipy, line 781: do we need additional unit tests for this part? or they are already covered? the Integration tests are good

auvipy (approved):

celery/app/base.py
Nusnus: The codecov marks these lines as covered, although I didn't add anything specific (makes sense though). I think the current tests for this PR should be enough, but I'll keep it in mind.

Nusnus:
> also could you find additional issues like this?

Sure!

And thanks for the review @auvipy ! 🤙

Issue #7874, opened 2022-11-02
Bare-Bones Python 3.8+ Asyncio Coroutine Support For Celery 5+


the-wondersmith:
# Checklist

- [X] I have checked the [issues list](https://github.com/celery/celery/issues?q=is%3Aissue+label%3A%22Issue+Type%3A+Enhancement%22+-label%3A%22Category%3A+Documentation%22)
  for similar or identical enhancement to an existing feature.
- [X] I have checked the [pull requests list](https://github.com/celery/celery/pulls?q=is%3Apr+label%3A%22Issue+Type%3A+Enhancement%22+-label%3A%22Category%3A+Documentation%22)
  for existing proposed enhancements.
- [X] I have checked the [commit log](https://github.com/celery/celery/commits/master)
  to find out if the if the same enhancement was already implemented in the
  master branch.
- [X] I have included all related issues and possible duplicate issues in this issue
      (If there are none, check this box anyway).

#### Related Issues

- [6552](https://github.com/celery/celery/issues/6552)
- [6483](https://github.com/celery/celery/issues/6483)

#### Possible Duplicates

- None

# Brief Summary

Building on the work initially done by @ShaheedHaque and @kai3341 in [celery-pool-asyncio](https://github.com/kai3341/celery-pool-asyncio), I've worked out [an absolutely bare-bones implementation](https://github.com/the-wondersmith/celery-pool-asyncio/tree/celery5) of a Python 3.8+ / Celery 5+ compatible worker pool class that enables the use of vanilla-Python async functions (coroutine functions) with Celery's standard `app.task` decorator. It does not interfere with or preclude the continued use of extant synchronous functions, nor does it force extant codebases to convert current code to async.

# Design

The implementation is a _heavily_ stripped back version of @kai3341's existing Celery 4 compatible `celery-pool-asyncio` package. In does not include _automatic_ monkey-patching and more, and although it does utilize monkey the only thing being patched in that manner is `celery.app.trace.build_tracer`. The applied patch does not change the existing implementation of the `celery.app.trace.build_tracer` utility, except to use replace direct calls of a task function or its associated "handler" methods (i.e. `on_error`) with the new `AsyncIOPool`'s `run` method instead.

[...]

PR #7925, opened 2022-11-27
Add annotations to minimise differences with celery-aio-pool's tracer.py.

ShaheedHaque:
## Description

As per #7874, add annotations to trace.py to improve it's documentation, and make it easier to compare with celery-aio-pool's tracer.py. See also #the-wondersmith/celery-aio-pool/pull/4.

(Note: unit tests are currently badly broken)

### DIFF
--- a/celery/app/trace.py
+++ b/celery/app/trace.py
@@ -8,6 +8,7 @@ import os
 import sys
 import time
 from collections import namedtuple
+from typing import Any, Callable, Dict, FrozenSet, Optional, Sequence, Tuple, Type, Union
 from warnings import warn
 
 from billiard.einfo import ExceptionInfo
@@ -16,6 +17,8 @@ from kombu.serialization import loads as loads_message
 from kombu.serialization import prepare_accept_content
 from kombu.utils.encoding import safe_repr, safe_str
 
+import celery
+import celery.loaders.app
 from celery import current_app, group, signals, states
 from celery._state import _task_stack
 from celery.app.task import Context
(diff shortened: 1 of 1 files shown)


celery/app/trace.py
```diff
@@ -369,7 +382,12 @@ def build_tracer(name, task, loader=None, hostname=None, store_errors=True,
     from celery import canvas
     signature = canvas.maybe_signature  # maybe_ does not clone if already
 
-    def on_error(request, exc, state=FAILURE, call_errbacks=True):
+    def on_error(
+            request: celery.app.task.Context,
+            exc: Union[Exception, Type[Exception]],
+            state: str = FAILURE,
+            call_errbacks: bool = True) -> tuple[Info, Any, Any, Any]:
```
clokep, line 389: I think you'll want to use `Tuple` from `typing` here to support Python < 3.9 (I think it was 3.9 which added that...)

celery/app/trace.py
```diff
@@ -378,7 +396,13 @@ def on_error(request, exc, state=FAILURE, call_errbacks=True):
         )
         return I, R, I.state, I.retval
 
-    def trace_task(uuid, args, kwargs, request=None):
+    def trace_task(
+            uuid: str,
+            args: Sequence[Any],
+            kwargs: dict[str, Any],
+            request: Optional[dict[str, Any]] = None) -> trace_ok_t:
```
clokep, line 403: Again, I think you need to use `Dict` here:

```suggestion
            kwargs: Dict[str, Any],
            request: Optional[Dict[str, Any]] = None) -> trace_ok_t:
```

celery/app/trace.py
ShaheedHaque: Good catch.

ShaheedHaque:
@auvipy I think I accidentally removed you as a reviewer, but I am not sure how to add you back.

celery/app/trace.py
```diff
@@ -369,7 +382,12 @@ def build_tracer(name, task, loader=None, hostname=None, store_errors=True,
     from celery import canvas
     signature = canvas.maybe_signature  # maybe_ does not clone if already
 
-    def on_error(request, exc, state=FAILURE, call_errbacks=True):
+    def on_error(
+            request: celery.app.task.Context,
+            exc: Union[Exception, Type[Exception]],
+            state: str = FAILURE,
+            call_errbacks: bool = True) -> Tuple[Info, Any, Any, Any]:
```
clokep, line 389: I believe that `I.state` is a string, but that can always be tightened up in the future.

```suggestion
            call_errbacks: bool = True) -> Tuple[Info, Any, str, Any]:
```

clokep (approved):
I don't see any issues with it.

Nusnus:
This PR was reverted as it introduced a bug, unfortunately.
At the time, it was impossible to know. Now we have the means.

The work needs to be done again, apologies & thank you 🙏

Nusnus:
> I have reviewed the code and am not able to see what caused the issue? AFAICS, this commit consists of annotations which are not supposed to have any run time effect. Did some supporting import go wrong perhaps?

My first attempt was just reverting single lines - anything that wasn’t pure type annotation change. I also added `from __future__ import annotations` and ran `pyupgrade`, which further cleaned the original contribution.
I am sure I missed something, because it didn’t work. I was so tired when I stumbled upon this, so I just reverted the entire PR, and the original bug was gone, so I went with it (due to being _just_ a documentation PR and the bug being an **urgent production-breaking issue**).

> @Nusnus OK, sorry for the inconvenience.

No worries - on the contrary though, you’re welcome to redo the work now that you can trust the CI to let you know if something breaks, so go ahead - and tag me for review 🤙

PR #7919, opened 2022-11-23
Fixed bug when chaining a chord with a group

Nusnus:
Reported at Issue #5958

Issue #5958, opened 2020-02-11
Chain/Group possibile BUG


Ixiodor:
# Checklist

* [x] I have verified that the issue exists against the `master` branch of Celery.
* [ ] This has already been asked to the [discussion group](https://groups.google.com/forum/#!forum/celery-users) first.
* [x] I have read the relevant section in the
  [contribution guide](http://docs.celeryproject.org/en/latest/contributing.html#other-bugs)
  on reporting bugs.
* [x] I have checked the [issues list](https://github.com/celery/celery/issues?q=is%3Aissue+label%3A%22Issue+Type%3A+Bug+Report%22+-label%3A%22Category%3A+Documentation%22)
  for similar or identical bug reports.
* [x] I have checked the [pull requests list](https://github.com/celery/celery/pulls?q=is%3Apr+label%3A%22PR+Type%3A+Bugfix%22+-label%3A%22Category%3A+Documentation%22)
  for existing proposed fixes.
* [x] I have checked the [commit log](https://github.com/celery/celery/commits/master)
  to find out if the bug was already fixed in the master branch.
* [x] I have included all related issues and possible duplicate issues
  in this issue (If there are none, check this box anyway).

## Mandatory Debugging Information
* [ ] I have included the output of ``celery -A proj report`` in the issue.
    (if you are not able to do this, then at least specify the Celery
     version affected).
* [x] I have verified that the issue exists against the `master` branch of Celery.
* [ ] I have included the contents of ``pip freeze`` in the issue.
* [ ] I have included all the versions of all the external dependencies required
  to reproduce this bug.

## Optional Debugging Information

* [ ] I have tried reproducing the issue on more than one Python version
  and/or implementation.
* [ ] I have tried reproducing the issue on more than one message broker and/or
  result backend.
* [ ] I have tried reproducing the issue on more than one version of the message
  broker and/or result backend.
* [ ] I have tried reproducing the issue on more than one operating system.
* [ ] I have tried reproducing the issue on more than one workers pool.
* [ ] I have tried reproducing the issue with autoscaling, retries,
  ETA/Countdown & rate limits disabled.
* [ ] I have tried reproducing the issue after downgrading
  and/or upgrading Celery and its dependencies.

#### Related Issues
* None

#### Possible Duplicates
* None

## Environment & Settings

**Celery version**: 4.4.0 (cliffs)

<details>
<summary><b><code>celery report</code> Output:</b></summary>
<p>

```
[...]

PR #7938, opened 2022-12-04
Fixed a bug in losing chain links when unchaining an inner chain with links

Nusnus:
Related to Issue #5958 per being part of the same flow this issue was reporting about.
PR Includes:
1. Fix in `unchain_tasks` to include the links in the cloned tasks list.
2. Added unit test to check the `options` are structured correctly.
3. Added integration test to verify the link is indeed called.

### TL;DR regarding #5958
The bug that broke the chain `__or__` operator, also broke the linking of the `__or__`'ed chain.
[The fix](https://github.com/celery/celery/issues/7919) to Issue #5958 did not take into account this new links bug.
This then (hopefully) sums up the major bugs of the `__or__` operator in chains.

### DIFF
--- a/celery/canvas.py
+++ b/celery/canvas.py
@@ -960,14 +960,17 @@ class _chain(Signature):
     def unchain_tasks(self):
         """Return a list of tasks in the chain.
 
-        The tasks list would be cloned from the chain's tasks,
-        and all of the tasks would be linked to the same error callback
+        The tasks list would be cloned from the chain's tasks.
+        All of the chain callbacks would be added to the last task in the (cloned) chain.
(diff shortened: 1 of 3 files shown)


auvipy:
OK let me review again

Nusnus:
> The bug that broke the chain `__or__` operator, also broke the linking of the `__or__`'ed chain.

Is not 100% accurate TBH.
Better say that this is another aspect of the general issue which is what the other issue discusses.

auvipy:
yeah no worries. I don't expect all this fixed at once :p but lets gradually keep fixing!

auvipy (approved):

PR #7942, opened 2022-12-05
Housekeeping for Canvas.py

Nusnus:
Small housekeeping for `canvas.py` for the 5.3 release.
Will keep it in Draft until ready for merge.
Comments are welcome for any commit in the PR.

### DIFF
--- a/celery/canvas.py
+++ b/celery/canvas.py
@@ -161,7 +161,6 @@ class StampingVisitor(metaclass=ABCMeta):
          Returns:
              Dict: headers to update.
          """
-        pass
 
     def on_chord_header_start(self, chord, **header) -> dict:
         """Method that is called on сhord header stamping start.
--- a/celery/canvas.py
+++ b/celery/canvas.py
@@ -2247,13 +2247,14 @@ class _chord(Signature):
             applied to the body.
         """
         if self.app.conf.task_allow_error_cb_on_chord_header:
-            # self.tasks can be a list of the chord header workflow.
-            if isinstance(self.tasks, (list, tuple)):
-                for task in self.tasks:
-                    task.link_error(errback)
-            else:
-                self.tasks.link_error(errback)
+            for task in self.tasks:
+                task.link_error(errback)
         else:
+            # Once this warning is removed, the whole method needs to be refactored to:
+            # 1. link the error callback to each task in the header
+            # 2. link the error callback to the body
+            # 3. return the error callback
+            # In summary, up to 4 lines of code + updating the method docstring.
             warnings.warn(
                 "task_allow_error_cb_on_chord_header=False is pending deprecation in "
                 "a future release of Celery.\n"
--- a/celery/canvas.py
+++ b/celery/canvas.py
@@ -92,7 +92,7 @@ def _merge_dictionaries(d1, d2):
             else:
                 if isinstance(value, (int, float, str)):
                     d1[key] = [value]
-                if isinstance(d2[key], list):
+                if isinstance(d2[key], list) and d1[key] is not None:
                     d1[key].extend(d2[key])
                 else:
                     if d1[key] is None:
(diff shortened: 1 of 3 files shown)


Nusnus:
PR is ready for merge from my side.

celery/canvas.py
```diff
@@ -2248,13 +2247,14 @@ def link_error(self, errback):
...
         """
         if self.app.conf.task_allow_error_cb_on_chord_header:
-            # self.tasks can be a list of the chord header workflow.
-            if isinstance(self.tasks, (list, tuple)):
-                for task in self.tasks:
-                    task.link_error(errback)
-            else:
-                self.tasks.link_error(errback)
+            for task in self.tasks:
+                task.link_error(errback)
```
thedrow, line 2251: Why did you remove this condition? `self.tasks` may be a list or a tuple and if it isn't, it's a group.

E.g.
```
chord([s1.s(), s2.s()], body.s())
chord(group(s1.s(), s2.s()), body.s())
```

We need to support both cases.

t/unit/tasks/test_canvas.py
```diff
@@ -454,6 +455,9 @@ def test_OR(self):
         assert isinstance(ax, _chain)
         assert len(ax.tasks), 3 == 'consolidates chain to chain'
 
+        with pytest.raises(TypeError):
+            assert signature('foo') | None
+
```
thedrow, line 460: Please add a subtest that explains this assert.
There's no need to add subtests for all the cases but if we're adding a new assert, we should describe why it is there.

Nusnus:
@auvipy If you have a patch to fix the linter this could be a good time to add it in, so we can close open ends for the upcoming release.

t/unit/tasks/test_canvas.py
```diff
@@ -2386,3 +2395,63 @@ def test_is_dict(self):
...
+        d2 = {'dict': {'b': 2}, 'list': [3, 4], 'set': {'a', 'b'}}
+        _merge_dictionaries(d1, d2)
+        assert d1 == {
+            'dict': {'a': 1, 'b': 2},
+            'list': [1, 2, 3, 4],
+            'tuple': (1, 2),
+            'set': {'a', 'b'}
+        }
+
+    @pytest.mark.parametrize('d1,d2,expected_result', [
```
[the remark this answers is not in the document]
thedrow: Looks good

thedrow (changes requested):

celery/canvas.py
Nusnus: There's no problem with that, `self.tasks` is accessed correctly with both examples.
To prove it, I've added the PR a test case with your both examples: `test_link_error_on_chord_header()`
Check it out to confirm and resolve conversation if you're satisfied @thedrow

t/unit/tasks/test_canvas.py
Nusnus: Done

Nusnus:
All comments have been addressed, PR is ready for merge from my side.

Nusnus:
> PR is ready for merge from my side

This only means I have responded to all of the open ends from the PR, to mark "I have nothing else to do here".
So no pressure, but if all of the comments are resolved and fixed, then we shouldn't wait for the sake of waiting :)

Nusnus:
I don’t understand what's wrong with the integration tests, they just broke and the tests are not even running, and I'm not sure how to fix it.
Do you have any advice?
@auvipy @thedrow

auvipy:
can you please check which version of tox the CI is using now? tox got a new major release very recently AFAIK.

Nusnus:
The integration tests were running an old version, you were right @auvipy .
I've added a fix, let's hope everything works now 🙏

auvipy:
will do a thorough review tomorrow

thedrow (approved):
LGTM

PR #7945, opened 2022-12-06
Add beat_cron_starting_deadline_seconds to prevent unwanted cron runs

### LINKED ISSUE #4806: celery.schedules.crontab is_due logic can trigger periodic celery beat tasks at arbitrary times unrelated to the crontab parameters when last_run_at value is sufficiently old
There's an issue in the implementation of `celery.schedules.crontab` method that can cause `crontab.is_due` to claim a schedule is due at a time that is completely unrelated to the given crontab parameters. This appears to happen in cases where the `last_run_at` value is older than the most recent feasible time the schedule could have run. It looks like that this issue was introduced as part of complex logic added nearly 8 years ago to improve the accuracy of time remaining estimates for `crontab` periodic tasks: 4ed89ec49582b540149cf06047f091ebd20fb300

## Checklist
Issue first observed in a celery deployment running celery v3.1.19

From inspecting the `celery.schedules.crontab` code it appears likely that this issue is present in all celery versions as far back as v2.1.0 through to v4.1.1 .

Unit tests (please see below) confirm issue is still present in master (b599b96960be9dd42b3dee82a58bd1d711df0317 at time of writing).

## Steps to reproduce
Please apply this patch to celery master branch, remove `@skip.todo` from the first new test, run unit tests, observe the first of these added unit tests fails:

```
diff --git a/t/unit/app/test_schedules.py b/t/unit/app/test_schedules.py
index a7b3025..0340461 100644
--- a/t/unit/app/test_schedules.py
+++ b/t/unit/app/test_schedules.py
@@ -26,6 +26,18 @@ def patch_crontab_nowfun(cls, retval):
         cls.nowfun = prev_nowfun
 
 
+def is_time_feasible_wrt_crontab_schedule(t, z):
+    # z : celery.schedules.crontab instance
+    t = z.maybe_make_aware(t)
+    return (
+        t.month in z.month_of_year and
+        (t.isoweekday() % 7) in z.day_of_week and
+        t.day in z.day_of_month and
+        t.hour in z.hour and
+        t.minute in z.minute
+    )
+
+
 @skip.unless_module('ephem')
 class test_solar:
 
@@ -803,3 +815,59 @@ class test_crontab_is_due:
             due, remaining = self.yearly.is_due(datetime(2009, 3, 12, 7, 30))
             assert not due
             assert remaining == 4 * 24 * 60 * 60 - 3 * 60 * 60
+
+    @skip.todo('FIXME crontab logic is defective when last_run_at is older than the most recent feasible time wrt schedule')
+    def test_daily_execution_if_last_run_at_was_days_ago_and_current_time_does_not_match_crontab_schedule_then_execution_is_not_due(self):
+        z = self.crontab(hour=7, minute=30)
+        last_run_at = datetime(2018, 6, 1, 7, 30)
+        now = datetime(2018, 6, 9, 23, 48)
+        expected_next_execution_time = datetime(2018, 6, 10, 7, 30)
+        expected_remaining = (expected_next_execution_time - now).total_seconds()
+        # check our assumptions
+        assert is_time_feasible_wrt_crontab_schedule(last_run_at, z)
+        assert not is_time_feasible_wrt_crontab_schedule(now, z)
+        assert is_time_feasible_wrt_crontab_schedule(expected_next_execution_time, z)
+        assert now < expected_next_execution_time
+        assert expected_remaining == (7 * 60 + 30 + 12) * 60
+        # test is_due
+        with patch_crontab_nowfun(z, now):
+            due, remaining = z.is_due(last_run_at=last_run_at)
+            assert remaining == expected_remaining
+            assert not due
+
+    def test_daily_execution_if_last_run_at_was_the_most_recent_feasible_time_wrt_schedule_in_past_and_current_time_does_not_match_crontab_schedule_then_execution_is_not_due(self):
+        z = self.crontab(hour=7, minute=30)
+        last_run_at = datetime(2018, 6, 9, 7, 30)
+        now = datetime(2018, 6, 9, 23, 48)
+        expected_next_execution_time = datetime(2018, 6, 10, 7, 30)
+        expected_remaining = (expected_next_execution_time - now).total_seconds()
+        # check our assumptions
+        assert is_time_feasible_wrt_crontab_schedule(last_run_at, z)
+        assert not is_time_feasible_wrt_crontab_schedule(now, z)
+        assert is_time_feasible_wrt_crontab_schedule(expected_next_execution_time, z)
+        assert now < expected_next_execution_time
+        assert expected_remaining == (7 * 60 + 30 + 12) * 60
+        # test is_due
+        with patch_crontab_nowfun(z, now):
+            due, remaining = z.is_due(last_run_at=last_run_at)
+            assert remaining == expected_remaining
+            assert not due
+
+    def test_daily_execution_if_last_run_at_was_more_recent_than_the_most_recent_feasible_time_wrt_schedule_in_past_and_current_time_does_not_match_crontab_schedule_then_execution_is_not_due(self):
+        z = self.crontab(hour=7, minute=30)
+        last_run_at = datetime(2018, 6, 9, 10, 30) # not feasible wrt to current schedule. case can happen if schedule is modified after a run
+        now = datetime(2018, 6, 9, 23, 48)
+        expected_next_execution_time = datetime(2018, 6, 10, 7, 30)
+        expected_remaining = (expected_next_execution_time - now).total_seconds()
+        # check our assumptions
+        assert not is_time_feasible_wrt_crontab_schedule(last_run_at, z)
+        assert not is_time_feasible_wrt_crontab_schedule(now, z)
+        assert is_time_feasible_wrt_crontab_schedule(expected_next_execution_time, z)
+        assert now < expected_next_execution_time
+        assert expected_remaining == (7 * 60 + 30 + 12) * 60
+        # test is_due
+        with patch_crontab_nowfun(z, now):
+            due, remaining = z.is_due(last_run_at=last_run_at)
+            assert remaining == expected_remaining
+            assert not due
+
```

## Expected behavior
No matter what the value of `last_run_at` is, `crontab.is_due(last_run_at)` should never return `schedstate(True, rem)` when the current time `now` is not feasible with respect to the given crontab parameters.

## Actual behavior
if `last_run_at` value is older than the most recent time that is feasible with respect to the given crontab parameters then `crontab.is_due(last_run_at)` returns `schedstate(True, rem)`  when the current time is not feasible with respect to the crontab parameters.

## Comments
This behaviour is surprising as it is an undocumented departure from cron-like behaviour. This behaviour is somewhat like an undocumented variant of uncontrollable `anacron` behaviour. "Uncontrollable" in the sense that unlike `anacrontab`'s `START_HOURS_RANGE` parameter - there is no control at all over when tasks will be run when a scheduled execution is missed.

We experience this issue in production using the venerable Celery version of 3.1.19 : we have a celery beat process that is backed by a database using custom scheduler code that is derived from the django-celery-beat scheduler. Suppose we stop this celery beat process for some reason such as scheduled maintenance or during a deployment that needs to redeploy code to our celery cluster. When we later restart celery beat again then some or all `celery.schedules.crontab` scheduled tasks may immediately execute if there was a time during the celery beat downtime period that matches the crontab parameters.

One hack that can mitigate this behaviour is updating the "last_run_at" value for all celery crontab scheduled periodic tasks to the current time when celery beat starts, before celery beat makes any calls to the `celery.schedules.crontab.is_due` method.

abs25:
## Description

Fixes #4806

Previously, when using cron, a task would fire immediately if the last date the cron should have run is between the last actual run date and now. This is a major issue for periodic tasks in Django, because just saving the periodic task would cause it to execute. We ran into this issue when updating a cron time in one of our periodic tasks, inadvertently sending a few thousand emails out.

This fix is adding a new setting `beat_cron_starting_deadline_seconds`. This is the number of seconds it will look back to allow a cron job that could feasibly have been executed to still be due. If the feasible datetime is beyond the deadline, it will not execute immediately, instead waiting for the next feasible time. It is `None` by default, so behavior will be unchanged unless one specifies a number.

### DIFF
--- a/celery/app/defaults.py
+++ b/celery/app/defaults.py
@@ -78,6 +78,7 @@ NAMESPACES = Namespace(
         scheduler=Option('celery.beat:PersistentScheduler'),
         schedule_filename=Option('celerybeat-schedule'),
         sync_every=Option(0, type='int'),
+        cron_starting_deadline=Option(None, type=int)
     ),
     broker=Namespace(
         url=Option(None, type='string'),
--- a/docs/userguide/configuration.rst
+++ b/docs/userguide/configuration.rst
@@ -3495,3 +3495,14 @@ changes to the schedule into account.
 Also when running Celery beat embedded (:option:`-B <celery worker -B>`)
 on Jython as a thread the max interval is overridden and set to 1 so
 that it's possible to shut down in a timely manner.
+
+.. setting:: beat_cron_starting_deadline_seconds
+
+``beat_cron_starting_deadline_seconds``
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Default: None.
+
+When using cron, the number of seconds :mod:`~celery.bin.beat` can look back
+when deciding whether a cron schedule is due. When set to `None`, cronjobs that
+are past due will always run immediately.
--- a/docs/userguide/configuration.rst
+++ b/docs/userguide/configuration.rst
@@ -3496,10 +3496,12 @@ Also when running Celery beat embedded (:option:`-B <celery worker -B>`)
 on Jython as a thread the max interval is overridden and set to 1 so
 that it's possible to shut down in a timely manner.
 
-.. setting:: beat_cron_starting_deadline_seconds
+.. setting:: beat_cron_starting_deadline
 
-``beat_cron_starting_deadline_seconds``
-~~~~~~~~~~~~~~~~~~~~~~~~~~
+``beat_cron_starting_deadline``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. versionadded:: 5.3
 
 Default: None.
 
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -291,3 +291,4 @@ Tizian Seehaus, 2022/02/09
 Oleh Romanovskyi, 2022/06/09
 JoonHwan Kim, 2022/08/01
 Kaustav Banerjee, 2022/11/10
+Austin Snoeyink 2022/12/06
--- a/celery/schedules.py
+++ b/celery/schedules.py
@@ -36,7 +36,6 @@ CRON_REPR = """\
 {0._orig_day_of_week} (m/h/dM/MY/d)>\
 """
 
-
 SOLAR_INVALID_LATITUDE = """\
 Argument latitude {lat} is invalid, must be between -90 and 90.\
 """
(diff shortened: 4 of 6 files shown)


celery/app/defaults.py
```diff
@@ -78,6 +78,7 @@ def __repr__(self):
         scheduler=Option('celery.beat:PersistentScheduler'),
         schedule_filename=Option('celerybeat-schedule'),
         sync_every=Option(0, type='int'),
+        cron_starting_deadline_seconds=Option(None, type=int)
```
auvipy, line 81: do we have any better naming scheme for this?

auvipy (changes requested):

celery/app/defaults.py
abs25: Kubernetes has a similar option which follows this naming scheme, but I am open to other ideas.

https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/#cron-job-limitations

auvipy: what about cron_starting_deadline instead of seconds?

auvipy (changes requested):

celery/app/defaults.py
abs25: Sounds good to me. Updated.

auvipy:
seems the codcov action need some attention

docs/userguide/configuration.rst
```diff
@@ -3495,3 +3495,14 @@ changes to the schedule into account.
 Also when running Celery beat embedded (:option:`-B <celery worker -B>`)
 on Jython as a thread the max interval is overridden and set to 1 so
 that it's possible to shut down in a timely manner.
+
+.. setting:: beat_cron_starting_deadline_seconds
+
+``beat_cron_starting_deadline_seconds``
```
auvipy, line 3501: we should mention Versionadded 5.3 annotation here

auvipy (changes requested):

abs25:
Looks like the pypy-3.8 on ubuntu-latest action is still having issues? I am also unsure why the linter is failing, but it seems to be failing on upstream/master as well.

auvipy:
i re started that

auvipy:
Celery / Unit (pypy-3.9, ubuntu-latest) (pull_request)  passed so it should be a network issue

auvipy (approved):

celery/app/defaults.py
```diff
@@ -78,6 +78,7 @@ def __repr__(self):
         scheduler=Option('celery.beat:PersistentScheduler'),
         schedule_filename=Option('celerybeat-schedule'),
         sync_every=Option(0, type='int'),
+        cron_starting_deadline=Option(None, type=int)
```
nagareproject, line 81: Should be the `'int'` string instead of the `int`class

auvipy: Thanks for noticing

PR #7974, opened 2022-12-21
Pin redis==4.3.4

auvipy:
4.4 release introduces few breaking changes


Nusnus:
Do we also need to pin it somewhere in the CI? @auvipy 
And great work finding the correct version 💯

auvipy:
I already did that on kombu first and saw normal integration tests are passing except pypy [new version/compat issue it seems]

Nusnus (approved):

thedrow:
Please avoid pinning to a specific version. This is a bad practice.
Pinning to a specific version would force our users to upgrade even if they don't want to.
It is enough to specify `redis>=4.2.2,<4.4.0` for us to disallow installing a version with breaking changes.

auvipy:
https://github.com/celery/celery/pull/7980

PR #7074, opened 2021-11-10
Change pytz>dev to a PEP 440 compliant pytz>0.dev.0

frantisekz:
*Note*: Before submitting this pull request, please review our [contributing
guidelines](https://docs.celeryproject.org/en/master/contributing.html).

## Description

pytz>dev version specification doesn't seem to be PEP 440 compliant. 
pytz>0.dev.0 should achieve the same in the compliant way.

Ref. https://bugzilla.redhat.com/show_bug.cgi?id=2019954#c1 and PEP 440: https://www.python.org/dev/peps/pep-0440/#developmental-releases

PR #8002, opened 2023-01-11
pytz

auvipy:


requirements/dev.txt
```diff
@@ -1,4 +1,4 @@
-pytz>dev
+pytz(>dev)
```
auvipy, line 1:
```suggestion
pytz>2022.7
```

edmorley:
(Cross-linking a few issues, since the pip 24.1+ changes are going to break quite a few older apps)

This was fixed in Celery 5.2.1+ too - see:
https://github.com/celery/celery/pull/7074

PR #8023, opened 2023-01-23
Fix few typos, provide configuration + workflow for codespell to catch any new

yarikoptic:
Individual commit messages might have extra information/motivation for a specific change.

### DIFF
--- a/.codespellrc
+++ b/.codespellrc
@@ -1,0 +1,4 @@
+[codespell]
+skip = .git,.venv,*.svg,package-lock.json,*.key
+# Some names and timezone (lower cased)
+ignore-words-list = gool,markey,sherif,wil,ist,fromm,brane,bu,nott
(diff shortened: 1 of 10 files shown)


yarikoptic:
- pre-commit.ci seems to freak out because of a multiline extend-ignore in setup.cfg  which I dit not even change in this PR.
- sphinx -- seems also happen to me if I try to build locally on `main` as well, so also seems not related to this PR
 
odd but seems to be consistent with fails in the other fresh PR: https://github.com/celery/celery/pull/8022

auvipy (approved):

Nusnus:
> * pre-commit.ci seems to freak out because of a multiline extend-ignore in setup.cfg  which I dit not even change in this PR.

You were right friend, thanks for the hint!
#8044

PR #7995, opened 2023-01-06
sqlalchemy==1.4.46

auvipy:

PR #8033, opened 2023-01-29
Revert "sqlalchemy==1.4.46 (#7995)"

Nusnus:
The change `sqlalchemy==1.4.46` broke the unit tests in `main` so I'm reverting it until we find a better fix because we can't have main broken.

I'm merging this PR now as we cannot have `main` broken.
This is an issue for me because I found a bug in the `Task.replace` and I have a fix that I can't merge because it fails for issues caused from the reverted commit.

FYI @auvipy


auvipy:
ok

PR #8034, opened 2023-01-29
Fixed bug with replacing a stamped task with a chain or a group (inc. links/errlinks)

Nusnus:


auvipy (approved):

celery/app/task.py
```diff
@@ -959,13 +959,34 @@ def replace(self, sig):
...
             })
             sig.options['stamped_headers'] = stamped_headers
             sig.options.update(stamps)
 
+            links = sig.options['link'] if 'link' in sig.options else []
+            links.extend(sig.options['link_error'] if 'link_error' in sig.options else [])
+
+            if hasattr(sig, "tasks"):
+                tasks = sig.tasks
+                if isinstance(tasks, group):
```
thedrow, line 975: What happens if `tasks` is a list or a tuple?
What happens if it is a `regen`?

We shouldn't ignore such cases.

t/unit/tasks/test_canvas.py
```diff
@@ -456,6 +456,116 @@ def on_replace(self, sig):
             assert headers['on_errback'] is True
             assert headers['header'] == 'value'
 
+    @pytest.mark.parametrize('sig_to_replace', [
+        group(signature('sig1'), signature('sig2')),
```
thedrow, line 460:
```suggestion
        group(signature(f'sig{i}') for i in range(2)),
        group([signature('sig1'), signature('sig2')]),
        group((signature('sig1'), signature('sig2'))),
        group(signature('sig1'), signature('sig2')),
```

celery/app/task.py
```diff
@@ -959,13 +959,34 @@ def replace(self, sig):
...
             stamped_headers.extend(sig.options.get('stamped_headers', []))
+            stamped_headers = list(set(stamped_headers))
             stamps.update({
                 stamp: value
                 for stamp, value in sig.options.items() if stamp in sig.options.get('stamped_headers', [])
             })
             sig.options['stamped_headers'] = stamped_headers
             sig.options.update(stamps)
 
+            links = sig.options['link'] if 'link' in sig.options else []
```
thedrow, line 970: We should probably add comments here explaining what we're doing

thedrow (changes requested):

celery/app/task.py
Nusnus: Your new test cases should cover this. Let's see if the PR tests pass with the new edge cases.

Nusnus: But it should also work as the canvas primitive already handles the `tuple` and `regen` cases, but the added matrix for the test will cover this 100%

Nusnus: @thedrow Yes, the new test cases are passing with this code which confirms your original question is resolved.
> What happens if tasks is a list or a tuple?

It works.

> What happens if it is a regen?

It works.

> We shouldn't ignore such cases.

We do not :)

thedrow (approved):

PR #8044, opened 2023-02-01
Fixed formatting in setup.cfg that caused flake8 to misbehave

Nusnus:


Nusnus:
@auvipy FYI I’ve fixed the flake8 issues.
The linter now passes on `main` again 😊

Issue #7683, opened 2022-08-04
`celery multi start` raises when used with preload options


maconfr:
Preload options do not work with `celery multi start` and `celery report`.

This can easily reproduced by the following steps.

Setup a test project like this:

````bash
$ cd /tmp/
$ python3 -m venv celery_venv
$ . celery_venv/bin/activate
$ git clone https://github.com/celery/celery.git
Cloning into 'celery'...
[...]
$ cd celery
$ pip install -e .
[...]
$ celery report

software -> celery:5.3.0b1 (dawn-chorus) kombu:5.3.0b1 py:3.9.2
            billiard:4.0.2 py-amqp:5.1.1
platform -> system:Linux arch:64bit, ELF
            kernel version:5.10.0-16-amd64 imp:CPython
loader   -> celery.loaders.default.Loader
settings -> transport:amqp results:disabled

deprecated_settings: None

````

Then create a `proj.py`. You can simply use the [Preload options](https://docs.celeryproject.org/en/stable/userguide/extending.html#preload-options)
example.

Starting a worker with the `--template` option works with `celery worker`:

````bash
$ celery -A proj worker --template foo
[2022-08-04 12:57:19,836: WARNING/MainProcess] No hostname was supplied. Reverting to default 'localhost'
 
 -------------- celery@d6a4 v5.3.0b1 (dawn-chorus)
--- ***** ----- 
-- ******* ---- Linux-5.10.0-16-amd64-x86_64-with-glibc2.31 2022-08-04 12:57:19
- *** --- * --- 
- ** ---------- [config]
- ** ---------- .> app:         __main__:0x7fcbd6c5a1c0
- ** ---------- .> transport:   amqp://guest:**@localhost:5672//
- ** ---------- .> results:     disabled://
- *** --- * --- .> concurrency: 8 (prefork)
-- ******* ---- .> task events: OFF (enable -E to monitor tasks in this worker)
--- ***** ----- 
 -------------- [queues]
                .> celery           exchange=celery(direct) key=celery
[...]
````

But starting a worker with the `--template` option fails with `celery multi start`:

````bash
$ celery -A proj multi start worker1 --template foo
Traceback (most recent call last):
  File "/tmp/celery_venv/bin/celery", line 33, in <module>
    sys.exit(load_entry_point('celery', 'console_scripts', 'celery')())
  File "/tmp/celery/celery/__main__.py", line 15, in main
    sys.exit(_main())
[...]

PR #8049, opened 2023-02-02
added missing kwargs arguments in some cli cmd

auvipy:
based on comment https://github.com/celery/celery/issues/7683#issuecomment-1372179683

### DIFF
--- a/celery/bin/celery.py
+++ b/celery/bin/celery.py
@@ -173,7 +173,7 @@ def celery(ctx, app, broker, result_backend, loader, config, workdir,
 
 @celery.command(cls=CeleryCommand)
 @click.pass_context
-def report(ctx):
+def report(ctx, **kwargs):
     """Shows information useful to include in bug-reports."""
     app = ctx.obj.app
     app.loader.import_default_modules()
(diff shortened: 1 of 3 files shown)


Nusnus:
Hey @auvipy , any updates on this PR?

auvipy:
I need manual verification from you or Omer

Nusnus:
> I need manual verification from you or Omer

Ok so I've been reading a bit about this change, and about the original discussion, and I'm not sure this will fix the issue.
Adding the `**kwargs` will only absorb those errors but if the actual `kwargs` isn't being used as well, then for example where they used `--template`, it will not be honored as we don't do anything else in this change beside accepting the new arguments.

The original use-case needs to be tested (manually would be fine too) before we can merge this change, in case I am missing something and the `kwargs` does get used somehow or that everything works if we just "ignore" this exception by absorbing the extra parameters into `kwargs`.

Do you have a way to validate this change indeed fixes the issue? @auvipy

Nusnus:
@thedrow Hey can you give it a quick look please?
You might be able to judge quickly if we can push/block for testing (as it might not be required).

Thanks!

auvipy (commented):
merging as temporary solution and it is used as convention in other parts of CLI

Issue #8052, opened 2023-02-04
schedules.crontab abnormally stop tasks for some timezone


pkyosx:
#### Related Issues

- [django-celery-beat](https://github.com/celery/django-celery-beat/issues/629)

## Environment & Settings

**Celery version**: 5.2.7 (dawn-chorus)

## Required Dependencies

- **Minimal Python Version**: 3.10
- **Minimal Celery Version**: 5.2.7
- **Minimal Kombu Version**: 5.2.4
- **Minimal Broker Version**: N/A or Unknown
- **Minimal Result Backend Version**: N/A or Unknown
- **Minimal OS and/or Kernel Version**: N/A or Unknown
- **Minimal Broker Client Version**: N/A or Unknown
- **Minimal Result Backend Client Version**: N/A or Unknown

### Python Packages

<details>
<summary>
<b><code>pip freeze</code> Output:</b>
<b><code>amqp==5.1.1</code></b>
<b><code>billiard==3.6.4.0</code></b>
<b><code>celery==5.2.7</code></b>
<b><code>click==8.1.3</code></b>
<b><code>click-didyoumean==0.3.0</code></b>
<b><code>click-plugins==1.1.1</code></b>
<b><code>click-repl==0.2.0</code></b>
<b><code>kombu==5.2.4</code></b>
<b><code>prompt-toolkit==3.0.36</code></b>
<b><code>pytz==2022.7.1</code></b>
<b><code>six==1.16.0</code></b>
<b><code>vine==5.0.0</code></b>
<b><code>wcwidth==0.2.6</code></b>
</summary>
<p>

```
```

</p>
</details>

## Minimally Reproducible Test Case

<details>
<p>

```python
import pytz
import datetime
from celery import schedules

tz = pytz.timezone('America/Los_Angeles')

def normalized_time(year, month, day, hour, min, sec):
    return tz.normalize(
            pytz.utc.localize(datetime.datetime(year, month, day, hour, min, sec))
        )


def run(last_run_at, now):
    class Foo(schedules.crontab):
        def __init__(self, *args, **kwargs):
            self.tz = tz
            super().__init__(*args, **kwargs)

        def now(self):
            return now

    foo = Foo(minute='0', hour='*', day_of_week='*', day_of_month='28-31', month_of_year='*')
    return foo.is_due(last_run_at=last_run_at)


for hour in range(0, 23):
    last_run_at = normalized_time(2023, 1, 29, hour, 0, 0)
[...]

PR #8053, opened 2023-02-05
Fix cronjob that use day of month and negative UTC timezone

pkyosx:
Before this patch, the cross day schedule jumps to the future and some tasks are skipped

## Description
Fix https://github.com/celery/celery/issues/8052

When we specify day_of_month of cronjob to non-wildcard and choose a negative UTC timezone such as "America/Los_Angeles", the roll_over function mistakenly skip valid candidates. 

The roll_over function use is_before_last_run function which does not properly take timezone, next_hour and next_minute into account.

### DIFF
--- a/celery/schedules.py
+++ b/celery/schedules.py
@@ -72,8 +72,8 @@ class BaseSchedule:
     def is_due(self, last_run_at):
         raise NotImplementedError()
 
-    def maybe_make_aware(self, dt):
-        return maybe_make_aware(dt, self.tz)
+    def maybe_make_aware(self, dt, naive_as_utc=True):
+        return maybe_make_aware(dt, self.tz, naive_as_utc=naive_as_utc)
 
     @property
     def app(self):
@@ -468,9 +468,8 @@ class crontab(BaseSchedule):
             return False
 
         def is_before_last_run(year, month, day):
-            return self.maybe_make_aware(datetime(year,
-                                                  month,
-                                                  day)) < last_run_at
+            return self.maybe_make_aware(datetime(year, month, day, next_hour, next_minute),
+                                         naive_as_utc=False) < last_run_at
 
         def roll_over():
             for _ in range(2000):
--- a/celery/utils/time.py
+++ b/celery/utils/time.py
@@ -305,10 +305,11 @@ def to_utc(dt):
     return make_aware(dt, timezone.utc)
 
 
-def maybe_make_aware(dt, tz=None):
+def maybe_make_aware(dt, tz=None, naive_as_utc=True):
     """Convert dt to aware datetime, do nothing if dt is already aware."""
     if is_naive(dt):
-        dt = to_utc(dt)
+        if naive_as_utc:
+            dt = to_utc(dt)
         return localize(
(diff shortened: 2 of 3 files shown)


celery/schedules.py
```diff
@@ -468,9 +468,9 @@ def day_out_of_range(year, month, day):
             return False
 
         def is_before_last_run(year, month, day):
-            return self.maybe_make_aware(datetime(year,
```
auvipy, line 471: can we make sure that maybe_make_aware removal won't create any other consequences?

pkyosx: is_before_last_run is introduced to prevent some issue that test_monthly_moy_execution_is_not_due2 is protecting. 

There are two issue with current implementation.
1. next_hour and next_minute should be taken into account. 
2. maybe_make_aware is not the correct time convertion function in this case.

The purpose of roll_over is to move forward candidate to a valid day of month and month of year. is_before_last_run should protect _delta_to_next from returning a time earlier than last_run_at. 

From the following return value, we can see that is_before_last_run should make sure 'year/month/day next_hour:next_minute:second microsecond' should not before last_run_at.
```
           return ffwd(year=datedata.year,
                    month=months_of_year[datedata.moy],
                    day=days_of_month[datedata.dom],
                    hour=next_hour,
                    minute=next_minute,
                    second=0,
                    microsecond=0)
```

However, there is a problem when we use maybe_make_aware(datetime(year, month, day)). It treat the year, month and day in UTC and then convert to actual tz. 

In our case, the candidate '2023/1/29 00:00:00' is interpreted as '2023/1/29T00:00:00Z' and converted to '2023/1/28T16:00:00-08:00' which is wrong.
```
def maybe_make_aware(dt, tz=None):
    """Convert dt to aware datetime, do nothing if dt is already aware."""
    if is_naive(dt):
        dt = to_utc(dt)
        return localize(
            dt, timezone.utc if tz is None else timezone.tz_or_local(tz),
        )
    return dt
```

There are two possible ways to mitigate the issue. First solution is that we can create another function maybe_make_aware_ex to replace self.maybe_make_aware. 

```
def maybe_make_aware_ex(dt, tz=None):
    if is_naive(dt):
        return localize(
            dt, timezone.utc if tz is None else timezone.tz_or_local(tz),
        )
    return dt
    
 def is_before_last_run(year, month, day):
            return self.maybe_make_aware_ex(datetime(year, month, day, next_hour, next_minute)) < last_run_at
```

However, people might be confused by maybe_make_aware_ex and maybe_make_aware. 
Another way to do it is simply treat year/month/day/next_hour/next_minutes as if they are numbers in the same timezone as last_run_at which is what I do here.

pkyosx:
The patch coverage diff is 100%, not sure why coverage dropped?
Also, the coverage report uploading seems broken for ubuntu?

auvipy:
lets see after another round of CI

pkyosx:
pypy seems to be fail to upload coverage report, others are good now.

auvipy:
i restarted both to trigger integration tests

auvipy (approved):

PR #8062, opened 2023-02-09
Added new canvas integration test for chains and groups

jberclaz:
This test is built to reproduce the issue mentioned on the GitHub discussion: https://github.com/celery/celery/discussions/8027

In chains starting with a group, the AsyncResult does not contain a full list of tasks. Instead, the group task is linked directly to the last task, and the other tasks are short-circuited.

Note that this test will fail with the current version of Celery.

### DIFF
--- a/t/integration/test_canvas.py
+++ b/t/integration/test_canvas.py
@@ -1031,6 +1031,35 @@ class test_chain:
         # Cleanup
         redis_connection.delete(redis_key, 'Done')
 
+    def test_chain_starting_with_group(self, manager, subtests):
+        """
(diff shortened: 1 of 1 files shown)


d-sasi (approved):

auvipy:
@d-sasi this is a failing test only, we still need the fix

Nusnus:
Thank you for the PR @jberclaz 
It's very useful to report a bug with a failing test 👏 

I'm currently fully invested in Celery 5.3 release blockers, but I will try to spare some time soon to investigate this.

Nusnus:
Fixed in `main`: #7919

PR #8064, opened 2023-02-13
Stamping Mechanism Refactoring 2

Nusnus:


t/unit/tasks/test_stamping.py
```diff
@@ -1227,81 +1221,3 @@ def test_chord_stamping_one_level(self, subtests):
 
         with subtests.test("sig_2_res has stamped_headers", stamped_headers=["stamp"]):
             assert sorted(sig_2_res._get_task_meta()["stamped_headers"]) == sorted(["stamp"])
-
-    @pytest.mark.usefixtures("depends_on_current_app")
```
auvipy, line 1231: don't we need these features anymore?

Nusnus: No - I implemented it based on a wrong understanding of what is `Task.replace()` thus adding a behavior that does not make sense.
So I'm removing some of the recently developed code I added (even though it works, it's not needed because its not correct that the replace will stamp the replaced task implicitly).

examples/stamping/myapp.py
```diff
@@ -16,14 +16,15 @@
    # The shell service is used to run the example.
     (window2)$ celery -A myapp shell
 
-   # Use (copy) the content of shell.py to run the workflow via the
+   # Use (copy) the content of the examples modules to run the workflow via the
    # shell service.
 
-   # Use one of two demo runs via the shell service:
+   # Use one of demo runs via the shell service:
```
auvipy, line 22:
```suggestion
   # Use one of the 3 demo runs via the shell service:
```

auvipy (changes requested):

celery/canvas.py
```diff
@@ -550,86 +550,97 @@ def set(self, immutable=None, **options):
...
         # to avoid implicitly casting the duplicated stamp into a list of both stamps from headers and
         # visitor_headers of the same key.
         # Example:
         #   headers = {"foo": "bar1"}
         #   visitor_headers = {"foo": "bar2"}
-        #   _merge_dictionaries(headers, visitor_headers, aggregate_duplicates=True)
+        #   _merge_dictionaries(headers, visitor_headers, append_stamps=True)
         #   headers["foo"] == ["bar1", "bar2"] -> The stamp is now a list
-        #   _merge_dictionaries(headers, visitor_headers, aggregate_duplicates=False)
+        #   _merge_dictionaries(headers, visitor_headers, append_stamps=False)
```
thedrow, line 581: You haven't changed the `_merge_dictionaries` signature to match this comment.
The argument name is still `aggregate_duplicates`, see this [line](https://github.com/celery/celery/pull/8064/files#diff-3a80ff45da16a11b96e26a63973d7d490187a68ddc1949e2dfd7fd090b208841R596) for example.

celery/canvas.py
```diff
@@ -550,86 +550,97 @@ def set(self, immutable=None, **options):
...
-        stamped_headers.update(self.options.get("stamped_headers", []))
-        headers["stamped_headers"] = list(stamped_headers)
-        for previous_header in stamped_headers:
-            if previous_header not in headers and previous_header in self.options:
-                headers[previous_header] = self.options[previous_header]
+        if self_headers:
+            stamped_headers = set(headers.get("stamped_headers", []))
+            stamped_headers.update(self.options.get("stamped_headers", []))
+            headers["stamped_headers"] = list(stamped_headers)
+            reducted_options = {k: v for k, v in self.options.items() if k in headers["stamped_headers"]}
```
thedrow, line 604: Did you mean `redacted_options`?

celery/canvas.py
```diff
@@ -550,86 +550,97 @@ def set(self, immutable=None, **options):
...
-            _merge_dictionaries(headers, self.options, aggregate_duplicates=aggregate_duplicates)
-
-        # Sync missing stamps from self.options (relevant for stamping during task replacement)
-        stamped_headers = set(headers.get("stamped_headers", []))
-        stamped_headers.update(self.options.get("stamped_headers", []))
-        headers["stamped_headers"] = list(stamped_headers)
-        for previous_header in stamped_headers:
-            if previous_header not in headers and previous_header in self.options:
-                headers[previous_header] = self.options[previous_header]
+        if self_headers:
```
thedrow, line 600: Why is merging from `self.options` now optional? This wasn't the case before.

celery/canvas.py
thedrow: Also, why are they redacted? What do you mean by that?

celery/canvas.py
```diff
@@ -641,8 +652,13 @@ def stamp_links(self, visitor, **headers):
             visitor_headers = None
             if visitor is not None:
                 visitor_headers = visitor.on_callback(link, **headers) or {}
-            headers = self._stamp_headers(visitor_headers, **headers)
-            link.stamp(visitor=visitor, **headers)
+            headers = self._stamp_headers(
+                visitor_headers=visitor_headers,
+                append_stamps=append_stamps,
+                self_headers=False,  # Don't stamp links with own stamps implicitly
```
thedrow, line 658: The comment doesn't explain why that's the case

thedrow (changes requested):

examples/stamping/myapp.py
Nusnus: The `examples.py` has 3 demos, the revoke example has 2 demos, so I don't want to commit to a number in the docs and just say "one of the demo runs"

celery/canvas.py
Nusnus: That's because `_merge_dictionaries` uses `aggregate_duplicates` and this uses `append_stamps`, which is sent to `aggregate_duplicates` of `_merge_dictionaries`.

`append_stamps` - logical meaning in the stamping context.
`aggregate_duplicates` - logical meaning when merging dicts.
They are matching when merging stamps dicts.

celery/canvas.py
Nusnus: > Did you mean `redacted_options`?

Late night typos :)
Will fix.

Nusnus: > Also, why are they redacted? What do you mean by that?

`{k: v for k, v in self.options.items() if k in headers["stamped_headers"]}`
Meaning `self.options` fields which are stamps.

celery/canvas.py
Nusnus: It is not optional, the `self_headers` isn't an external param but an internal flag to use.
The behavior has not changed in regard to how the stamping works, just the code readability and organization in this case.

celery/canvas.py
thedrow: You're not following me. This comment contains code that will not work when copy/pasted.

Nusnus: Apologies, I understand now 🙏 
Fix pushed.

celery/canvas.py
Nusnus: I've added a better explanation - fix pushed.

celery/canvas.py
Nusnus: Added a comment to clarify + fixed typo - fix pushed

Nusnus:
GitHub is having issues with running the CI.
I've canceled the run and will reinitiate it in a few minutes to let it "rest a bit".

thedrow (approved):

PR #8071, opened 2023-02-16
Fix command for starting flower with specified broker URL

ShukantPal:
The documentation specifies the command to start flower as

```
celery flower --broker="redis://localhost:6379/0"
```

Flower now gives a warning and ignores the `--broker` argument.

```
You have incorrectly specified the following celery arguments after flower command: ['--broker']. Please specify them after celery command instead following this template: celery [celery args] flower [flower args].
```

This is because `--broker` is an argument for Celery not Flower.

## Description

This PR fixes the aforementioned issue in the documentation by moving the `--broker` argument to the correct place.

### DIFF
--- a/docs/userguide/monitoring.rst
+++ b/docs/userguide/monitoring.rst
@@ -299,9 +299,9 @@ Broker URL can also be passed through the
 
 .. code-block:: console
 
-    $ celery flower --broker=amqp://guest:guest@localhost:5672//
+    $ celery --broker=amqp://guest:guest@localhost:5672// flower
     or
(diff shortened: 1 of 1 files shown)


auvipy (approved):

auvipy:
thanks

ErnestinaQiu:
so,  what is the correct command in win10? I use `celery -A tasks worker -l debug -c 4 -P eventlet  flower --address --port=5555  > ./log.log  2>&1& `  but it raises error as `Usage: celery worker [OPTIONS]
Try 'celery worker --help' for help.

Error: No such option: --address
`

PR #8108, opened 2023-03-05
Type annotations for `celery/apps/beat.py`

max-muoto:

## Description

Type annotations for `celery/apps/beat.py` as a part of https://github.com/celery/celery/issues/7394.

### DIFF
--- a/celery/apps/beat.py
+++ b/celery/apps/beat.py
@@ -16,7 +16,7 @@ from signal import Signals
 from types import FrameType
 from typing import Any
 
-from celery import VERSION_BANNER, Celery, platforms, beat
+from celery import VERSION_BANNER, Celery, beat, platforms
 from celery.beat import Scheduler
 from celery.utils.imports import qualname
 from celery.utils.log import LOG_LEVELS, get_logger
--- a/celery/apps/beat.py
+++ b/celery/apps/beat.py
@@ -80,7 +80,7 @@ class Beat:
     def app(self) -> Celery:
         if self._app:
             return self._app
-        raise ValueError("App not provided.")
+        raise ValueError("Beat must provided a Celery app.")
 
     @app.setter
     def app(self, app: Celery) -> None:
--- a/celery/apps/beat.py
+++ b/celery/apps/beat.py
@@ -17,7 +17,7 @@ from types import FrameType
 from typing import Any
 
 from celery import VERSION_BANNER, Celery, beat, platforms
-from celery.beat import Service
+from celery.beat import Scheduler, Service
 from celery.utils.imports import qualname
 from celery.utils.log import LOG_LEVELS, get_logger
 from celery.utils.time import humanize_seconds
@@ -46,8 +46,8 @@ class Beat:
     def __init__(self, max_interval: int | None = None, app: Celery | None = None,
                  socket_timeout: int = 30, pidfile: str | None = None, no_color: bool | None = None,
                  loglevel: str = 'WARN', logfile: str | None = None, schedule: str | None = None,
-                 scheduler: beat.Scheduler | None = None,
-                 scheduler_cls: beat.Scheduler | None = None,  # XXX use scheduler
+                 scheduler: Scheduler | None = None,
+                 scheduler_cls: Scheduler | None = None,  # XXX use scheduler
                  redirect_stdouts: bool | None = None,
                  redirect_stdouts_level: str | None = None,
                  quiet: bool = False, **kwargs: Any) -> None:
--- a/celery/apps/beat.py
+++ b/celery/apps/beat.py
@@ -16,7 +16,7 @@ from signal import Signals
 from types import FrameType
 from typing import Any
 
-from celery import VERSION_BANNER, Celery, beat, platforms
+from celery import VERSION_BANNER, Celery, platforms
 from celery.beat import Scheduler, Service
 from celery.utils.imports import qualname
 from celery.utils.log import LOG_LEVELS, get_logger
--- a/celery/apps/beat.py
+++ b/celery/apps/beat.py
@@ -41,7 +41,7 @@ logger = get_logger('celery.beat')
 class Beat:
     """Beat as a service."""
 
-    app: Celery | None = None
+    _app: Celery | None = None
 
     def __init__(self, max_interval: int | None = None, app: Celery | None = None,
                  socket_timeout: int = 30, pidfile: str | None = None, no_color: bool | None = None,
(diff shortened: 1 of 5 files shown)


celery/apps/beat.py
```diff
@@ -142,15 +147,15 @@ def startup_info(self, service):
...
         arg_start = 'manage' in sys.argv[0] and 2 or 1
         platforms.set_process_title(
             'celery beat', info=' '.join(sys.argv[arg_start:]),
         )
 
-    def install_sync_handler(self, service):
+    def install_sync_handler(self, service: Service):
         """Install a `SIGTERM` + `SIGINT` handler saving the schedule."""
-        def _sync(signum, frame):
+        def _sync(signum: Signals, frame: FrameType):
```
max-muoto, line 158: For some reason the pre-commiter prevents me from quoting type hints. Would prefer to do that in order to conditionally import using `if TYPE_CHECKING`. Is this intended?

celery/apps/beat.py
```diff
@@ -71,25 +76,35 @@ def __init__(self, max_interval=None, app=None,
         if not isinstance(self.loglevel, numbers.Integral):
             self.loglevel = LOG_LEVELS[self.loglevel.upper()]
 
-    def run(self):
+    @property
```
max-muoto, line 79: Using this pattern to accommodate the required type safety from MyPy (obviously this doesn't really provide type safety). Would you prefer that I refactor the class to ensure actual type safety?

celery/apps/beat.py
```diff
@@ -142,15 +157,15 @@ def startup_info(self, service):
...
 
-    def set_process_title(self):
+    def set_process_title(self) -> None:
         arg_start = 'manage' in sys.argv[0] and 2 or 1
         platforms.set_process_title(
             'celery beat', info=' '.join(sys.argv[arg_start:]),
         )
 
-    def install_sync_handler(self, service):
+    def install_sync_handler(self, service: Service) -> None:
```
max-muoto, line 166: Pre-commit was replacing quoted type hints. Is this intended? I think it would be useful to allow them, so that I could conditionally import types with `if TYPE_CHECKING`.

Nusnus:
The CI was stuck - on it

max-muoto:
> The CI was stuck - on it

Hmm, not able to replicate locally. Do you have any debugging advice? Never really seen a CI just get stuck.

Nusnus:
> > The CI was stuck - on it
> 
> Hmm, not able to replicate locally. Do you have any debugging advice? Never really seen a CI just get stuck.

Tbh "stuck" isn't accurate, it was pending/waiting due to a time limit if I'm not mistaken.

I've stopped the run and will re-run soon. It should allow everything to run smoothly.

max-muoto:
I see, thanks!

Nusnus:
Thank You friend!
I personally like annotations so it's a great PR IMHO.

And don't worry, I got my eyes on the CI. If something happens I'll comment/fix etc.

celery/apps/beat.py
Nusnus: This appears to be out of scope so you can leave it like that for now.
Maybe in a different PR if you wish.

Nice idea though!

celery/apps/beat.py
```diff
@@ -71,22 +77,32 @@ def __init__(self, max_interval=None, app=None,
         if not isinstance(self.loglevel, numbers.Integral):
             self.loglevel = LOG_LEVELS[self.loglevel.upper()]
 
-    def run(self):
+    @property
+    def app(self) -> Celery:
+        if self._app:
+            return self._app
+        raise ValueError("Beat must provided a Celery app.")
```
Nusnus, line 84: > Added lines #L83 - L84 were not covered by tests

Can you add a unit test for this please?

celery/apps/beat.py
```diff
@@ -71,22 +77,32 @@ def __init__(self, max_interval=None, app=None,
...
+
+    @app.setter
+    def app(self, app: Celery) -> None:
+        self._app = app
```
Nusnus, line 88: > Added line #L88 was not covered by tests

Can you add a unit test for this please?

Nusnus (approved):

celery/apps/beat.py
max-muoto: Sounds good!

celery/apps/beat.py
max-muoto: Will do for both!

max-muoto: Fixed a type hint (`scheduler`/`scheduler_cls`), and added a test that covers both.

Nusnus: 👍

max-muoto: I seem to have some trouble running unit tests locally, pytest must be caching (or something else, not sure) as the new field on the `Beat` class isn't even picked up when inspecting with `pdb`. Do you have any ideas here?

Nusnus:
https://github.com/celery/celery/actions/runs/4338258344/jobs/7574969608#step:8:32984

CleanShot 2023-03-06 at 00 58 58@2x

max-muoto:
@Nusnus Modified it a way that doesn't require a setter/getter so removed the test, if you want to take another look.

It turns out there was no need to explicitly set the `app` field in the class to None, since it gets overridden anyway when grabbing it from the app.

celery/apps/beat.py
```diff
@@ -36,16 +41,16 @@ class Beat:
...
-
-    def __init__(self, max_interval=None, app=None,
-                 socket_timeout=30, pidfile=None, no_color=None,
-                 loglevel='WARN', logfile=None, schedule=None,
-                 scheduler=None,
-                 scheduler_cls=None,  # XXX use scheduler
-                 redirect_stdouts=None,
-                 redirect_stdouts_level=None,
-                 quiet=False, **kwargs):
+    app: Celery
```
auvipy, line 44:
```suggestion
    app: Celery = None
```

auvipy: I need to know, why should be remove app = None while trying to add type hints

auvipy (changes requested):

celery/apps/beat.py
max-muoto: The reason why we need to remove `None` is to make the methods in the class type safe for mypy. As if it can be `None` here, and it can be `None` in the `init` method, then you can't call a single method on app without first checking if exists

max-muoto: App gets overridden anyway when grabbing it from the app. If it isn't set, and no app was provided for the init, then it doesn't make a difference since it would have run into an attribute error trying to call a method on app anyway in the `init` method anway.

Nusnus (approved):
@auvipy LGTM

auvipy (approved):

PR #8120, opened 2023-03-11
Fixed bug where retrying a task loses its stamps

Nusnus:


celery/app/task.py
```diff
@@ -145,6 +145,11 @@ def as_execution_options(self):
             'replaced_task_nesting': self.replaced_task_nesting,
             'origin': self.origin,
         }
+        if hasattr(self, 'stamps') and hasattr(self, 'stamped_headers'):
```
auvipy, line 148: after the integration tests, can some unit tests added here?

auvipy (changes requested):

celery/app/task.py
Nusnus: Done - unit test pushed.
Thanks for reviewing quickly! ❤️

auvipy (approved):
I am OK with attribute checks for stamps and related tests. if the test coverage can be improved or any other better suggestion from Omer I will leave to his reviews

thedrow (approved):

Issue #6185, opened 2020-06-23
unpickleable exceptions cause queue to crash


caleb15:
# Checklist

* [x] I have verified that the issue exists against the `master` branch of Celery. (not against latest commit specifically but against latest version 4.4.5)
* [x] This has already been asked to the [discussion group](https://groups.google.com/forum/#!forum/celery-users) first.
* [x] I have read the relevant section in the
  [contribution guide](http://docs.celeryproject.org/en/latest/contributing.html#other-bugs)
  on reporting bugs.
* [x] I have checked the [issues list](https://github.com/celery/celery/issues?q=is%3Aissue+label%3A%22Issue+Type%3A+Bug+Report%22+-label%3A%22Category%3A+Documentation%22)
  for similar or identical bug reports.
* [x] I have checked the [pull requests list](https://github.com/celery/celery/pulls?q=is%3Apr+label%3A%22PR+Type%3A+Bugfix%22+-label%3A%22Category%3A+Documentation%22)
  for existing proposed fixes.
* [x] I have checked the [commit log](https://github.com/celery/celery/commits/master)
  to find out if the bug was already fixed in the master branch.
* [x] I have included all related issues and possible duplicate issues
  in this issue (If there are none, check this box anyway).

## Mandatory Debugging Information
* [x] I have included the output of ``celery -A proj report`` in the issue.
    (if you are not able to do this, then at least specify the Celery
     version affected). 
* [x] I have verified that the issue exists against the `master` branch of Celery.
* [x] I have included the contents of ``pip freeze`` in the issue.
* [x] I have included all the versions of all the external dependencies required
  to reproduce this bug.

## Optional Debugging Information

* [x] I have tried reproducing the issue on more than one Python version
  and/or implementation.
* [ ] I have tried reproducing the issue on more than one message broker and/or
  result backend.
* [ ] I have tried reproducing the issue on more than one version of the message
  broker and/or result backend.
* [ ] I have tried reproducing the issue on more than one operating system.
* [ ] I have tried reproducing the issue on more than one workers pool.
* [ ] I have tried reproducing the issue with autoscaling, retries,
  ETA/Countdown & rate limits disabled.
* [x] I have tried reproducing the issue after downgrading
  and/or upgrading Celery and its dependencies.

#### Related Issues
* None

#### Possible Duplicates
* None

## Environment & Settings

**Celery version**: 4.4.5 (cliffs)

Celery Report:
https://pastebin.com/0x13Xjiq

# Steps to Reproduce
[...]

Issue #6990, opened 2021-10-08
Worker crash when providing un-un-pickleable Exception to retry arg


amcgavin:
# Checklist

* [x] I have verified that the issue exists against the `master` branch of Celery.
* [ ] This has already been asked to the [discussion group](https://groups.google.com/forum/#!forum/celery-users) first.
* [x] I have read the relevant section in the
  [contribution guide](http://docs.celeryproject.org/en/latest/contributing.html#other-bugs)
  on reporting bugs.
* [x] I have checked the [issues list](https://github.com/celery/celery/issues?q=is%3Aissue+label%3A%22Issue+Type%3A+Bug+Report%22+-label%3A%22Category%3A+Documentation%22)
  for similar or identical bug reports.
* [x] I have checked the [pull requests list](https://github.com/celery/celery/pulls?q=is%3Apr+label%3A%22PR+Type%3A+Bugfix%22+-label%3A%22Category%3A+Documentation%22)
  for existing proposed fixes.
* [x] I have checked the [commit log](https://github.com/celery/celery/commits/master)
  to find out if the bug was already fixed in the master branch.
* [x] I have included all related issues and possible duplicate issues
  in this issue (If there are none, check this box anyway).

## Mandatory Debugging Information
* [x] I have included the output of ``celery -A proj report`` in the issue.
    (if you are not able to do this, then at least specify the Celery
     version affected).
* [x] I have verified that the issue exists against the `master` branch of Celery.
* [x] I have included the contents of ``pip freeze`` in the issue.
* [x] I have included all the versions of all the external dependencies required
  to reproduce this bug.

## Optional Debugging Information

* [ ] I have tried reproducing the issue on more than one Python version
  and/or implementation.
* [ ] I have tried reproducing the issue on more than one message broker and/or
  result backend.
* [ ] I have tried reproducing the issue on more than one version of the message
  broker and/or result backend.
* [ ] I have tried reproducing the issue on more than one operating system.
* [ ] I have tried reproducing the issue on more than one workers pool.
* [ ] I have tried reproducing the issue with autoscaling, retries,
  ETA/Countdown & rate limits disabled.
* [x] I have tried reproducing the issue after downgrading
  and/or upgrading Celery and its dependencies.

#### Related Issues
* None

#### Possible Duplicates
* None

## Environment & Settings

**Celery version**:

<details>
<summary><b><code>celery report</code> Output:</b>
</summary>
<p>

```

[...]

PR #8132, opened 2023-03-16
Fix worker crash on un-pickleable exceptions

alessio-b2c2:
## Description

Fix celery worker crash when a `retry` is issued with an unpickleable exception.

Fixes https://github.com/celery/celery/issues/6185 and https://github.com/celery/celery/issues/6990.


auvipy:
> @alessio-b2c2 why was this closed ?

another PR was contributed in the main branch which is based on this

Issue #8078, opened 2023-02-20
Deprecate pytz and use zoneinfo


wongcht:
# Checklist

- [x] I have checked the [issues list](https://github.com/celery/celery/issues?q=is%3Aissue+label%3A%22Issue+Type%3A+Enhancement%22+-label%3A%22Category%3A+Documentation%22)
  for similar or identical enhancement to an existing feature.
- [x] I have checked the [pull requests list](https://github.com/celery/celery/pulls?q=is%3Apr+label%3A%22Issue+Type%3A+Enhancement%22+-label%3A%22Category%3A+Documentation%22)
  for existing proposed enhancements.
- [x] I have checked the [commit log](https://github.com/celery/celery/commits/main)
  to find out if the if the same enhancement was already implemented in the
  main branch.
- [x] I have included all related issues and possible duplicate issues in this issue
      (If there are none, check this box anyway).

#### Related Issues

- None

#### Possible Duplicates

- None

# Brief Summary

Considering `pytz` is deprecating in Django and will be removed in 5.0, suggest follow approaches in Django
- use `zoneinfo` to replace `pytz`
- backport by `backports.zoneinfo`

Ref: https://docs.djangoproject.com/en/4.1/releases/4.0/

# Design

## Architectural Considerations

None

## Proposed Behavior

No changes

## Proposed UI/UX

No changes

## Diagrams

N/A

## Alternatives

None

wongcht:
I'm happy to help develop, thanks!

auvipy:
ping me with your draft PR please when it is in progress

PR #8159, opened 2023-03-27
Deprecate pytz and use zoneinfo

max-muoto:

## Description

Fixes https://github.com/celery/celery/issues/8078. Fully remove pytz and replace it with ZoneInfo. Add backports for ZoneInfo to get support for versions of Python earlier than 3.9.


max-muoto:
Looks like I need to add some additional test coverage, but wanted to throw this out here  @auvipy @Nusnus.

max-muoto:
@auvipy Fixed issues on Windows.

auvipy:
good job. I will look into this tomorrow

t/unit/utils/test_time.py
```diff
@@ -278,7 +271,7 @@ def test_localize_aware_dt_idempotent(self):
         assert localized_utc_offset == local_zone.utcoffset(local_time)
 
 
-@pytest.mark.parametrize('s,expected', [
+@ pytest.mark.parametrize('s,expected', [
```
Nusnus, line 274: There are a few spaces that sneaked into this file :)
Check below as well

Nusnus:
> good job. I will look into this tomorrow

So far LGTM2 :)

Nusnus:
> good job. I will look into this tomorrow

Indeed good job @max-muoto !
LGTM so far (check out the comment I left)

t/unit/utils/test_time.py
max-muoto: Fixed!

requirements/default.txt
```diff
@@ -7,3 +6,5 @@ click-didyoumean>=0.3.0
 click-repl>=0.2.0
 click-plugins>=1.1.1
 importlib-metadata>=3.6; python_version < '3.8'
+backports.zoneinfo>=0.2.1; python_version < '3.9'
+tzdata>=2022.7
```
max-muoto, line 10: Needed for Windows, as it doesn't have system time zone data that zoneinfo can use by default.

Details: https://tzdata.readthedocs.io/en/latest/

auvipy: also for the deprecation period you can keep this.

max-muoto:
> do we need any deprecation message or additional docs change for this? I'm bit more careful before merging

We shouldn't need any deprecration warnings/messages, as I simply porting over the functionality over our existing functions to use `zoneinfo` and/or built-in Python date utils that mimic the previous functionality. I did move over `parse_iso8601` to stop use `zoneinfo`, but this function already had a deprecation warning in favor of `fromisoformat`.

 More than happy to improve the documentation around zoneinfo and how the date utils use it.

auvipy:
> More than happy to improve the documentation around zoneinfo and how the date utils use it - if that works for you.

that would be really great. you have already done pretty good job

max-muoto:
> > More than happy to improve the documentation around zoneinfo and how the date utils use it - if that works for you.
> 
> that would be really great. you have already done pretty good job

Sounds good, will aim to get that done later today.

requirements/default.txt
max-muoto: Can definitely add it back - but since we've fully removed it's usage, do we possibly want it removed just to ensure it doesn't get added back?

max-muoto:
@auvipy Added nearly fully type-hinting to `utils/time.py` to improve generally readability (can wrap it up, and get the file added to the `pyproject.toml` in a separate PR), but went ahead and added some additional documentation around ZoneInfo and the usage of the various functions in regards to it throughout the utils. Let me know if what I added suffices, or if there is anything else that I could do in terms of expanding documentation. (Doesn't seem like it's a general practice to put documentation in markdown files in the codebase, but can do that if it would be helpful). Apologies for the delay.

max-muoto:
> just for sanity check can you please check which other celery projects are using pytz? I have to handle them as well after this is merged

Happy to put out PRs for other repos. The ones that actively use it are:

- kombu

Ones that simply depend on it:

- jumpstarter
- Celery-Kubernetes-Operator
- django-celery-results

auvipy:
IMHO, we need change in kombu first before merging this as celery depends on kombu

max-muoto:
> IMHO, we need change in kombu first before merging this as celery depends on kombu

I can do that, taking a look it should be a super quick PR.

auvipy:
is this cleaned up https://docs.celeryq.dev/en/stable/userguide/configuration.html#std-setting-enable_utc ?

max-muoto:
> is this cleaned up https://docs.celeryq.dev/en/stable/userguide/configuration.html#std-setting-enable_utc ?

Thanks for the callout! Just updated the documentation here to reflect that you can choose from `ZoneInfo` timezones as opposed to `pytz` timezones.

celery/utils/time.py
```diff
@@ -259,52 +291,58 @@ def maybe_iso8601(dt):
...
 
-def make_aware(dt, tz):
+def _can_detect_ambiguous(tz: tzinfo) -> bool:
+    """Helper function to determine if a timezone can detect ambiguous times using dateutil."""
+
+    return isinstance(tz, ZoneInfo) or hasattr(tz, "is_ambiguous")
+
+
+def _is_ambigious(dt: datetime, tz: tzinfo) -> bool:
+    """Helper function to determine if a timezone is ambigious using dateutil.
```
auvipy, line 306:
```suggestion
    """Helper function to determine if a timezone is ambiguous using python-dateutil.
```

celery/utils/time.py
```diff
@@ -259,52 +291,58 @@ def maybe_iso8601(dt):
...
+
+    Returns False if the timezone cannot detect ambiguity, or if there is no abiguity, otherwise True.
```
auvipy, line 308:
```suggestion
    Returns False if the timezone cannot detect ambiguity, or if there is no ambiguity, otherwise True.
```

celery/utils/time.py
```diff
@@ -259,52 +291,58 @@ def maybe_iso8601(dt):
...
+
+    In order to detect ambiguous datetimes, the timezone must be built using ZoneInfo, or have an is_ambiguous
+    method. Previously, pytz timezones would throw an AmbiguousTimeError if the localized dt was ambigious,
```
auvipy, line 311:
```suggestion
    method. Previously, pytz timezones would throw an AmbiguousTimeError if the localized dt was ambiguous,
```

auvipy (changes requested):

celery/utils/time.py
```diff
@@ -102,53 +112,69 @@ def _isdst(self, dt):
 
 
 class _Zone:
+    """Timezone class that provides the timezone for the application.
+    If `enable_utc` is disabled, LocalTimezone is provided as the timezone provider through local().
+    Otherwise, this class provides UTC ZoneInfo instance as the timezone provider for the application.
```
auvipy, line 117:
```suggestion
    Otherwise, this class provides a UTC ZoneInfo instance as the timezone provider for the application.
```

auvipy (changes requested):

celery/utils/time.py
```diff
@@ -259,52 +291,58 @@ def maybe_iso8601(dt):
...
 
-def make_aware(dt, tz):
+def _can_detect_ambiguous(tz: tzinfo) -> bool:
+    """Helper function to determine if a timezone can detect ambiguous times using dateutil."""
+
+    return isinstance(tz, ZoneInfo) or hasattr(tz, "is_ambiguous")
+
+
+def _is_ambigious(dt: datetime, tz: tzinfo) -> bool:
+    """Helper function to determine if a timezone is ambiguous using python-dateutil.
```
auvipy, line 306:
```suggestion
    """Helper function to determine if a timezone is ambiguous using python's dateutil module.
```

auvipy (changes requested):

auvipy (approved):

auvipy:
@max-muoto did we accidentally introduced https://pypi.org/project/python-dateutil/ here?

max-muoto:
> @max-muoto did we accidentally introduced https://pypi.org/project/python-dateutil/ here?

Apologies for not clarifying there during the PR review. Yes, I added that in to simplify error checking (e.g. ambiguous dates) in regards to `ZoneInfo`. We can move away from this library if we develop our own utils for checking things such as this. 

Happy to put a follow-up PR removing the dependency and taking the approach of building out our own utils if you think that's the best path. For example, we could have something like this: https://github.com/encode/django-rest-framework/blob/7bebe9772488fcf67949e1221e70af642ed0cb74/rest_framework/utils/timezone.py#L12

auvipy:
> > @max-muoto did we accidentally introduced https://pypi.org/project/python-dateutil/ here?
> 
> Apologies for not clarifying there during the PR review. Yes, I added that in to simplify error checking (e.g. ambiguous dates) in regards to `ZoneInfo`. We can move away from this library if we develop our own utils for checking things such as this.
> 
> Happy to put a follow-up PR removing the dependency and taking the approach of building out our own utils if you think that's the best path. For example, we could have something like this: https://github.com/encode/django-rest-framework/blob/7bebe9772488fcf67949e1221e70af642ed0cb74/rest_framework/utils/timezone.py#L12

it would be great if  We can move away from this library if we develop our own utils for checking things such as this. please come with a PR and ping me

max-muoto:
> > > @max-muoto did we accidentally introduced https://pypi.org/project/python-dateutil/ here?
> 
> > 
> 
> > Apologies for not clarifying there during the PR review. Yes, I added that in to simplify error checking (e.g. ambiguous dates) in regards to `ZoneInfo`. We can move away from this library if we develop our own utils for checking things such as this.
> 
> > 
> 
> > Happy to put a follow-up PR removing the dependency and taking the approach of building out our own utils if you think that's the best path. For example, we could have something like this: https://github.com/encode/django-rest-framework/blob/7bebe9772488fcf67949e1221e70af642ed0cb74/rest_framework/utils/timezone.py#L12
> 
> 
> 
> it would be great if  We can move away from this library if we develop our own utils for checking things such as this. please come with a PR and ping me

Will do! I'll get a PR out.

max-muoto:
@auvipy I'm actually starting to see some things that make me think we should either A) adopt tzdata as a dependency or B) copy over the relevant functions it provides giving credit. After doing more research, the `datetime_exists` it provides is flawed, https://pytz-deprecation-shim.readthedocs.io/en/latest/migration.html and isn't straightforward to fix. For example, if you look here: https://github.com/pganssle/pytz-deprecation-shim/issues/29


Do you have any thoughts on best next steps?

auvipy:
we can keep python-dateutils but should we get rid of pytz-deprecation shim?

max-muoto:
> we can keep python-dateutils but should we get rid of pytz-deprecation shim?

Apologies for the hiatus, I have been pretty busy these last couple of weeks.

Currently we have no dependency on `pytz_deprecation_shim`. However, I recommend we go in add `python-dateutils` for DRF since there I opted to add our own date utils there that will have the aforementioned problems, so it would be good to take a more similar approach to Celery.

auvipy:
OK cool. thanks

PR #8161, opened 2023-03-28
Update redis.txt 4.3.6

auvipy:


auvipy:
should be updated after https://github.com/redis/redis-py/issues/2665 is resolved

auvipy (commented):
as celery is 3.7+ only this version should not be affected

PR #8164, opened 2023-03-28
Update & simplify GHA pip caching

stumpylog:

## Description

This PR accomplishes two things for the workflow.  

One, it will resolve the niggling warnings about using `set-output`, which is deprecated and going to be removed eventually.

Two (and this is the large one), it changes the action to utilize the `pip` caching from `actions/setup-python`.  This frees the project from needing to maintain caching and get any benefits from the upstream's improvements to the caching.  The cache key will already include the OS, OS version, Python version and has been updated to include the `setup.py` as the previous code did.  This could also extend to the requirement text files easily, if desired.

You can compare the timing of the [first run](https://github.com/stumpylog/celery/actions/runs/4547884225) without caching to a [second run](https://github.com/stumpylog/celery/actions/runs/4548157700) where the cache keys were hit.  38m 36s down to 31m 17s isn't a huge decrease, but it's a little better.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -103,6 +103,8 @@ jobs:
               uses: actions/setup-python@v4
               with:
                   python-version: ${{ matrix.python-version }}
+                  cache: 'pip'
+                  cache-dependency-path: '**/setup.py'
           -   name: Install tox
               run: python -m pip install --upgrade pip tox tox-gh-actions
           -   name: >
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -46,19 +46,8 @@ jobs:
       uses: actions/setup-python@v4
       with:
         python-version: ${{ matrix.python-version }}
-
-    - name: Get pip cache dir
-      id: pip-cache
-      run: |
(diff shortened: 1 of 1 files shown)


auvipy (commented):
can you please point out to relevant docs please?

auvipy (approved):

auvipy:
> This could also extend to the requirement text files easily, if desired.

can you please elaborate more on this? can this be handled in this PR?

auvipy (changes requested):
https://github.com/celery/celery/pull/8164#issuecomment-1488188892

stumpylog:
The basic setup docs: https://github.com/actions/setup-python#caching-packages-dependencies

The more advanced docs, with a lot of examples, including how to include other files are part of the cache key: https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#caching-packages

Currently, besides the OS, platform and Python version, the `setup.py` files are included as part of the cache key, so the cache will break if any of these change.  I included the setup files as the original did.  The cache key could also be updated to inclue catching changes in the .txt files inside the requirements folder, either alongside or in place of, the setup files.  I was not sure if that would be useful or not, but wanted to highlight it as an option.

auvipy (approved):

PR #8206, opened 2023-04-18
start removing codecov from pypi

auvipy:
this was hidden due to cache


auvipy (commented):
it seems to be redundant

Issue #8207, opened 2023-04-18
Time drift is calculated by comparing the sender's and receiver's local times


monai:
# Checklist

- [ ] I have verified that the issue exists against the `main` branch of Celery.
- [ ] This has already been asked to the [discussions forum](https://github.com/celery/celery/discussions) first.
- [ ] I have read the relevant section in the
  [contribution guide](https://docs.celeryq.dev/en/main/contributing.html#other-bugs)
  on reporting bugs.
- [x] I have checked the [issues list](https://github.com/celery/celery/issues?q=is%3Aissue+label%3A%22Issue+Type%3A+Bug+Report%22+-label%3A%22Category%3A+Documentation%22)
  for similar or identical bug reports.
- [x] I have checked the [pull requests list](https://github.com/celery/celery/pulls?q=is%3Apr+label%3A%22PR+Type%3A+Bugfix%22+-label%3A%22Category%3A+Documentation%22)
  for existing proposed fixes.
- [ ] I have checked the [commit log](https://github.com/celery/celery/commits/main)
  to find out if the bug was already fixed in the main branch.
- [ ] I have included all related issues and possible duplicate issues
  in this issue (If there are none, check this box anyway).

## Mandatory Debugging Information

- [ ] I have included the output of ``celery -A proj report`` in the issue.
    (if you are not able to do this, then at least specify the Celery
     version affected).
- [ ] I have verified that the issue exists against the `main` branch of Celery.
- [ ] I have included the contents of ``pip freeze`` in the issue.
- [ ] I have included all the versions of all the external dependencies required
  to reproduce this bug.

## Optional Debugging Information

- [ ] I have tried reproducing the issue on more than one Python version
  and/or implementation.
- [ ] I have tried reproducing the issue on more than one message broker and/or
  result backend.
- [ ] I have tried reproducing the issue on more than one version of the message
  broker and/or result backend.
- [ ] I have tried reproducing the issue on more than one operating system.
- [ ] I have tried reproducing the issue on more than one workers pool.
- [ ] I have tried reproducing the issue with autoscaling, retries,
  ETA/Countdown & rate limits disabled.
- [ ] I have tried reproducing the issue after downgrading
  and/or upgrading Celery and its dependencies.

#### Related Issues

- #6213

#### Possible Duplicates

- None

# Steps to Reproduce

## Minimally Reproducible Test Case

Have two nodes in different time zones, i.e., having different `TZ` environment variable values.

# Expected Behavior

[...]

PR #8236, opened 2023-05-08
Fix Time Drift Calculation

max-muoto:

## Description

Fix bug discussed in https://github.com/celery/celery/issues/8207 but doing the time drift calculation in events in UTC as opposed to local times.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
   - repo: https://github.com/asottile/pyupgrade
-    rev: v3.3.1
+    rev: v3.3.2
     hooks:
       - id: pyupgrade
         args: ["--py37-plus"]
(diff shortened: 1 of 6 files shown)


max-muoto:
@auvipy Still working on resolving some things here, but will ping you once it's ready.

auvipy:
I still can't see any fix except the introduction of a newly introduced variable

max-muoto:
@auvipy Not totally done (also need to add test coverages), but the idea is that `non_adjusted_timestamp` is the `timestamp` except it never gets localized with `utcoffset`.

`body['timestamp'] = adjust_timestamp(timestamp, offset)` adjusts the `timestamp` with the provided `utcoffset`.

max-muoto:
@auvipy Basic implementation is done, let me know if what I said makes sense. Will add unit tests tomorrow if you think it looks good.

auvipy (commented):
were you able to fix it?

PR #8238, opened 2023-05-08
chore(build): switch to Poetry

stegayet:

## Description

WIP

This is a proposal about switching from scattered build and requirements files to PEP-517 build system using `poetry` and `poetry-core` with `pyproject.toml`.
The work is not done yet, I'd like to get some feedback about the idea first before moving forward.

TODO

- [ ] Check that we have no regression in wheel content
- [ ] Check that dependencies version and extras are the same as before
- [ ] Update GH actions accordingly

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,3 +1,106 @@
+[tool.poetry]
+name = "celery"
+version = "5.3.0b2"
+description = "Distributed Task Queue."
+license = "BSD-3-Clause"
+authors = ["Ask Solem <auvipy@gmail.com>"]
+homepage = "https://docs.celeryq.dev/"
+keywords = ["task", "queue", "job", "async", "rabbitmq", "amqp", "redis", "python", "distributed", "actor"] 
+classifiers = []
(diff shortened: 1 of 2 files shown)


Nusnus:
Hey @stegayet , thank you for taking the lead on such proposal.
I had on my mind upgrading us to Poetry after 5.3 release, as the upcoming `pytest-celery` also uses poetry and I think it would benefit Celery as well.

That being said, it means that it will be interesting after the release which unfortunately is still delaying.
Also, I believe a proper upgrade should be on more repos and not just Celery.

@thedrow @auvipy - What do you think?

stegayet:
I can do the same move for other repositories related to Celery :+1: 
We are not in a rush, this can definitely wait after 5.3 release.

Do you have an idea when 5.3 release is going to land ?

auvipy:
> Do you have an idea when 5.3 release is going to land ?

we plan to release it in this may

Viicos:
What about keeping setuptools and just migrating to `pyproject`? Poetry can be tricky w.r.t. dependencies for libraries (see [this article](https://iscinumpy.dev/post/bound-version-constraints/)) and things might change with https://github.com/python-poetry/poetry/pull/9135, meaning more work will have to be done once PEP 621 gets supported by Poetry.

PR #8250, opened 2023-05-14
Class method to process crontab string

padgy:
Convenience method for creating a crontab object from a string. Many configurations in systems are written as a cron string, so adding a helper function for this would make reading schedules easier.


auvipy (changes requested):
I would like to know more context and examples and unit tests for the proposed change

auvipy:
we need proper unit tests for considering that, also wait for feedback from more users

celery/schedules.py
```diff
@@ -408,6 +408,14 @@ def __init__(self, minute: str = '*', hour: str = '*', day_of_week: str = '*',
         self.month_of_year = self._expand_cronspec(month_of_year, 12, 1)
         super().__init__(**kwargs)
 
+    @classmethod
+    def from_string(cls, crontab: str):
+        """
+        Create a Crontab from a string. For example ``crontab='* * * * *'``.
+        """
+        minute, hour, day_of_week, day_of_month, month_of_year = crontab.split(" ")
```
jayeff, line 416: I would suggest to switch the order of fields to match offical cron format
```suggestion
        minute, hour, day_of_month, month_of_year, day_of_week = crontab.split(" ")
```

```
# ┌───────────── minute (0–59)
# │ ┌───────────── hour (0–23)
# │ │ ┌───────────── day of the month (1–31)
# │ │ │ ┌───────────── month (1–12)
# │ │ │ │ ┌───────────── day of the week (0–6) (Sunday to Saturday;
# │ │ │ │ │                                   7 is also Sunday on some systems)
# │ │ │ │ │
# │ │ │ │ │
# * * * * * <command to execute>
```
Source: https://en.wikipedia.org/wiki/Cron

jayeff (commented):
I too think having a convenience method to get a `crontab` from a cron expression would be helpful in cases where you use configration (e.g. enviroment variable) to define a schedule for some tasks. We use this quite a lot to have different schedules for different runtime enviroments

I would suggest to use the official cron format though as per suggestion

@auvipy In case you give a go ahead I could update the format and add some unit tests to this PR so that it can be merged

auvipy:
> @auvipy In case you give a go ahead I could update the format and add some unit tests to this PR so that it can be merged

sorry for the late response. Please give it a shot if you can!

Nusnus (changes requested):
Please also take note of the failing lint error:
```console
mypy.....................................................................Failed
- hook id: mypy
- exit code: 1

celery/schedules.py:415: error: Function is missing a return type annotation  [no-untyped-def]
Found 1 error in 1 file (checked 10 source files)
```

Thank you 🙏

jayeff:
I created a new PR https://github.com/celery/celery/pull/9079 which include tests and fixes the mypy error

Nusnus:
Thank you @padgy for the original contribution ♥️
It was finished and merged here: https://github.com/celery/celery/pull/9079

PR #8254, opened 2023-05-17
update old hard-coded URLs from celeryq.org to celeryq.dev

infinitewarp:
## Description

I recently encountered the `E_WOULDBLOCK` message at runtime, and when I followed the suggested link, I was taken to a site that _does **not**_ appear to belong to the actual Celery dev org, but instead some generic programming blog that _is not_ hosting Celery docs.

```
*** RuntimeError: Never call result.get() within a task!
```

This is a simple change to update these old hard-coded URLs to use the current correct domain.

These URLs were originally added back in 2016 with cf04c93d77. I remember the domain name ownership kerfuffle that happened a year ago. Perhaps this was a casualty of that event, or perhaps whoever owned the `celeryq.org` domain decided to stop redirecting to the correct docs? 🤷

### DIFF
--- a/celery/result.py
+++ b/celery/result.py
@@ -28,7 +28,7 @@ __all__ = (
 
 E_WOULDBLOCK = """\
 Never call result.get() within a task!
-See http://docs.celeryq.org/en/latest/userguide/tasks.html\
+See http://docs.celeryq.dev/en/latest/userguide/tasks.html\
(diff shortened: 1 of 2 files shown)


infinitewarp:
Oh, I guess someone else spotted this recently too and already fixed it. Sorry for the noise.

PR #8149, opened 2023-03-23
Fix `Task.handle_ignore` not wrapping exceptions properly

youtux:
## Description

Fix `Task.handle_failure` not wrapping the exception correctly in case of unpicklealbe exceptions.

Other changes:
* add the traceback object too when wrapping an unpickleable exception
* added tests to ensure that `Task.handle_failure` and `Task.handle_retry` work as expected

This work is a continuation of the fixes I made in https://github.com/celery/celery/pull/8133.

PR #8257, opened 2023-05-22
Fix exc_type being exception instance rather than type

Mapiarz:
This PR https://github.com/celery/celery/pull/8149 introduced a regression which will cause some log handlers to crash. In my particular case, it is `structlog` with `ConsoleRenderer` (which uses `rich` under the hood). The reason for that I believe is https://github.com/celery/celery/commit/3ce5b85806104e14f75a377fadc4de3e50038396#diff-fb194522c454315cd87ccb1cf4a2a53aafa49124c85abc01dcca4310cf19d6d3R225 which passes the exception instance rather than the exception type. Some handlers will inspect and expect there to be a type and that will lead to crashes.

### DIFF
--- a/celery/app/trace.py
+++ b/celery/app/trace.py
@@ -222,7 +222,7 @@ class TraceInfo:
             # a traceback.
             _, _, exc.__traceback__ = sys.exc_info()
 
-        exc_type = get_pickleable_etype(orig_exc)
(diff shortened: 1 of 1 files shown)


auvipy:
@youtuxcan you verify, please?

youtux:
Ah yes, that makes sense!

auvipy (approved):

PR #8295, opened 2023-06-03
fixed a small float value of retry_backoff

ipakeev:

## Description
I encountered an unexpected behavior: if I set retry_backoff to 0.0 < {value} < 1.0, all retries will run after default_retry_delay (180 seconds). 
I don't think it makes much sense to allow a floating retry_backoff value, so I propose just rounding the {value} up to 1.

### DIFF
--- a/celery/app/autoretry.py
+++ b/celery/app/autoretry.py
@@ -18,7 +18,7 @@ def add_autoretry_behaviour(task, **options):
     retry_kwargs = options.get(
         'retry_kwargs', getattr(task, 'retry_kwargs', {})
     )
-    retry_backoff = int(
+    retry_backoff = float(
         options.get('retry_backoff',
                     getattr(task, 'retry_backoff', False))
     )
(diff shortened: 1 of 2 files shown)


auvipy (changes requested):
can you please add additional test cases for this change?

ipakeev:
@auvipy Added tests.

auvipy:
it would be great to add additional integration tests

ipakeev:
Added additional tests.

@auvipy what kind of integration tests do you expect? I've not found any integration cases in which retry_backoff is used.

celery/app/autoretry.py
```diff
@@ -48,7 +48,7 @@ def run(*args, **kwargs):
                 if retry_backoff:
                     retry_kwargs['countdown'] = \
                         get_exponential_backoff_interval(
-                            factor=retry_backoff,
+                            factor=int(max(1.0, retry_backoff)),
```
auvipy, line 51: are we going to turn everything into 1 only or multiple of 1.0?

ipakeev: All values less then 1 turn into 1 (except zero). For the other values, the behavior remains the same. I'm using 1.0 (not 1) just for linter.

auvipy (approved):

auvipy:
merging with the hope that no regression was introduced! if we got any report, please try to collaborate if needed. thanks for your contribution.

PR #8296, opened 2023-06-03
Avoid librabbitmq problem in python 3.11

lspvic:

## Description

1. It seems `librabbitmq` is no longer maintained, and the `librabbitmq` package has 
bugs in python 3.11, as stated in https://github.com/celery/librabbitmq/issues/167.
If install requires with `celery[librabbitmq]` of some other packages or system migrating
to 3.11 will lead to errors. As `librabbitmq` is not suggested, we can restricted
install require python versions for bakcward compatibility and further 3.11.
```
librabbitmq>=2.0.0; python_version < '3.11'
```
2. The bundle `celery[amqp]` in readme does not exist actually. There's a warning
`WARNING: celery 5.*.* does not provide the extra 'amqp'` when installing with it.
This might be confusing. Shall we use some others to show as an example 

@auvipy

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -260,9 +260,9 @@ separating them by commas.
 ::
 
 
-    $ pip install "celery[amqp]"
+    $ pip install "celery[gevent]"
 
-    $ pip install "celery[amqp,redis,auth,msgpack]"
+    $ pip install "celery[gevent,redis,auth,msgpack]"
 
 The following bundles are available:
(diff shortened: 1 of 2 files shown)


auvipy:
> 2\. Shall we use some others to show as an example

yes please

auvipy (commented):
merged. thanks

PR #8305, opened 2023-06-08
ChainMap observers fix

shahar-lev:
Observers should not be shared across different instances. Aside from unwanted behavior, this can lead to object leaks (like celery app objects not being garbage collected).

### DIFF
--- a/t/unit/utils/test_collections.py
+++ b/t/unit/utils/test_collections.py
@@ -2,14 +2,14 @@ import pickle
 from collections.abc import Mapping
 from itertools import count
 from time import monotonic
+from unittest.mock import Mock
 
 import pytest
-from unittest.mock import Mock
 from billiard.einfo import ExceptionInfo
 
 import t.skip
-from celery.utils.collections import (AttributeDict, BufferMap, ConfigurationView, DictAttribute, LimitedSet,
-                                      Messagebuffer, ChainMap)
+from celery.utils.collections import (AttributeDict, BufferMap, ChainMap, ConfigurationView, DictAttribute,
+                                      LimitedSet, Messagebuffer)
 from celery.utils.objects import Bunch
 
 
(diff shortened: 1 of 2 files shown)


t/unit/utils/test_collections.py
```diff
@@ -448,3 +449,16 @@ def test_pop_empty_no_default(self):
...
+
+class test_ChainMap:
+
+    def test_observers_not_shared(self):
+        a = ChainMap()
+        b = ChainMap()
+        callback = Mock()
+        a.bind_to(callback)
+        b.update(x=1)
+        callback.assert_not_called()
```
auvipy, line 462: can we lean to pytest like syntax?

auvipy (changes requested):

t/unit/utils/test_collections.py
shahar-lev: is there an alternative to `assert_not_called` and `assert_called_once_with` in pytest?
I see they are used extensively in celery unit tests.

auvipy: ugh sorry my mistake!

shahar-lev: do you need anything else from me?

auvipy: is it possible to increase test coverage and if possible to add integration tests as well?

shahar-lev: It's weird that I reduced coverage by this commit - It's just a bugfix with a new unit test that checks it actually fixed the bug (I didn't add any new "logic" to the code itself).
It also feels like a low-level bug that doesn't suit an integration test.
WDYT?

auvipy: codecov sometimes do mistakes :D

auvipy (approved):

PR #8308, opened 2023-06-12
fix supported versions in docs

kviktor:
## Description

Updated the documentation about the new versions supported.

Got the PyPy one from the README https://github.com/celery/celery/blob/main/README.rst?plain=1#L64

Django one from the release notes: https://github.com/celery/celery/releases/tag/v5.3.0 (technically between 1.11 and 2.2 in theory 5.0.x - 5.2.x range is supported but maybe its better to nudge ppl to the latest one)

### DIFF
--- a/docs/django/first-steps-with-django.rst
+++ b/docs/django/first-steps-with-django.rst
@@ -19,8 +19,8 @@ Using Celery with Django
 
 .. note::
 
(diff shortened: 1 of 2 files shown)


auvipy (approved):
interestingly i missed them! thanks

PR #8312, opened 2023-06-13
Fixes #8224 Fixes worker command optimizations flag

### LINKED ISSUE #8224: Why has the CLI worker command flag `optimizations` changed from `--optimizations` in 5.*?
This question is referencing this code:

https://github.com/celery/celery/blob/3b1c768c39f641a70d3086bd5a1079a421f94344/celery/bin/worker.py#L169
```
@click.option('optimization',
              '-O',
              default='default',
              cls=CeleryOption,
              type=click.Choice(('default', 'fair')),
              help_group="Worker Options",
              help="Apply optimization profile.")
```
Why has the flag removed the double hyphen? It seems like a bug, but I assume there was a reason for it? The reason I'm asking is that this change breaks a bunch of code we have, so I'd like to understand the change.

stuart-bradley:
## Description

This PR fixes #8224 wherein the `optimizations` flag of the worker command lost its double `-`, resulting in unintended breaking behaviour in 5.*.


auvipy (commented):
did you check the celery 5.0.0 release notes and migration guide?

stuart-bradley:
@auvipy I did, this change is not referenced.

auvipy:
Celery 5.0 introduces a new CLI implementation which isn’t completely backwards compatible.

The global options can no longer be positioned after the sub-command. Instead, they must be positioned as an option for the celery command like so:

auvipy:
@Nusnus it would be great if you could take  a feedback from Omer

Nusnus (approved):

Nusnus:
@auvipy 
> @Nusnus it would be great if you could take a feedback from Omer

He's OOO so to speak for some time and available only for critical/urgent issues so I prefer to not reach him with that at the moment and help myself instead.

So I've reviewed the changes myself, and it makes sense to me.
This definitely looks like a typo, and the fix looks correct.

Thank you @stuart-bradley!
Added my approval.

celery/bin/worker.py
```diff
@@ -166,8 +166,8 @@ def detach(path, argv, logfile=None, pidfile=None, uid=None,
               type=LOG_LEVEL,
               help_group="Worker Options",
               help="Logging level.")
-@click.option('optimization',
-              '-O',
+@click.option('-O',
```
auvipy, line 169: -O before or after? just for double check

Nusnus: Before :)

Nusnus: See code above/below, the pattern is consistent being first `-O`, then `--performance`

auvipy (approved):

PR #8321, opened 2023-06-17
removed extraneous line in trace code

eukreign:
This is a very simple, one line change, to delete an extraneous line in `celery/app/trace.py`.

### DIFF
--- a/celery/app/trace.py
+++ b/celery/app/trace.py
@@ -341,7 +341,6 @@ def build_tracer(
 
     loader = loader or app.loader
     ignore_result = task.ignore_result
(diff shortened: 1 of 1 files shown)


celery/app/trace.py
```diff
@@ -341,7 +341,6 @@ def build_tracer(
 
     loader = loader or app.loader
     ignore_result = task.ignore_result
-    track_started = task.track_started
```
auvipy, line 344: any extra tests to cover it to avoid any regression?

auvipy (changes requested):

PR #8339, opened 2023-06-24
Bugfix: Removed unecessary stamping code from _chord.run()

Nusnus:
Stamping is for signatures, not for tasks, so it is **never** expected to receive it via `options`.


auvipy (commented):
can we consider this for 5.3.x release?

Nusnus:
> can we consider this for 5.3.x release?

I tagged 5.4 but you're right, this is definitely a 5.3.x, so yeah !

auvipy (approved):

PR #8340, opened 2023-06-24
Remove as it seems to be buggy.

auvipy:
After pushing a tag, we can now use github release notes generator so it is better to remove now. when it was added, github didn't had the release note generator.


Nusnus:
I think I found the issue.
The script isn't _bugged_ per say, it's just defined to always override retroactively and matches *all of the* GitHub releases to `Changelog.rst`, so if we release a version manually, when this file is changed everything else will be changed also in GitHub releases.

Nusnus:
This means that to release a new GitHub release, we need to just push a commit to this file.
That means, to get 5.3.0 message back we need to create a PR to Changelog.rst and it should be fixed automatically @auvipy

Nusnus:
So I suggest closing the PR and leaving this script intact, it makes sense to me and it does work.

That being said, this PR is great for documentation of this issue for future cases ❤️

Nusnus:
We can use the release notes generator manually when we're committing a new version in `Changelog.rst` and copy it, adding our personal note as well that will be kept as it's in the rst file

auvipy:
good idea

auvipy (commented):
I think it is not needed for our case. we can find something better or any new approach.

Issue #7039, opened 2021-11-03
Error using update_state with Arangodb backend


pikachupichu25:
# Checklist

* [x] I have verified that the issue exists against the `master` branch of Celery.
* [ ] This has already been asked to the [discussions forum](https://github.com/celery/celery/discussions) first.
* [ ] I have read the relevant section in the
  [contribution guide](http://docs.celeryproject.org/en/latest/contributing.html#other-bugs)
  on reporting bugs.
* [x] I have checked the [issues list](https://github.com/celery/celery/issues?q=is%3Aissue+label%3A%22Issue+Type%3A+Bug+Report%22+-label%3A%22Category%3A+Documentation%22)
  for similar or identical bug reports.
* [ ] I have checked the [pull requests list](https://github.com/celery/celery/pulls?q=is%3Apr+label%3A%22PR+Type%3A+Bugfix%22+-label%3A%22Category%3A+Documentation%22)
  for existing proposed fixes.
* [ ] I have checked the [commit log](https://github.com/celery/celery/commits/master)
  to find out if the bug was already fixed in the master branch.
* [ ] I have included all related issues and possible duplicate issues
  in this issue (If there are none, check this box anyway).

## Mandatory Debugging Information
* [ ] I have included the output of ``celery -A proj report`` in the issue.
    (if you are not able to do this, then at least specify the Celery
     version affected).
* [ ] I have verified that the issue exists against the `master` branch of Celery.
* [ ] I have included the contents of ``pip freeze`` in the issue.
* [ ] I have included all the versions of all the external dependencies required
  to reproduce this bug.

## Optional Debugging Information

* [ ] I have tried reproducing the issue on more than one Python version
  and/or implementation.
* [ ] I have tried reproducing the issue on more than one message broker and/or
  result backend.
* [ ] I have tried reproducing the issue on more than one version of the message
  broker and/or result backend.
* [ ] I have tried reproducing the issue on more than one operating system.
* [ ] I have tried reproducing the issue on more than one workers pool.
* [ ] I have tried reproducing the issue with autoscaling, retries,
  ETA/Countdown & rate limits disabled.
* [ ] I have tried reproducing the issue after downgrading
  and/or upgrading Celery and its dependencies.

#### Related Issues
* None

#### Possible Duplicates
* None

## Environment & Settings

**Celery version**: 5.0.0

<details>
<summary><b><code>celery report</code> Output:</b></summary>
<p>

```
[...]

PR #8351, opened 2023-07-01
Implement safer ArangoDB queries

othieno:
The AQL queries used in the ArangoDbBackend's implementation are potentially
vulnerable to injections because no sanity checks are performed on the
arguments used to build the query strings.

This is particularly evident when using a database collection with a dash in
its name, e.g. "celery-task-results". The query string generated by the set
method is 'INSERT {task: v}, _key: "k"} INTO celery-task-results' instead of
'INSERT {task: v}, _key: "k"} INTO `celery-task-results`' (backticks
surrounding collection name). The former is evaluated as a substraction
(celery - task - results) and is therefore an illegal collection name,
while the latter is evaluated as a string.

This commit re-implements the setter and getters using bind parameters,
which performs the necessary safety checks. Furthermore, the new query used
in the set method accounts for updates to existing keys, resolving https://github.com/celery/celery/issues/7039.

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -292,3 +292,4 @@ Oleh Romanovskyi, 2022/06/09
 JoonHwan Kim, 2022/08/01
 Kaustav Banerjee, 2022/11/10
 Austin Snoeyink 2022/12/06
+Jeremy Z. Othieno 2023/07/27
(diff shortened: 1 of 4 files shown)


auvipy (commented):
having improved test coverage for all the methods would be great

auvipy:
are the new changes backward compatible?

auvipy (approved):

PR #8383, opened 2023-07-20
Fix eager tasks does not populate name field

### LINKED ISSUE #7715: EagerResult doesn't seem to poplate name
Sometimes I run my tasks with `CELERY_TASK_ALWAYS_EAGER` to aid debugging, it seems like the `name` property of EagerResult isn't populated, which makes this sort of investigation more tricky.

I have some code I use to list my tasks in a django app, and part of this is to grab the task name.

When running eagerly, the EagerResult task status is SUCCESS, so I would have expected `name` to be available at this point.

KOliver94:
*Note*: Before submitting this pull request, please review our [contributing guidelines](https://docs.celeryq.dev/en/main/contributing.html).

## Description

Multiple issues were reported both here (https://github.com/celery/celery/issues/7715) and in django-celery-results (https://github.com/celery/django-celery-results/issues/289 & https://github.com/celery/django-celery-results/issues/362) that when running in eager mode name of the tasks are not populated. After my investigation it seems to me that this field was just missed as it's not critical in terms of functionality. I added it and it seems to work.

Let me know what do you think.

Fixes #7715

### DIFF
--- a/celery/app/task.py
+++ b/celery/app/task.py
@@ -825,7 +825,7 @@ class Task:
         if isinstance(retval, Retry) and retval.sig is not None:
             return retval.sig.apply(retries=retries + 1)
         state = states.SUCCESS if ret.info is None else ret.info.state
-        return EagerResult(task_id, retval, state, traceback=tb)
+        return EagerResult(task_id, self.name, retval, state, traceback=tb)
 
     def AsyncResult(self, task_id, **kwargs):
         """Get AsyncResult instance for the specified task.
--- a/celery/app/task.py
+++ b/celery/app/task.py
@@ -788,6 +788,7 @@ class Task:
 
         request = {
             'id': task_id,
+            'task': self.name,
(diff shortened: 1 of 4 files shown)


auvipy (approved):
that looks good to me! can we add some sort of integration test for this as well? some manual acknowledgement would be also great

KOliver94:
Thank you @auvipy! I took a look at the existing tests but I didn't find any for similar case so I'm not sure what type of test can we provide for this. If you can show me some which you think can be used as starter and specify what you think should be tested, I'm happy to write some.

auvipy:
I think we are safe to go as is for now

shifenhutu:
can  you release a version ?

auvipy:
it will be included in next point release very soon

auvipy:
we got a bug / regression report here https://github.com/celery/celery/issues/8472 can you please verify?

Issue #8372, opened 2023-07-13
result.id is causing celery beat to fail when using transaction.on_commit()


Raghavsalotra:
https://github.com/celery/celery/blob/811ed96edbf7d7ae0681ae67ced63e6994a6e63a/celery/beat.py#L285

Reason for asking this question:
We have transaction.on_commit wrapper for out celery tasks which defers the execution of task until the transaction is committed to database. In that case the result returned will be None and celery beat will fail with this error:

`[celery.redirected] level=WARNING MainProcess (2023-07-13 09:29:06,184 trace_id="short_id:TEih9A3Q2mEm4J;" [log.write:232] process=96566 thread=8631279104)  message="self.apply_entry(entry, producer=self.producer)"
[celery.redirected] level=WARNING MainProcess (2023-07-13 09:29:06,184 trace_id="short_id:TEih9A3Q2mEm4J;" [log.write:232] process=96566 thread=8631279104)  message="  File "/lib/python3.8/site-packages/celery/beat.py", line 292, in apply_entry"
[celery.redirected] level=WARNING MainProcess (2023-07-13 09:29:06,184 trace_id="short_id:TEih9A3Q2mEm4J;" [log.write:232] process=96566 thread=8631279104)  message="    "
[celery.redirected] level=WARNING MainProcess (2023-07-13 09:29:06,184 trace_id="short_id:TEih9A3Q2mEm4J;" [log.write:232] process=96566 thread=8631279104)  message="debug('%s sent. id->%s', entry.task, result.id)"
[celery.redirected] level=WARNING MainProcess (2023-07-13 09:29:06,184 trace_id="short_id:TEih9A3Q2mEm4J;" [log.write:232] process=96566 thread=8631279104)  message="AttributeError"
[celery.redirected] level=WARNING MainProcess (2023-07-13 09:29:06,184 trace_id="short_id:TEih9A3Q2mEm4J;" [log.write:232] process=96566 thread=8631279104)  message=": "
[celery.redirected] level=WARNING MainProcess (2023-07-13 09:29:06,184 trace_id="short_id:TEih9A3Q2mEm4J;" [log.write:232] process=96566 thread=8631279104)  message="'NoneType' object has no attribute 'id'"
`

Is there a way using configuration parameter so that we do not end up in this issue or We might need to have a fork just for this change.

bmrobin:
i get this error too;  to get unblocked i just don't use `transaction.on_commit` with Celery tasks invoked via Beat.

[...]

PR #8428, opened 2023-08-08
fix(beat): debug statement should only log AsyncResult.id if it exists

bmrobin:
## Description
`celery.beat.apply_entry()` assumes that `apply_async()` will return an `AsyncResult` containing an ID.  this is not always the case:  for example when Django's `transaction.on_commit()` wraps the invoked task. this proposed fix simply modifies the `debug()` logging statement to more safely handle the result that is being logged after the task is created.

Fixes https://github.com/celery/celery/issues/8372

### DIFF
--- a/t/unit/app/test_beat.py
+++ b/t/unit/app/test_beat.py
@@ -301,6 +301,29 @@ class test_Scheduler:
         scheduler = mScheduler(app=self.app)
         assert isinstance(scheduler.info, str)
 
+    def test_apply_entry_handles_empty_result(self):
+        s = mScheduler(app=self.app)
+        entry = s.Entry(name='a name', task='foo', app=self.app)
+
+        with patch.object(s, 'apply_async') as mock_apply_async:
+            with patch("celery.beat.debug") as mock_debug:
+                mock_apply_async.return_value = None
+                s.apply_entry(entry)
+        mock_debug.assert_called_once_with('%s sent.', entry.task)
+
+        with patch.object(s, 'apply_async') as mock_apply_async:
+            with patch("celery.beat.debug") as mock_debug:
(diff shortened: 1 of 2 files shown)


Nusnus:
You can run it locally to make sure it works like this:
`tox -e 3.11-unit -- -k test_apply_entry_handles_empty_result `

Nusnus:
Seems to be an issue with Python 3.8 not supporting the given syntax, but except for this small issue LGTM ✅

bmrobin:
@Nusnus noted.  i tried what the "contributing" guide suggested with `make test` but i got a bunch of assertion failures when running on `main` (not my branch), so i decided to see what your CI on Github Actions would result in.

i ran the test i wrote locally on py3.11 but i'll check with `tox` now

t/unit/app/test_beat.py
```diff
@@ -301,6 +301,24 @@ def test_info(self):
...
+        entry = s.Entry(name='a name', task='foo', app=self.app)
+        with (patch.object(s, 'apply_async') as mock_apply_async,
+              patch("celery.beat.debug") as mock_debug):
+            mock_apply_async.return_value = None
+            s.apply_entry(entry)
+        mock_debug.assert_called_once_with('%s sent.', entry.task)
+
+        task_id = 'taskId123456'
+        with (patch.object(s, 'apply_async') as mock_apply_async,
+              patch("celery.beat.debug") as mock_debug):
```
bmrobin, line 315: oh woops - this syntax (parenthesized context managers) was not available in older python versions. fixing now...

Nusnus (approved):

Nusnus:
Thank you ❤️

PR #8435, opened 2023-08-13
Update auth.txt

auvipy:
new fix released


auvipy (commented):
this can be merged as is

PR #8442, opened 2023-08-17
Limit redis client to 4.x since 5.x fails the test suite

thedrow:
*Note*: Before submitting this pull request, please review our [contributing
guidelines](https://docs.celeryq.dev/en/main/contributing.html).

## Description

It seems that redis-py 5.0.0 fixes the failure of `test_chain.test_nested_chain_group_lone` in the integration tests and now the test [unexpectedly passes](https://github.com/celery/celery/actions/runs/5878301999/job/15940467006#step:7:1072) instead of timing out. However, we don't know what fixed the bug that caused a timeout to begin with.
In the meantime, let's limit upgrading redis-py to 5.x until we have time to investigate.

PR #8443, opened 2023-08-17
Limit tox to < 4.9

Nusnus:
Tox has [released 4.9.0](https://github.com/tox-dev/tox/releases/tag/4.9.0) which fixed a bug where non existing environments did NOT fail, so since tox 4.9.0 they now DO fail: https://github.com/tox-dev/tox/issues/2858

We are relying on the previous behavior (the bugged one) so until we can fix it, I'm limiting tox as it breaks every PR right now which is worse than the limitation itself (which is not recommended).

We need to fix this soon so we can remove this limitation @auvipy @thedrow

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -51,7 +51,7 @@ jobs:
         cache-dependency-path: '**/setup.py'
 
     - name: Install tox
-      run: python -m pip install --upgrade pip tox tox-gh-actions
+      run: python -m pip install --upgrade pip 'tox<4.9' tox-gh-actions
     - name: >
         Run tox for
         "${{ matrix.python-version }}-unit"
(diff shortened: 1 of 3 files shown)


auvipy:
might be useful https://tox.wiki/en/latest/changelog.html and should we consider finding out alternative?

auvipy (commented):
seems working

Nusnus:
I've been investigating this issue & [redis 5.x issue](https://github.com/celery/celery/pull/8442) in the last few days, what timing hehe..
Anyways, @auvipy & @thedrow, FYI, we're now limited to tox <4.9 and Redis < 5.

These are not major issues _at the moment_ but 1) be aware, 2) we need to remove these limitations soon as the time will come when upgrading will be mandatory so let's keep that in mind.

Nusnus:
> We are relying on the previous behavior (the bugged one) ...

Meaning we lack integration tests support in tox.ini under `[gh-actions]` section

auvipy (approved):
I'm going to merge this for now. to keep the CI working. we have to rethink the way we configured integration tests in tox.

Nusnus:
@auvipy 
> I'm going to merge this for now. to keep the CI working. we have to rethink the way we configured integration tests in tox.

I've improved my fix and we also have 100% passing tests so this is good to go indeed.

Regarding the future, I guess I'll be able to get to it once I go back to developing the smoke tests PR as it will suffer the same issues as the integration tests, or maybe earlier, we'll see.

PR #8464, opened 2023-08-30
Remove tox version limit

Nusnus:
This reverts #8443.
tox v4.11 now works normally.
We're potentially now dependent on 4.11> but I'll leave it without restrictions for now.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -51,7 +51,7 @@ jobs:
         cache-dependency-path: '**/setup.py'
 
     - name: Install tox
-      run: python -m pip install --upgrade pip 'tox<4.9' tox-gh-actions
+      run: python -m pip install --upgrade pip 'tox' tox-gh-actions
     - name: >
         Run tox for
         "${{ matrix.python-version }}-unit"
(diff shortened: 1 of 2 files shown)


requirements/pkgutils.txt
```diff
@@ -4,7 +4,7 @@ flake8>=3.8.3
 flakeplus>=1.1
 flake8-docstrings>=1.7.0
 pydocstyle==6.3.0
-tox>=3.8.4
+tox>=3.8.4,<5.9.0
```
auvipy, line 7: is it 4.9 or 5.9? I saw that the latest release is 4.11 https://pypi.org/project/tox/

auvipy (changes requested):

requirements/pkgutils.txt
auvipy: but seems tests are passing!

Nusnus: > but seems tests are passing!

I was just lazy checking tox 4.11 - that tmp commit is removed and the whole PR is refactored @auvipy 

Let's hope it fixes the issue! 🙏

PR #8466, opened 2023-08-30
Add option to S3 backend to partition by date

petebachant:
## Description

Adds a new configuration parameter to the S3 backend to enable partitioning by date, which is useful for my application. Can be enabled with `app.conf.s3_partition_by_date = True`.

### DIFF
--- a/celery/backends/s3.py
+++ b/celery/backends/s3.py
@@ -1,5 +1,7 @@
 """s3 result store backend."""
 
+from datetime import datetime
+
 from kombu.utils.encoding import bytes_to_str
 
 from celery.exceptions import ImproperlyConfigured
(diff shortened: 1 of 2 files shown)


celery/backends/s3.py
```diff
@@ -48,9 +50,13 @@ def __init__(self, **kwargs):
 
         self.base_path = conf.get('s3_base_path', None)
 
+        self.partition_by_date = conf.get('s3_partition_by_date', False)
+
         self._s3_resource = self._connect_to_s3()
 
     def _get_s3_object(self, key):
+        if self.partition_by_date:
```
auvipy, line 58: can you please add proper tests to the proposed change? it is hard to verify for us without having any test cases. also can you please point to the S3 docs relevant to the change?

auvipy (changes requested):

celery/backends/s3.py
```diff
@@ -48,9 +50,13 @@ def __init__(self, **kwargs):
...
+            key = "date=" + str(datetime.utcnow().date()) + "/" + key
```
petebachant, line 59: This should probably use the `date_done` attribute of the task metadata so the key can be looked up.

PR #8474, opened 2023-09-02
Update Changelog

mlissner:
Currently the changelog is pretty impossible to read. It has Markdown formatting instead of RST. This fixes it a bit.

### DIFF
--- a/Changelog.rst
+++ b/Changelog.rst
@@ -26,7 +26,8 @@ v5.3.2
 :release-date: 2023-08-31 1:30 P.M GMT+2
 :release-by: Tomer Nosrati
 
-## What's Changed
+What's Changed
+--------------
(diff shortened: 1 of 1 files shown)


Nusnus:
We have a [GitHub workflow](https://github.com/celery/celery/blob/main/.github/workflows/changerelease.yml) that creates a GitHub release based on the Changelog.rst.
GitHub Release uses Markdown. The rst does not need it.
This means we're kinda deadlocked here because if we remove markdown we fix the rst formatting but lose the GitHub release pretty message and if we set the markdown syntax the rst file gets messy.

v5.3.2/3 has caused multiple issues unfortunately - I am now reviewing all of the failing points to come up with a general solution.

Thank you for the report & PR! - I want to review the release process again before moving forward with this PR so I can be wiser before I take further actions.

auvipy:
we might consider moving to markdown from rst for the changelog?

auvipy (approved):

PR #8338, opened 2023-06-23
store children with database backend

aaronst:
Attempt to fix #8336. The test passes with `assert meta['children'] is None`, but need to test further with a task that actually has children.

PR #8475, opened 2023-09-02
Revert "store children with database backend"

auvipy:
Reverts celery/celery#8338

@Nusnus we might revert this for now and reintroduce in future with a proper migration support. for migration support I can work

### DIFF
--- a/celery/backends/database/models.py
+++ b/celery/backends/database/models.py
@@ -25,7 +25,6 @@ class Task(ResultModelBase):
     date_done = sa.Column(sa.DateTime, default=datetime.utcnow,
                           onupdate=datetime.utcnow, nullable=True)
     traceback = sa.Column(sa.Text, nullable=True)
-    children = sa.Column(PickleType, nullable=True)
 
     def __init__(self, task_id):
         self.task_id = task_id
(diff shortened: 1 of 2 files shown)


Nusnus (approved):
Agreed. This was a good change but unfortunately not complete. Once we have the remaining changes we can welcome this back.

Nusnus:
@aaronst we're reverting the change in `main` as we need to release an urgent patch version and we'd be happy to welcome this back once completed.

Ping me if you need any help/review and I'll prioritize it for me due to this situation.

aaronst:
@Nusnus @auvipy I suggest updating the documentation so that users are aware the database backends are not feature complete. The parent/child relationship between tasks is important and currently broken for any database backend.

Nusnus:
> @Nusnus @auvipy I suggest updating the documentation so that users are aware the database backends are not feature complete. The parent/child relationship between tasks is important and currently broken for any database backend.

@aaronst Is it reasonable to ask you please for the remaining changes (the migration) so we can complete it instead of documenting the gap?

aaronst:
@Nusnus I’m happy to help with the migration, but this bug still needs to be documented if the fix is being reverted.

auvipy:
I will be happy to accept a 

> @Nusnus I’m happy to help with the migration, but this bug still needs to be documented if the fix is being reverted.

I'm still open to a doc fix, until the code is fixed

PR #8498, opened 2023-09-12
Fix typos in test_canvas.py

monteiro-renato:


t/integration/test_canvas.py
```diff
@@ -3132,12 +3132,12 @@ def task_received_handler(request=None, **kwargs):
...
                         stamped_header in link.options
                         for stamped_header in link.options["stamped_headers"]
-                        if link  # the link itself doensn't have a link
+                        if link  # the link itself doesn't have a link
                     ],
                     [
                         stamped_header in link_error.options
                         for stamped_header in link_error.options["stamped_headers"]
-                        if link_error  # the link_error itself doensn't have a link
+                        if link_error  # the link_error itself doesn't have a link
```
monteiro-renato, line 3140:
```suggestion
                        if link_error  # the link_error itself doesn't have a link_error
```
This might also be a copy past typo, but since I'm still trying to understand the code I didn't change it.

t/integration/test_canvas.py
```diff
@@ -1877,7 +1877,7 @@ def test_chord_on_error(self, manager):
         backend = fail.app.backend
         j_key = backend.get_key_for_group(original_group_id, '.j')
         redis_connection = get_redis_connection()
-        # The redis key is either a list or zset depending on configuration
+        # The redis key is either a list or a set depending on configuration
```
auvipy, line 1880:
```suggestion
        # The redis key is either a list or zset depending on configuration
```

auvipy: from search: The Redis zset function is used to store the key type and value. In a redis, zset keys are unique, and values are limited by the number of floating point operations. The unique property in it is used to access the members. We can also access the items by using sorted orders and order score values.

auvipy (changes requested):

t/integration/test_canvas.py
monteiro-renato: I've pushed a new commit adding more context as to what a zset is.
There was an issue on the Redis project Github (# 4024) asking why the sorted set commands had the z prefix.

I can revert back to what is used to be if you want.

monteiro-renato: I struggled to find references to "zset" in the Redis documentation. Not sure if it has fallen into disuse (the term itself).

It's used extensively in this book tho > https://redis.com/ebook/part-2-core-concepts/chapter-3-commands-in-redis/3-5-sorted-sets/

auvipy (approved):

t/integration/test_canvas.py
auvipy: I got this 
ZADD zset 10 a
ZADD zset 5 b
ZADD zset 12.55 c
ZRANGE zset 0 -1 => b,a,c

from https://redis.io/docs/manual/patterns/twitter-clone/

Nusnus (approved):

PR #8527, opened 2023-09-20
Fixed indentation in Dockerfile for Python 3.11

Nusnus:


auvipy:
It is always better to run them locally when new versions are added or removed in docker O:)

PR #8532, opened 2023-09-21
intial schema migrations setup for sqlalchemy models 

auvipy:
using alembic

### DIFF
--- a/celery/backends/database/alembic.ini
+++ b/celery/backends/database/alembic.ini
@@ -1,0 +1,116 @@
+# A generic, single database configuration.
+
+[alembic]
+# path to migration scripts
+script_location = migrations
+
+# template used to generate migration file names; The default value is %%(rev)s_%%(slug)s
+# Uncomment the line below if you want the files to be prepended with date and time
+# see https://alembic.sqlalchemy.org/en/latest/tutorial.html#editing-the-ini-file
+# for all available tokens
+# file_template = %%(year)d_%%(month).2d_%%(day).2d_%%(hour).2d%%(minute).2d-%%(rev)s_%%(slug)s
+
+# sys.path path, will be prepended to sys.path if present.
+# defaults to the current working directory.
+prepend_sys_path = .
+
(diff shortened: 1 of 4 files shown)


Nusnus:
👑

auvipy:
I need to get my hands dirty with this for a few more days to make it fully ready.

ThiefMaster:
Is this really something that should be on the celery side?

- If I do not use alembic and just use `create_all()` to create my database tables, then I just need to make my SA aware of the models (ie by calling a function from celery that creates them using my application's declarative base)
- If I use alembic, I'm already maintaining my own revisions, so I would do the same as above, and use alembic to autogenerate the revision to create the celery-related models (if that's even needed: It's only needed when adding celery at a later stage, or when using alembic to create the initial database instead of letting SA's `create_all()` do this)

auvipy:
I will share more context here.

auvipy:
> Is this really something that should be on the celery side?
> 
>     * If I do not use alembic and just use `create_all()` to create my database tables, then I just need to make my SA aware of the models (ie by calling a function from celery that creates them using my application's declarative base)
> 
>     * If I use alembic, I'm already maintaining my own revisions, so I would do the same as above, and use alembic to autogenerate the revision to create the celery-related models (if that's even needed: It's only needed when adding celery at a later stage, or when using alembic to create the initial database instead of letting SA's `create_all()` do this)

we got a regression report and had to revert https://github.com/celery/celery/pull/8475#issuecomment-1703935320 so thought that adding migration is the safer bet. what do you think?

ThiefMaster:
I think DB changes should be clearly documented in the changelog. But executing them extremely dependent on the environment.

Note: This is assuming that you do not maintain your completely separate sqlalchemy environment. If you do, then migrations in celery make sense (because it's completely independent of the application's database usage). But if you defer this to the application's sqlalchemy environment, then IMHO what I mentioned in my previous comment applies.

auvipy:
is using multi db schema migrations a feasible option?

PR #8555, opened 2023-10-06
Add 🥲 (Smiling Face with Tear) to output.

Kaligule:
The output ("all alone") sounds a little bit emotional, so here is a way to emphasize that.

### DIFF
--- a/celery/worker/consumer/mingle.py
+++ b/celery/worker/consumer/mingle.py
@@ -46,7 +46,7 @@ class Mingle(bootsteps.StartStopStep):
              for nodename, reply in replies.items() if reply]
             info('mingle: sync complete')
         else:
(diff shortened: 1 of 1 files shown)


50-Course (commented):
Interesting PR 😄 , If this ever makes it to prod, can't for them reviews. By the way, I think we are doing just fine; the smiley can infer interpretations.

Nusnus:
If it was a personal project this could have been nice/cute, but unfortunately this is less consistent with the current tone so I prefer to not support the change.

That being said, if the community wants to push a tone change we can discuss about it first 🙂

50-Course:
> ...but unfortunately this is less consistent with the current tone so I prefer to not support the change.

The reason I said, we were doing just fine.

Issue #8517, opened 2023-09-18
`config_from_object` causes infinite recursion when given a nested module that does not exist


BraedonLeonard:
# Checklist

- [x] I have verified that the issue exists against the `main` branch of Celery.
- [ ] This has already been asked to the [discussions forum](https://github.com/celery/celery/discussions) first.
- [x] I have read the relevant section in the
  [contribution guide](https://docs.celeryq.dev/en/main/contributing.html#other-bugs)
  on reporting bugs.
- [x] I have checked the [issues list](https://github.com/celery/celery/issues?q=is%3Aissue+label%3A%22Issue+Type%3A+Bug+Report%22+-label%3A%22Category%3A+Documentation%22)
  for similar or identical bug reports.
- [x] I have checked the [pull requests list](https://github.com/celery/celery/pulls?q=is%3Apr+label%3A%22PR+Type%3A+Bugfix%22+-label%3A%22Category%3A+Documentation%22)
  for existing proposed fixes.
- [x] I have checked the [commit log](https://github.com/celery/celery/commits/main)
  to find out if the bug was already fixed in the main branch.
- [x] I have included all related issues and possible duplicate issues
  in this issue (If there are none, check this box anyway).

## Mandatory Debugging Information

- [x] I have included the output of ``celery -A proj report`` in the issue.
    (if you are not able to do this, then at least specify the Celery
     version affected).
- [x] I have verified that the issue exists against the `main` branch of Celery.
- [x] I have included the contents of ``pip freeze`` in the issue.
- [x] I have included all the versions of all the external dependencies required
  to reproduce this bug.

## Optional Debugging Information

- [ ] I have tried reproducing the issue on more than one Python version
  and/or implementation.
- [x] I have tried reproducing the issue on more than one message broker and/or
  result backend.
- [ ] I have tried reproducing the issue on more than one version of the message
  broker and/or result backend.
- [ ] I have tried reproducing the issue on more than one operating system.
- [ ] I have tried reproducing the issue on more than one workers pool.
- [ ] I have tried reproducing the issue with autoscaling, retries,
  ETA/Countdown & rate limits disabled.
- [ ] I have tried reproducing the issue after downgrading
  and/or upgrading Celery and its dependencies.

#### Related Issues

- None

#### Possible Duplicates

- None

## Environment & Settings

**Celery version**: 5.3.4 (emerald-rush)

If the line causing the bug is left in, then `celery -A app.tasks report` fails with the same recursion error, but otherwise the result is:

[...]

PR #8619, opened 2023-11-07
Fixed RecursionError caused by giving `config_from_object` nested mod…

frolenkov-nikita:
Fixed RecursionError caused by giving `config_from_object` nested module that does not exist (#8517)

## Description

The problem is described by new unit test: t/unit/app/test_app.py::test_config_form_object__module_attr_does_not_exist

When you provide a string with valid module name but invalid submodule (attr) to `app.config_from_object` the 3rd party util `kombu.utils.imports.symbol_by_name` would raise `AttributeError` which would be "handled" by `PendingConfiguration(UserDict)` and would cause infinite recursion. This happens during the "lazy initialization" of the `app.conf` when `app.conf` attr is accessed.

The fix is to catch `AttributeError` and raise `ModuleNotFoundError` which would be propagated correctly and would provide the real error message to the end user.

I don't feel very good re: this fix, I am new to this project, so your criticism is highly desired.

### DIFF
--- a/celery/app/base.py
+++ b/celery/app/base.py
@@ -975,7 +975,14 @@ class Celery:
             This is used by PendingConfiguration:
                 as soon as you access a key the configuration is read.
         """
-        conf = self._conf = self._load_config()
+        try:
+            conf = self._conf = self._load_config()
+        except AttributeError as e:
+            # AttributeError is not propagated, it is "handled" by
+            # PendingConfiguration parent class. This causes
+            # confusing RecursionError.
+            raise ModuleNotFoundError(*e.args)
+
         return conf
 
     def _load_config(self):
--- a/celery/app/base.py
+++ b/celery/app/base.py
@@ -977,11 +977,11 @@ class Celery:
         """
         try:
             conf = self._conf = self._load_config()
-        except AttributeError as e:
+        except AttributeError as err:
             # AttributeError is not propagated, it is "handled" by
             # PendingConfiguration parent class. This causes
             # confusing RecursionError.
-            raise ModuleNotFoundError(*e.args)
+            raise ModuleNotFoundError(*err.args) from err
 
         return conf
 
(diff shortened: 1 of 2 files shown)


t/unit/app/test_app.py
```diff
@@ -696,6 +696,13 @@ class Config:
             assert exc.args[0].startswith('task_default_delivery_mode')
             assert 'CELERY_DEFAULT_DELIVERY_MODE' in exc.args[0]
 
+    def test_config_form_object__module_attr_does_not_exist(self):
+        with pytest.raises(ModuleNotFoundError) as exc:
+            self.app.config_from_object(f'{__name__}.bar')
+            # the module must exist, but it should not have the config attr
+            assert self.app.conf.broker_url is None
```
Nusnus, line 703: Shouldn't this and the next assertion be outside the `with` scope?

Nusnus (commented):
Nice work!
Check out my comment

celery/app/base.py
```diff
@@ -975,7 +975,14 @@ def _finalize_pending_conf(self):
...
                 as soon as you access a key the configuration is read.
         """
-        conf = self._conf = self._load_config()
+        try:
+            conf = self._conf = self._load_config()
+        except AttributeError as e:
+            # AttributeError is not propagated, it is "handled" by
+            # PendingConfiguration parent class. This causes
+            # confusing RecursionError.
+            raise ModuleNotFoundError(*e.args)
```
Nusnus, line 984: You can use `raise from` e.g raise Module..() from e

celery/app/base.py
```diff
@@ -975,7 +975,14 @@ def _finalize_pending_conf(self):
             This is used by PendingConfiguration:
                 as soon as you access a key the configuration is read.
         """
-        conf = self._conf = self._load_config()
+        try:
+            conf = self._conf = self._load_config()
+        except AttributeError as e:
```
Nusnus, line 980: Not critical, but more common `err` than just `e`

frolenkov-nikita:
Thank you very much, I've incorporated your feedback.

t/unit/app/test_app.py
auvipy: is this resolved? @frolenkov-nikita

auvipy (commented):
seems better now. waiting for another round of review from Nusnus

auvipy (approved):

Nusnus (approved):
Sorry for the delay, good work!

Issue #8625, opened 2023-11-08
Task is requeued with compression header but the body is not compressed


mathieu-lemay:
# Checklist

- [x] I have verified that the issue exists against the `main` branch of Celery.
- [x] I have read the relevant section in the
  [contribution guide](https://docs.celeryq.dev/en/main/contributing.html#other-bugs)
  on reporting bugs.
- [x] I have checked the [issues list](https://github.com/celery/celery/issues?q=is%3Aissue+label%3A%22Issue+Type%3A+Bug+Report%22+-label%3A%22Category%3A+Documentation%22)
  for similar or identical bug reports.
- [x] I have checked the [pull requests list](https://github.com/celery/celery/pulls?q=is%3Apr+label%3A%22PR+Type%3A+Bugfix%22+-label%3A%22Category%3A+Documentation%22)
  for existing proposed fixes.
- [x] I have checked the [commit log](https://github.com/celery/celery/commits/main)
  to find out if the bug was already fixed in the main branch.
- [x] I have included all related issues and possible duplicate issues
  in this issue (If there are none, check this box anyway).

## Mandatory Debugging Information

- [x] I have included the output of ``celery -A proj report`` in the issue.
    (if you are not able to do this, then at least specify the Celery
     version affected).
- [x] I have verified that the issue exists against the `main` branch of Celery.
- [x] I have included the contents of ``pip freeze`` in the issue.
- [x] I have included all the versions of all the external dependencies required
  to reproduce this bug.

## Optional Debugging Information

- [ ] I have tried reproducing the issue on more than one Python version
  and/or implementation.
- [x] I have tried reproducing the issue on more than one message broker and/or
  result backend.
- [ ] I have tried reproducing the issue on more than one version of the message
  broker and/or result backend.
- [ ] I have tried reproducing the issue on more than one operating system.
- [ ] I have tried reproducing the issue on more than one workers pool.
- [ ] I have tried reproducing the issue with autoscaling, retries,
  ETA/Countdown & rate limits disabled.
- [x] I have tried reproducing the issue after downgrading
  and/or upgrading Celery and its dependencies.

#### Related Issues

- None

#### Possible Duplicates

- None

## Environment & Settings

**Celery version**: 5.3.4

<details>
<summary><b><code>celery report</code> Output:</b></summary>
<p>

```
software -> celery:5.3.4 (emerald-rush) kombu:5.3.3 py:3.11.5
            billiard:4.2.0 redis:4.6.0
[...]

PR #8633, opened 2023-11-11
Update task.py get_custom_headers missing 'compression' key

auvipy:
to address https://github.com/celery/celery/issues/8625


Nusnus:
Seems legit but can we add some tests maybe?

auvipy:
> Seems legit but can we add some tests maybe?

All modified and coverable lines are covered by tests ✅  but I'm not fully sure what test to add. have to check tomorrow

Nusnus (approved):

PR #8640, opened 2023-11-14
Added "Serverless" section to Redis doc (redis.rst)

Nusnus:
New documentation for serverless backend/broker option with Redis.

**Not to be merged before we're ready to merge v5.4 PRs**


auvipy:
we should tell them that we have experimental Kafka support as well and people are using it without much issues recently. I think we should document that. may be I / you should look into it later

Nusnus:
> we should tell them that we have experimental Kafka support as well and people are using it without much issues recently. I think we should document that. may be I / you should look into it later

Yes I'll definitely check this out - thank you, good catch!

PR #8632, opened 2023-11-10
Propagates more ImportErrors during autodiscovery

johnjameswhitman:
# Problem

The gist of the problem is that the current implementation of autodiscovering tasks will hide `ImportErrors` that originate outside of `importlib.import_module`.

For details see discussion in https://github.com/celery/celery/discussions/8620.

# Solution

Narrow the caught-exceptions to ones that:
1. Are `ModuleNotFoundError` (which is a subclass of `ImportError` that `importlib` seems to raise these days), and
2. Have a `name` attribute that matches the task being "guessed" (previously the code would default to the task's name if the exception lacked a `name` attribute).

Also, refactored the tests into what seem like the specific edge cases `find_related_module` is trying to handle.

PR #8660, opened 2023-11-22
Re-raise ModuleNotFoundError unless for guessed task

johnjameswhitman:
## Description

We just merged #8632, which should propagate import-errors that originate outside of the `importlib.import_module` call. After reading through the code again today, I realized my change may swallow some other import errors that should fail.

I've updated the integration test with clearer cases, and modified `celery.loaders.base.find_related_module` so that it only returns none if the exact task-module being guessed is what caused the `ModuleNotFoundError`.

### DIFF
--- a/celery/loaders/base.py
+++ b/celery/loaders/base.py
@@ -271,7 +271,7 @@ def find_related_module(package, related_name):
     except ModuleNotFoundError as e:
         import_exc_name = getattr(e, 'name', None)
         # If candidate does not exist, then return None.
-        if import_exc_name and module_name.startswith(import_exc_name):
+        if import_exc_name and module_name == import_exc_name:
             return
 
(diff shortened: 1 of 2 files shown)


Nusnus (approved):

celery/loaders/base.py
```diff
@@ -271,7 +271,7 @@ def find_related_module(package, related_name):
     except ModuleNotFoundError as e:
         import_exc_name = getattr(e, 'name', None)
         # If candidate does not exist, then return None.
-        if import_exc_name and module_name.startswith(import_exc_name):
+        if import_exc_name and module_name == import_exc_name:
```
auvipy, line 274: should we adjust the unit tests as well? with this new change?

auvipy (approved):
except the unit test the Int tests seems good

PR #8719, opened 2023-12-12
[Smoke Tests only] Using pytest-xdist config: --dist=loadscope

Nusnus:
https://github.com/pytest-dev/pytest-xdist/blob/master/CHANGELOG.rst#pytest-xdist-350-2023-11-21
> pytest-xdist 3.5.0 (2023-11-21)
> 
> Features
> 
> https://github.com/pytest-dev/pytest-xdist/issues/632: --dist=loadscope now sorts scopes by number of tests to assign largest scopes early -- in many cases this should improve overall test session running time, as there is less chance of a large scope being left to be processed near the end of the session, leaving other workers idle.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -11,6 +11,7 @@ on:
         - '**.txt'
         - '.github/workflows/python-package.yml'
         - '**.toml'
+        - "tox.ini"
   pull_request:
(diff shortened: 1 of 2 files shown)


Nusnus:
3.11 Env is broken due to GitHub.
Yesterday was all fine.

Nusnus:
3.11 Env is broken regardless of this PR

PR #8743, opened 2023-12-28
Remove `new` adjective in docs

Viicos:
Django 1.4 is not that _new_ nowadays.

Would be also nice to add Django docs inventory to intersphinx mapping so that these settings can be referenced?


auvipy (approved):

auvipy (commented):
yes that would be a good way

Viicos:
I'll add it in a separate PR if possible

PR #8750, opened 2024-01-03
add type annotation to celery/utils/iso8601.py

em1le:
Relates to #7394

### DIFF
--- a/celery/utils/iso8601.py
+++ b/celery/utils/iso8601.py
@@ -50,7 +50,7 @@ TIMEZONE_REGEX = re.compile(
 )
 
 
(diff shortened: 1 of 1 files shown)


Nusnus (approved):

celery/utils/iso8601.py
```diff
@@ -50,7 +50,7 @@
 )
 
 
-def parse_iso8601(datestring):
+def parse_iso8601(datestring: str) -> str:
```
lelit, line 53: Why `str`? Doesn't it return a `datetime` instance?

PR #8753, opened 2024-01-04
Updated concurrency docs page.

robotrapta:
## Description

This is a docs page I wish had existed when I started using celery a couple years ago.  The current page seems to imply your only two choices are `gevent` and `eventlet` and that led me down a path of unnecessary operational pain.

This page is far from comprehensive, but I don't think it's misleading like the current page is.  Also I suspect it's probably a good idea to tell people to stick with prefork unless they have reason not to, especially since other models lack certain features - a fact which AFAICT isn't currently documented at all.

### DIFF
--- a/docs/userguide/concurrency/index.rst
+++ b/docs/userguide/concurrency/index.rst
@@ -7,8 +7,40 @@
 :Release: |version|
 :Date: |today|
 
+Concurrency in Celery enables the parallel execution of tasks. The default
+model, `prefork`, is well-suited for many scenarios and generally recommended
+for most users.  In fact, switching to another mode will silently disable
+certain features like `soft_timeout` and `max_tasks_per_child`.
+
+This page gives a quick overview of the available options which you can pick
+between using the `--pool` option when starting the worker.
+
+Overview of Concurrency Options
+-------------------------------
+
+- `prefork`: The default option, ideal for CPU-bound tasks and most use cases.
+  It is robust and recommended unless there's a specific need for another model.
+- `eventlet` and `gevent`: Designed for IO-bound tasks, these models use
+  greenlets for high concurrency. Note that certain features, like `soft_timeout`,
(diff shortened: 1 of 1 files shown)


auvipy (commented):
this looks good to me

docs/userguide/concurrency/index.rst
```diff
@@ -7,8 +7,39 @@
...
+  greenlets for high concurrency. Note that certain features, like `soft_timeout`,
+  are not available in these modes.  These have detailed documentation pages
+  linked below.
+- `solo`: Executes tasks sequentially in the main thread.
+- `threads`: Utilizes threading for concurrency, available if the
+  `concurrent.futures` module is present.
+- `custom`: Enables specifying a custom worker pool implementation through
+  environment variables.
+
+For a detailed look at these options, there is an excellent overview at [Celery Worker Pools
```
Nusnus, line 33: The article seems very detailed and of high quality, but it would be incorrect to include a personal link in the official doc. Please remove this part.

Nusnus: That being said, it would be **highly appreciated** if you incorporate the knowledge of that article into the official documentation. I would personally be grateful and will help speed up the review process to make it happen.

Nusnus (changes requested):
Very nice addition! Thank you very much!
Please review my comment and tag me for any assistance - it would be an impactful effort improving this aspect of the documentation!

docs/userguide/concurrency/index.rst
robotrapta: I think I understand why there's such a policy in the internal docs, and so removed the link, but it does make the project a lot harder to document.

I'm not sure when I'm gonna have time to write a better summary of that doc to include here.

Nusnus (approved):
Adding a link for the good guide for ref, at least in the PR:
https://celery.school/celery-worker-pools

PR #8838, opened 2024-02-09
Support moto 5.0

s-t-e-v-e-n-k:
## Description

moto 5.0 has been released, and the major change is to pull all of the seperate mock calls into one -- mock_aws. Continue to support moto 4, since it's easy to do so.


Nusnus (approved):
Great fix - thank you very much!

Issue #8840, opened 2024-02-12
`AttributeError: 'dict' object has no attribute 'clone'` in `group.link_error()`


murrple-1:
# Checklist

- [x] I have verified that the issue exists against the `main` branch of Celery.
- [ ] This has already been asked to the [discussions forum](https://github.com/celery/celery/discussions) first.
- [x] I have read the relevant section in the
  [contribution guide](https://docs.celeryq.dev/en/main/contributing.html#other-bugs)
  on reporting bugs.
- [x] I have checked the [issues list](https://github.com/celery/celery/issues?q=is%3Aissue+label%3A%22Issue+Type%3A+Bug+Report%22+-label%3A%22Category%3A+Documentation%22)
  for similar or identical bug reports.
- [x] I have checked the [pull requests list](https://github.com/celery/celery/pulls?q=is%3Apr+label%3A%22PR+Type%3A+Bugfix%22+-label%3A%22Category%3A+Documentation%22)
  for existing proposed fixes.
- [x] I have checked the [commit log](https://github.com/celery/celery/commits/main)
  to find out if the bug was already fixed in the main branch.
- [x] I have included all related issues and possible duplicate issues
  in this issue (If there are none, check this box anyway).

## Mandatory Debugging Information

- [ ] I have included the output of ``celery -A proj report`` in the issue.
    (if you are not able to do this, then at least specify the Celery
     version affected).
- [x] I have verified that the issue exists against the `main` branch of Celery.
- [ ] I have included the contents of ``pip freeze`` in the issue.
- [ ] I have included all the versions of all the external dependencies required
  to reproduce this bug.

## Optional Debugging Information

- [ ] I have tried reproducing the issue on more than one Python version
  and/or implementation.
- [ ] I have tried reproducing the issue on more than one message broker and/or
  result backend.
- [ ] I have tried reproducing the issue on more than one version of the message
  broker and/or result backend.
- [ ] I have tried reproducing the issue on more than one operating system.
- [ ] I have tried reproducing the issue on more than one workers pool.
- [ ] I have tried reproducing the issue with autoscaling, retries,
  ETA/Countdown & rate limits disabled.
- [x] I have tried reproducing the issue after downgrading
  and/or upgrading Celery and its dependencies.

#### Related Issues

- https://github.com/celery/celery/issues/8678

#### Possible Duplicates

- None

## Environment & Settings

**Celery version**: 5.4.0rc1 (opalescent)

<details>
[...]

Issue #8678, opened 2023-11-28
task with an errback throws `AttributeError` when replaced with a chord, when `task_allow_error_cb_on_chord_header` is set


robertgalloway-nobleai:
# Checklist

- [x] I have verified that the issue exists against the `main` branch of Celery.
- [ ] This has already been asked to the [discussions forum](https://github.com/celery/celery/discussions) first.
- [x] I have read the relevant section in the
  [contribution guide](https://docs.celeryq.dev/en/main/contributing.html#other-bugs)
  on reporting bugs.
- [x] I have checked the [issues list](https://github.com/celery/celery/issues?q=is%3Aissue+label%3A%22Issue+Type%3A+Bug+Report%22+-label%3A%22Category%3A+Documentation%22)
  for similar or identical bug reports.
- [x] I have checked the [pull requests list](https://github.com/celery/celery/pulls?q=is%3Apr+label%3A%22PR+Type%3A+Bugfix%22+-label%3A%22Category%3A+Documentation%22)
  for existing proposed fixes.
- [x] I have checked the [commit log](https://github.com/celery/celery/commits/main)
  to find out if the bug was already fixed in the main branch.
- [x] I have included all related issues and possible duplicate issues
  in this issue (If there are none, check this box anyway).

## Mandatory Debugging Information

- [x] I have included the output of ``celery -A proj report`` in the issue.
    (if you are not able to do this, then at least specify the Celery
     version affected).
- [x] I have verified that the issue exists against the `main` branch of Celery.
- [x] I have included the contents of ``pip freeze`` in the issue.
- [x] I have included all the versions of all the external dependencies required
  to reproduce this bug.

## Optional Debugging Information

- [ ] I have tried reproducing the issue on more than one Python version
  and/or implementation.
- [ ] I have tried reproducing the issue on more than one message broker and/or
  result backend.
- [ ] I have tried reproducing the issue on more than one version of the message
  broker and/or result backend.
- [ ] I have tried reproducing the issue on more than one operating system.
- [x] I have tried reproducing the issue on more than one workers pool.
- [ ] I have tried reproducing the issue with autoscaling, retries,
  ETA/Countdown & rate limits disabled.
- [x] I have tried reproducing the issue after downgrading
  and/or upgrading Celery and its dependencies.

#### Related Issues

https://github.com/celery/celery/issues/8456

#### Possible Duplicates

- None

## Environment & Settings

**Celery version**: 5.3.6 (emerald-rush)

<details>
[...]

PR #8841, opened 2024-02-12
Another fix for `link_error` signatures being `dict`s instead of `Signature`s

murrple-1:
Related to https://github.com/celery/celery/issues/8678

Fix for https://github.com/celery/celery/issues/8840

## Description

The most recent release ([v5.4.0rc1](https://github.com/celery/celery/releases/tag/v5.4.0rc1)) includes https://github.com/celery/celery/pull/8702, which fixes an issue where the signature can be a `dict` in the case of a `chord.link_error()`. I don't know enough about Celery to know what the causes the different code paths, but the fix is almost exactly the same as the earlier PR, just for `group.link_error()`.

### DIFF
--- a/celery/canvas.py
+++ b/celery/canvas.py
@@ -722,7 +722,7 @@ class Signature(dict):
         """
         return self.append_to_list_option('link', callback)
 
-    def link_error(self, errback):
+    (self, errback):
         """Add callback task to be applied on error in task execution.
 
         Returns:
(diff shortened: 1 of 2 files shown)


Nusnus (changes requested):
Thanks for the fix!
Can you please add a unit test to verify it?

Nusnus (approved):
Thanks!

PR #8939, opened 2024-04-02
TMP TEST 

Nusnus:
!!! DO NOT MERGE !!!

Testing if something works in the CI or not due to using Mac as a local dev machine where it does work.


celery/worker/components.py
```diff
@@ -236,5 +236,33 @@ def create(self, w):
             worker_options=w.options,
             disable_rate_limits=w.disable_rate_limits,
             prefetch_multiplier=w.prefetch_multiplier,
-        )
+            url=w.app.conf.broker_url.split(';')[0]
+        ),
+            self.instantiate(
```
anilaratna2, line 241: registers to event loop

celery/worker/components.py
```diff
@@ -236,5 +236,33 @@ def create(self, w):
...
+                worker_options=w.options,
+                disable_rate_limits=w.disable_rate_limits,
+                prefetch_multiplier=w.prefetch_multiplier,
+                url=w.app.conf.broker_url.split(';')[1]
+            )]
         return c
+
+    def start(self, parent):
+        if self.obj:
+            if isinstance(self.obj, list):
```
anilaratna2, line 261: self.obj == c on line 257

celery/worker/worker.py
```diff
@@ -158,7 +158,7 @@ def on_close(self):
 
     def on_stopped(self):
         self.timer.stop()
-        self.consumer.shutdown()
+        self.consumer[0].shutdown()
```
anilaratna2, line 161: just ot make it compile

celery/worker/components.py
```diff
@@ -222,7 +222,7 @@ def create(self, w):
             prefetch_count = max(w.max_concurrency, 1) * w.prefetch_multiplier
         else:
             prefetch_count = w.concurrency * w.prefetch_multiplier
-        c = w.consumer = self.instantiate(
+        c = w.consumer = [self.instantiate(
```
anilaratna2, line 225: we also discussed maybe to test with consumermixin having 2 consumers with different URLs and then debugging. its not used in a worker but is a smaller case to find race conditions and fix kombu.hub and other components that may be failing eg amqp transport frame

anilaratna2 (commented):
.

PR #8964, opened 2024-04-14
TMP TEST 2

Nusnus:
!!! DO NOT MERGE !!!

Testing if something works in the CI or not due to using Mac as a local dev machine where it does work.


celery/worker/components.py
```diff
@@ -95,6 +95,47 @@ def _patch_thread_primitives(self, w):
...
+        w.hub2 = None
+        super().__init__(w, **kwargs)
+
+    def include_if(self, w):
+        return w.use_eventloop
+
+    def create(self, w):
+        if w.hub2 is None:
+            required_hub = getattr(w._conninfo, 'requires_hub', None)
+            w.hub2 = set_event_loop((
```
Nusnus, line 113: Shouldn’t use the `set_event_loop` because it uses the global. It needs to create a new instance.

Nusnus: It needs to create a hub per consumer (e.g. from `w.consumers`)

Nusnus: Integrate logic into Hub class:
1. If `w.consumer` is NOT none, go as usual.
2. if `w.consumers` is NOT none, apply same logic for each consumer (to get its own hub (loop) instance).

celery/worker/components.py
```diff
@@ -137,7 +178,8 @@ def create(self, w):
         max_restarts = None
         if w.app.conf.worker_pool in GREEN_POOLS:  # pragma: no cover
             warnings.warn(UserWarning(W_POOL_SETTING))
-        threaded = not w.use_eventloop or IS_WINDOWS
```
Nusnus, line 140: `not w.use_eventloop or IS_WINDOWS or app.new_flag is not None`

celery/worker/components.py
```diff
@@ -215,14 +257,16 @@ def create(self, w):
 class Consumer(bootsteps.StartStopStep):
     """Bootstep starting the Consumer blueprint."""
 
+    requires = ('celery.worker.components:Hub2',)
```
Nusnus, line 260: This will not be needed then

celery/worker/components.py
```diff
@@ -215,14 +257,16 @@ def create(self, w):
...
+
     last = True
 
     def create(self, w):
         if w.max_concurrency:
             prefetch_count = max(w.max_concurrency, 1) * w.prefetch_multiplier
         else:
             prefetch_count = w.concurrency * w.prefetch_multiplier
-        c = w.consumer = self.instantiate(
+        c = w.consumer = [self.instantiate(
```
Nusnus, line 269: If new flag True:
1. `w.consumer` = None
2. `w.consumers` = List

Else:
1. `w.consumer` = Normal usage
2. `w.consumers` = None

celery/worker/consumer/consumer.py
```diff
@@ -181,7 +181,8 @@ def __init__(self, on_task_request,
                  pool=None, app=None,
                  timer=None, controller=None, hub=None, amqheartbeat=None,
                  worker_options=None, disable_rate_limits=False,
-                 initial_prefetch_count=2, prefetch_multiplier=1, **kwargs):
+                 initial_prefetch_count=2, prefetch_multiplier=1, url=None, **kwargs):
```
Nusnus, line 184: Add new parameter to allow passing the url directly when instantiating a new Consumer but use None for the default behavior

celery/worker/worker.py
```diff
@@ -158,7 +158,11 @@ def on_close(self):
 
     def on_stopped(self):
         self.timer.stop()
-        self.consumer.shutdown()
+        if isinstance(self.consumer, list):
```
Nusnus, line 161: Instead of adjusting the `w.consumer` to be a list, use the new flag to choose which consumer to act upon

PR #8965, opened 2024-04-14
Allow local dynamodb to be installed on another host than localhost

peerjakobsen:

## Description
The previous version of dynamodb backend only allowed to have local dynamodb on your local dev machine. I changed it so that you can have local dynamodb running on an different machine and use an ip address instead of localhost:8000 . I use 127.0.0.1 as ip address in the unit test but it works with any ip address.


celery/backends/dynamodb.py
```diff
@@ -153,6 +154,14 @@ def __init__(self, url=None, table_name=None, *args, **kwargs):
                 secret_access_key=aws_secret_access_key
             )
 
+    @staticmethod
+    def _is_valid_ip(ip):
```
Nusnus, line 158: I wonder if that's the best way to check it.. or if it's worth delaying for 🤔

Anyway, should not be merged before v5.4 release. We're in RC2 at the moment.

Nusnus:
The docker-docs failure is not related btw

celery/backends/dynamodb.py
peerjakobsen: I guess I could do 
pattern = re.compile(r'^[a-z]{2}-[a-z]+-[0-9]{1}$')
    if pattern.match(input_string):
      do real dynamodb
    else:
      do local dynamodb
if we trust the AWS region names keeps following the current format of xx-yyyy-n
That would also allow to use non ip hosts names.

thedrow (approved):

Issue #8997, opened 2024-05-01
celery killing itself?


lewijw:
### Discussed in https://github.com/celery/celery/discussions/8994

<div type='discussions-op-text'>

<sup>Originally posted by **lewijw** April 30, 2024</sup>
I'm running celery in a pod in k8s, and I am seeing a situation where it starts as PID 1, writes the pid file, and seems to crash for some reason.  Then, k8s restarts the process; it starts again as PID 1; but since the pid file has a 1 in it, I am wondering if it is killing itself?   I'm not a python expert, but this code seems to need a comparison of os.getpid() with the contents of the pid file before killing the process.
https://github.com/celery/celery/blob/main/celery/platforms.py#L191</div>

baseplate-admin:
I think this can be closed?

Due to merging of 

- https://github.com/celery/celery/pull/8998

PR #8998, opened 2024-05-02
don't kill if pid same as file (#8997)

lewijw:

## Description

When celery runs in a pod in k8s, it runs as PID 1.  If it crashes, k8s restarts the process but since the pid file has a 1 in it, it kills itself.  With this change, if the pid in the file is the same as the pid of the process, it will not kill the process.


auvipy (changes requested):
this will need some unit tests to avoid any regression

t/unit/utils/test_platforms.py
```diff
@@ -689,6 +689,15 @@ def test_remove_if_stale_no_pidfile(self):
         assert p.remove_if_stale()
         p.remove.assert_called_with()
```
auvipy, line 691: is the change of file permission 100644 → 100755  needed here? or it was done by accident?

auvipy (changes requested):

t/unit/utils/test_platforms.py
lewijw: Mistake corrected.

auvipy (approved):

PR #9007, opened 2024-05-06
don't kill if pid same as file (#8997) (#8998)

lewijw:
The pid file needs to be deleted.

### DIFF
--- a/celery/platforms.py
+++ b/celery/platforms.py
@@ -188,6 +188,7 @@ class Pidfile:
             return True
         if pid == os.getpid():
             # this can be common in k8s pod with PID of 1 - don't kill
+            self.remove()
             return True
 
         try:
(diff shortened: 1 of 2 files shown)


auvipy (commented):
can you please better explain the new change? do you think an integration test is possible to add for this part?

lewijw:
@auvipy and @Nusnus, I understand your concern, and I am not sure why I did not see this issue earlier because as soon as remove_if_stale returns, the acquire() function is called which tries to write the pid file.  So, why I am just now getting a "file exists" failure is a bit confusing to me.  However, I do believe that a revert would not bring the codebase to a better state because the code kills the process that it is running in.  I am also very certain that there will be no changes after this commit.
As far as creating an integrated test, that would be very difficult because you have to make sure the process id remains the same between runs, and as far as I know, the only way to make that happen is to run the process in a k8s pod.

lewijw:
Note also that every time remove_if_stale returns True, it removes the pid file first.  I was reluctant to do the same at first until I was certain it was required.  Now it looks like it is required.

auvipy (approved):

PR #9010, opened 2024-05-07
SecuritySerializer: ensure pack separator will not be conflicted with serialized fields

shirsa:

## Description

This PR fixes a bug in `SecuritySerializer` module, where the serialized data may contain the value of the `sep` used to distinguish between the serialized fields. In that case, the fields split will be incorrect and the signature verify will fail.
Here, I encode binary values to base64, to prevent the data will contain the  separator

### DIFF
--- a/celery/security/serialization.py
+++ b/celery/security/serialization.py
@@ -55,37 +55,20 @@ class SecureSerializer:
     def _pack(self, body, content_type, content_encoding, signer, signature,
               sep=str_to_bytes('\x00\x01')):
         fields = sep.join(
-            ensure_bytes(s) for s in [signer, signature, content_type,
-                                      content_encoding, body]
+            ensure_bytes(s) for s in [b64encode(signer), b64encode(signature),
+                                      content_type, content_encoding, body]
         )
         return b64encode(fields)
 
     def _unpack(self, payload, sep=str_to_bytes('\x00\x01')):
         raw_payload = b64decode(ensure_bytes(payload))
-        first_sep = raw_payload.find(sep)
-
-        signer = raw_payload[:first_sep]
-        signer_cert = self._cert_store[signer]
-
-        # shift 3 bits right to get signature length
-        # 2048bit rsa key has a signature length of 256
-        # 4096bit rsa key has a signature length of 512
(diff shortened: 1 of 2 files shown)


shirsa:
@auvipy @Nusnus 
Can you check please?
Thanks :)

celery/security/serialization.py
```diff
@@ -55,37 +55,20 @@ def deserialize(self, data):
...
-            ensure_bytes(s) for s in [signer, signature, content_type,
-                                      content_encoding, body]
+            ensure_bytes(s) for s in [b64encode(signer), b64encode(signature),
+                                      content_type, content_encoding, body]
         )
         return b64encode(fields)
 
     def _unpack(self, payload, sep=str_to_bytes('\x00\x01')):
         raw_payload = b64decode(ensure_bytes(payload))
-        first_sep = raw_payload.find(sep)
```
auvipy, line 65: I would like to know bit more detail about the changes proposed, care to provide bit more?

shirsa: Sure @auvipy . This is the logic that we have today for unpacking the serialized data:
1. extract signer and `signature_start_index` using first appearance of `sep`
2. get signature bytes according to `signature_start_index` and signature length. Note, that we assume. here that the signature is decoded as-is in the raw_payload, so we can use the signature length to calculate `signature_end_index`
3. extract all other fields from remaining bytes using split by `sep`

The proposed solution simplifies this process, by assuming that the `sep` bytes can be present in the payload only as separators (and not randomly be as part of signature bytes, for example). Since we guarantee that when we encode the binary data using `base64` (which will not contain `\x00\x01`).
With that assumption, we can just split all fields according to `sep` and get all fields at once.
The only exception here is the `body` itself, which may contain the `sep` bytes inside. Since the body is encoded to the end of the payload, we can limit the number of splits according to `sep`, to prevent more splits in that case.
(We can also encode the body as base64 and then we can assure `sep` will not be in `body` as well, but I like this better since it is more performant).

shirsa:
@auvipy I added a small commit

auvipy (approved):

PR #9037, opened 2024-05-21
Fix certificate validity check

SPKorhonen:
Use 'not_valid_after_utc' instead of 'not_valid_after' when checking for certificate validity to prevent errors with avare/naive timedates

Fixes error:

File "<>\site-packages\celery\security\certificate.py", line 46, in has_expired
    return datetime.datetime.now(datetime.timezone.utc) >= self._cert.not_valid_after
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: can't compare offset-naive and offset-aware datetimes


auvipy (changes requested):
can you please update the relevant tests as well?

SPKorhonen:
The tests _test_has_expired_mock_ and _test_has_not_expired_mock_ which are the two tests using _not_valid_after_call have been changed to use the _not_valid_after_call_utc_ variant instead of original

auvipy (approved):

PR #8491, opened 2023-09-08
Add a Task class specialised for Django

browniebroke:
*Note*: Before submitting this pull request, please review our [contributing
guidelines](https://docs.celeryq.dev/en/main/contributing.html).

## Description

Add a task specialised to use with Django. Offers an API to call `.delay()` on tasks at the end of the database transaction. 

- [x] Add task to the codebase
- [x] Add docs
- [x] Add tests 

## Rationale

A common pitfall using Celery with Django is to call `.delay()` in a middle of a database transaction and have the task start executing before the transaction is committed. I think it's partly due to the fact that doing the call properly is a bit cumbersome. 

This PR adds a specialised task that aim at offering a nicer API for it. I've added the initial code and some docs to explain how it would be used, but before going further, I'd like to know whether this would be a welcome addition or if it's considered too out of scope.

PR #9038, opened 2024-05-22
Fix detection of custom task set as class attribute with Django

browniebroke:

## Description

Fix for feature added in 

- https://github.com/celery/celery/pull/8491

As reported in [a comment on the pull request
](https://github.com/celery/celery/pull/8491#issuecomment-2119191201), this feature breaks in case the user customised their task via a class attribute, and have Django installed:

```python
class CeleryApp(Celery):
    registry_cls = 'tenant_schemas_celery.registry:TenantTaskRegistry'
    task_cls = 'tenant_schemas_celery.task:TenantTask'
```

- [x] Add a test to cover this use case
- [x] When such task is defined, don't use the built-in `DjangoTask`

### DIFF
--- a/celery/app/base.py
+++ b/celery/app/base.py
@@ -240,7 +240,12 @@ class Celery:
         self.loader_cls = loader or self._get_default_loader()
         self.log_cls = log or self.log_cls
         self.control_cls = control or self.control_cls
-        self._custom_task_cls_used = bool(task_cls)
+        self._custom_task_cls_used = (
+            # Custom task class provided as argument
+            bool(task_cls)
+            # subclass of Celery with a task_cls attribute
+            or self.__class__ is not Celery and hasattr(self.__class__, 'task_cls')
+        )
         self.task_cls = task_cls or self.task_cls
         self.set_as_current = set_as_current
         self.registry_cls = symbol_by_name(self.registry_cls)
--- a/t/unit/fixups/test_django.py
+++ b/t/unit/fixups/test_django.py
@@ -103,7 +103,7 @@ class test_DjangoFixup(FixupCase):
             self.sigs.worker_init.connect.assert_called_with(f.on_worker_init)
             assert self.app.loader.now == f.now
 
-            # Specialized Task class is used
+            # Specialized DjangoTask class is used
             assert self.app.task_cls == 'celery.contrib.django.task:DjangoTask'
             from celery.contrib.django.task import DjangoTask
             assert issubclass(f.app.Task, DjangoTask)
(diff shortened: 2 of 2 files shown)


celery/app/base.py
```diff
@@ -240,7 +240,12 @@ def __init__(self, main=None, loader=None, backend=None,
         self.loader_cls = loader or self._get_default_loader()
         self.log_cls = log or self.log_cls
         self.control_cls = control or self.control_cls
-        self._custom_task_cls_used = bool(task_cls)
+        self._custom_task_cls_used = (
```
maciej-gol, line 243: What about dropping the whole attribute here, and instead doing the following comparison in the fixup code:

```python
from celery.app import Celery
if self.task_cls == Celery.task_cls:
    <task cls wasn't overriden, proceed with task override>
```
?

This should, I think, significantly simplify the code.

browniebroke: What if the user wants to override Celery to not use the DjangoTask, but the default Calary task?

maciej-gol: There is no functional difference between `Task` and `DjangoTask` (DT is a strict functional superset of T), so the only cases where this matters are subclass/instance checks which already sound shady and trying to support these sounds like a can of worms, hm? 

If they _really_ want `Task` and not `DT`, then maybe they are also considering disabling the whole django fixup.

browniebroke: Yes it _should_ be fine, and I'm probably being extra careful. 

I'm trying to make a change that has the minimum amount of side effects, as any side effects could be considered a breaking change. With a project the size of Celery, it's hard to know what folks are using and how... As your own use case proves (I didn't forsee it coming).

maciej-gol: I agree that its best to avoid breaking changes. Having said that, reducing complexity in a project the size of Celery is something I would also keep in consideration :) Especially when: 1) we don't know at all if there are people that want to opt-out from the DT injection 2) there is an easy workaround, as described above.

I believe you will choose whatever solution is best for the project, and I don't have any strong opinions here (nor I have any time pressure as I have fixed this issue on my side).

auvipy: for v.5.4.x we need to keep the backward compatibility. in future major versions, we can think of improving it and deprecating un needed API.

auvipy:
Brian, do you intend to do more changes here?

browniebroke:
Assuming you meant Bruno (me) 😁, no I don't intend to add any more changes. Should be good to go from my perspective.

auvipy:
Sorry sorry! less attention to detail problem! :D

auvipy (approved):

PR #9089, opened 2024-06-21
Don't override explicitly set configuration with env variables

### LINKED ISSUE #4284: Environment variable for Broker URL is prioritized over app settings

soceanainn:

## Description

Fixes #4284.

To give a concrete example where this is causing issues:

1. My organisation runs a large Python monolithic backend with many Celery apps/workers (which currently all sharing a single RabbitMQ broker).
2. We recently started work to split our most critical use cases into separate isolated Rabbit MQ brokers for better isolation.
3. We came across this issue because setting specific broker URLs for specific Celery apps in our code base isn't working.  Our backend is sending everything to the value configured in `CELERY_BROKER_URL` env variable, instead of the specific value configured for the Celery app.

I believe that this behaviour is extremely surprising for developers (that explicitly defined settings are being overwritten by env variables).


celery/app/utils.py
```diff
@@ -84,31 +84,31 @@ def __init__(self, *args, deprecated_settings=None, **kwargs):
...
             self.broker_url
         )
 
     @property
     def broker_write_url(self):
         return (
-            os.environ.get('CELERY_BROKER_WRITE_URL') or
             self.get('broker_write_url') or
+            os.environ.get('CELERY_BROKER_WRITE_URL') or
             self.broker_url
```
soceanainn, line 97: This behaviour for `broker_read_url` / `broker_write_url` may be surprising, as I think it's less clear cut whether an explicitly defined `broker_url` should be used ahead of specific env vars for read/write URLs.

In this case, I think that env vars for read/write URLs are probably more specific than explicit setting for `broker_url`, so I set the priority accordingly. 

Happy to adjust if necessary

soceanainn:
@auvipy tagging you here as you had volunteered to review any contributions to the original issue (although I understand that was 18 months ago)

auvipy:
All the tests are failing. Can you check?

SebastianDix:
So it seems that test_custom_remote_command[control-custom_cmd0] FAILED

and then it continued running test_custom_remote_command[control-custom_cmd1] until the 30 minute timeout

celery/app/utils.py
SebastianDix: @soceanainn Thank you for your contribution. 

I think this might however break a lot of Django projects with stray environment variables, so this should be announced in some change log I think, not sure if it needs to be part of this PR.

And also we should make this super explicit in the docs.

soceanainn:
Okay I see the issue in tests now:
1. Celery CLI is actually configured by explicitly setting these env variables: https://github.com/celery/celery/blob/1aff856ea37a477639ef2c8883f121d6670f72e0/celery/bin/celery.py#L152-L162
2. This is causing issues in tests due to my change, since we pass `broker` [CLI option in the test](https://github.com/celery/celery/blob/582e169b9d1750fd416cfa94ecfd26f43568f7b2/t/unit/bin/test_control.py), but it's using an explicitly configured AMQP broker_url from [t.integration.test_worker_config](https://github.com/celery/celery/blob/main/t/integration/test_worker_config.py) when defining the [test app](https://github.com/celery/celery/blob/main/t/unit/bin/proj/app_with_custom_cmds.py):

auvipy (changes requested):
would you mind adding new tests for the proposed changes? My understanding so far is that you are trying to change the precedence mostly?

Nusnus:
@thedrow can you give this a look please?

t/unit/bin/test_control.py
```diff
@@ -12,14 +10,6 @@
 _INSPECT_OPTIONS = ['--timeout', '0']  # Avoid waiting for the zero workers to reply
 
 
-@pytest.fixture(autouse=True)
```
auvipy, line 15: why removing this test? and why no new one?

auvipy (changes requested):

auvipy (changes requested):
can't add / consider this without additional tests

PR #9121, opened 2024-07-07
Added support for Quorum Queues

### LINKED ISSUE #6067: support for quorum queues
# Checklist

* [x] I have checked the [issues list](https://github.com/celery/celery/issues?utf8=%E2%9C%93&q=is%3Aissue+label%3A%22Issue+Type%3A+Feature+Request%22+)
* [ ] I have checked the [pull requests list](https://github.com/celery/celery/pulls?utf8=%E2%9C%93&q=is%3Apr+label%3A%22PR+Type%3A+Feature%22+)
  master branch.

#### Related Issues
* [py-amqp](https://github.com/celery/py-amqp/issues/302)

#### Possible Duplicates
* None

# Brief Summary

support for quorum queues.

# Design

None

## Diagrams

N/A

## Alternatives

None

Nusnus:
Fixes #6067
1. Added implementation without handling ETAs for now.
2. Includes tests.
3. With documentation + new example.

The bash scripts of the example are used to set up a local RabbitMQ 3-nodes cluster to run the example against.

Additional official RabbitMQ documentation related to the changes of this PR:
1. [Quorum Queues](https://www.rabbitmq.com/docs/quorum-queues)
1.1. [Global QoS](https://www.rabbitmq.com/docs/quorum-queues#global-qos)
2. [Consumer Prefetch](https://www.rabbitmq.com/docs/consumer-prefetch)
3. [Consumer Acknowledgements and Publisher Confirms](https://www.rabbitmq.com/docs/confirms)
3.1. [Channel Prefetch Setting (QoS)](https://www.rabbitmq.com/docs/confirms#channel-qos-prefetch)
4. [Migrate your RabbitMQ Mirrored Classic Queues to Quorum Queues](https://www.rabbitmq.com/docs/migrate-mcq-to-qq)

### DIFF
--- a/celery/app/defaults.py
+++ b/celery/app/defaults.py
@@ -346,6 +346,7 @@ NAMESPACES = Namespace(
         task_log_format=Option(DEFAULT_TASK_LOG_FMT),
         timer=Option(type='string'),
         timer_precision=Option(1.0, type='float'),
+        detect_quorum_queues=Option(True, type='bool'),
     ),
 )
 
--- a/celery/app/defaults.py
+++ b/celery/app/defaults.py
@@ -261,6 +261,7 @@ NAMESPACES = Namespace(
         inherit_parent_priority=Option(False, type='bool'),
         default_delivery_mode=Option(2, type='string'),
         default_queue=Option('celery'),
+        default_queue_type=Option('classic', type='string'),
         default_exchange=Option(None, type='string'),  # taken from queue
         default_exchange_type=Option('direct'),
         default_routing_key=Option(None, type='string'),  # taken from queue
--- a/celery/app/amqp.py
+++ b/celery/app/amqp.py
@@ -249,9 +249,13 @@ class AMQP:
         if max_priority is None:
             max_priority = conf.task_queue_max_priority
         if not queues and conf.task_default_queue:
+            queue_arguments = None
+            if conf.task_default_queue_type == 'quorum':
+                queue_arguments = {'x-queue-type': 'quorum'}
             queues = (Queue(conf.task_default_queue,
                             exchange=self.default_exchange,
-                            routing_key=default_routing_key),)
+                            routing_key=default_routing_key,
+                            queue_arguments=queue_arguments),)
         autoexchange = (self.autoexchange if autoexchange is None
                         else autoexchange)
         return self.queues_cls(
--- a/celery/worker/consumer/tasks.py
+++ b/celery/worker/consumer/tasks.py
@@ -1,7 +1,13 @@
 """Worker Task Consumer Bootstep."""
+
+from __future__ import annotations
+
+import warnings
+
 from kombu.common import QoS, ignore_errors
 
 from celery import bootsteps
+from celery.exceptions import CeleryWarning
 from celery.utils.log import get_logger
 
 from .mingle import Mingle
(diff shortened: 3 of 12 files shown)


Ducky2048:
I might have found a bug in this. If setting `TASK_QUEUES = (kombu.Queue(name="hello"),)` then the check here https://github.com/celery/celery/pull/9121/files#diff-891d2ad741a068f4a7fb765bf7583291b6b7ccfd61d5b90169e5abbf8beafa84R116 fails to detect a quorum queue because queue_arguments is empty - even though such a quorum queue already exists in the rabbitmq.

Nusnus:
> I might have found a bug in this. If setting `TASK_QUEUES = (kombu.Queue(name="hello"),)` then the check here https://github.com/celery/celery/pull/9121/files#diff-891d2ad741a068f4a7fb765bf7583291b6b7ccfd61d5b90169e5abbf8beafa84R116 fails to detect a quorum queue because queue_arguments is empty - even though such a quorum queue already exists in the rabbitmq. Later on we hit the `amqp.exceptions.AMQPNotImplementedError: Basic.consume: (540) NOT_IMPLEMENTED` error because we didn't disable global qos.

When we define quorum queues, we want an explicit definition of the queues, which is how we detect them, as you mentioned. Is it possible to resolve your issue by just adapting the queue args?

If not, would you like to contribute a patch to query the broker in the detection flow to see if such a queue is already defined (and then mark the detection as True)?

Ducky2048:
> When we define quorum queues, we want an explicit definition of the queues, which is how we detect them, as you mentioned. Is it possible to resolve your issue by just adapting the queue args?

Yes, this is how I solved it. I didn't know whether this was the expected way for things to work or if it was an oversight, this is why I wanted to let you know. 

> If not, would you like to contribute a patch to query the broker in the detection flow to see if such a queue is already defined (and then mark the detection as True)?

Thank you for your trust, but I don't feel confident enough for that.

Nusnus:
> > When we define quorum queues, we want an explicit definition of the queues, which is how we detect them, as you mentioned. Is it possible to resolve your issue by just adapting the queue args?
> 
> Yes, this is how I solved it. I didn't know whether this was the expected way for things to work or if it was an oversight, this is why I wanted to let you know. 
> 
> > If not, would you like to contribute a patch to query the broker in the detection flow to see if such a queue is already defined (and then mark the detection as True)?
> 
> Thank you for your trust, but I don't feel confident enough for that. 
> 
> 

Good point. Thanks for mentioning it.
If more ppl would ask for it, we can prioritize the effort.

And don't be afraid to try. Worst case you fail.
[October](https://hacktoberfest.com/) is a good month to try out too :)

celery/app/defaults.py
```diff
@@ -261,6 +261,7 @@ def __repr__(self):
         inherit_parent_priority=Option(False, type='bool'),
         default_delivery_mode=Option(2, type='string'),
         default_queue=Option('celery'),
+        default_queue_type=Option('classic', type='string'),
```
ulgens, line 264: @Nusnus Hey, do you remember why you defaulted to `classic`?

auvipy: for backward compatibility

auvipy: so that the existing deployments dont break. and can be switched in feature gracefully...

ulgens: I was asking because RabbitMQ's default is quorum, and realized Celery is overriding it. Is there a possibility to deprecate that behaviour gradually?

auvipy: there are many versions of rabbitmq where qourum is not default

ulgens: @auvipy I'm missing your point. In past releases, yes. It became the recommended queue type in 2024.

auvipy: > @auvipy I'm missing your point. In past releases, yes. It became the recommended queue type in 2024.

it is for legacy celery projects in the wild and enterprise environment. you don't need to worry about it.

ulgens: I am not sure why you are trying to shut me down but we can do better than this. You don't need respond if you have nothing to share. Thanks for your time tho.

@Nusnus do you have any additional info you can share? Thanks in advance.

auvipy: no one is trying to shutting you down. I think you have an understanding problem. or lack of understanding on broader level.

PR #9168, opened 2024-07-31
Option to use and kill process group in prefork

LSerranoPEReN:

## Description

I have a project where I launch background subprocesses during `worker_process_init` signal and kill them during `worker_process_shutdown`. Sometimes, `worker_process_init`  takes more time than allowed by `worker_proc_alive_timeout`, and the problematic worker process is then killed by celery.
The background subprocesses are not killed when this happens, only the main worker process.

This PR adds a new option to create a new process group when starting a prefork pool process, and use `kill_pg` when killing processes that exceed the deadline. This at least solve my use case, but **please let me know if this is a good idea or not.**

Still missing ATM:
- [ ] Handle windows case (no support for `kill_pg`)
- [ ] Add some tests (maybe hard to test properly)

### MWE
You can change `worker_proc_use_process_group` to `False` to see that fork are not killed by default.
Also, you need to send a hello task, because `worker_proc_alive_timeout` it's not enforced the first time.
(I don't know if it's a bug or not).

```python
from celery import Celery, signals
from time import sleep
import os
import sys

class Config:
    worker_proc_alive_timeout=1.5
    worker_max_tasks_per_child=1
    worker_proc_use_process_group=True

app = Celery('hello', broker='redis://localhost')
app.config_from_object(Config())

@signals.worker_process_init.connect()
def init_worker_process(**kwargs):
    pid = os.fork()
    if pid == 0:
        for i in range(5):
            sleep(1)
            print(i)
        sys.exit(0)
    sleep(2)

@app.task
def hello():
    return 'hello world'
```

### DIFF
--- a/celery/app/defaults.py
+++ b/celery/app/defaults.py
@@ -341,6 +341,7 @@ NAMESPACES = Namespace(
         pool_putlocks=Option(True, type='bool'),
         pool_restarts=Option(False, type='bool'),
         proc_alive_timeout=Option(4.0, type='float'),
+        proc_use_process_group=Option(False, type='bool'),
         prefetch_multiplier=Option(4, type='int'),
         eta_task_limit=Option(None, type='int'),
         enable_prefetch_count_reduction=Option(True, type='bool'),
--- a/celery/concurrency/asynpool.py
+++ b/celery/concurrency/asynpool.py
@@ -425,7 +425,7 @@ class AsynPool(_pool.Pool):
 
     def __init__(self, processes=None, synack=False,
                  sched_strategy=None, proc_alive_timeout=None,
-                 *args, **kwargs):
+                 proc_use_process_group=None, *args, **kwargs):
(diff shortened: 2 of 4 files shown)


Nusnus (changes requested):
Hey there, thank you for the PR.

I see you have 2 unchecked items for the PR, so I will change it to draft to properly communicate the current state. Please return it to ready for review once the items are done.

That being said, we also need proper tests please.
- Unit tests will cover the code coverage part.
- Integration tests - add if reasonable.
- Smoke tests - Simulates a production-like environment using the new [pytest-celery v1.x.x](https://docs.celeryq.dev/projects/pytest-celery/en/latest/) so it should be possible to test even shutdowns and killing processes.

auvipy (changes requested):
we can ignore windows support and focus on some unit and integration tests, which shows this features are actually working

PR #9023, opened 2024-05-14
Add support for Pydantic model validation/serialization (fixes #8751)

mathiasertl:
Allow validation of arguments using Pydantic, as well as result serialization. This PR would fix (IMHO) #8751.

With this PR, you can pass `pydantic=True` to a task decorator. If set, any argument type hinted as a Pydantic model will be validated (note: this means data validation *and* loading into the model class Pydantic speak) to that model, and any result matching a return type annotation will automatically be converted to a dict (using `model_dump()`).

With this feature, tasks look a lot more like FastAPI endpoints, where query arguments, request and response bodies are validated/serialized using Pydantic. The advantage is of course automatic data validation and safer typing. mypy will catch type hinting errors, and IDEs auto-complete values in the models.

PR #9171, opened 2024-08-03
update python versions and docker compose

mathiasertl:
Update Python versions in `CONTRIBUTING.rst`, and use `docker compose` instead of `docker-compose`. This originally comes from #9023 .


Nusnus (approved):
🔥

PR #9207, opened 2024-08-25
Native Delayed Delivery in RabbitMQ

### LINKED ISSUE #9149: Improved ETA support for quorum queues
> One issue was that ETA may saturate de prefetch queue and block incoming new job.
> 
> You should have a test that generate a lots of ETA tasks, and then still try to schedule new tasks to see if the prefetch count is updated accordingly and the task is executed (which was what the global_qos was doing).

Thank you. That was a hint in the right direction.

I’ve created a [PR](https://github.com/celery/celery/pull/9121) that adds support to quorum queues without ETA tasks support, but with proper documentation and an automatic warning in the worker log to ensure it won’t go unnoticed.

> While this solution works for us, we experienced a task loss issue with RabbitMQ cluster using quorum queue. We did some digging and it turned out that we also want to add `app.conf.broker_transport_options = {"confirm_publish": True}` possibly due to the [implementation differences](https://groups.google.com/g/rabbitmq-users/c/LAflf1l4LtY#:~:text=A%20quorum%20queue%20ignores%20messages%20published%20by%20a%20closed%20channel%20and%20a%20classic%20queue%20does%20not.%20That%27s%20an%20implementation%20detail.) of quorum queue and classic queue when handling messages from closed channel.
> 
> refs - #5410, https://groups.google.com/g/rabbitmq-users/c/LAflf1l4LtY

I’ve also added a warning per this comment from @cbwchuck, so you’ll know you need to enable confirm publish when using quorum queues.

---

My implementation basically adds two new feature flags:
1. `worker_detect_quorum_queues` which is set by default to `True`: Auto-detect if at least one queue (including the default queue) is set to quorum and disables the global QoS.
2. `task_default_queue_type` which is set by default to `”classic"`: Can be set to `”quorum"` to automatically change the default `celery` queue type to quorum-queue. This means that if you don’t provide your own queue (using the queue argument accordingly), you can just do `app.conf.task_default_queue_type = “quorum”` (not forgetting `app.conf.broker_transport_options = {"confirm_publish": True}`), and the worker will create a quorum default queue and turn off the global QoS (per flag no 1).

ETA support will need to be developed separately to allow progressive support for quorum queues as to not hold back the currently proposed solution which might be enough for many users.

For more details see the PR changes.
It includes proper documentation for these flags, a new example project and 100% test coverage, so it should be clear what was changed from all angles.

EDIT:
Here are screenshots of the documentation from the PR’s build.
CleanShot 2024-07-08 at 00 27 56@2x
CleanShot 2024-07-08 at 00 28 07@2x

_Originally posted by @Nusnus in https://github.com/celery/celery/issues/6067#issuecomment-2212577881_

thedrow:

## Description

Fixes #9149.
Since ETA tasks will block the worker when using quorum queues we had to provide an alternative solution that will support ETA tasks on RabbitMQ.
This PR provides a solution based on NServiceBus' [implementation](https://docs.particular.net/transports/rabbitmq/delayed-delivery) of native delayed delivery.
The solution is based on a series of exchanges and queues as detailed in NServiceBus' documentation.
To use this feature you need to enable the `broker_native_delayed_delivery` configuration setting when using quorum queues.

### DIFF
--- a/celery/app/base.py
+++ b/celery/app/base.py
@@ -513,6 +513,7 @@ class Celery:
                 if shared:
                     def cons(app):
                         return app._task_from_fun(fun, **opts)
+
                     cons.__name__ = fun.__name__
                     connect_on_app_finalize(cons)
                 if not lazy or self.finalized:
@@ -848,10 +849,12 @@ class Celery:
                     options['routing_key'] = routing_key
                     options['exchange'] = exchange
         elif is_native_delayed_delivery and options['queue'].exchange.type == 'direct':
-            logger.warn("Direct exchanges are not supported with native delayed delivery.\n"
-                        f"{options['queue'].exchange.name} is a direct exchange but should be a topic exchange or "
-                        f"a fanout exchange in order for native delayed delivery to work properly.\n"
-                        f"If quorum queues are used, this task may block the worker process until the ETA arrives.")
+            logger.warning(
+                'Direct exchanges are not supported with native delayed delivery.\n'
+                f'{options["queue"].exchange.name} is a direct exchange but should be a topic exchange or '
+                'a fanout exchange in order for native delayed delivery to work properly.\n'
+                'If quorum queues are used, this task may block the worker process until the ETA arrives.'
+            )
 
         if expires is not None:
             if isinstance(expires, datetime):
@@ -1013,6 +1016,7 @@ class Celery:
                 'broker_connection_timeout', connect_timeout
             ),
         )
+
     broker_connection = connection
 
     def _acquire_connection(self, pool=True):
@@ -1032,6 +1036,7 @@ class Celery:
                 will be acquired from the connection pool.
         """
         return FallbackContext(connection, self._acquire_connection, pool=pool)
+
     default_connection = connection_or_acquire  # XXX compat
 
     def producer_or_acquire(self, producer=None):
@@ -1047,6 +1052,7 @@ class Celery:
         return FallbackContext(
             producer, self.producer_pool.acquire, block=True,
         )
+
     default_producer = producer_or_acquire  # XXX compat
 
     def prepare_config(self, c):
--- a/celery/app/base.py
+++ b/celery/app/base.py
@@ -834,6 +834,8 @@ class Celery:
             or conf.broker_url.startswith('py-amqp://')
         ):
             if eta:
+                if isinstance(eta, str):
+                    eta = isoparse(eta)
                 countdown = (maybe_make_aware(eta) - self.now()).total_seconds()
 
             if countdown:
--- a/celery/app/base.py
+++ b/celery/app/base.py
@@ -829,33 +829,33 @@ class Celery:
         options = router.route(
             options, route_name or name, args, kwargs, task_type)
 
-        if conf.broker_native_delayed_delivery and (
-            conf.broker_url.startswith('amqp://')
-            or conf.broker_url.startswith('py-amqp://')
-        ):
+        is_native_delayed_delivery = conf.broker_native_delayed_delivery and (
+            conf.broker_url.startswith('amqp://') or conf.broker_url.startswith('py-amqp://'))
+        if is_native_delayed_delivery and options['queue'].exchange.type != 'direct':
             if eta:
                 if isinstance(eta, str):
                     eta = isoparse(eta)
                 countdown = (maybe_make_aware(eta) - self.now()).total_seconds()
 
             if countdown:
-                routing_key = '.'.join(list(f'{int(countdown):028b}')) + f'.{options["queue"].routing_key}'
-                exchange = Exchange(
-                    'celery_delayed_27',
-                    type='topic',
-                )
-                options['queue'] = Queue(
-                    'celery_delayed_27',
-                    exchange=exchange,
-                    routing_key=routing_key,
-                    queue_arguments={
-                        "x-queue-type": "quorum",
-                        "x-dead-letter-strategy": "at-least-once",
-                        "x-overflow": "reject-publish",
-                        "x-message-ttl": pow(2, 27) * 1000,
-                        "x-dead-letter-exchange": 'celery_delayed_26',
-                    }
-                )
+                if countdown > 0:
+                    routing_key = '.'.join(list(f'{int(countdown):028b}')) + f'.{options["queue"].routing_key}'
+                    exchange = Exchange(
+                        'celery_delayed_27',
+                        type='topic',
+                    )
+                    options['queue'] = Queue(
+                        'celery_delayed_27',
+                        exchange=exchange,
+                        routing_key=routing_key,
+                        queue_arguments={
+                            "x-queue-type": "quorum",
+                            "x-dead-letter-strategy": "at-least-once",
+                            "x-overflow": "reject-publish",
+                            "x-message-ttl": pow(2, 27) * 1000,
+                            "x-dead-letter-exchange": 'celery_delayed_26',
+                        }
+                    )
 
         if expires is not None:
             if isinstance(expires, datetime):
--- a/celery/app/base.py
+++ b/celery/app/base.py
@@ -832,7 +832,7 @@ class Celery:
             options, route_name or name, args, kwargs, task_type)
 
         is_native_delayed_delivery = conf.broker_native_delayed_delivery and (
-            conf.broker_url.startswith('amqp://') or conf.broker_url.startswith('py-amqp://'))
+            self.producer_pool.connections.connection == 'amqp')
         if is_native_delayed_delivery and options['queue'].exchange.type != 'direct':
             if eta:
                 if isinstance(eta, str):
--- a/celery/app/base.py
+++ b/celery/app/base.py
@@ -14,7 +14,7 @@ from operator import attrgetter
 
 from click.exceptions import Exit
 from dateutil.parser import isoparse
-from kombu import Exchange, Queue, pools
+from kombu import Exchange, pools
 from kombu.clocks import LamportClock
 from kombu.common import oid_from
 from kombu.utils.compat import register_after_fork
@@ -844,18 +844,9 @@ class Celery:
                         'celery_delayed_27',
                         type='topic',
                     )
-                    options['queue'] = Queue(
-                        'celery_delayed_27',
-                        exchange=exchange,
-                        routing_key=routing_key,
-                        queue_arguments={
-                            "x-queue-type": "quorum",
-                            "x-dead-letter-strategy": "at-least-once",
-                            "x-overflow": "reject-publish",
-                            "x-message-ttl": pow(2, 27) * 1000,
-                            "x-dead-letter-exchange": 'celery_delayed_26',
-                        }
-                    )
+                    del options['queue']
+                    options['routing_key'] = routing_key
+                    options['exchange'] = exchange
         elif is_native_delayed_delivery and options['queue'].exchange.type == 'direct':
             logger.warn("Direct exchanges are not supported with native delayed delivery.\n"
                         f"{options['queue'].exchange.name} is a direct exchange but should be a topic exchange or "
--- a/celery/app/base.py
+++ b/celery/app/base.py
@@ -14,7 +14,7 @@ from operator import attrgetter
 
 from click.exceptions import Exit
 from dateutil.parser import isoparse
-from kombu import pools
+from kombu import pools, Queue, Exchange
 from kombu.clocks import LamportClock
 from kombu.common import oid_from
 from kombu.utils.compat import register_after_fork
@@ -828,6 +828,33 @@ class Celery:
         ignore_result = options.pop('ignore_result', False)
         options = router.route(
             options, route_name or name, args, kwargs, task_type)
+
+        if conf.broker_native_delayed_delivery and (
+            conf.broker_url.startswith('amqp://')
+            or conf.broker_url.startswith('py-amqp://')
+        ):
+            if eta:
+                countdown = (maybe_make_aware(eta) - self.now()).total_seconds()
+
+            if countdown:
+                routing_key = '.'.join(list(f'{int(countdown):028b}')) + f'.{options["queue"].routing_key}'
+                exchange = Exchange(
+                    'celery_delayed_27',
+                    type='topic',
+                )
+                options['queue'] = Queue(
+                    'celery_delayed_27',
+                    exchange=exchange,
+                    routing_key=routing_key,
+                    queue_arguments={
+                        "x-queue-type": "quorum",
+                        "x-dead-letter-strategy": "at-least-once",
+                        "x-overflow": "reject-publish",
+                        "x-message-ttl": pow(2, 27) * 1000,
+                        "x-dead-letter-exchange": 'celery_delayed_26',
+                    }
+                )
+
         if expires is not None:
             if isinstance(expires, datetime):
                 expires_s = (maybe_make_aware(
--- a/celery/app/base.py
+++ b/celery/app/base.py
@@ -39,6 +39,7 @@ from celery.utils.objects import FallbackContext, mro_lookup
 from celery.utils.time import maybe_make_aware, timezone, to_utc
 
 from ..utils.annotations import annotation_is_class, annotation_issubclass, get_optional_arg
+from ..utils.quorum_queues import detect_quorum_queues
 # Load all builtin tasks
 from . import backends, builtins  # noqa
 from .annotations import prepare as prepare_annotations
@@ -831,8 +832,8 @@ class Celery:
         options = router.route(
             options, route_name or name, args, kwargs, task_type)
 
-        is_native_delayed_delivery = conf.broker_native_delayed_delivery and (
-            self.producer_pool.connections.connection == 'amqp')
+        is_native_delayed_delivery = detect_quorum_queues(self,
+                                                          self.producer_pool.connections.connection.transport_cls)[0]
         if is_native_delayed_delivery and options['queue'].exchange.type != 'direct':
             if eta:
                 if isinstance(eta, str):
--- a/celery/app/base.py
+++ b/celery/app/base.py
@@ -856,6 +856,11 @@ class Celery:
                             "x-dead-letter-exchange": 'celery_delayed_26',
                         }
                     )
+        elif is_native_delayed_delivery and options['queue'].exchange.type == 'direct':
+            logger.warn("Direct exchanges are not supported with native delayed delivery.\n"
+                        f"{options['queue'].exchange.name} is a direct exchange but should be a topic exchange or "
+                        f"a fanout exchange in order for native delayed delivery to work properly.\n"
+                        f"If quorum queues are used, this task may block the worker process until the ETA arrives.")
 
         if expires is not None:
             if isinstance(expires, datetime):
--- a/celery/app/base.py
+++ b/celery/app/base.py
@@ -14,7 +14,7 @@ from operator import attrgetter
 
 from click.exceptions import Exit
 from dateutil.parser import isoparse
-from kombu import pools, Queue, Exchange
+from kombu import Exchange, Queue, pools
 from kombu.clocks import LamportClock
 from kombu.common import oid_from
 from kombu.utils.compat import register_after_fork
--- a/celery/app/base.py
+++ b/celery/app/base.py
@@ -17,6 +17,7 @@ from dateutil.parser import isoparse
 from kombu import Exchange, pools
 from kombu.clocks import LamportClock
 from kombu.common import oid_from
+from kombu.transport.native_delayed_delivery import calculate_routing_key
 from kombu.utils.compat import register_after_fork
 from kombu.utils.objects import cached_property
 from kombu.utils.uuid import uuid
@@ -840,7 +841,7 @@ class Celery:
 
             if countdown:
                 if countdown > 0:
-                    routing_key = '.'.join(list(f'{int(countdown):028b}')) + f'.{options["queue"].routing_key}'
+                    routing_key = calculate_routing_key(int(countdown), options["queue"].routing_key)
                     exchange = Exchange(
                         'celery_delayed_27',
                         type='topic',
--- a/celery/app/defaults.py
+++ b/celery/app/defaults.py
@@ -96,7 +96,7 @@ NAMESPACES = Namespace(
         heartbeat_checkrate=Option(3.0, type='int'),
         login_method=Option(None, type='string'),
         native_delayed_delivery=Option(False, type='bool'),
-        native_delayed_delivery_queue_type=Option(default='classic', type='string'),
+        native_delayed_delivery_queue_type=Option(default='quorum', type='string'),
         pool_limit=Option(10, type='int'),
         use_ssl=Option(False, type='bool'),
 
--- a/celery/app/defaults.py
+++ b/celery/app/defaults.py
@@ -96,7 +96,6 @@ NAMESPACES = Namespace(
         heartbeat_checkrate=Option(3.0, type='int'),
         login_method=Option(None, type='string'),
         native_delayed_delivery=Option(False, type='bool'),
-        native_delayed_delivery_queue_type=Option(default='classic', type='string'),
         pool_limit=Option(10, type='int'),
         use_ssl=Option(False, type='bool'),
 
--- a/celery/app/defaults.py
+++ b/celery/app/defaults.py
@@ -96,6 +96,7 @@ NAMESPACES = Namespace(
         heartbeat_checkrate=Option(3.0, type='int'),
         login_method=Option(None, type='string'),
         native_delayed_delivery=Option(False, type='bool'),
+        native_delayed_delivery_queue_type=Option(default='classic', type='string'),
         pool_limit=Option(10, type='int'),
         use_ssl=Option(False, type='bool'),
 
--- a/celery/app/defaults.py
+++ b/celery/app/defaults.py
@@ -95,7 +95,6 @@ NAMESPACES = Namespace(
         heartbeat=Option(120, type='int'),
         heartbeat_checkrate=Option(3.0, type='int'),
         login_method=Option(None, type='string'),
-        native_delayed_delivery=Option(False, type='bool'),
         native_delayed_delivery_queue_type=Option(default='quorum', type='string'),
         pool_limit=Option(10, type='int'),
         use_ssl=Option(False, type='bool'),
--- a/celery/app/defaults.py
+++ b/celery/app/defaults.py
@@ -95,6 +95,7 @@ NAMESPACES = Namespace(
         heartbeat=Option(120, type='int'),
         heartbeat_checkrate=Option(3.0, type='int'),
         login_method=Option(None, type='string'),
+        native_delayed_delivery=Option(False, type='bool'),
         pool_limit=Option(10, type='int'),
         use_ssl=Option(False, type='bool'),
 
--- a/celery/worker/consumer/consumer.py
+++ b/celery/worker/consumer/consumer.py
@@ -169,6 +169,7 @@ class Consumer:
             'celery.worker.consumer.heart:Heart',
             'celery.worker.consumer.control:Control',
             'celery.worker.consumer.tasks:Tasks',
+            'celery.worker.consumer.delayed_delivery:DelayedDelivery',
             'celery.worker.consumer.consumer:Evloop',
             'celery.worker.consumer.agent:Agent',
         ]
--- a/celery/worker/consumer/consumer.py
+++ b/celery/worker/consumer/consumer.py
@@ -478,9 +478,9 @@ class Consumer:
         return self.ensure_connected(
             self.app.connection_for_read(heartbeat=heartbeat))
 
-    def connection_for_write(self, heartbeat=None):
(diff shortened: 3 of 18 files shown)


thedrow:
@Nusnus Any idea why the quorum queues smoke tests are failing because the RabbitMQ container is not ready yet?

thedrow:
> > @Nusnus Any idea why the quorum queues smoke tests are failing because the RabbitMQ container is not ready yet?
> 
> Does it pass locally?

No it does not. I haven't done anything that can destabilize the test suite so I'm not sure what could be wrong.

celery/app/base.py
```diff
@@ -816,6 +817,33 @@ def send_task(self, name, args=None, kwargs=None, countdown=None,
         ignore_result = options.pop('ignore_result', False)
         options = router.route(
             options, route_name or name, args, kwargs, task_type)
+
+        is_native_delayed_delivery = conf.broker_native_delayed_delivery and (
+            conf.broker_url.startswith('amqp://') or conf.broker_url.startswith('py-amqp://'))
```
Nusnus, line 822: Please fix like in https://github.com/celery/celery/pull/9294

It was originally my code I know. This PR improves upon it.

celery/worker/consumer/delayed_delivery.py
```diff
@@ -0,0 +1,82 @@
...
+from celery import Celery, bootsteps
+from celery.utils.log import get_logger
+from celery.worker.consumer import Consumer, Tasks
+
+__all__ = ('DelayedDelivery',)
+
+logger = get_logger(__name__)
+
+
+class DelayedDelivery(bootsteps.StartStopStep):
```
Nusnus, line 12: Please add documentation to this class indicating how it implements the interface etc.

celery/worker/consumer/tasks.py
```diff
@@ -91,12 +90,17 @@ def qos_global(self, c) -> bool:
 
         if c.app.conf.worker_detect_quorum_queues:
             using_quorum_queues, qname = self.detect_quorum_queues(c)
+
+            if c.app.conf.broker_native_delayed_delivery is True:
```
Nusnus, line 94: Does this mean the new flag implicitly enables quorum queue mode a.k.a global equal false?

Because the purpose of the original code is to validate against the queues args.

Won't it be confusing?

Thinking out loud

Nusnus (changes requested):
Optional comment: Add assertion error messages.
The only reason I bring this up is due to the nature of this change, which is relying on an algorithm to implement an ETA mechanism, so these errors could be crucial, and as the author of the changes, you will probably come up with the best error messages as you understand the protocol.

Food for thought.

Except that, please see my other comments. Something needs fixing too.

Nusnus:
Also please make sure the CI fully passes 🙏

celery/worker/consumer/tasks.py
thedrow: You're right. That's no longer the case.

thedrow:
CI fails because I'm installing Kombu from Git. @Nusnus Once you release Kombu's RC I can change the requirements to that RC.

Nusnus (changes requested):
@thedrow 
> CI fails because I'm installing Kombu from Git. @Nusnus Once you release Kombu's RC I can change the requirements to that RC.

Done. Please rebase on main.
The requirements are already updated.

Ping me if the CI fully passes for a review.
There have been many changes since my last review.

I'll do it today if the CI is green of course.

thedrow:
@Nusnus Ping.

docs/getting-started/backends-and-brokers/rabbitmq.rst
```diff
@@ -172,3 +172,62 @@ but rather use the :command:`rabbitmqctl` command:
     $ sudo rabbitmqctl stop
 
 When the server is running, you can continue reading `Setting up RabbitMQ`_.
+
+.. _using-quorum-queues:
+
+Using Quorum Queues
+===================
+
```
Nusnus, line 180: Missing version string (v5.5)

docs/getting-started/backends-and-brokers/rabbitmq.rst
```diff
@@ -172,3 +172,62 @@ but rather use the :command:`rabbitmqctl` command:
...
+    task_queues = [Queue('my-queue', queue_arguments={'x-queue-type': 'quorum'})]
+    broker_transport_options = {"confirm_publish": True}
+
+If you'd like to change the type of the default queue, set the :setting:`task_default_queue_type` setting to ``quorum``.
+
+Celery automatically detects if quorum queues are used using the :setting:`worker_detect_quorum_queues` setting.
+We recommend to keep the default behavior turned on.
+
+.. _`Quorum Queues`: https://www.rabbitmq.com/docs/quorum-queues
```
Nusnus, line 200: Please also add a link to this article [Migrating from Mirrored Classic Queues to Quorum Queues](https://www.rabbitmq.com/blog/2023/03/02/quorum-queues-migration)

docs/getting-started/backends-and-brokers/rabbitmq.rst
```diff
@@ -172,3 +172,62 @@ but rather use the :command:`rabbitmqctl` command:
...
+Autoscaling relies on increasing and decreasing the prefetch count whenever a new process is instantiated
+or terminated so it won't work when Quorum Queues are detected.
+
+Similarly, the :setting:`worker_enable_prefetch_count_reduction` setting will be a no-op even when set to ``True``
+when Quorum Queues are detected.
+
+In addition, :ref:`ETA/Countdown <calling-eta>` will block the worker when received until the ETA arrives since
+we can no longer increase the prefetch count and fetch another task from the queue.
+
+In order to properly schedule ETA/Countdown tasks you need to enable :ref:`Native Delayed Delivery <native-delayed-delivery>`.
```
Nusnus, line 219: Again, the ETA support should be detected and enabled automatically, so this doc requires fixing as well to say that when quorum quques are used, the ETA mechanism will be powered by Native Delayed Delivery of RabbitMQ automatically, instead of using the loop timers when not using quorum queues

docs/getting-started/backends-and-brokers/rabbitmq.rst
```diff
@@ -172,3 +172,62 @@ but rather use the :command:`rabbitmqctl` command:
...
+-----------------------
+
+Since tasks with ETA/Countdown will block the worker until they are scheduled for execution,
+we need to use RabbitMQ's native capabilities to schedule the execution of tasks.
+
+The design is borrowed from NServiceBus. If you are interested in the implementation details, refer to their `documentation`_.
+
+.. _documentation: https://docs.particular.net/transports/rabbitmq/delayed-delivery
+
+To enable Native Delayed Delivery set the :setting:`broker_native_delayed_delivery` setting to ``True``.
```
Nusnus, line 233: Again, all mentions of the flag need to be adjusted. It has to be automatic, or else the design will be error-prone and will cause either headaches or waste a lot of time reporting and investigating issues.

**People will miss this flag, trust me bro.** @thedrow

t/smoke/tests/quorum_queues/test_native_delayed_delivery.py
```diff
@@ -0,0 +1,169 @@
...
+        self,
+        celery_setup: CeleryTestSetup,
+        default_worker_app: Celery
+    ):
+        broker: RabbitMQManagementBroker = celery_setup.broker
+        api = broker.get_management_url() + "/api/queues"
+        response = requests.get(api, auth=HTTPBasicAuth("guest", "guest"))
+        assert response.status_code == 200
+        res = response.json()
+        assert isinstance(res, list)
```
Nusnus, line 38: Move this into a fixture please. It will also reduce code duplication.

t/smoke/tests/quorum_queues/test_native_delayed_delivery.py
```diff
@@ -0,0 +1,169 @@
...
+            conf = default_worker_app.conf
+            assert queue_arguments["x-queue-type"] == conf.broker_native_delayed_delivery_queue_type
+
+    def test_native_delayed_delivery_exchange_configuration(self, celery_setup: CeleryTestSetup):
+        broker: RabbitMQManagementBroker = celery_setup.broker
+        api = broker.get_management_url() + "/api/exchanges"
+        response = requests.get(api, auth=HTTPBasicAuth("guest", "guest"))
+        assert response.status_code == 200
+        res = response.json()
+        assert isinstance(res, list)
```
Nusnus, line 64: Replace with fixture

t/smoke/tests/quorum_queues/test_native_delayed_delivery.py
```diff
@@ -0,0 +1,169 @@
...
+        self,
+        celery_setup: CeleryTestSetup,
+        default_worker_app: Celery
+    ):
+        broker: RabbitMQManagementBroker = celery_setup.broker
+        api = broker.get_management_url() + "/api/queues"
+        response = requests.get(api, auth=HTTPBasicAuth("guest", "guest"))
+        assert response.status_code == 200
+        res = response.json()
+        assert isinstance(res, list)
```
Nusnus, line 97: Same

t/smoke/tests/quorum_queues/test_native_delayed_delivery.py
```diff
@@ -0,0 +1,169 @@
...
+            assert queue_arguments["x-message-ttl"] == pow(2, queue_level) * 1000
+            assert queue_arguments["x-queue-type"] == 'classic'
+
+    def test_native_delayed_delivery_exchange_configuration(self, celery_setup: CeleryTestSetup):
+        broker: RabbitMQManagementBroker = celery_setup.broker
+        api = broker.get_management_url() + "/api/exchanges"
+        response = requests.get(api, auth=HTTPBasicAuth("guest", "guest"))
+        assert response.status_code == 200
+        res = response.json()
+        assert isinstance(res, list)
```
Nusnus, line 121: Same

t/smoke/tests/quorum_queues/test_native_delayed_delivery.py
```diff
@@ -0,0 +1,169 @@
+from datetime import timedelta
+from datetime import timezone as datetime_timezone
+
+import pytest
+import requests
+from future.backports.datetime import datetime
+from pytest_celery import CeleryTestSetup
```
Nusnus, line 7:
```suggestion
from pytest_celery import CeleryTestSetup, RESULT_TIMEOUT
```

t/smoke/tests/quorum_queues/test_native_delayed_delivery.py
```diff
@@ -0,0 +1,169 @@
...
+        app.conf.task_default_routing_key = 'celery'
+
+        return app
+
+    def test_countdown(self, celery_setup: CeleryTestSetup):
+        s = noop.s().set(queue=celery_setup.worker.worker_queue)
+
+        result = s.apply_async(countdown=5)
+
+        result.get(timeout=10)
```
Nusnus, line 148:
```suggestion
        result.get(timeout=RESULT_TIMEOUT)
```

t/smoke/tests/quorum_queues/test_native_delayed_delivery.py
```diff
@@ -0,0 +1,169 @@
...
+
+    def test_eta(self, celery_setup: CeleryTestSetup):
+        s = noop.s().set(queue=celery_setup.worker.worker_queue)
+
+        result = s.apply_async(eta=datetime.now(datetime_timezone.utc) + timedelta(0, 5))
+
+        result.get(timeout=10)
```
Nusnus, line 155:
```suggestion
        result.get(timeout=RESULT_TIMEOUT)
```

t/smoke/tests/quorum_queues/test_native_delayed_delivery.py
```diff
@@ -0,0 +1,169 @@
...
+
+    def test_eta_str(self, celery_setup: CeleryTestSetup):
+        s = noop.s().set(queue=celery_setup.worker.worker_queue)
+
+        result = s.apply_async(eta=(datetime.now(datetime_timezone.utc) + timedelta(0, 5)).isoformat())
+
+        result.get(timeout=10)
```
Nusnus, line 162:
```suggestion
        result.get(timeout=RESULT_TIMEOUT)
```

t/smoke/tests/quorum_queues/test_native_delayed_delivery.py
```diff
@@ -0,0 +1,169 @@
...
+
+    def test_eta_in_the_past(self, celery_setup: CeleryTestSetup):
+        s = noop.s().set(queue=celery_setup.worker.worker_queue)
+
+        result = s.apply_async(eta=(datetime.now(datetime_timezone.utc) - timedelta(0, 5)).isoformat())
+
+        result.get(timeout=10)
```
Nusnus, line 169:
```suggestion
        result.get(timeout=RESULT_TIMEOUT)
```

t/unit/app/test_app.py
```diff
@@ -1399,6 +1401,190 @@ def test_send_task_expire_as_string(self):
         except TypeError as e:
             pytest.fail(f'raise unexcepted error {e}')
 
+    def test_native_delayed_delivery_countdown(self):
```
Nusnus, line 1404: I wonder if you should maybe move the delayed delivery unit tests to their own module, like the smoke tests also have their own test module for this grand feature.

If you’re already fixing the above comments, I highly recommend moving this to a dedicated test file to keep the test environment/codebase clean and easy to maintain.

I hate our super long test files, it’s very hard to locate stuff, so let’s not complicate it more.

celery/worker/consumer/tasks.py
```diff
@@ -22,9 +22,8 @@
 Detected quorum queue "%r", disabling global QoS.
 With global QoS disabled, ETA tasks may not function as expected. Instead of adjusting
 the prefetch count dynamically, ETA tasks will occupy the prefetch buffer, potentially
-blocking other tasks from being consumed. To mitigate this, either set a high prefetch
-count or avoid using quorum queues until the ETA mechanism is updated to support a
-disabled global QoS, which is required for quorum queues.
+blocking other tasks from being consumed.
+Please enable native delayed delivery so that ETA tasks will not block the worker.
```
Nusnus, line 26: After reading the rest of the comment, consider removing this warning altogether (with the tests etc.)
It is not needed anymore. The ETA feature is now fully supported both with/without quorum queues.

t/unit/worker/test_consumer.py
```diff
@@ -637,6 +637,7 @@ def test_qos_global_worker_detect_quorum_queues_true_with_quorum_queues(self):
     def test_qos_global_eta_warning(self):
```
Nusnus, line 637: Well, basically, if we enable ETA via native delayed delivery automatically, then we don’t need any warning anymore. The feature is complete.

It’s better not to include extra code “just because it was done”, it increases the maintenance complexity.
Please remove the warning or present a valid argument to maintain it, as I don’t see a reason anymore.

Nusnus: Let me be more clear - the warning message itself isn’t needed anymore, not the warning code.
It requires rephrasing that only the global QoS is disabled and how it effect prefetch count, e.g., `Detected quorum queue "%r", disabling global QoS.` and that’s it plus minus.

The ETA is no longer a problem to report the user about, it will just work now.

t/unit/app/test_app.py
Nusnus: I see now that you did add such a file - why didn’t you include these tests there then?

Cosmetic issue though.

Jean-Daniel:
Out of curiosity, what are the benefit of the NServiceBus implementation versus using [per-message TTL](https://www.rabbitmq.com/docs/ttl#per-message-ttl-in-publishers) by specifying the `expiration` field when sending a delayed message ?

Nusnus (changes requested):
@thedrow
> CI fails because I'm installing Kombu from Git. @Nusnus Once you release Kombu's RC I can change the requirements to that RC.

Please take note the new Kombu RC release, which is a requirement for this PR, has introduced a new breaking bug as described in https://github.com/celery/kombu/issues/2157.
Obviously, we do not want to revert any changes, but if we merge this PR without fixing Kombu (first IMHO), then we’ll introduce a dependency in Celery v5.5 to a confirmed broken (RC) version of Kombu.

Please review the Kombu’s report by @woutdenolf in addition to the PR comments from yesterday, as per my explanation above, it is closely related.

celery/app/defaults.py
```diff
@@ -95,6 +95,8 @@ def __repr__(self):
         heartbeat=Option(120, type='int'),
         heartbeat_checkrate=Option(3.0, type='int'),
         login_method=Option(None, type='string'),
+        native_delayed_delivery=Option(False, type='bool'),
+        native_delayed_delivery_queue_type=Option(default='classic', type='string'),
```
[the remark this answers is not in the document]
thedrow: My client wants to use this before migrating to quorum queues so the default is set to classic.

celery/app/defaults.py
```diff
@@ -95,6 +95,8 @@ def __repr__(self):
         heartbeat=Option(120, type='int'),
         heartbeat_checkrate=Option(3.0, type='int'),
         login_method=Option(None, type='string'),
+        native_delayed_delivery=Option(False, type='bool'),
```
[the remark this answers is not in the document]
thedrow: That's what the warning is for. We need this feature flag so that this can be turned on manually when all those exchanges and queues are declared only when needed.
People may chose to leave this turned off for their own reasons.

celery/worker/consumer/consumer.py
```diff
@@ -476,9 +477,9 @@ def connection_for_read(self, heartbeat=None):
         return self.ensure_connected(
             self.app.connection_for_read(heartbeat=heartbeat))
 
-    def connection_for_write(self, heartbeat=None):
+    def connection_for_write(self, url=None, heartbeat=None):
```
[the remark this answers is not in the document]
thedrow: We need this to declare the exchanges and queues on failover brokers as well.

docs/getting-started/backends-and-brokers/rabbitmq.rst
thedrow: People won't miss this flag because of the warning I print when using quorum queues without this feature enabled.

t/smoke/tests/quorum_queues/test_native_delayed_delivery.py
thedrow: I want a shorter timeout because the task should end after the countdown. This is not flaky.

t/smoke/tests/quorum_queues/test_native_delayed_delivery.py
thedrow: I want a shorter timeout because the task should end after the countdown. This is not flaky.

t/smoke/tests/quorum_queues/test_native_delayed_delivery.py
thedrow: I want a shorter timeout because the task should end after the countdown. This is not flaky.

celery/worker/consumer/tasks.py
thedrow: We need this warning so that people will enable native delayed delivery if quorum queues are used.

thedrow:
> Out of curiosity, what are the benefit of the NServiceBus implementation versus using [per-message TTL](https://www.rabbitmq.com/docs/ttl#per-message-ttl-in-publishers) by specifying the `expiration` field when sending a delayed message ?

I haven't considered other implementations. This just works so there's no need to find other solutions.

t/unit/worker/test_consumer.py
thedrow: We need the warning since we're keeping the feature flag.

celery/app/defaults.py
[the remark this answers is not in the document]
Nusnus: > My client wants to use this before migrating to quorum queues so the default is set to classic.

I understand, but why does it make sense?
It makes sense to me to be default `quorum` because that’s the only flow it is used.
If not, you force the user to 1) be aware of this flag and 2) set it.
Instead, setting it to `quorum` gives an automatic behavior, which is why `classic` doesn’t make sense to me.

You may bring this up with your client to ask for clarifications, but we want to reach an agreement based on what’s best for the code.

t/unit/app/test_app.py
Nusnus: Make sense, thanks for explaining!

t/unit/worker/test_consumer.py
Nusnus: > We need the warning since we're keeping the feature flag.

Let’s agree for now that the warning is indeed dependent on this decision.
If the flag remains, so will the warning, and vice versa.

celery/app/defaults.py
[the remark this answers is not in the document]
Nusnus: Related to: https://github.com/celery/celery/pull/9207/files#r1801792448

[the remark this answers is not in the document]
Nusnus: > People may choose to leave this turned off for their own reasons.

Give me one example that it is useful as `False`.
Any case anyone can benefit from it?

I argue that the `False` path is not useful for any use-case, which is why it should be `True`, or actually not even used and assumed `True` automatically if quorum queues are detected, per no use case at all of the `False` option.

docs/getting-started/backends-and-brokers/rabbitmq.rst
Nusnus: So the way I see it,
Option 1) Log warning and expect user to see, read, understand, look for doc and adjust/fix code, even if it’s a simple line or two of change.
Option 2) No warning, automatic detection of state and automatic configuration accordingly. The user knows nothing, needs to configure nothing, and the behavior with/without quorum queues is the same from the POV the user/API, which is the definition of ETA support for quorum queues. Requiring to enable more flags creates a design dependency which affect the readability of the feature, breaks compatibility with the standard ETA feature (which doesn’t requires any flags) and by nature and definition, is error-prone (automatic/manual approach).

How come Option 2 doesn’t appeal to you @thedrow ?

t/smoke/tests/quorum_queues/test_native_delayed_delivery.py
Nusnus: Cool

Nusnus (changes requested):
Replied to what was answered so far.
Please review my new comments and the rest of yesterday's comments, which are still awaiting your response/fix.

Ping me when you’re done.

P.S
Merge/Rebase on `main` to make sure you’re up to date.

docs/getting-started/backends-and-brokers/rabbitmq.rst
thedrow: I added the link.

docs/getting-started/backends-and-brokers/rabbitmq.rst
thedrow: Added a version string

docs/getting-started/backends-and-brokers/rabbitmq.rst
thedrow: We have a warning when quorum queues are detected that will notify users about this option.
My client wants to migrate to native delayed delivery before migrating to quorum queues.

celery/worker/consumer/delayed_delivery.py
```diff
@@ -0,0 +1,37 @@
...
+
+    def include_if(self, c):
+        return (c.app.conf.broker_native_delayed_delivery
+                and c.connection_for_write().transport.driver_type == 'amqp')
+
+    def start(self, c: Consumer):
+        app: Celery = c.app
+
+        for broker_url in app.conf.broker_url.split(';'):
+            try:
```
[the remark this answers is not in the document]
Nusnus: I know I suggested this fix but on second thought, what happens if the failover broker (2nd broker) has a valid connection loss.. we want to ensure the connection in this case.

So I think my fix suggestion is bugged.

Wdyt?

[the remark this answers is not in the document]
thedrow: In that case, failover have already occurred and we're running the bootstep again so I think it's fine.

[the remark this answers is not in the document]
thedrow: I think the existing failover smoke tests already cover this usecase.
What more would you like to add?

celery/app/defaults.py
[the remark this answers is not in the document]
thedrow: You may want to keep the old, timer based solution when not using quorum queues. Hence, the sensible default is `False`.
You may also want to use native delayed delivery without quorum queues, in which case the flag should be set to `True`. That's exactly what my client wants to do first.
They want to test this feature without using quorum queues first and later on migrate to quorum queues.

celery/app/defaults.py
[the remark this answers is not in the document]
thedrow: In a busy production environment, using quorum queues for all the native delayed delivery queues may overload RabbitMQ.
In addition, my client wants to test this feature without using quorum queues at first.

docs/getting-started/backends-and-brokers/rabbitmq.rst
thedrow: Because my client wants to test this feature without using quorum queues at first.
I think we can assume developers see this warning and that's just fine by me.

t/smoke/tests/quorum_queues/test_native_delayed_delivery.py
thedrow: I tried moving this to a fixture but I got an error due to a bug in pytest-celery.
I think this doesn't matter much.

t/smoke/tests/quorum_queues/test_native_delayed_delivery.py
thedrow: I tried moving this to a fixture but I got an error due to a bug in pytest-celery.
I think this doesn't matter much.

t/smoke/tests/quorum_queues/test_native_delayed_delivery.py
thedrow: I tried moving this to a fixture but I got an error due to a bug in pytest-celery.
I think this doesn't matter much.

t/smoke/tests/quorum_queues/test_native_delayed_delivery.py
thedrow: I tried moving this to a fixture but I got an error due to a bug in pytest-celery.
I think this doesn't matter much.

Nusnus (changes requested):
Reviewed again. Don’t forget to rebase on `main`.
See new comments.

celery/app/defaults.py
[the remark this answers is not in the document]
thedrow: I changed the default to quorum.

celery/worker/consumer/tasks.py
thedrow: I removed the warning.

docs/getting-started/backends-and-brokers/rabbitmq.rst
thedrow: Native delayed delivery is now enabled automatically when quorum queues are detected.

thedrow:
@Nusnus I addressed most of your concerns excluding the bug which I don't know how to fix.
Please re-review.

Nusnus:
> @Nusnus I addressed most of your concerns excluding the bug which I don't know how to fix.
> Please re-review.

Thanks! 🙏
I'll review it again tomorrow/next day!

auvipy (approved):

celery/worker/consumer/delayed_delivery.py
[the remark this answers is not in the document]
Nusnus: > We can't tell that we're in the middle of a failover, can't we?

Good point.

> That will be most cases.

As an optimistic, I want to believe; it also makes sense as you present it.

—

We’ve invested a lot of effort on this issue so far, so we can go with our current fix and let the community report if this becomes an issue.

I will tag you if I see anything related, for now, we will assume low probability.

Thank you!

docs/getting-started/backends-and-brokers/rabbitmq.rst
```diff
@@ -172,3 +172,67 @@ but rather use the :command:`rabbitmqctl` command:
...
+or terminated so it won't work when Quorum Queues are detected.
+
+Similarly, the :setting:`worker_enable_prefetch_count_reduction` setting will be a no-op even when set to ``True``
+when Quorum Queues are detected.
+
+In addition, :ref:`ETA/Countdown <calling-eta>` will block the worker when received until the ETA arrives since
+we can no longer increase the prefetch count and fetch another task from the queue.
+
+In order to properly schedule ETA/Countdown tasks we automatically detect if quorum queues are used
+and in case they are, Celery automatically enables :ref:`Native Delayed Delivery <native-delayed-delivery>`.
```
Nusnus, line 224: Beautiful 😍

Nusnus (approved):
Well done @thedrow 👏
Thank you for attending to all of the review issues during the work on this new core mechanism in Celery. The community is going to appreciate it a lot, trust me bro 😜

O25Z

PR #9213, opened 2024-08-27
Added Soft Shutdown Mechanism

Nusnus:
Celery v5.5 introduces a new concept in the worker-stopping mechanism, the “Soft Shutdown”.
The soft shutdown provides a time-limited warm shutdown that is initiated just before the cold shutdown. This allows for a slight delay before the cold shutdown, which improves the gracefulness of the worker teardown procedure. It is most useful with Redis and SQS due to their visibility-timeout mechanism, as it allows the worker enough time to acknowledge successful messages during shutdown and restore the visibility timeout to zero (``[WARNING/MainProcess] Restoring N unacknowledged message(s)``) before the worker process terminates completely. Another benefit is allowing a worker who runs both short and long tasks to finish a few more short tasks before re-queuing the remaining long-running tasks. You will need to experiment to find the right delay value, recommended values can be 10, 30, 60 seconds. Too high value can lead to a long waiting time before the worker is terminated and trigger a KILL signal to terminate the worker by the host system forcefully.

The soft shutdown is disabled by default to maintain backward compatibility with the cold shutdown terminating immediately. To enable it, set the [worker_soft_shutdown_timeout](https://celery--9213.org.readthedocs.build/en/9213/userguide/configuration.html#worker-soft-shutdown-timeout) to the desired delay. The soft shutdown will be initiated with `SIGQUIT` (or `SIGTERM` if `REMAP_SIGTERM=SIGQUIT`), and then the cold shutdown will be initiated automatically after the soft shutdown timeout expires.

To learn more about the different shutdown types, see [new docs](https://celery--9213.org.readthedocs.build/en/9213/userguide/workers.html#worker-shutdown).

In addition, #8282 has been fixed with a new “Hard Shutdown” which is helpful for development purposes where the worker is stuck. You can now spam Ctrl+C to force the process to die, which was previously bugged if the warm shutdown was already initiated.

Lastly, the REMAP_SIGTERM=SIGQUIT “hidden feature” was documented + smoke tested.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -157,7 +157,7 @@ jobs:
         -   name: >
                 Run tox for
                 "${{ matrix.python-version }}-smoke"
-            timeout-minutes: 20
+            timeout-minutes: 30
             run: >
                 tox --verbose --verbose -e
                 "${{ matrix.python-version }}-smoke" -- -n auto -k failover
@@ -231,7 +231,7 @@ jobs:
         -   name: >
                 Run tox for
                 "${{ matrix.python-version }}-smoke"
-            timeout-minutes: 20
+            timeout-minutes: 30
             run: >
                 tox --verbose --verbose -e
                 "${{ matrix.python-version }}-smoke" -- -n auto -k stamping
@@ -268,7 +268,7 @@ jobs:
         -   name: >
                 Run tox for
                 "${{ matrix.python-version }}-smoke"
-            timeout-minutes: 20
+            timeout-minutes: 30
             run: >
                 tox --verbose --verbose -e
                 "${{ matrix.python-version }}-smoke" -- -n auto -k test_canvas.py
@@ -305,7 +305,7 @@ jobs:
         -   name: >
                 Run tox for
                 "${{ matrix.python-version }}-smoke"
-            timeout-minutes: 20
+            timeout-minutes: 30
             run: >
                 tox --verbose --verbose -e
                 "${{ matrix.python-version }}-smoke" -- -n auto -k test_consumer.py
@@ -342,7 +342,7 @@ jobs:
         -   name: >
                 Run tox for
                 "${{ matrix.python-version }}-smoke"
-            timeout-minutes: 20
+            timeout-minutes: 30
             run: >
                 tox --verbose --verbose -e
                 "${{ matrix.python-version }}-smoke" -- -n auto -k test_control.py
@@ -379,7 +379,7 @@ jobs:
         -   name: >
                 Run tox for
                 "${{ matrix.python-version }}-smoke"
-            timeout-minutes: 20
+            timeout-minutes: 30
             run: >
                 tox --verbose --verbose -e
                 "${{ matrix.python-version }}-smoke" -- -n auto -k test_signals.py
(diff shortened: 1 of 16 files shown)


fjsj:
@Nusnus the PR description LGTM, but I think the docs (and maybe the tests) should also explain/check the behavior when `task_reject_on_worker_lost` is True or False. The current [docs say](https://docs.celeryq.dev/en/stable/userguide/configuration.html#task-reject-on-worker-lost):
> Even if [task_acks_late](https://docs.celeryq.dev/en/stable/userguide/configuration.html#std-setting-task_acks_late) is enabled, the worker will acknowledge tasks when the worker process executing them abruptly exits or is signaled (e.g., KILL/INT, etc).
> Setting this to true allows the message to be re-queued instead, so that the task will execute again by the same worker, or another worker.

Nusnus:
> @Nusnus the PR description LGTM, but I think the docs (and maybe the tests) should also explain/check the behavior when `task_reject_on_worker_lost` is True or False. The current [docs say](https://docs.celeryq.dev/en/stable/userguide/configuration.html#task-reject-on-worker-lost):
> 
> > Even if [task_acks_late](https://docs.celeryq.dev/en/stable/userguide/configuration.html#std-setting-task_acks_late) is enabled, the worker will acknowledge tasks when the worker process executing them abruptly exits or is signaled (e.g., KILL/INT, etc).
> > Setting this to true allows the message to be re-queued instead, so that the task will execute again by the same worker, or another worker.

Thank you. Adding `reject_on_worker_lost` smoke tests can be a good contribution, but it is less related to the purpose of the changes of this PR. The soft shutdown (new) mechanism does not change based on the `reject_on_worker_lost` flag. The tests for the `reject_on_worker_lost` later _can_ also test the soft shutdown procedure though.

thedrow (changes requested):
Please ensure to raise test coverage before we merge this PR.

thedrow:
I think we should mention this setting [here](https://github.com/celery/celery/blob/main/docs/getting-started/backends-and-brokers/redis.rst) and [here](https://github.com/celery/celery/blob/main/docs/getting-started/backends-and-brokers/sqs.rst).

Nusnus:
> I think we should mention this setting [here](https://github.com/celery/celery/blob/main/docs/getting-started/backends-and-brokers/redis.rst) and [here](https://github.com/celery/celery/blob/main/docs/getting-started/backends-and-brokers/sqs.rst).

Good point, I'll add it

Nusnus:
> > I think we should mention this setting [here](https://github.com/celery/celery/blob/main/docs/getting-started/backends-and-brokers/redis.rst) and [here](https://github.com/celery/celery/blob/main/docs/getting-started/backends-and-brokers/sqs.rst).
> 
> Good point, I'll add it

Done.
Redis: https://celery--9213.org.readthedocs.build/en/9213/getting-started/backends-and-brokers/redis.html#soft-shutdown
SQS: https://celery--9213.org.readthedocs.build/en/9213/getting-started/backends-and-brokers/sqs.html#caveats

Nusnus:
[Celery v5.5.0b3](https://github.com/celery/celery/releases/tag/v5.5.0b3) released.
@fjsj FYI

PR #9227, opened 2024-09-02
Show broker_connection_retry_on_startup warning only if it evaluates as False

### LINKED ISSUE #9225: Confusing warning about `broker_connection_retry_on_startup`
# Checklist

  for existing proposed fixes.

#### Related Issues

- None

#### Possible Duplicates

- None

# Description

We just update Celery and we get this warning:

```
python-3.11/lib/python3.11/site-packages/celery/worker/consumer/consumer.py:508: 
CPendingDeprecationWarning: The broker_connection_retry configuration setting will no longer determine
whether broker connection retries are made during startup in Celery 6.0 and above.
If you wish to retain the existing behavior for retrying connections on startup,
you should set broker_connection_retry_on_startup to True.
```

Since in our configuration we didn't touch either (`broker_connection_retry` nor `broker_connection_retry_on_startup`), it took me some time to understand what is going on.

Also, from the message it looks like that the default for `broker_connection_retry_on_startup` will be `False`, but the [docs](https://docs.celeryq.dev/en/stable/userguide/configuration.html#broker-connection-retry-on-startup) specify that's enabled by default.

# Suggestions

Assuming that the default in Celery 6 will be to retry every time (both `True` by default), I think that the deprecation warning should not appear if `broker_connection_retry is True`, as no change would result from the deprecation going into action.

To recap, that's my proposition:
- `broker_connection_retry` unset or `True`, `broker_connection_retry_on_startup` unset: no warning (everything evaluated as `True`) <= suggested change
- `broker_connection_retry=False`, `broker_connection_retry_on_startup` unset: warning as it is now
- both `broker_connection_retry` and `broker_connection_retry_on_startup` set: no warning, as it is now

Let me know what are your thoughts, I can prepare a little PR if you agree.

serl:
## Description

Fixes #9225. As said in the issue, the warning is confusing if the user is sticking to the defaults already (as there's nothing to do). With this PR, the warning is shown only if the configuration would actually change with the 6.0 update.

### DIFF
--- a/celery/worker/consumer/consumer.py
+++ b/celery/worker/consumer/consumer.py
@@ -505,13 +505,14 @@ class Consumer:
             # to determine whether connection retries are disabled.
             retry_disabled = not self.app.conf.broker_connection_retry
 
-            warnings.warn(
-                CPendingDeprecationWarning(
-                    f"The broker_connection_retry configuration setting will no longer determine\n"
-                    f"whether broker connection retries are made during startup in Celery 6.0 and above.\n"
-                    f"If you wish to retain the existing behavior for retrying connections on startup,\n"
-                    f"you should set broker_connection_retry_on_startup to {self.app.conf.broker_connection_retry}.")
-            )
+            if retry_disabled:
+                warnings.warn(
+                    CPendingDeprecationWarning(
+                        "The broker_connection_retry configuration setting will no longer determine\n"
+                        "whether broker connection retries are made during startup in Celery 6.0 and above.\n"
+                        "If you wish to refrain from retrying connections on startup,\n"
+                        "you should set broker_connection_retry_on_startup to False instead.")
+                )
         else:
             if self.first_connection_attempt:
                 retry_disabled = not self.app.conf.broker_connection_retry_on_startup
--- a/t/unit/worker/test_consumer.py
+++ b/t/unit/worker/test_consumer.py
@@ -452,12 +452,12 @@ class test_Consumer(ConsumerTestCase):
         c.app.conf.broker_connection_retry_on_startup = broker_connection_retry_on_startup
         c.app.conf.broker_connection_retry = broker_connection_retry
 
-        if broker_connection_retry_on_startup is None:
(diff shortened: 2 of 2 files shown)


Nusnus (changes requested):
Please fix failing tests, it appears to be a real bug.

auvipy:
the test seems passing, isn't it? what is actually failing?

Nusnus:
> the test seems passing, isn't it? what is actually failing?

Ohhh.. now I see where I commented this hahah
It was a mistake. I typed it from my phone in the wrong PR 😅

auvipy (approved):

celery/worker/consumer/consumer.py
```diff
@@ -505,13 +505,14 @@ def _error_handler(exc, interval, next_step=CONNECTION_RETRY_STEP):
...
-                CPendingDeprecationWarning(
-                    f"The broker_connection_retry configuration setting will no longer determine\n"
-                    f"whether broker connection retries are made during startup in Celery 6.0 and above.\n"
-                    f"If you wish to retain the existing behavior for retrying connections on startup,\n"
-                    f"you should set broker_connection_retry_on_startup to {self.app.conf.broker_connection_retry}.")
-            )
+            if retry_disabled:
+                warnings.warn(
+                    CPendingDeprecationWarning(
+                        "The broker_connection_retry configuration setting will no longer determine\n"
```
auvipy, line 511: should we keep using the f"string" here as before?

serl: You mean with the `{self.app.conf.broker_connection_retry}`? I don't think so, given that I also changed "retain the existing behavior" to "refrain from retrying connections".
Actually, I copied the warning in the [doc](https://docs.celeryq.dev/en/stable/userguide/configuration.html#broker-connection-retry), I think having the exact same words here and there make a lot of sense

(Or if you were just talking about the `f` before the `"`, the linter complained about it being useless given that there's no variables inside anymore)

thedrow (approved):
LGTM

Issue #1810, opened 2014-01-21
ETA tasks lost with redis


scott-coates:
I've been trying for a couple of weeks to fix this issue I'm getting where eta tasks are being intermittently dropped.

```
software -> celery:3.1.8 (Cipater) kombu:3.0.10 py:2.7.4
            billiard:3.3.0.14 py-amqp:1.4.1
platform -> system:Linux arch:64bit, ELF imp:CPython
loader   -> celery.loaders.default.Loader
settings -> transport:amqp results:disabled
```

I'm using redis for the transport.

I've narrowed it down and can reproduce it. When I open up flower, I see several (20+) tasks scheduled. On Heroku, I issue the `SIGTERM` command and the worker begins to shutdown. If I issue this shutdown while tasks are still running, then the issue will arrise. 

Upon booting up the worker, I can go into flower and will see 0 (or a few) of my eta tasks. I believe these scheduled tasks which reside in the `unacked` HASH are never successfully written to the `celery` HASH in redis.

I am sure the redis instance is not going down abruptly. I am using RedisLive to monitor it and it has been up for several days. Similarily, I'm sure Heroku is not going down or running out of memory.

Going through the code, I believe this is an issue with kombu retrieving the tasks from `unacked` and celery not putting them back into `celery`.

morrissimo:
Saw this behavior today:
 1. task enqueued with countdown=30:
```python
my_task.apply_async((arg1, arg2, arg3), countdown=30)
```
 2. celery restart issued
 3. task never runs :/

Versions and settings:
* celery 3.1.11
* django 1.6.11
* redis 2.8.19 (on AWS ElastiCache)
* celery settings:
```
CELERY_ACKS_LATE = True
CELERY_RESULT_BACKEND = <redis_url>
BROKER_URL = <redis_url>
BROKER_TRANSPORT_OPTIONS = {'fanout_prefix': True}
CELERYD_HIJACK_ROOT_LOGGER = False
CELERYD_TASK_TIME_LIMIT = 300 
CELERYD_SOFT_TASK_TIME_LIMIT = 250  # Keep this less than the hard limit
CELERY_ALWAYS_EAGER = True
CELERY_ACCEPT_CONTENT = ['pickle', 'json', 'msgpack', 'yaml']
CELERY_SEND_TASK_ERROR_EMAILS = True
```

Those commenters that switched to RabbitMQ: was that the only change you made to address this issue? Have you seen any more occurrences since you made that switch?

[...]

PR #9231, opened 2024-09-03
Added worker_enable_soft_shutdown_on_idle

Nusnus:
Allows the [soft shutdown](https://docs.celeryq.dev/en/latest/userguide/workers.html#soft-shutdown) to be enabled even if no tasks are running.
This is useful when ETA tasks were received but still didn’t start and there’s a cold shutdown.
The soft shutdown timer should allow enough time for ETA tasks to be re-queued, just like with normal running tasks during soft shutdown that were canceled when the timeout expired.

Without enabling this flag but still setting a soft shutdown timeout, when ETA tasks are received, and there’s a cold shutdown and there are no tasks running, the worker will terminate immediately and might not re-queue the ETA tasks on time. This may be relevant for #1810 (and https://github.com/celery/kombu/issues/533).

### DIFF
--- a/celery/app/defaults.py
+++ b/celery/app/defaults.py
@@ -310,6 +310,7 @@ NAMESPACES = Namespace(
             False, type='bool'
         ),
         soft_shutdown_timeout=Option(0.0, type='float'),
+        enable_soft_shutdown_on_idle=Option(False, type='bool'),
         concurrency=Option(None, type='int'),
         consumer=Option('celery.worker.consumer:Consumer', type='string'),
         direct=Option(False, type='bool', old={'celery_worker_direct'}),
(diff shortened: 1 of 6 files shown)


kbiedrzycki:
@Nusnus will this really address issue with ETA task loss? I thought that they are not being a part of `active_requests` (https://github.com/celery/celery/pull/9213/files#diff-15905920605b6295205bfe3661dec9557f310e0cad1d14d17fe060e895323cd0R739), so they won't be really cancelled properly. Why would that PR help with ETA task loss?

Nusnus:
> @Nusnus will this really address issue with ETA task loss? I thought that they are not being a part of `active_requests` (https://github.com/celery/celery/pull/9213/files#diff-15905920605b6295205bfe3661dec9557f310e0cad1d14d17fe060e895323cd0R739), so they won't be really cancelled properly. Why would that PR help with ETA task loss?

For not being part of the `active_requests`, the soft shutdown assumes there are no running tasks and skips the waiting time of the cold shutdown. The cold shutdown by itself does cancel the ETA tasks, but without this flag enabled, it will still shut down too fast for the message restoration to complete.

PR #9241, opened 2024-09-06
fixing rpc message reporting

### LINKED ISSUE #4084: rpc backend doesn't seem to report message status correctly.
This is version 4.0.2
If I use rabbitmq, the rpc result backend, and a custom queue the message's status never seems to change and stays 'PENDING', even if the logging in the worker reports that it has been successfully executed until I do a get() of some sorts when the status changes to 'SUCCESS'. When I change the backend to amqp this the system works as expected, giving SUCCESS before doing the get(). Also the redis backend doesn't give this problem, so it seems to be rpc specific.

Note that not setting a custom queue, so using the default also works as expected just like the other backends!

I've got in my tasks.py:
```python
# tasks.py
from celery import Celery

app = Celery('tasks', 
broker='amqp://',
backend='rpc://' # here I use the rpc backend
)
@app.task
def test_task(s):
    return s
```
I start this with `celery -A tasks worker --loglevel=info -Q myqueue

On the other side I do:
```python
>>> import tasks
>>> result = tasks.test_task.apply_async(queue='myqueue', args=['blaat',]) # set custom queue

>>> result.ready() # worker already executed, so should give True
False
>>> result.status # with RPC backend it returned False while it should have been True in previous statement
'PENDING'
>>> print(result.get())
'This is a string'
>>> result.ready()
True
>>> result.status
'SUCCESS'
```

If I start this qithout the -Q option: 'celery -A tasks worker --loglevel=info' I get:
```python
>>> import tasks
>>> result = tasks.test_task.apply_async(args=['blaat',]) # set no custom queue

>>> result.ready() # worker already executed, so should give True
True
>>> result.status # previous seemed to work
'SUCCESS'
>>> print(result.get()) # SUCCESS was also reported
'This is a string'
>>> result.ready()
True
>>> result.status
'SUCCESS'
```

Thesnowyegret:
when result.get() is used it creates a consumer that is supposed to be stopped when the on_result_fulfilled() is called, but because the method overrides the original implementation with a pass, this consumer is never stopped, and the AsyncResult._get_task_meta() stops working.

Closes #4084

### DIFF
--- a/celery/backends/rpc.py
+++ b/celery/backends/rpc.py
@@ -184,11 +184,6 @@ class RPCBackend(base.Backend, AsyncBackendMixin):
         # By default we don't have to declare anything when sending a result.
         pass
 
-    def on_result_fulfilled(self, result):
-        # This usually cancels the queue after the result is received,
-        # but we don't have to cancel since we have one queue per process.
-        pass
-
     def as_uri(self, include_password=True):
         return 'rpc://'
 


auvipy (changes requested):
it would be nice to have some tests to avoid regression

Thesnowyegret:
> it would be nice to have some tests to avoid regression

I only started working with celery a few weeks ago and I am not familiar with the testing framework of the project. I only traced and fixed the bug because it blocked progress in a project I am working on. 

an adequate test case is described in my comment in https://github.com/celery/celery/issues/4084#issuecomment-2333665432 

Unfortunately I don't have the time to also study and understand the testing framework that is used in the project, I would be happy to do it if you can guide me through the process of doing it.

thedrow:
The [integration tests](https://github.com/celery/celery/actions/runs/10810128849/job/29987040992?pr=9241#step:7:1076) are failing with this change. It seems like overriding `on_result_fulfilled()` has some significance.

Thesnowyegret:
it seems this change, only works with AsyncResult, and it destroys the functionality of the GroupResult.

thedrow:
Is there a workaround that will resolve both the bug and avoid the issue with GroupResult?

PR #9173, opened 2024-08-04
Add check for soft_time_limit and time_limit values

ashm-dev:
*Note*: Before submitting this pull request, please review our [contributing
guidelines](https://docs.celeryq.dev/en/main/contributing.html).

## Description
Fixes #9125

PR #9246, opened 2024-09-09
Correct the error description in exception message when validate soft_time_limit

narasux:
## Description

I noticed that in PR #9173 a check was added to ensure that `soft_time_limit `must be less than or equal to `time_limit`. However, the current exception message conflicts with the checking logic and may mislead Celery users. So I submitting a PR to correct these.

### DIFF
--- a/celery/app/task.py
+++ b/celery/app/task.py
@@ -543,7 +543,7 @@ class Task:
             TypeError: If not enough arguments are passed, or too many
                 arguments are passed.  Note that signature checks may
                 be disabled by specifying ``@task(typing=False)``.
-            ValueError: If soft_time_limit and time_limit both set
+            ValueError: If soft_time_limit and time_limit both are set
                 but soft_time_limit is greater than time_limit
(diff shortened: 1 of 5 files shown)


Nusnus (approved):
Good catch!
We missed it during review.

Thanks!

celery/app/task.py
```diff
@@ -543,8 +543,8 @@ def apply_async(self, args=None, kwargs=None, task_id=None, producer=None,
             TypeError: If not enough arguments are passed, or too many
                 arguments are passed.  Note that signature checks may
                 be disabled by specifying ``@task(typing=False)``.
-            ValueError: If soft_time_limit and time_limit are set,
-                and soft_time_limit is less than time_limit
+            ValueError: If soft_time_limit and time_limit both set
```
auvipy, line 546:
```suggestion
            ValueError: If soft_time_limit and time_limit both are set
```

auvipy (changes requested):

Nusnus (approved):

PR #9252, opened 2024-09-10
Reduced Smoke Tests to min/max supported python (3.8/3.12)

Nusnus:


.github/workflows/python-package.yml
```diff
@@ -133,7 +133,7 @@ jobs:
     strategy:
         fail-fast: false
         matrix:
-            python-version: ['3.8', '3.9', '3.10', '3.11', '3.12']
+            python-version: ['3.8', '3.12']
```
auvipy, line 136: smart move

auvipy (approved):

.github/workflows/python-package.yml
Nusnus: Thanks bro!
I’m dying inside from all of these CI failures haha.. We need a clean CI and I won’t let it go until we have it 💪

PR #9263, opened 2024-09-13
Run all smoke tests CI jobs together

Nusnus:
With Blacksmith and using only min/max Python versions for the smoke tests, we can undo some optimizations I did before to try and make it more stable.


Nusnus:
No false negatives.. cool

Issue #7651, opened 2022-07-25
Visibility timeout set in celery config does not work, celery still using the default 1 hour 


la-luo-unit21:
# Checklist

- [x] I have checked the [issues list](https://github.com/celery/celery/issues?utf8=%E2%9C%93&q=is%3Aissue+label%3A%22Category%3A+Documentation%22+)
  for similar or identical bug reports.
- [x] I have checked the [pull requests list](https://github.com/celery/celery/pulls?q=is%3Apr+label%3A%22Category%3A+Documentation%22)
  for existing proposed fixes.
- [x] I have checked the [commit log](https://github.com/celery/celery/commits/master)
  to find out if the bug was already fixed in the master branch.
- [x] I have included all related issues and possible duplicate issues in this issue
      (If there are none, check this box anyway).

#### Related Issues

Here is how we write the configs file:
```
result_expires = 60
task_acks_late = True  # enables retry on task failure
worker_prefetch_multiplier = 1  # reserving one task at a time
accept_content = ['json']
result_extended = True
broker_transport_options = {
    'queue_order_strategy': 'round_robin',
    'visibility_timeout': 7200  # by seconds, 2 hours
}
task_send_sent_event = True
```
Here is how we configure the worker:
```
        celery = Celery(
            __name__,
            broker=config.CELERY_BROKER_URL,
            backend=config.CELERY_RESULT_BACKEND
        )

        os.environ.setdefault(
            'CELERY_CONFIG_MODULE',
            'celeryconfigs.default'
        )
        celery.config_from_envvar('CELERY_CONFIG_MODULE')
        celery.conf.ONCE = {
            'backend': 'celery_once.backends.Redis',
            'settings': {
                'url': config.CELERY_BROKER_URL,
                'default_timeout': 60 * 60
            }
        }
```
However, tasks running long still get redelivered after 1 hour as the default visibility timeout, not 2 hours as we set. We are following the docs https://docs.celeryq.dev/en/stable/userguide/configuration.html#broker-transport-options What could we possibly miss?
#### Possible Duplicates

- None

# Description

# Suggestions

andyzheung:
do you solve this problem? @la-luo-unit21

la-luo-unit21:
> do you solve this problem? @la-luo-unit21

Hi we are still experiencing this and see if someone can shed some light here

auvipy:
https://github.com/celery/celery/issues/7696 is a duplicate

auvipy:
[...]

Issue #5935, opened 2020-01-24
Long running jobs redelivering after broker visibility timeout with celery and redis


sujithgangaraju:
I am using celery 4.3 + Redis + flower. I have a few long-running jobs with acks_late=True and task_reject_on_worker_lost=True. I am using celery grouping to group jobs run parallelly and append result and use in parent job.

In this scenario, my few jobs will run more than an hour, after every one hour the same child jobs are redelivering again to the worker.

The sample jobs as below.
```
@app.task(queue='q1', bind=True, acks_late=True, task_reject_on_worker_lost=True, max_retries=3)
def job_1(self):
  do_something()
  task_group = group(job_2.s(batch) for batch in range(0, len([1,2,3,4,5,6]), 3))
  result_group = task_group.apply_async()

@app.task(queue='q1', bind=True, acks_late=True, task_reject_on_worker_lost=True, max_retries=3)
def job_2(self, batch):
  do_something()
  return result
```
The above job_2 will run more than hour and after one hour the same is redelivering again to the worker.

My celery setup and config as shown below:
```
c = Celery(app.import_name,
           backend=app.config['CELERY_RESULT_BACKEND'],
           broker=app.config['CELERY_BROKER_URL'])
config.py

CELERY_BROKER_URL = os.environ['CELERY_BROKER_URL']
CELERY_RESULT_BACKEND = os.environ['CELERY_RESULT_BACKEND']
CELERY_BROKER_TRANSPORT_OPTIONS = {'visibility_timeout': 36000}
```
I tried to increase visibility timeout to 10 hours after the redelivering issue like above in configuration. but it looks like not working.

Please help with this issue and let me know if any solution.

auvipy:
try celery==4.4.0 & report again

sujithgangaraju:
Sure will try and let you know.

With the same configurations ? Or any change in the above ?

sujithgangaraju:
@auvipy I updated the version to 4.4.0 and i see the issue remains, with acks_late=True the job is redelivering after one hour even I increase visibility_timeout to 10hours.

```
@app.task(queue='test_task', bind=True, acks_late=True, task_reject_on_worker_lost=True, max_retries=3)
def test_long_run(self):
    print('came here')
    n = 8
    for i in range(0,n):
        print(i)
        sleep(900)
    return "completed job"
```
Please find the screenshot and help on this issue if any soultion and resolve this.

yanwencheng:
[...]

PR #9264, opened 2024-09-13
Improve documentation on visibility timeout

kylez-ithaka:
## Description

Updated documentation to make it more obvious and clear how to set visibility timeout for Redis broker.

Need to set ALL the values (not clear in documentation still) for it to take effect, and conflicts between apps cause issues

Per https://github.com/celery/celery/issues/7651 and https://github.com/celery/celery/issues/5935


Nusnus (approved):
Important fix!

Thank you 🙏

PR #9271, opened 2024-09-15
Improved Redis container stability in the smoke tests

Nusnus:
My goal with all of the recent smoke test changes is to eliminate the false positives and allow a stable and deterministic CI to maintain a 100% passing rate on `main.`

Can’t guarantee this is the last fix, but I do hope so..

### DIFF
--- a/t/smoke/conftest.py
+++ b/t/smoke/conftest.py
@@ -3,7 +3,7 @@ import os
 import pytest
 from pytest_celery import (LOCALSTACK_CREDS, REDIS_CONTAINER_TIMEOUT, REDIS_ENV, REDIS_IMAGE, REDIS_PORTS,
                            RedisContainer)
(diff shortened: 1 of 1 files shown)


t/smoke/conftest.py
```diff
@@ -54,6 +54,7 @@ def default_worker_tasks(default_worker_tasks: set) -> set:
     network="{default_pytest_celery_network.name}",
     wrapper_class=RedisContainer,
     timeout=REDIS_CONTAINER_TIMEOUT,
+    command=fxtr("default_redis_broker_command"),
```
Nusnus, line 57: The `default_redis_broker_command` fixture configures Redis to boot with proper configurations.
Without it, the default config will collapse under the heavy use of the CI

auvipy (approved):

auvipy (approved):

PR #9294, opened 2024-09-17
Updated quorum queue detection to handle multiple broker instances

bkienker:
This fixes a small issue with #6067 where a list of brokers (instead of a single string) would raise an exception during the `startswith()` call. Now it will check the connection driver type directly.

I also updated the unit tests as needed.

Ref discussion #9289


Nusnus:
Did it work with your failover broker urls?

bkienker:
> Did it work with your failover broker urls?

Yes it did, as soon as I applied the patch the worker started up normally, including the Global QoS warning in the logs.

PR #9309, opened 2024-09-21
Added Python 3.13 Support

Nusnus:
Let’s see if it works out of the box


Nusnus:
Appears related: https://github.com/celery/billiard/issues/403

PR #9333, opened 2024-10-02
Update user guide to recommend delay_on_commit

browniebroke:
## Description

Update task user guide to mention the `delay_on_commit` shortcut that was added in https://github.com/celery/celery/pull/8491

### DIFF
--- a/docs/userguide/tasks.rst
+++ b/docs/userguide/tasks.rst
@@ -2043,8 +2043,8 @@ There's a race condition if the task starts executing
 before the transaction has been committed; The database object doesn't exist
 yet!
 
-The solution is to use the ``on_commit`` callback to launch your Celery task
-once all transactions have been committed successfully.
+The solution is to use
+:meth:`~celery.contrib.django.task.DjangoTask.delay_on_commit` instead:
 
 .. code-block:: python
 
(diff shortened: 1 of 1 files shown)


docs/userguide/tasks.rst
```diff
@@ -2054,7 +2054,31 @@ once all transactions have been committed successfully.
...
-        transaction.on_commit(lambda: expand_abbreviations.delay(article.pk))
+        expand_abbreviations.delay_on_commit(article.pk)
+        return HttpResponseRedirect('/articles/')
+
+This method was added in Celery 5.4. It's a shortcut that uses Django's
+``on_commit`` callback to launch your Celery task once all transactions
+have been committed successfully.
+
+With Celery <5.4
+~~~~~~~~~~~~~~~~
```
browniebroke, line 2065: Left the old instructions for earlier versions of Celery, but perhaps we can entirely remove the old way since the docs for the older versions are accessible as e.g. https://docs.celeryq.dev/en/v5.3.6/userguide/tasks.html

Nusnus: It’s not bad to say that in an older version it behaves differently, but I agree with your POV/logic.

Nusnus (approved):
As I see no problem with leaving the comment about older versions, I approve this change.
@browniebroke LMK if you want to commit a cleanup. If not, I’ll merge the PR as-is.

browniebroke:
> As I see no problem with leaving the comment about older versions, I approve this change.

Cool, that's good with me then 👍🏻

PR #9380, opened 2024-10-29
Enable support for multiple inheritance when creating Task classes.

TheWangDaolong:
This pull request enhances the app.task decorator to allow the base parameter to accept a tuple of classes. This change increases the ability to inherit from multiple parent classes, simplifying the creation of complex tasks.

Background:
While implementing an abortable singleton task using celery.contrib.abortable.AbortableTask and celery_once.QueueOnce from https://github.com/cameronmaske/celery-once, I discovered that this change allows for a more straightforward implementation.

Before the Change:
```
class DoSomethingTask(AbortableTask, QueueOnce):
    def run(self, *args, **kwargs):
        pass

app.tasks.register(DoSomethingTask())
```
After the Change:
```
@app.task(bind=True, queue='test', base=(AbortableTask, QueueOnce))
def do_something(self, *args, **kwargs):
    pass
```

Thank you for considering this improvement!


auvipy (changes requested):
can you please also add tests to verify the change?

auvipy:
some sort of additional test coverage is needed for accepting this change

PR #9407, opened 2024-11-12
add performance optimizing bytecode

evildojo666:
This bytecode improves performance by over 5000% at runtime. For more information please review https://docs.python.org/3/library/dis.html


Nusnus:
> This bytecode improves performance by over 5000% at runtime. For more information please review https://docs.python.org/3/library/dis.html

I’d love to see this claim proven :)

PR #9432, opened 2024-11-25
Prevent event_mask from being overwritten.

Androidown:
## Description

Current implementation of _select_imp has a limitation that, for example, when same `fd` appears in both argument `writers` and `err`, the event_mask of `fd` will end up being `POLLERR` instead of `POLLOUT | POLLERR`.

This may lead to celery not consuming task after app restart.

If an error occurs while writing job to celery worker and at the same time celery app restarts, `AsyncPool._active_writers` will contain the writing job and method `AsyncPool._flush_writer` will be triggered, providing the same `fds` for both `writers` and `err` to `_select -> _select_imp`. If the worker proc is still alive and related fd is writable (i.e. no error), the variable `writable`, `readable` returned by `_select` call will both be empty and `again` will be 0, and the loop `while fds:` will never ends
```python
    def _flush_writer(self, proc, writer):
        fds = {proc.inq._writer}
        try:
            while fds:
                if not proc._is_alive():
                    break  # process exited
                readable, writable, again = _select(
                    writers=fds, err=fds, timeout=0.5,
                )
                if not again and (writable or readable):
                    try:
                        next(writer)
                    except (StopIteration, OSError, EOFError):
                        break
        finally:
            self._active_writers.discard(writer)
```

### DIFF
--- a/celery/concurrency/asynpool.py
+++ b/celery/concurrency/asynpool.py
@@ -116,7 +116,7 @@ if hasattr(select, 'poll'):
         fd_to_mask = {}
 
         if readers:
-            for fd in map(_ensure_integral_fd, writers):
+            for fd in map(_ensure_integral_fd, readers):
                 fd_to_mask[fd] = fd_to_mask.get(fd, 0) | POLLIN
         if writers:
             for fd in map(_ensure_integral_fd, writers):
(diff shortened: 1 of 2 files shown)


Nusnus (approved):
LGTM, thank you!

PR #9495, opened 2025-01-15
Fix typo for default of sig

daveisfera:
Fix a typo so the default of `sig` will match the uses of the function

### DIFF
--- a/celery/apps/worker.py
+++ b/celery/apps/worker.py
@@ -279,7 +279,7 @@ class Worker(WorkController):
         )
 
 
-def _shutdown_handler(worker: Worker, sig='TERM', how='Warm', callback=None, exitcode=EX_OK, verbose=True):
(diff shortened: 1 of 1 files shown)


Nusnus (commented):
Was there an issue / bug that needed this change?

Nusnus:
> Was there an issue / bug that needed this change?

Is this related?
https://github.com/celery/celery/issues/9505

daveisfera:
> > Was there an issue / bug that needed this change?
> 
> Is this related? #9505

I noticed it while looking into that issue, but it wasn't impacting anything I noticed because all existing callers I could see specify that value, so it was just a fix for completeness.

auvipy (changes requested):
some verification test possible?

daveisfera:
> some verification test possible?

All of the callers of this function are in this same file and you can see that they all have `SIGXXXX` for the value passed to `sig` so the default of `TERM` is incorrect and should either be changed to `SIGTERM` or removed

auvipy (approved):

Nusnus:
Thank you for the fix!

PR #9533, opened 2025-01-29
Use safe_repr for req.eta in logging extra context

xzkostyan:
## Description

Users can define more suitable `logging.Formatter` for logging in applications. `"data"` in `extra` should be safe for logging in different formats like keys `argsrepr` and `kwargsrepr`.

We use json formatter for logging and `json.dumps` fails with `TypeError: Object of type datetime is not JSON serializable`.

Of course we can deal with it with

```python

def custom_json(obj):
    if isinstance(obj, datetime):
        return obj.isoformat() 
    raise TypeError(f'Cannot serialize object of {type(obj)}')

json.dumps(..., default=custom_json)
```

But it's not consistent with `argsrepr` and `kwargsrepr`.


xzkostyan:
Is anything I can do with failed checks?

auvipy (changes requested):
you would need to add tests and fix the failing ones, due to this proposed change, also we have to figure out if this is the right choice we are going to merge

xzkostyan:
@auvipy did you to figured this out?

auvipy:
You have to make the CI green first for your PR

xzkostyan:
> also we have to figure out if this is the right choice we are going to merge

Why should I write test if this PR won't be merged?

auvipy:
I didn’t mean that, pr with tests have a higher chance to get merged as they are easier to review

PR #9534, opened 2025-01-30
Unroll group when chaining a group with a single item to another signature when using the | operator

omerkatzkcg:

## Description

Before this change the following group would be upgraded to a chord:

```python
>>> from celery import chain, group
>>> from tasks import add
>>> group(add.s(1, 1)) | add.s(2)
%add([add(1, 1)], 2)
```

After this change the group is correctly unrolled into a single signature:

``` python
>>> from celery import chain, group
>>> from tasks import add
>>> group(add.s(1, 1)) | add.s(2)
add(1, 1) | add(2)
```

This is a breaking change but it is a necessary one since chain(group(add.s(1,2)), add.s(1)) correctly unrolls the group into a single signature. The breaking change is documented.

### DIFF
--- a/celery/canvas.py
+++ b/celery/canvas.py
@@ -1583,6 +1583,10 @@ class group(Signature):
 
     def __or__(self, other):
         # group() | task -> chord
+        # If the group is unrolled, return a chain instead
+        g = maybe_unroll_group(self)
+        if not isinstance(g, group):
+            return g | other
         return chord(self, body=other, app=self._app)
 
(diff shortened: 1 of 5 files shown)


auvipy (changes requested):
There are integration tests failure and some warnings. please fix them for next round of review. I'm surprised to see this was forced merged before.

omerkatzkcg:
@auvipy Please re-run CI. The tests should pass now.

PR #9535, opened 2025-01-30
Replacing a task with a chain which contains a group now returns a result instead of hanging

omerkatzkcg:

## Description

The following task

```python
@shared_task(bind=True)
def replace_with_chain_which_contains_a_group(self):
    return self.replace(chain(add.s(1, 2), group(add.s(1), add.s(1))))
```

did not return a result. This PR fixes the issue and now a result is returned correctly instead of hanging until timeout occurs.


auvipy (changes requested):
Integration tests are failing here as well. can you fix them?

omerkatzkcg:
@auvipy Can you re-run CI please?

PR #9540, opened 2025-02-02
Update install-pyenv.sh

auvipy:
seems this part was not aligned with other updates


auvipy:
merged as all were green

PR #9545, opened 2025-02-04
Revert "[pre-commit.ci] pre-commit autoupdate"

Nusnus:
Reverts celery/celery#9542

Implicitly upgrades isort instead of via #9543 with proper tests.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -17,7 +17,7 @@ repos:
         exclude: ^celery/app/task\.py$|^celery/backends/cache\.py$
 
   - repo: https://github.com/codespell-project/codespell
-    rev: v2.4.1
+    rev: v2.4.0
     hooks:
       - id: codespell # See pyproject.toml for args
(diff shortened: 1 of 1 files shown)


Nusnus:
@auvipy FYI
Also, please consider every merge to `main`.
If it’s not critical, please add the 5.5.1 or 5.6.0 milestone on the issue and we’ll get to it after the release.
We’ll have a RC5 of course before the actual release due to recent events, but I want as little changes as possible even for pre-commit or stuff like that so I can manage reviewing every single change in the release and make sure nothing either slipped away or worse, we have introduced changes that should not have been part of 5.5. (Celery/Kombu)

Lastly, I want to avoid any changes that may affect our dev environment, like updating tox/pre-commit or whatever that can cause even the slightest delay (e.g, deps that dropped py38 and we’re still relying on 3.8).

Let me know if it makes/doesn’t make sense bro.

auvipy:
Interestingly I blocked some of this 😆

PR #9548, opened 2025-02-05
remove Python 3.8 from CI as EOL

auvipy:
After celery v5.5.0/v5.5.1 is released in PyPI


.github/workflows/python-package.yml
```diff
@@ -136,7 +136,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ['3.8', '3.12', '3.13']
+        python-version: ['3.9', '3.12', '3.13']
```
bashir-abdelwahed, line 139: why are we skipping testing for python 3.10/3.11 ?

bashir-abdelwahed (approved):

.github/workflows/python-package.yml
Nusnus: Due to instability with the CI. This is a technical debt we need to solve.

I might reconsider this though, good question :)

auvipy: it was due to reduce number of CI build. technically if something pass in python 3.9 and 3.13 it should pass in 3.10, 11 and 12... this was the reasoning that time

bashir-abdelwahed: why do we want to reduce CI ? is it because celery pays for it (to blacksmith) ?
If yes how much does it cost ?

Nusnus: > why do we want to reduce CI ? is it because celery pays for it (to blacksmith) ? If yes how much does it cost ?

We were reaching GitHub limits back then. Blacksmith unlocked those limits. I will handle this @bashir-abdelwahed, thank you for noticing!

auvipy: handled in another PR https://github.com/celery/celery/pull/9621

Nusnus: > Due to instability with the CI. This is a technical debt we need to solve.

@auvipy I prefer to handle it myself as I mentioned, because one of the reasons we did it was due to overly flaky state, requiring many re-runs on PRs etc., hurting our operational performance. Just reverting it will not be enough unfortunately. I want to fix the root cause of the flakyness and then enable it back so every CI run will consistently pass with rare cases of flakiness.

auvipy:
done in a separate PR

PR #9551, opened 2025-02-07
fix(crontab): resolve constructor type conflicts

pcrock-thmdo:
## Description

The `crontab` class constructor parameter types conflict with actual functionality. The parameter cron spec values are marked as `str`, but can actually be `int | str | Iterable[int]`.

This creates a type alias called `Cronspec` to represent the correct types, and then uses it to fix typing inconsistencies.

### DIFF
--- a/celery/schedules.py
+++ b/celery/schedules.py
@@ -4,9 +4,8 @@ from __future__ import annotations
 import re
 from bisect import bisect, bisect_left
 from collections import namedtuple
-from collections.abc import Iterable
 from datetime import datetime, timedelta, tzinfo
-from typing import Any, Callable, Mapping, Sequence, Union
+from typing import Any, Callable, Iterable, Mapping, Sequence, Union
 
 from kombu.utils.objects import cached_property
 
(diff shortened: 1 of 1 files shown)


celery/schedules.py
```diff
@@ -555,7 +557,7 @@ def roll_over() -> None:
     def __repr__(self) -> str:
         return CRON_REPR.format(self)
 
-    def __reduce__(self) -> tuple[type, tuple[str, str, str, str, str], Any]:
+    def __reduce__(self) -> tuple[type, tuple[Cronspec, Cronspec, Cronspec, Cronspec, Cronspec], Any]:
```
auvipy, line 560: should we update/add additional tests to verify this changes?

auvipy (changes requested):

celery/schedules.py
pcrock-thmdo: since no actual logic has changed (only type annotations), i personally don't see a reason to change any tests.

for what it's worth, i was prompted to change this method signature because mypy failed CI. so in a way, it is already being tested by mypy.

auvipy (approved):

PR #9552, opened 2025-02-07
Fix the send_task method when detecting if the native delayed delivery approach is available

mikhaillazko:
## Description

The native delayed delivery mechanism for RabbitMQ fails when setting `broker_url` with SSL. This occurs because an incorrect property was passed to the `detect_quorum_queues` method.

I didn't report this issue, no links to attach.

### DIFF
--- a/celery/app/base.py
+++ b/celery/app/base.py
@@ -832,8 +832,8 @@ class Celery:
         options = router.route(
             options, route_name or name, args, kwargs, task_type)
 
-        is_native_delayed_delivery = detect_quorum_queues(self,
-                                                          self.producer_pool.connections.connection.transport_cls)[0]
+        driver_type = self.producer_pool.connections.connection.transport.driver_type
+        is_native_delayed_delivery = detect_quorum_queues(self, driver_type)[0]
         if is_native_delayed_delivery and options['queue'].exchange.type != 'direct':
             if eta:
                 if isinstance(eta, str):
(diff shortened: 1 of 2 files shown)


Nusnus (commented):
Great find.
I'll review it when the CI fully passes.

Thank you!

celery/app/base.py
```diff
@@ -832,8 +832,8 @@ def send_task(self, name, args=None, kwargs=None, countdown=None,
         options = router.route(
             options, route_name or name, args, kwargs, task_type)
 
-        is_native_delayed_delivery = detect_quorum_queues(self,
-                                                          self.producer_pool.connections.connection.transport_cls)[0]
+        driver_type = self.producer_pool.connections.connection.transport.driver_type
```
auvipy, line 835: is it possible to add additional Integration tests for the change?

auvipy (approved):
some smoke tests are failing due to time out issues mainly, may be we could overlook them

Nusnus:
> some smoke tests are failing due to time out issues mainly, may be we could overlook them

I've started them again.

Take note stabilizing the smoke tests is on my checklist as well.

Nusnus (approved):
I remembered there was another place this was relevant, but it appears to be correct there already:
https://github.com/celery/celery/blob/b1b886d1dde479feaf54874a660ddec4a97e2442/celery/worker/consumer/delayed_delivery.py#L19

Great fix!
Thank you very much!

PR #9553, opened 2025-02-09
worker_max_memory_per_child: kilobyte is 1024 bytes

mksm:
## Description

The max RSS value, [provided](https://github.com/celery/billiard/blob/main/billiard/compat.py#L242-L279) and [used](https://github.com/celery/billiard/blob/main/billiard/pool.py#L378-L385) by billiard to determine if a worker process should be recycled based on `worker_max_memory_per_child` is in multiples of 1024 bytes, technically, kibibytes This PR clarifies this in the option's documentation, without introducing the `kibibyte` term, to avoid broader discussion at this point.

### DIFF
--- a/docs/userguide/configuration.rst
+++ b/docs/userguide/configuration.rst
@@ -3207,16 +3207,16 @@ it's replaced with a new one. Default is no limit.
 Default: No limit.
 Type: int (kilobytes)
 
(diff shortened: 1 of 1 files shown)


auvipy (approved):
make sense

PR #9576, opened 2025-02-24
Clearify docs about infinite max_retries for tasks

ya-pekatoros:
## Description

This changes makes docs more clear about infinity retries for tasks.

The changes add the example of passing max_retries=None to the task decorator to build the task with infinity retries. 

Also the changes clarify the docs string of the retry method of the Task removing "first" word, because to make you task retries forever you should only pass the max_retries=None to the Task class itself.


Nusnus (changes requested):
Waiting on https://github.com/celery/celery/pull/9577#pullrequestreview-2637298878

ya-pekatoros:
No need of changes because these changes are already in https://github.com/celery/celery/pull/9577

PR #9577, opened 2025-02-24
Improve docs about acks_on_failure_or_timeout

ya-pekatoros:
## Description

This changes makes docs more clear about infinity retries for tasks and acks_on_failure_or_timeout.

The changes add the example of passing max_retries=None to the task decorator to build the task with infinity retries.

Also the changes clarify the docs string of the retry method of the Task removing "first" word, because to make you task retries forever you should only pass the max_retries=None to the Task class itself.

Actually the value of attribute 'acks_on_failure_or_timeout' for Task is 'None' by default. 

Yes, if acks_late=False, this logic is disabled, but this information is already in docs. So I suggest to clarify the actual value of this attribute.

### DIFF
--- a/docs/userguide/configuration.rst
+++ b/docs/userguide/configuration.rst
@@ -610,12 +610,12 @@ has been executed, not *right before* (the default behavior).
 ``task_acks_on_failure_or_timeout``
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Default: Enabled
+Default: Disabled
 
 When enabled messages for all tasks will be acknowledged even if they
 fail or time out.
 
-Configuring this setting only applies to tasks that are
+This setting only applies to tasks that are
 acknowledged **after** they have been executed and only if
 :setting:`task_acks_late` is enabled.
 
--- a/celery/app/task.py
+++ b/celery/app/task.py
@@ -691,8 +691,8 @@ class Task:
                 this execution.  Changes to this parameter don't propagate to
                 subsequent task retry attempts.  A value of :const:`None`,
                 means "use the default", so if you want infinite retries you'd
-                have to set the :attr:`max_retries` attribute of the task to
-                :const:`None` first.
+                have to set the :attr:`max_retries` attribute of the task class to
+                :const:`None`.
             time_limit (int): If set, overrides the default time limit.
             soft_time_limit (int): If set, overrides the default soft
                 time limit.
--- a/docs/userguide/tasks.rst
+++ b/docs/userguide/tasks.rst
@@ -734,6 +734,17 @@ avoid having all the tasks run at the same moment. It will also cap the
 maximum backoff delay to 10 minutes. All these settings can be customized
 via options documented below.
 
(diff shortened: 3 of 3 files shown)


ya-pekatoros:
In addition, I wonder about actually ack in case of task fail. I see, what this attribute leads to call reject but with reque=False, and actually change the acknowledged flag to True in case on call "on_failure".

Nusnus (dismissed):
This PR appears to have duplicated / similar changes from #9576, so it needs to be cleared first. Please review the other PR and close it if not relevant, so I’ll know where to focus.

Thanks!

ya-pekatoros:
> This PR appears to have duplicated / similar changes from #9576, so it needs to be cleared first. Please review the other PR and close it if not relevant, so I’ll know where to focus.
> 
> Thanks!

Yes, if you agree with the proposal - I can merge both changes. 

And I also wonder about this question:
https://github.com/celery/celery/pull/9577#issuecomment-2678462259
If it's true, may be the acks_on_failure_or_timeout is not the best name of the attribute.

ya-pekatoros:
@Nusnus Hi again! I've closed other PR. Also, please, could you re-run checks above?

auvipy (commented):
are the failing tests related to this? may be not right?

auvipy (approved):

auvipy:
Thanks

docs/userguide/configuration.rst
```diff
@@ -610,12 +610,12 @@ has been executed, not *right before* (the default behavior).
 ``task_acks_on_failure_or_timeout``
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Default: Enabled
+Default: Disabled
 
 When enabled messages for all tasks will be acknowledged even if they
 fail or time out.
 
-Configuring this setting only applies to tasks that are
```
Nusnus, line 618: Was this phrasing confusing? What was wrong with it?

Nusnus (commented):
I want to avoid reverting this PR so I left a few comments to better understand the changes. Take note @ya-pekatoros as this PR was merged during a later RC phase and every change is super sensitive so I want to have very clear motivation for merged changes at the moment or I will be required to revert it to proceed with the release cycle.

I hope you understand, thank you.

docs/userguide/configuration.rst
ya-pekatoros: @Nusnus I think that 'this setting' is more concise and clear variant, because we apply settings. The setting is the subject here.

docs/userguide/configuration.rst
```diff
@@ -610,12 +610,12 @@ has been executed, not *right before* (the default behavior).
 ``task_acks_on_failure_or_timeout``
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Default: Enabled
+Default: Disabled
```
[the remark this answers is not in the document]
auvipy: ughs! thought just as an docs update. shouldn't hurry merging this

PR #9590, opened 2025-03-01
Fix tests with Click 8.2

cjwatson:
## Description

https://github.com/pallets/click/pull/2523 introduced changes to `click.testing.Result` that broke a few unit tests in celery.  Although this Click version hasn't been fully released yet, this adjusts Celery to work with both old and new versions.


Nusnus (approved):
Thank you!!

PR #9604, opened 2025-03-11
Replacing a task with a chain which contains a group now returns a result instead of hanging

auvipy:
## Description

The following task

```python
@shared_task(bind=True)
def replace_with_chain_which_contains_a_group(self):
    return self.replace(chain(add.s(1, 2), group(add.s(1), add.s(1))))
```

did not return a result. This PR fixes the issue and now a result is returned correctly instead of hanging until timeout occurs.


auvipy (commented):
may be we should also add some unit test for this?

auvipy:
@Nusnus boss, need some help here, can you point me where exactly to adjust the unit test for this :D forgot many thing

auvipy:
Integration tests and docker build failures are not related to this specific pr

Nusnus (commented):
@auvipy 
> @Nusnus boss, need some help here, can you point me where exactly to adjust the unit test for this :D forgot many thing

Np, I got your back!

Check this: https://github.com/celery/celery/blob/main/t/unit/tasks/test_canvas.py

auvipy:
I am going to fix this in upcoming commits

PR #9642, opened 2025-03-28
Add max retry check in task rejection to prevent infinite requeue

sandeep-kesarwani:

### DIFF
--- a/celery/worker/request.py
+++ b/celery/worker/request.py
@@ -19,7 +19,7 @@ from celery.app.task import Context
 from celery.app.trace import fast_trace_task, trace_task, trace_task_ret
 from celery.concurrency.base import BasePool
 from celery.exceptions import (Ignore, InvalidTaskError, Reject, Retry, TaskRevokedError, Terminated,
-                               TimeLimitExceeded, WorkerLostError)
+                               TimeLimitExceeded, WorkerLostError, MaxRetriesExceededError)
 from celery.platforms import signals as _signals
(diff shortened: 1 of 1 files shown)


auvipy:
can you please also link the related issues?

chrisbarber:
After exceeding the retries, will the task `on_failure` method end up getting invoked?

We would like to direct failures (including worker lost and hard timeout) to failure queues on redis that  are not listened to, sort of like dead letter.

I was also considering that there could be an option in Celery to specify a queue to requeue into when rejecting, rather than requeuing always to the same one. But if this change will mean we can access the task during `on_failure` then we can do a `retry` from there to a specific queue and not lose anything if the failure was part of a chain etc.

chrisbarber:
> After exceeding the retries, will the task `on_failure` method end up getting invoked?
> 
> We would like to direct failures (including worker lost and hard timeout) to failure queues on redis that are not listened to, sort of like dead letter.
> 
> I was also considering that there could be an option in Celery to specify a queue to requeue into when rejecting, rather than requeuing always to the same one. But if this change will mean we can access the task during `on_failure` then we can do a `retry` from there to a specific queue and not lose anything if the failure was part of a chain etc.

I tested by applying the body of the `reject` method from this PR to an overridden (custom) `Request` class.

The rejection count comes back as always 0 i.e. this PR seems to have no effect / not do what it claims to.

Nor does unconditionally running the `mark_as_failure` section (as if the rejection count had been exceeded) run the task `on_failure` method as I was interested in.

PR #9687, opened 2025-05-05
Make wheel metadata reproducible

cjwatson:
## Description

The various `Provides-Extra` sections in the built wheel's `METADATA` file were shuffled arbitrarily between different builds, which made it more difficult to detect other reproducibility issues (see https://reproducible-builds.org/).  There doesn't seem any reason for `EXTENSIONS` here to be a set; a tuple will work just as well, and adds an ordering guarantee.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -9,7 +9,7 @@ NAME = 'celery'
 
 # -*- Extras -*-
 
-EXTENSIONS = {
+EXTENSIONS = (
     'arangodb',
     'auth',
     'azureblockblob',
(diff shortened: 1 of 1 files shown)


Nusnus (approved):
The CI errors are not related, something happened on `main` with redis.

Thank you!
The PR makes sense

auvipy (approved):
yes, the redis v8 release might have broke the CI

Nusnus:
> yes, the redis v8 release might have broke the CI

redis v5.3.0/6.0.0 😅
https://pypi.org/project/redis/#history

PR #9702, opened 2025-05-12
feature: add celery_beat_task header in task options

alirafiei75:
## Description

This pull request adds a new custom header `celery_beat_task` to task options, enabling reliable identification of tasks originating from Celery Beat. This enhancement addresses the challenge discussed in one of django-celery-beat issues (number 854).

## Changes

* Added `celery_beat_task` header to task options when tasks are sent by Celery Beat
* Added unit tests for the new header

## Related Issues

https://github.com/celery/django-celery-beat/issues/854


simonepozzoli-pix4d:
I already see a `periodic_task_name` in the task headers when scheduling through celery beat.
Try with:
```python
periodic_task_name = getattr(self.request, "periodic_task_name", None)
```
The name of periodic tasks should be unique in celery beat, you can use it to fetch the entry from the db.

auvipy:
> I already see a `periodic_task_name` in the task headers when scheduling through celery beat. Try with:
> 
> ```python
> periodic_task_name = getattr(self.request, "periodic_task_name", None)
> ```
> 
> The name of periodic tasks should be unique in celery beat, you can use it to fetch the entry from the db.

can you please elaborate more?

simonepozzoli-pix4d:
> > I already see a `periodic_task_name` in the task headers when scheduling through celery beat. Try with:
> > ```python
> > periodic_task_name = getattr(self.request, "periodic_task_name", None)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The name of periodic tasks should be unique in celery beat, you can use it to fetch the entry from the db.
> 
> can you please elaborate more?

I'm using the database scheduler in my project. When a periodic task is scheduled, I can see a `periodic_task_name` header is set to it.
After looking at the code, it seems like it's only set by the DatabaseScheduler https://github.com/celery/django-celery-beat/blob/ff4a5f79a34e1e27d5222881ea3c74bbf0ecfdc7/django_celery_beat/schedulers.py#L89. If the standard Scheduler is used, the header is not present.

auvipy:
@alirafiei75 can you please check the comments here?

alirafiei75:
> @alirafiei75 can you please check the comments here?

@auvipy I reviewed the comments. As the commenter noted in their second reply, periodic_task_name is only present when using django-celery-beat's DatabaseScheduler.

It is not present when using standard core Celery Beat (e.g., configuring beat_schedule directly with default schedulers).

Setting celery_beat_task: True directly in core celery/beat.py ensures that all tasks dispatched by Beat, regardless of whether a third-party scheduler package like django-celery-beat is used, have a consistent way to be identified downstream.

auvipy (approved):

PR #9718, opened 2025-05-22
feat(cli): add --json option to celery report for structured JSON output

yiyu0x:
## Description

This pull request adds a new `--json` (`-j`) option to the `celery report` command, allowing users to output bugreport information in structured JSON format.

With this enhancement:
- Each line containing `->` is treated as a new key in the JSON object.
- The value after `->` (and any subsequent indented lines) is grouped into a list if multi-line, or as a string if single-line.
- Indented lines following a key are included as part of that key's value.
- Any lines not matching the above are grouped under a `details` key.
- Blank lines are ignored.

This feature improves the automation and integration of bug reporting, making it easier to parse and analyze Celery environment information programmatically.


auvipy (changes requested):
can you please check why the CI was failing?

PR #9746, opened 2025-06-07
Improve user-friendly error messages for result backend in CLI

rogerforlife:
1. Show clear error when result_backend is not configured or cannot connect (e.g., Redis not running)
2. Catch redis.exceptions.ConnectionError for better compatibility
3. Make CLI output more helpful for end users


auvipy (changes requested):
please add unit tests for the changes

PR #9752, opened 2025-06-10
remove futures extra as it is not needed in python 3

auvipy:


requirements/extras/thread.txt
```diff
@@ -1 +1 @@
-futures>=3.1.1; python_version < '3.0'
+
```
ulgens, line 1: Shouldn't we delete the file and remove references?

auvipy: yes

PR #9765, opened 2025-06-17
Add xfail test for default queue/exchange fallback ignoring task_default_* settings

maccinza:
## Description

This PR adds a regression-style test to demonstrate that Celery falls back to creating a celery queue of type classic and an exchange of type direct, even when task_default_queue_type and task_default_exchange_type are explicitly set.

The test is marked with @pytest.mark.xfail(strict=True) to:
  - Document the current undesired behavior
    - celery exchange defaults to type "direct" (instead of task_default_exchange_type)
    - celery queue defaults to type "classic" (instead of task_default_queue_type)
  - Automatically surface if the behavior changes in future versions
  
### Related Discussion

This issue was raised in the Celery GitHub discussion: https://github.com/celery/celery/discussions/9742

### Why it matters

When using quorum queues or non-direct exchanges (e.g. topic), users expect Celery to respect the configured defaults. This test makes the fallback behavior visible and prevents silent reintroduction if fixed upstream.

Fixing this likely involves changes in Kombu’s handling of default queue declarations when tasks are published without explicit routing.

### DIFF
--- a/t/integration/test_rabbitmq_default_queue_type_fallback.py
+++ b/t/integration/test_rabbitmq_default_queue_type_fallback.py
@@ -1,0 +1,57 @@
+# t/integration/test_rabbitmq_default_queue_type_fallback.py
+
+import pytest
+import time
+from kombu import Connection
+from celery import Celery
+
+
+@pytest.fixture()
+def app():
+    return Celery(
+        "test_app",
+        broker="amqp://guest:guest@rabbit:5672//",
+        backend="redis://redis:6379/0",
+        include=["t.integration.test_rabbitmq_default_queue_type_fallback"],
+    )
+
+
+@pytest.fixture()
+def ping(app):
+    @app.task(name="ping")
+    def ping():
+        return "pong"
+    return ping
+
+
+@pytest.mark.amqp
+@pytest.mark.timeout(30)
+@pytest.mark.xfail(
+    reason=(
+        "Celery does not respect task_default_exchange_type/queue_type "
+        "when using implicit routing to the 'celery' queue. It creates "
+        "a classic queue and direct exchange instead."
+    ),
+    strict=True,
+)
+def test_fallback_to_classic_queue_and_direct_exchange(app, ping):
+    from celery.contrib.testing.worker import start_worker
+
+    # Start worker and submit task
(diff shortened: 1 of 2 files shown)


t/integration/test_rabbitmq_default_queue_type_fallback.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+@pytest.fixture()
+def app(rabbitmq):
+    wait_for_port(rabbitmq.hostname, rabbitmq.ports[5672])
+
+    return Celery(
+        "test_app",
+        broker=f"pyamqp://guest:guest@{rabbitmq.hostname}:{rabbitmq.ports[5672]}/",
+        backend="redis://redis:6379/0",  # You may also want to make redis dynamic if needed
```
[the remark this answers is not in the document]
auvipy: what do you think about this suggestion?

auvipy:
We need to figure out to fix this intermittent  issue later....

raise TimeoutError(f"Timeout of {timeout}s exceeded")
  E       pytest_docker_tools.exceptions.TimeoutError: Timeout of 60s exceeded
  
  .tox/3.9-smoke/lib/python3.9/site-packages/pytest_docker_tools/utils.py:37: TimeoutError

t/integration/test_rabbitmq_default_queue_type_fallback.py
```diff
@@ -0,0 +1,86 @@
+import socket
+import time
+
+import pytest
+from kombu import Connection
+
+from celery import Celery
+
+
+def wait_for_port(host, port, timeout=30.0):
```
auvipy:
```suggestion
def wait_for_port(host, port, timeout=60.0):
```

t/integration/test_rabbitmq_default_queue_type_fallback.py
```diff
@@ -0,0 +1,86 @@
...
+@pytest.fixture()
+def ping(app):
+    @app.task(name="ping")
+    def ping():
+        return "pong"
+    return ping
+
+
+@pytest.mark.amqp
+@pytest.mark.timeout(30)
```
auvipy:
```suggestion
@pytest.mark.timeout(60)
```

auvipy (changes requested):
how about increasing the timeout?

PR #9766, opened 2025-06-17
Adds integration test for chord_unlock bug when routed to quorum/topic queue

maccinza:
## Description
This PR introduces an integration test that reproduces the issue described in [Discussion #9742](https://github.com/celery/celery/discussions/9742), where:

celery.chord_unlock tasks are routed to a declared quorum queue (bound to a topic exchange), but are either never consumed or endlessly retried, with rare occasions when it is successfully run.

### ✅ Summary of the test:

Chords are submitted before any worker is running.
chord_unlock is explicitly routed to a quorum queue via a topic exchange.
A worker is later started that consumes from both the header and callback queues.
The test verifies that all chord results complete successfully.

### ❌ Observed behavior:

A significant number of chord results remain pending or fail due to the chord_unlock task being ignored or retried indefinitely.

### ✅ Expected behavior:

The chord_unlock task should be routed correctly and consumed by the worker since the queue is explicitly declared and consumed.


auvipy:
interestingly, the tests are passing!

maccinza:
> interestingly, the tests are passing!

I guess it is actually `XFailing`, but the issue is intermittent, it can rarely work as expected with all chords being correctly run.

t/integration/test_rabbitmq_chord_unlock_routing.py
```diff
@@ -0,0 +1,152 @@
+import time
```
auvipy, line 2:
```suggestion
import logging
import time
```

Issue #9792, opened 2025-07-03
DelayedDelivery Step Connection is left open without closing


bityob:
# Checklist

- [x] I have verified that the issue exists against the `main` branch of Celery.
- [ ] This has already been asked to the [discussions forum](https://github.com/celery/celery/discussions) first.
- [x] I have read the relevant section in the
  [contribution guide](https://docs.celeryq.dev/en/main/contributing.html#other-bugs)
  on reporting bugs.
- [x] I have checked the [issues list](https://github.com/celery/celery/issues?q=is%3Aissue+label%3A%22Issue+Type%3A+Bug+Report%22+-label%3A%22Category%3A+Documentation%22)
  for similar or identical bug reports.
- [x] I have checked the [pull requests list](https://github.com/celery/celery/pulls?q=is%3Apr+label%3A%22PR+Type%3A+Bugfix%22+-label%3A%22Category%3A+Documentation%22)
  for existing proposed fixes.
- [x] I have checked the [commit log](https://github.com/celery/celery/commits/main)
  to find out if the bug was already fixed in the main branch.
- [x] I have included all related issues and possible duplicate issues
  in this issue (If there are none, check this box anyway).
- [x] I have tried to reproduce the issue with [pytest-celery](https://docs.celeryq.dev/projects/pytest-celery/en/latest/userguide/celery-bug-report.html) and added the reproduction script below.

## Mandatory Debugging Information

- [x] I have included the output of ``celery -A proj report`` in the issue.
    (if you are not able to do this, then at least specify the Celery
     version affected).
- [x] I have verified that the issue exists against the `main` branch of Celery.
- [x] I have included the contents of ``pip freeze`` in the issue.
- [x] I have included all the versions of all the external dependencies required
  to reproduce this bug.

## Optional Debugging Information

- [x] I have tried reproducing the issue on more than one Python version
  and/or implementation.
- [ ] I have tried reproducing the issue on more than one message broker and/or
  result backend.
- [ ] I have tried reproducing the issue on more than one version of the message
  broker and/or result backend.
- [ ] I have tried reproducing the issue on more than one operating system.
- [ ] I have tried reproducing the issue on more than one workers pool.
- [ ] I have tried reproducing the issue with autoscaling, retries,
  ETA/Countdown & rate limits disabled.
- [x] I have tried reproducing the issue after downgrading
  and/or upgrading Celery and its dependencies.

#### Related Issues

- None

#### Possible Duplicates

- None

## Environment & Settings

[...]

PR #9793, opened 2025-07-03
Replace DelayedDelivery connection creation to use context manger

bityob:
## Description

- Replaced `DelayedDelivery` step connection creation to use context manager, so the the connection will be closed when the step is completed. 

Fixing an issue I found - https://github.com/celery/celery/issues/9792.

## Verifications 

Running Celery with my fix, created the exchange and queues needed with binding, and closed the connection. 

```
uvx --no-cache --with /home/bit/Code/celery celery -A app worker --loglevel=debug --without-gossip --without-mingle --without-heartbeat --prefetch-multiplier=1 --concurrency=1
```

Logs - 

image

Rabbit connections - only 1 left as expected, for consuming tasks - 

image

Rabbit channels - only 1 channel -

image

Exchanges were created successfully - 

image

Queues were created successfully -

image

### DIFF
--- a/celery/worker/consumer/delayed_delivery.py
+++ b/celery/worker/consumer/delayed_delivery.py
@@ -114,33 +114,33 @@ class DelayedDelivery(bootsteps.StartStopStep):
             OSError: If there are network-related issues
             Exception: For other unexpected errors during setup
         """
-        connection: Connection = c.app.connection_for_write(url=broker_url)
-        queue_type = c.app.conf.broker_native_delayed_delivery_queue_type
-        logger.debug(
-            "Setting up delayed delivery for broker %r with queue type %r",
-            broker_url, queue_type
-        )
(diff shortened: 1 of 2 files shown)


bityob:
> the code changes are making the CI fail. can you please add/update tests to verify this change?

- Those tests failed because of the usage of connection context manager that is not support seamlessly in `Mock`. 
Fixed by changing those cases to use `MagicMock` instead of `Mock`. 
- Also modified `test_start_native_delayed_delivery_topic_exchange` test, to verify the connection context was called

Verified locally by running - 
```
cd docker
docker compose run -it --rm celery bash
pyenv exec python3.12 -m pytest t/unit/worker/test_native_delayed_delivery.py -v
```

image

auvipy (approved):
thats good now

Issue #9759, opened 2025-06-16
Support setting exclusive: true for celeryev.* event queues to avoid RabbitMQ deprecation


marekvesely-direct:
RabbitMQ 4.x deprecates queues that are durable: false and exclusive: false (transient_nonxcl_queues).
Currently, celeryev.* event queues are created as classic queues with these exact parameters, triggering the deprecation warning and blocking future upgrades.

Please add a configuration option to allow setting exclusive: true or durable: true on event queues (e.g., when using EventReceiver or pidbox). This would ensure compatibility with newer RabbitMQ versions. Reference: https://www.rabbitmq.com/blog/2021/08/21/4.0-deprecation-announcements#removal-of-transient-non-exclusive-queues

auvipy:
I can see this as a duplicate of an existing issue. but need to cross check....

marekvesely-direct:
> I can see this as a duplicate of an existing issue. but need to cross check....

I didnt found any similar issue.

Dima-Diachenko:
+1, we're waiting for this feature as well to not see warning message and be able to update to RabbitMQ version 4.2+

auvipy:
@maccinza did you face this as well?

MarcosJunior250714:
This lib is creating automatic the queue, so, if I manual fix on Admin, creating the queue, when celery start this error appears: PreconditionFailed: Queue.declare: (406) PRECONDITION_FAILED - inequivalent arg 'durable' for queue 'name_queue' in vhost 'name_vhost': received 'true' but current is 'false'.
If I delete all queues, when start, create with deprecated defaults values.

ghirailghiro:
Hey @auvipy, could I jump on this?

Here’s what I’ve done so far:

* Added two settings under the `event` namespace: `event_queue_exclusive` and `event_queue_durable`.
* Patched `celery/events/receiver.py` to use them and prevent setting both to `True`.
* Added a couple of tests.
* Added matching settings under the `control` namespace, but I haven’t touched Kombu yet.

The pidbox queue, though, is created in Kombu (`kombu/pidbox.py`) and still comes up as `durable=False, exclusive=False`. Would you be okay with a small Kombu patch to read `control_queue_exclusive/_durable` and apply the same logic?

Let me know if I’ve misunderstood anything. Thanks!

auvipy:
yeah sure, i saw this github.com/celery/celery/pull/9815 here. feel free to come with a patch in kombu! lets continue the discussion there....

PR #9750, opened 2025-06-10
Use durable=True by default everywhere

trianglesis:
## Description

Trying the variant where all queues and exchanges are durable by default.

Discussed at issue [2237](https://github.com/celery/kombu/issues/2237)
RabbitMQ recommendation: [DOC](https://www.rabbitmq.com/docs/queues#durability)
Related to the `kombu` [PR](https://github.com/celery/kombu/pull/2313).

Conf: `cat /etc/rabbitmq/rabbitmq.conf`

```
# https://www.rabbitmq.com/docs/deprecated-features
deprecated_features.permit.transient_nonexcl_queues = false
```

A pretty simple `fix` just changed booleans to `durable=True`

Tested in my DEV environment.
Now running at prod version.

No issues with tasks, queues, managing and pinging/inspecting queues and workers.

I didn't run bundled tests, however. 
Only relied upon my envs.

<details>
  <summary>My celery conf</summary>

```python
def celery_setup():
    # Get creds and hosts
    setup_celery_env()
    # Setup django project
    django.setup()
    #  The backend is specified via the backend argument to Celery
    global app
    if app is None:
        app = Celery('server',
                     broker=RabbitMQCreds.BROKER,
                     backend=backend,
                     )

    app.conf.timezone = 'UTC'
    app.conf.enable_utc = True
    # General config:
    app.conf.update(
        result_backend=result_backend,
        accept_content=['pickle'],
        task_serializer='pickle',
        result_serializer='pickle',
        result_extended=True,
        task_track_started=True,
        beat_scheduler='django_celery_beat.schedulers:DatabaseScheduler',
        database_engine_options={'pool_timeout': 90},
        worker_prefetch_multiplier=1,
        worker_concurrency=1,
        worker_timer_precision=1.0,
        broker_heartbeat=10.0,
        broker_heartbeat_checkrate=2.0,
    )

    app.autodiscover_tasks()
    app.conf.update(
        worker_max_memory_per_child=1024 * 100,
        worker_max_tasks_per_child=100,
        worker_proc_alive_timeout=5.0,
        task_acks_late=False,
        task_acks_on_failure_or_timeout=True,
        worker_cancel_long_running_tasks_on_connection_loss=True,
        task_reject_on_worker_lost=False,
        worker_pool_restarts=True,
        worker_enable_remote_control=True,
        worker_lost_wait=20,
        broker_connection_retry_on_startup=True,
        broker_connection_retry=True,
        broker_connection_max_retries=0,
[...]

PR #9815, opened 2025-07-15
Choose queue type and exchange type when creating missing queues (fix #9671)

### LINKED ISSUE #9671: Adding task_create_missing_queues_type in order to autocreate missing queues with quorum type
- [x] I have checked the [issues list](https://github.com/celery/celery/issues?utf8=%E2%9C%93&q=is%3Aissue+label%3A%22Issue+Type%3A+Feature+Request%22+)
- [x] I have checked the [pull requests list](https://github.com/celery/celery/pulls?utf8=%E2%9C%93&q=is%3Apr+label%3A%22PR+Type%3A+Feature%22+)
- [ ] I have checked the [commit log](https://github.com/celery/celery/commits/main)
  to find out if the same feature was already implemented in the
  main branch.

# Brief Summary
The setting [task_create_missing_queues](https://docs.celeryq.dev/en/stable/userguide/configuration.html#task-create-missing-queues) gives ability to autocreate queues and exchanges. Celery 5.5 adds ability to use quorum queues in RabbitMQ. But I haven't found any ability to autocreate missing queues with `'x-queue-type': 'quorum'`. I can only set [task_default_queue_type](https://docs.celeryq.dev/en/stable/userguide/configuration.html#task-default-queue-type) which affects only default queue.
To my mind, it would be important to have this ability, as classic mirrored queues are [removed in RabbitMQ 4.0+](https://www.rabbitmq.com/blog/2023/03/02/quorum-queues-migration).

## Proposed Behavior
Add a setting `task_create_missing_queue_type` which declares queue type when creating missing queues (classic/quorum). Or, alternatively add `task_create_missing_queue_arguments` setting in order to edit arguments for autocreated queues in a much extandable manner.

ghirailghiro:
## Description

This PR adds **two new optional settings** that control how Celery auto-creates
queues when `task_create_missing_queues = True`:

| Setting | Default | What it does |
|---------|---------|--------------|
| `task_create_missing_queue_type` | `"classic"` | Select the AMQP queue kind: keep the default `"classic"`  or choose the type you need, e.g. `"quorum"`). |
| `task_create_missing_queue_exchange_type` | `None` | Override the exchange type used for the auto-exchange: set to `"topic"`, `"fanout"`, etc. |

## Changes(Fixes #9671)

Add two new settings that apply when Celery autogenerates a queue
(`task_create_missing_queues=True`):

- `task_create_missing_queue_type`

- `task_create_missing_queue_exchange_type`

Backwards compatibility: default behaviour (classic queue + direct exchange) is unchanged.

### DIFF
--- a/celery/app/amqp.py
+++ b/celery/app/amqp.py
@@ -46,6 +46,13 @@ class Queues(dict):
         create_missing (bool): By default any unknown queues will be
             added automatically, but if this flag is disabled the occurrence
             of unknown queues in `wanted` will raise :exc:`KeyError`.
+        create_missing_queue_type (str): Type of queue to create for missing queues.
+            Must be either 'classic' (default) or 'quorum'. If set to 'quorum',
+            the broker will declare new queues using the quorum type.
+        create_missing_queue_exchange_type (str): Type of exchange to use
+            when creating missing queues. If not set, the default exchange type
+            will be used. If set, the exchange type will be set to this value
+            when creating missing queues.
         max_priority (int): Default x-max-priority for queues with none set.
     """
 
@@ -53,14 +60,19 @@ class Queues(dict):
     #: The rest of the queues are then used for routing only.
     _consume_from = None
 
-    def __init__(self, queues=None, default_exchange=None,
-                 create_missing=True, autoexchange=None,
-                 max_priority=None, default_routing_key=None):
+    def __init__(
+            self, queues=None, default_exchange=None,
+            create_missing=True, create_missing_queue_type=None,
+            create_missing_queue_exchange_type=None, autoexchange=None,
+            max_priority=None, default_routing_key=None,
+    ):
         super().__init__()
         self.aliases = WeakValueDictionary()
         self.default_exchange = default_exchange
         self.default_routing_key = default_routing_key
         self.create_missing = create_missing
+        self.create_missing_queue_type = create_missing_queue_type
+        self.create_missing_queue_exchange_type = create_missing_queue_exchange_type
         self.autoexchange = Exchange if autoexchange is None else autoexchange
         self.max_priority = max_priority
         if queues is not None and not isinstance(queues, Mapping):
@@ -181,7 +193,16 @@ class Queues(dict):
                 self._consume_from.pop(queue, None)
 
     def new_missing(self, name):
-        return Queue(name, self.autoexchange(name), name)
+        queue_arguments = None
+        if self.create_missing_queue_type:
+            queue_arguments = {"x-queue-type": self.create_missing_queue_type}
+
+        if self.create_missing_queue_exchange_type:
+            exchange = Exchange(name, self.create_missing_queue_exchange_type)
+        else:
+            exchange = self.autoexchange(name)
+
+        return Queue(name, exchange, name, queue_arguments=queue_arguments)
 
     @property
     def consume_from(self):
@@ -238,14 +259,18 @@ class AMQP:
     def send_task_message(self):
         return self._create_task_sender()
 
-    def Queues(self, queues, create_missing=None,
-               autoexchange=None, max_priority=None):
+    def Queues(self, queues, create_missing=None, create_missing_queue_type=None,
+               create_missing_queue_exchange_type=None, autoexchange=None, max_priority=None):
         # Create new :class:`Queues` instance, using queue defaults
         # from the current configuration.
         conf = self.app.conf
         default_routing_key = conf.task_default_routing_key
         if create_missing is None:
             create_missing = conf.task_create_missing_queues
+            if create_missing_queue_type is None:
+                create_missing_queue_type = conf.task_create_missing_queue_type
+            if create_missing_queue_exchange_type is None:
+                create_missing_queue_exchange_type = conf.task_create_missing_queue_exchange_type
         if max_priority is None:
             max_priority = conf.task_queue_max_priority
         if not queues and conf.task_default_queue:
@@ -259,8 +284,14 @@ class AMQP:
         autoexchange = (self.autoexchange if autoexchange is None
                         else autoexchange)
         return self.queues_cls(
-            queues, self.default_exchange, create_missing,
-            autoexchange, max_priority, default_routing_key,
+            queues,
+            default_exchange=self.default_exchange,
+            create_missing=create_missing,
+            create_missing_queue_type=create_missing_queue_type,
+            create_missing_queue_exchange_type=create_missing_queue_exchange_type,
+            autoexchange=autoexchange,
+            max_priority=max_priority,
+            default_routing_key=default_routing_key,
         )
 
     def Router(self, queues=None, create_missing=None):
--- a/docs/userguide/configuration.rst
+++ b/docs/userguide/configuration.rst
@@ -134,6 +134,8 @@ have been moved into a new  ``task_`` prefix.
 ``CELERY_ANNOTATIONS``                     :setting:`task_annotations`
 ``CELERY_COMPRESSION``                     :setting:`task_compression`
 ``CELERY_CREATE_MISSING_QUEUES``           :setting:`task_create_missing_queues`
+``CELERY_CREATE_MISSING_QUEUE_TYPE``       :setting:`task_create_missing_queue_type`
+``CELERY_CREATE_MISSING_QUEUE_EXCHANGE_TYPE`` :setting:`task_create_missing_queue_exchange_type`
 ``CELERY_DEFAULT_DELIVERY_MODE``           :setting:`task_default_delivery_mode`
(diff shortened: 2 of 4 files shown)


auvipy (commented):
lets also fix the kombu front as well as described here (https://github.com/celery/celery/issues/9759#issuecomment-3081079711)

ghirailghiro:
Hi @auvipy I just open the PR on Kombu

I added to the control.py file on the celery part the two parameter control_queue_durable and control_queue_exclusive, after kombu is fixed i will make the commit with these modifications.

auvipy:
yeah thanks. saw that.

auvipy:
can you please also look at the approach in this PR? https://github.com/celery/kombu/pull/2313 and https://github.com/celery/celery/pull/9750 ? are they related to what you are trying to do?

ghirailghiro:
I checked both of those drafts. They flip every queue to `durable=True`, so the RabbitMQ warning disappears. For what I undestand pidbox queues are meant for short-lived messages that you don't need to keep after the broker is restarted.

My patch keeps the current behaviour unless the user asks for something different. It adds two flags: you can set the pidbox queue to exclusive (still transient, but now accepted by RabbitMQ 4.x) or, if you really need persistence, you can set it to durable. If you do nothing, it stays exactly as it is today.

If the project would rather move straight to the “everything durable” solution, I can drop this and follow that path o set is as default as `durable=True`

Please let me know if I’ve misunderstood anything.

auvipy:
your approach seems better so im ok with it.

ghirailghiro:
Hi @auvipy

I’ve added support for configuring durable and exclusive options on both event queues and control queue:

- event_queue_durable, event_queue_exclusive
- control_queue_durable, control_queue_exclusive

There’s also validation to prevent both being set to True, plus I added docs and tests. This is meant to fix #9759 on the celery part.

That said, I can’t fully test the changes because they depend on feature that we introduced in [Kombu PR 2338](https://github.com/celery/kombu/pull/2338)

Just wondering what’s the best way to proceed here, i need some guidance on how i can test it or if is necessary to test with the right version of kombu.

Thanks!

auvipy:
can you check why the unit tests are failing, please?

ghirailghiro:
My bad, I’ve just pushed a follow‑up commit that adds the missing guard clause to the failing test I introduced.

Also figured out how to pin my local env to the unreleased kombu main, so I run the full test‑suite and the CI should be green.

ghirailghiro:
For what i can see now the tests are failing because the tests are not pointing to the main branch of kombu but it's last release. The solution that i propose need to use the new parameters that we merged in the [Kombu PR 2338](https://github.com/celery/kombu/pull/2338) on the Mailbox class.

ghirailghiro:
I think the tests are now failing because they are not pointing to the main of Kombu, I think i need some guidance here to understand how to fix them.

auvipy:
yes kombu new version is required.

auvipy:
kombu has been updated, so lets see

auvipy:
the redis failures are not related to this rabbitmq bug fix pr

celery/events/receiver.py
```diff
@@ -34,25 +35,39 @@ class EventReceiver(ConsumerMixin):
...
+            raise ImproperlyConfigured(
+                'Queue cannot be both exclusive and durable, '
+                'choose one or the other.'
+            )
         self.queue = Queue(
             '.'.join([self.queue_prefix, self.node_id]),
             exchange=self.exchange,
             routing_key=self.routing_key,
-            auto_delete=True, durable=False,
+            auto_delete=True and (not queue_durable),
```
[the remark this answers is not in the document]
auvipy: can you cross check this suggestion, please?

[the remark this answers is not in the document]
ghirailghiro: I think Copilot is right, that True is useless

celery/events/receiver.py
```diff
@@ -34,25 +35,39 @@ class EventReceiver(ConsumerMixin):
...
+                 accept=None, queue_ttl=None, queue_expires=None,
+                 queue_exclusive=None,
+                 queue_durable=None):
         self.app = app_or_default(app or self.app)
         self.channel = maybe_channel(channel)
         self.handlers = {} if handlers is None else handlers
         self.routing_key = routing_key
         self.node_id = node_id or uuid()
-        self.queue_prefix = queue_prefix or self.app.conf.event_queue_prefix
+        cfg = self.app.conf
```
auvipy: why this is not self.cfg but only cfg?

ghirailghiro: My idea was to  just using it on the __init__

auvipy: @Nusnus whats your take on this?

Nusnus: Tbh I think the full `self.app.conf` should be used instead to make the code cleaner.

auvipy:
@ghirailghiro please address the review comments...

celery/events/receiver.py
ghirailghiro: For me that's ok, i will push the modifications

Nusnus (changes requested):
Nice work!
Very clean PR.

One change and we can merge it, please fix the missing `.. versionadded::` in the docs!

docs/userguide/monitoring.rst
```diff
@@ -814,3 +814,24 @@ worker-offline
...
 
 The worker has disconnected from the broker.
+
+Mailbox Configuration (Advanced)
+--------------------------------
+
+Celery uses `kombu.pidbox.Mailbox` internally to send control and broadcast commands
+to workers.
+
+.. versionadded:: Kombu 5.6.0
```
Nusnus, line 824: Kombu?

Nusnus: On second thought, even though we don’t use that notion anywhere else, it does make sense and maybe we should actually adopt mentioning which kombu version introduced a relevant change.

Interesting.

auvipy: yes, this feature is kombu 5.6 dependent---

docs/userguide/configuration.rst
```diff
@@ -2619,6 +2623,48 @@ If enabled (default), any queues specified that aren't defined in
 :setting:`task_queues` will be automatically created. See
 :ref:`routing-automatic`.
 
+.. setting:: task_create_missing_queue_type
+
+``task_create_missing_queue_type``
```
auvipy, line 2628: please fix the missing .. versionadded:: annotation in the docs.

docs/userguide/configuration.rst
```diff
@@ -2619,6 +2623,48 @@ If enabled (default), any queues specified that aren't defined in
...
+``task_create_missing_queues`` is enabled), this setting defines what type
+of RabbitMQ queue to create.
+
+- ``"classic"`` (default): declares a standard classic queue.
+- ``"quorum"``: declares a RabbitMQ quorum queue (adds ``x-queue-type: quorum``).
+
+.. setting:: task_create_missing_queue_exchange_type
+
+``task_create_missing_queue_exchange_type``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
auvipy:
```suggestion
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

docs/userguide/configuration.rst
ghirailghiro: added to all the new parameters with the last commit

celery/app/amqp.py
```diff
@@ -181,7 +193,21 @@ def deselect(self, exclude):
...
+                    "Valid types are 'classic' and 'quorum'."
+                )
+            queue_arguments = {"x-queue-type": self.create_missing_queue_type}
+
+        if self.create_missing_queue_exchange_type:
+            exchange = Exchange(name, self.create_missing_queue_exchange_type)
+        else:
+            exchange = self.autoexchange(name)
+
+        return Queue(name, exchange, name, queue_arguments=queue_arguments)
```
auvipy, line 210: why returning name twice in Queue class?

ghirailghiro: I think we can eliminate that is redundant

auvipy:
unit tests seems to be failing after the last commit. can you please cross check and fix them?

ghirailghiro:
I checked the celery/app/amqp.py on [main](https://github.com/celery/celery/blob/3703beb4589b4d590b84a1b08a8d6558149b6f4d/celery/app/amqp.py#L184C9-L184C15) and i think the right way is to explicit set the routing_key. I think we should roll back the last commit for keeping compatibility.

Nusnus:
🚀🚀

ghirailghiro:
Thank you guys for the support and sorry for the delay on the response!

Issue #5483, opened 2019-04-24
Celery puts django database connections into invalid state after fork


Chronial:
Django db connections are not fork-safe. For this reason, celery contains cleanup code to handle this after a fork: https://github.com/celery/celery/blob/3d387704d4a18db5aea758e9a26c1ee4d71659df/celery/fixups/django.py#L141-L151

That code is supposed to do two things:
1. Manually close the file handles for django's db connections
2. Tell django to close the db connections

The first step is necessary so the second step doesn't actually close the db connections of the parent process that they where inherited from.

Unfortunately, the second step was broken by #4292, with this change: https://github.com/celery/celery/pull/4292/files#diff-7da7bceb78d87096e818d34c80115d18L186.

It is important that the connections are actually closed – otherwise they are left in an invalid state (remember – we force-closed the file descriptor). But `close_if_unusable_or_obsolete` only sometimes closes the connection.

To elaborate: In the django default config, `CONN_MAX_AGE` is set to `0`, and then `close_if_unusable_or_obsolete` actually always closes the connection. See https://github.com/django/django/blob/76990cbb/django/db/backends/base/base.py#L513 in combination with https://github.com/django/django/blob/76990cbb/django/db/backends/base/base.py#L190.

But if `CONN_MAX_AGE` is set to anything but `0`, `close_if_unusable_or_obsolete` will not always close the connection. That leaves the connection in an invalid state, causing these bugs: https://github.com/celery/django-celery-results/issues/58 and https://github.com/celery/celery/issues/4878.

auvipy:
thanks for the report!

Jokairui:
so, what exactly should we deal with this 'mysql server has gone away' bug? @Chronial  @auvipy

Chronial:
@Jokairui you need to undo the changes of #4116. Until this is fixed in celery, you can either set `CONN_MAX_AGE` to `0` to avoid triggering the bug or you can do correct db handling yourself. This should do it:

```py
def fix_django_db(**kwargs):
    # Calling db.close() on some DB connections will cause the inherited DB
    # conn to also get broken in the parent process so we need to remove it
    # without triggering any network IO that close() might cause.
    for c in django.db.connections.all():
        if c and c.connection:
            try:
                os.close(c.connection.fileno())
            except (AttributeError, OSError, TypeError,
                    django.db.InterfaceError):
                pass
        try:
            c.close()
        except django.db.InterfaceError:
            pass
        except django.db.DatabaseError as exc:
[...]

Issue #4878, opened 2018-07-04
db connection closed and worker hangs with celery 4.2+


craigds:
This might be two separate bugs, but they might be related.

## Checklist

- [x] I have included the output of ``celery -A proj report`` in the issue.
- [x] I have verified that the issue exists against the `master` branch of Celery.

## Steps to reproduce

Run multiple celery workers with `celery multi`, each with its own queue, then pass tasks to them.

## Expected behavior

tasks should run on all workers.

### Most workers explode:

Most nodes give this, for every task:

```
[2018-07-04 02:52:21,280: WARNING/ForkPoolWorker-1] Request: <Context: {'origin': 'gen19542@vm', u'args': [], u'chain': None, 'root_id': '0c008f5a-507d-4b08-8082-0dcc0ab3495d', 'expires': None, u'is_eager': False, u'correlation_id': '0c008f5a-507d-4b08-8082-0dcc0ab3495d', u'chord': None, u'reply_to': '5e48e3d9-0b67-3122-b0c5-9b83efd9e5ca', 'id': '0c008f5a-507d-4b08-8082-0dcc0ab3495d', 'kwargsrepr': '{}', 'lang': 'py', 'retries': 0, 'task': 'dandelion.geostore.celery.debug_task', 'group': None, 'timelimit': [None, None], u'delivery_info': {u'priority': 0, u'redelivered': False, u'routing_key': u'celery', u'exchange': u''}, u'hostname': u'celery@vm', 'called_directly': False, 'parent_id': None, 'argsrepr': '()', u'errbacks': None, u'callbacks': None, u'kwargs': {}, 'eta': None, '_protected': 1}>
[2018-07-04 02:52:21,288: WARNING/ForkPoolWorker-1] /kx/dandelion/venv/local/lib/python2.7/site-packages/celery/app/trace.py:561: RuntimeWarning: Exception raised outside body: OperationalError('could not receive data from server: Bad file descriptor\n',):
Traceback (most recent call last):
  File "/kx/dandelion/venv/local/lib/python2.7/site-packages/celery/app/trace.py", line 442, in trace_task
    uuid, retval, task_request, publish_result,
  File "/kx/dandelion/venv/local/lib/python2.7/site-packages/celery/backends/base.py", line 144, in mark_as_done
    self.store_result(task_id, result, state, request=request)
  File "/kx/dandelion/venv/local/lib/python2.7/site-packages/celery/backends/base.py", line 316, in store_result
    request=request, **kwargs)
[...]

PR #9824, opened 2025-07-21
Revert "Use Django DB max age connection setting"

### LINKED ISSUE #4116: Django celery fixup doesn't respect Django settings for PostgreSQL connections

danlamanna:
This reverts commit f0c9b40bd4aa7228afa20f589e50f2e4225d804e (cc @whtsky @auvipy)

This reverts PR #6134 and stops using the close_if_unusable_or_obsolete API since there are edge cases where it's unable to detect if a connection if actually unusable, see the relevant Django issue which is marked as `wontfix`: https://code.djangoproject.com/ticket/30646 (see also https://forum.djangoproject.com/t/close-if-unusable-or-obsolete-fails-to-close-unusable-connections/41900).

A celery-specific way to introduce one of these edge cases is to run a task with a timeout that gets interrupted via `SIGALRM` while a query is executing. The connection will be unusable but considered usable by Django, causing the next invocation of a task on that worker to fail.

Since this is an optimization that can't be reliably used, Celery ought to close the connection after each task instead of trying to manage connections in a way similar to how the Django application does.

## Description

Sorry for the verbose timeline but this optimization has a bit of a history in Celery.

#### June 2017 
#4116 is filed citing that Celery doesn't reuse database connections the same way Django does
#### Sept 2017
#4292 is added to start reusing Django connections via `close_if_unusable_or_obsolete`
#### July 2018 - Apr 2019
#4878, #5483, and https://github.com/celery/django-celery-results/issues/58 crop up as a result of #4292
#### May 2019 
The optimization gets reverted in #5515 
#### June 2020
#6134 reintroduces this optimization to attempt to fix #4116 once and for all

### DIFF
--- a/celery/fixups/django.py
+++ b/celery/fixups/django.py
@@ -168,7 +168,7 @@ class DjangoWorkerFixup:
                 self._maybe_close_db_fd(c)
 
         # use the _ version to avoid DB_REUSE preventing the conn.close() call
-        self._close_database(force=True)
+        self._close_database()
         self.close_cache()
 
     def _maybe_close_db_fd(self, c: "BaseDatabaseWrapper") -> None:
@@ -197,13 +197,10 @@ class DjangoWorkerFixup:
             self._close_database()
         self._db_recycles += 1
 
-    def _close_database(self, force: bool = False) -> None:
+    def _close_database(self) -> None:
         for conn in self._db.connections.all():
             try:
-                if force:
-                    conn.close()
-                else:
-                    conn.close_if_unusable_or_obsolete()
+                conn.close()
             except self.interface_errors:
                 pass
             except self.DatabaseError as exc:
--- a/t/unit/fixups/test_django.py
+++ b/t/unit/fixups/test_django.py
@@ -274,6 +274,16 @@ class test_DjangoWorkerFixup(FixupCase):
             with pytest.raises(KeyError):
                 f._close_database()
 
+    def test_close_database_always_closes_connections(self):
+        with self.fixup_context(self.app) as (f, _, _):
+            conn = Mock()
+            f._db.connections.all = Mock(return_value=[conn])
+            f.close_database()
+            conn.close.assert_called_once_with()
+            # close_if_unusable_or_obsolete is not safe to call in all conditions, so avoid using
+            # it to optimize connection handling.
+            conn.close_if_unusable_or_obsolete.assert_not_called()
+
     def test_close_cache(self):
         with self.fixup_context(self.app) as (f, _, _):
             f.close_cache()
--- a/t/unit/fixups/test_django.py
+++ b/t/unit/fixups/test_django.py
@@ -284,6 +284,11 @@ class test_DjangoWorkerFixup(FixupCase):
             # it to optimize connection handling.
             conn.close_if_unusable_or_obsolete.assert_not_called()
 
+    def test_close_cache_raises_error(self):
+        with self.fixup_context(self.app) as (f, _, _):
+            f._cache.close_caches.side_effect = AttributeError
+            f.close_cache()
+
     def test_close_cache(self):
         with self.fixup_context(self.app) as (f, _, _):
             f.close_cache()
--- a/t/unit/fixups/test_django.py
+++ b/t/unit/fixups/test_django.py
@@ -196,7 +196,7 @@ class test_DjangoWorkerFixup(FixupCase):
                         f.on_worker_process_init()
                         mcf.assert_called_with(conns[1].connection)
                         f.close_cache.assert_called_with()
-                        f._close_database.assert_called_with(force=True)
+                        f._close_database.assert_called_with()
 
                         f.validate_models = Mock(name='validate_models')
                         patching.setenv('FORKED_BY_MULTIPROCESSING', '1')
@@ -264,35 +264,13 @@ class test_DjangoWorkerFixup(FixupCase):
             f._db.connections = Mock()  # ConnectionHandler
             f._db.connections.all.side_effect = lambda: conns
 
-            f._close_database(force=True)
-            conns[0].close.assert_called_with()
-            conns[0].close_if_unusable_or_obsolete.assert_not_called()
-            conns[1].close.assert_called_with()
-            conns[1].close_if_unusable_or_obsolete.assert_not_called()
-            conns[2].close.assert_called_with()
-            conns[2].close_if_unusable_or_obsolete.assert_not_called()
-
-            for conn in conns:
-                conn.reset_mock()
-
             f._close_database()
-            conns[0].close.assert_not_called()
-            conns[0].close_if_unusable_or_obsolete.assert_called_with()
-            conns[1].close.assert_not_called()
-            conns[1].close_if_unusable_or_obsolete.assert_called_with()
-            conns[2].close.assert_not_called()
-            conns[2].close_if_unusable_or_obsolete.assert_called_with()
+            conns[0].close.assert_called_with()
+            conns[1].close.assert_called_with()
+            conns[2].close.assert_called_with()
 
             conns[1].close.side_effect = KeyError(
                 'omg')
-            f._close_database()
-            with pytest.raises(KeyError):
-                f._close_database(force=True)
-
-            conns[1].close.side_effect = None
-            conns[1].close_if_unusable_or_obsolete.side_effect = KeyError(
-                'omg')
-            f._close_database(force=True)
             with pytest.raises(KeyError):
                 f._close_database()
 


danlamanna:
> I can revert this. but what will be consequences?

Before the connection was only closed based on a heuristic, now it will always be closed at the end of the task.

> can we add additional integration tests to make sure there will be no regression at all?

I can't see a way to add an integration test without adding postgres and a django application. If you're just looking for a big sign that steers people away from re-implementing something similar again, I can add a unit test that asserts `close_if_unusable_or_obsolete` isn't called, since using that API is the fundamental problem here. Does that work?

auvipy:
yeah sure, that would be great

danlamanna:
> yeah sure, that would be great

All set.

auvipy:
I suggest try part of the following test cases to improve test coverage.

`

# --- Edge Case: Simulate SIGALRM or OSError on DB close ---
def test__close_database_handles_oserror(self):
    with self.fixup_context(self.app) as (f, _, _):
        conns = [Mock()]
        # Simulate an OSError (e.g., interrupted system call/SIGALRM)
        conns[0].close.side_effect = OSError('interrupted system call')
        f.DatabaseError = OSError
        f.interface_errors = ()
        f._db.connections = Mock()
        f._db.connections.all.return_value = conns
        # Should not raise
        f._close_database(force=True)
        conns[0].close.assert_called_with()

# --- Multi-DB: All connections are forcibly closed ---
def test__close_database_multi_db(self):
    with self.fixup_context(self.app) as (f, _, _):
        conns = [Mock(), Mock(), Mock()]
        f.DatabaseError = KeyError
        f.interface_errors = ()
        f._db.connections = Mock()
        f._db.connections.all.return_value = conns
        f._close_database(force=True)
        for c in conns:
            c.close.assert_called_with()

# --- Connection already closed and other DB exceptions are handled gracefully ---
@pytest.mark.parametrize('exc', [KeyError('closed'), Exception('generic')])
def test__close_database_handles_various_db_errors(self, exc):
    with self.fixup_context(self.app) as (f, _, _):
        conns = [Mock()]
        conns[0].close.side_effect = exc
        f.DatabaseError = type(exc)
        f.interface_errors = ()
        f._db.connections = Mock()
        f._db.connections.all.return_value = conns
        # Should not propagate
        try:
            f._close_database(force=True)
        except Exception as raised:
            # Only propagate if not 'closed' or 'not connected'
            if 'closed' not in str(exc) and 'not connected' not in str(exc):
                assert raised is exc
            else:
                pytest.fail("Should not raise for 'closed' or 'not connected'")

# --- DB_REUSE_MAX: Ensure close is NOT called until threshold ---
def test_close_database_reuse_max_behavior(self):
    with self.fixup_context(self.app) as (f, _, _):
        f.db_reuse_max = 3
        f._db_recycles = 2
        with patch.object(f, '_close_database') as _close:
            f.close_database()
            _close.assert_not_called()
            f._db_recycles = 3
            f.close_database()
            _close.assert_called_with()

# --- Forbid use of CONN_MAX_AGE ---
def test_close_database_ignores_conn_max_age(self):
    with self.fixup_context(self.app) as (f, _, _):
        # Simulate a Django connection with CONN_MAX_AGE
        f._db.connections.settings_dict = {'CONN_MAX_AGE': 999}
        with patch.object(f, '_close_database') as _close:
            f.close_database()
            _close.assert_called_with()  # Should not check or use CONN_MAX_AGE

# --- Worker/process shutdown closes all connections ---
def test_on_worker_process_init_closes_connections(self):
    with self.fixup_context(self.app) as (f, _, _):
        conns = [Mock(connection=object()), Mock(connection=object())]
        for c in conns:
            c.connection = object()
        f._db.connections = Mock()
        f._db.connections.all.return_value = conns
        with patch.object(f, '_maybe_close_db_fd') as maybe_close:
            with patch.object(f, '_close_database') as close_db, \
                 patch.object(f, 'close_cache') as close_cache:
                f.on_worker_process_init()
                assert maybe_close.call_count == len(conns)
                close_db.assert_called_with(force=True)
                close_cache.assert_called()

# --- Concurrency: Multiple tasks finish simultaneously, closing only once ---
def test_close_database_concurrent_tasks(self):
    with self.fixup_context(self.app) as (f, _, _):
        with patch.object(f, '_close_database') as _close:
            f.db_reuse_max = 2
            f._db_recycles = 1
            f.close_database()
            f.close_database()
            # Only one call until threshold
            _close.assert_not_called()
            f._db_recycles = 2
            f.close_database()
            _close.assert_called_with()

# --- Dead code assertion: 'close_if_unusable_or_obsolete' is not used ---
def test_close_database_does_not_call_close_if_unusable_or_obsolete(self):
    with self.fixup_context(self.app) as (f, _, _):
        conns = [Mock()]
        conns[0].close_if_unusable_or_obsolete = MagicMock()
        f.DatabaseError = Exception
        f.interface_errors = ()
        f._db.connections = Mock()
        f._db.connections.all.return_value = conns
        f._close_database(force=True)
        conns[0].close_if_unusable_or_obsolete.assert_not_called()
`

danlamanna:
@auvipy I added a test case for `close_cache` since that was the one area of uncovered code. AFAICT this PR increases coverage and reduces cyclomatic complexity.

auvipy (approved):
merging this for now as it is a revert of an old feature which is not valid anymore. the pr add some extra tests to prevent possible regression.

auvipy:
btw, do we also need to clean or update the relevant docs as well?

danlamanna:
AFAICT, there were never any docs added for this optimization, so I don't think there need to be any additional changes.

auvipy:
ok thanks

orhanhenrik:
Hi, this is a big performance regression. Celery will now always close DB connections between every task, instead of respecting the django `CONN_MAX_AGE` setting. 

We had to revert our upgrade to 5.6.0 because it was adding a huge load on our database, since creating connections is relatively expensive.

auvipy:
OK. I tried to revert this but seems that is not possible. a manual revert is needed. can you take on this?

auvipy:
https://forum.djangoproject.com/t/close-if-unusable-or-obsolete-fails-to-close-unusable-connections/41900).

after reading the reply of a django core team member, it seems we should not revert this at all. instead we could try different improvement.

orhanhenrik:
> it seems we should not revert this at all.

As in, not revert the original code, or not revert this PR?

The way I understand the comment from the django forum, it sounds like this PR is not the correct solution. So it would make sense to revert this PR (revert #9824).

auvipy:
> The way I understand the comment from the django forum, it sounds like this PR is not the correct solution. So it would make sense to revert this PR (revert #9824).

we got a manual revert request here. as per the forum discussion, this PR should not be merged in the first place. so we have to revert it now. thanks for your report

danlamanna:
> > The way I understand the comment from the django forum, it sounds like this PR is not the correct solution. So it would make sense to revert this PR (revert #9824).
> 
> we got a manual revert request here. as per the forum discussion, this PR should not be merged in the first place. so we have to revert it now. thanks for your report

@auvipy 
I'll respond to that forum post - I don't think his first comment is correct. His comment is also about correctness, not performance. This PR strictly makes celery do the more correct thing at the cost of not reusing connections - because that was causing problems for us (repro is in the forum post and is effectively how celery timeouts function).

auvipy:
we also got this PR https://github.com/celery/celery/pull/10015 an you please take a look?

shtlrs:
+1 on [this comment](https://github.com/celery/celery/pull/9824#issuecomment-3625809205) from @orhanhenrik 

We have had to revert back to 5.5.3 twice now because our db user auth attemps were going through the roof.

We saw this with both 5.6.0 and 5.6.2, and was wondering what the next steps would be, if there are any.

auvipy:
> +1 on [this comment](https://github.com/celery/celery/pull/9824#issuecomment-3625809205) from @orhanhenrik
> 
> We have had to revert back to 5.5.3 twice now because our db user auth attemps were going through the roof.
> 
> We saw this with both 5.6.0 and 5.6.2, and was wondering what the next steps would be, if there are any.

can you please elaborate

shtlrs:
@auvipy sure thing.

We've tried to migrate to 5.6.x two times, the first time to 5.6.0 and the second to 5.6.2

The first time we noticed a huge increase in the number of db connections which lead to very big spikes in our connection latency graphs

We first thought it was because of https://github.com/celery/celery/pull/9953 and that it'd be fixed in 5.6.2, so we tried upgrading again and started noticing the exact same symptoms.

We can process up to 23M celery tasks in some of our environments, which connect to the db, and explain the explosion of connection attempts and latencies.

This is currently blocking a huge project we because we need the ability of defining the default exchange and queue types that comes with 5.6.2.


image

auvipy:
do you have any suggested fix? or any bandwidth to handle this?

shtlrs:
What I did was reintroduce the `close_if_unusable_or_obsolete` behavior via a custom fixup, as it wasn't really causing any of the listed errors for us.

I don't have bandwidth to focus solely on this now, but i think i will look into a better solution at some point if nothing comes up before.

auvipy:
did you check this comment https://github.com/celery/celery/pull/10021#issuecomment-3627179955 ? we tried a manual revert.

shtlrs:
Yes, and i saw that the PR was closed a long time ago and based on the comments, i figured that the change won't be put back any time soon, so i kinda did the same via a custom fixup.

auvipy:
can you share your findings to improve and fix this? we opened the PR again in a draft mode. you can add your review comments there as prospective suggestions.

PR #9842, opened 2025-08-01
Disable pytest-xdist for smoke tests and increase retries (CI ONLY)

Nusnus:


auvipy (approved):
this seems to be working

Nusnus:
Hopefully we're done with the flaky smoke tests failures 🙏
I suspect it has something to do with Docker, so removing `-n auto` should be the main fix.

Fortunately, Blacksmith's performance is so good we don't need `-n auto` anymore in the CI to keep the run short (few minutes per test module).

Let's see if it keeps stable for now, so we can focus on getting the integration tests stable too @auvipy

PR #9786, opened 2025-06-28
Vendor click-plugins, PyPI package no longer maintained.

sebastic:
[click-plugins](https://github.com/click-contrib/click-plugins) now recommends users to vendor it as the PyPI package is no longer maintained (see: https://github.com/click-contrib/click-plugins/pull/40).

PR #9878, opened 2025-08-29
Vendor click_plugins 2.0 since it is not distributed on pypi anymore

wosc:
See https://github.com/click-contrib/click-plugins/blob/main/README.rst

Also, the 1.x version available on pypi uses the deprecated pkg_resources package which is slated for removal from setuptools for end of 2025.


auvipy (commented):
duplicate of https://github.com/celery/celery/pull/9786

PR #9889, opened 2025-09-12
Fix gevent timeout race condition

ernestobits:

## Description

Makes it so that Timeout only wraps the task itself. See https://github.com/celery/celery/discussions/9888


celery/concurrency/base.py
```diff
@@ -21,13 +21,17 @@
 
 def apply_target(target, args=(), kwargs=None, callback=None,
                  accept_callback=None, pid=None, getpid=os.getpid,
-                 propagate=(), monotonic=time.monotonic, **_):
+                 propagate=(), timeout=None, Timeout=None, monotonic=time.monotonic, **_):
```
auvipy, line 24: why two time out? can you please explain

ernestobits: The first one is the timeout (int), the second is the Timeout class. This is the same as how ya'll currently call `apply_timeout`: https://github.com/celery/celery/blob/7c75fa738885315180f0194da04cf5105994fe13/celery/concurrency/gevent.py#L28-L31

Issue #9637, opened 2025-03-25
Celery + MSSQL + Windows - multiple issues


Abhijit-Revamp:
# Checklist

- [x] I have verified that the issue exists against the `main` branch of Celery.
- [ ] This has already been asked to the [discussions forum](https://github.com/celery/celery/discussions) first.
- [ ] I have read the relevant section in the
  [contribution guide](https://docs.celeryq.dev/en/main/contributing.html#other-bugs)
  on reporting bugs.
- [x] I have checked the [issues list](https://github.com/celery/celery/issues?q=is%3Aissue+label%3A%22Issue+Type%3A+Bug+Report%22+-label%3A%22Category%3A+Documentation%22)
  for similar or identical bug reports.
- [ ] I have checked the [pull requests list](https://github.com/celery/celery/pulls?q=is%3Apr+label%3A%22PR+Type%3A+Bugfix%22+-label%3A%22Category%3A+Documentation%22)
  for existing proposed fixes.
- [ ] I have checked the [commit log](https://github.com/celery/celery/commits/main)
  to find out if the bug was already fixed in the main branch.
- [ ] I have included all related issues and possible duplicate issues
  in this issue (If there are none, check this box anyway).
- [ ] I have tried to reproduce the issue with [pytest-celery](https://docs.celeryq.dev/projects/pytest-celery/en/latest/userguide/celery-bug-report.html) and added the reproduction script below.

## Mandatory Debugging Information

- [x] I have included the output of ``celery -A proj report`` in the issue.
    (if you are not able to do this, then at least specify the Celery
     version affected).
- [ ] I have verified that the issue exists against the `main` branch of Celery.
- [ ] I have included the contents of ``pip freeze`` in the issue.
- [ ] I have included all the versions of all the external dependencies required
  to reproduce this bug.

## Optional Debugging Information

- [ ] I have tried reproducing the issue on more than one Python version
  and/or implementation.
- [ ] I have tried reproducing the issue on more than one message broker and/or
  result backend.
- [ ] I have tried reproducing the issue on more than one version of the message
  broker and/or result backend.
- [ ] I have tried reproducing the issue on more than one operating system.
- [ ] I have tried reproducing the issue on more than one workers pool.
- [ ] I have tried reproducing the issue with autoscaling, retries,
  ETA/Countdown & rate limits disabled.
- [ ] I have tried reproducing the issue after downgrading
  and/or upgrading Celery and its dependencies.

#### Related Issues

- None

#### Possible Duplicates

[...]

Issue #9501, opened 2025-01-17
Support for computed columns and indexes for MSSQL backend JSON-based results column


dxdc:
# Checklist

- [x] I have checked the [issues list](https://github.com/celery/celery/issues?q=is%3Aissue+label%3A%22Issue+Type%3A+Enhancement%22+-label%3A%22Category%3A+Documentation%22)
  for similar or identical enhancement to an existing feature.
- [x] I have checked the [pull requests list](https://github.com/celery/celery/pulls?q=is%3Apr+label%3A%22Issue+Type%3A+Enhancement%22+-label%3A%22Category%3A+Documentation%22)
  for existing proposed enhancements.
- [x] I have checked the [commit log](https://github.com/celery/celery/commits/main)
  to find out if the same enhancement was already implemented in the
  main branch.
- [x] I have included all related issues and possible duplicate issues in this issue
      (If there are none, check this box anyway).

#### Related Issues

- [#8634](https://github.com/celery/celery/issues/8634)

#### Possible Duplicates

- None

# Brief Summary
I’m currently using Celery with MSSQL as the result backend, and I’ve modified the results column to store JSON instead of binary (see [#8634](https://github.com/celery/celery/issues/8634)). I’d like to add computed columns to help index and query the JSON data more efficiently, but I’m unsure if this is officially supported within the Celery framework or if it could break anything.

Could you advise on whether:

- Adding computed columns is a supported or recommended approach?
- This change might require a feature request or additional Celery configuration?

# Design

## Architectural Considerations

None

## Diagrams

N/A

## Alternatives

None

dxdc:
thx @auvipy ... I was thinking probably the database models need to be converted into something more extensible, while still keeping the needed fields in place.

do you have any suggestions how to approach this?

I'm also not sure if the schema is checked each time and aligned, in case of upgrades or changes for example.

https://github.com/celery/celery/blob/main/celery/backends/database/models.py

dxdc:
for example, wondering about something like this, but not sure if this is a proper way to override the backend. Maybe there is a better way to do it.

```py
# celery_custom.py

from celery import Celery
from celery.backends.database import DatabaseBackend
from celery.backends.database.models import Task
from sqlalchemy import Column, String, Integer

# 1. Define a mixin that supplies additional fields.
class AdditionalFieldsMixin:
    # Add any extra fields you require.
    extra_field = Column(String(255), nullable=True)
    another_field = Column(Integer, default=0, nullable=False)

[...]

PR #9907, opened 2025-09-23
Db json result

vladborovtsov:
## Description

Not sure if it has to be considered as draft. 

This change introduces a new database backend that's supposed to store results in JSON field. The main idea behind this is to have it in the codebase, avoid fragile monkey patching. 

Usage is simple: 
```
backend_url = (
    f'db-json+mssql+pyodbc://sa:{mssql_password}@localhost:1433/celery'
    '?driver=ODBC+Driver+18+for+SQL+Server&TrustServerCertificate=yes'
)
```
Use "db-json" prefix instead of just "db". 

The change is mainly for MSSQL, though it must work with modern MySQL as well as with Postgres, wherever SQLAlchemy supports the JSON field.

Perhaps, some tests still need to be written to cover the new backend. Though, it seems to be functional now. Will be happy to commit more time into this, but happy to see some feedback. 

Related issues: 
- https://github.com/celery/celery/issues/9637
- https://github.com/celery/celery/issues/9501

### DIFF
--- a/celery/backends/database/models.py
+++ b/celery/backends/database/models.py
@@ -42,14 +42,11 @@ class TaskBase(ResultModelBase):
     def __repr__(self):
         return '<Task {0.task_id} state: {0.status}>'.format(self)
 
-
     @classmethod
     def configure(cls, schema=None, name=None):
         cls.__table__.schema = schema
         cls.id.default.schema = schema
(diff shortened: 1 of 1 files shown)


vladborovtsov:
Closing this in favor of #9908. Extensions approach looks definitely better!

PR #9908, opened 2025-09-24
Add schema extension support for database result backend tables

dxdc:
## Motivation

Currently, customizing the schema of Celery’s SQLAlchemy result backend requires monkey-patching the models (`Task`, `TaskSet`, or `TaskExtended`) directly. This makes it hard to:

- Change column types (e.g., use `JSON` instead of `PickleType`)
- Add indexes or constraints
- Apply dialect-specific adjustments

This PR introduces a clean, explicit extension mechanism that allows applications to declare schema modifications in a **backwards-compatible** way.

See:
- [Issue #9501](https://github.com/celery/celery/issues/9501)  
- [PR #9907](https://github.com/celery/celery/pull/9907) (earlier attempt at JSON fields)

---

## Design

A new concept: **Schema Extensions**.

- Schema extensions are classes implementing:

  ```python
  class SchemaExtension(ABC):
      def extend(self, table: Table, metadata: MetaData) -> None:
          ...
  ```

- Passed into `DatabaseBackend` via a new `schema_extensions` argument.
- Keys in `schema_extensions` must be `"task"` or `"group"`, mapping to a list of extensions for each table.
- Extensions are applied once during model configuration, before tables are created.

This keeps the core code minimal while offering flexibility for production users who need advanced schema control.

---

## Example Usage

```python
from sqlalchemy import Index
from celery.backends.database.extensions import SchemaExtension, JsonResultExtension
from celery.backends.database import DatabaseBackend

class StatusIndexExtension(SchemaExtension):
    def extend(self, table: Table, metadata: MetaData) -> None:
        if table.name == "celery_taskmeta":
            Index("idx_taskmeta_status", table.c.status, unique=False)

backend = DatabaseBackend(
    url="postgresql://...",
    app=app,
    schema_extensions={
        "task": [JsonResultExtension(), StatusIndexExtension()],
        "group": [],
    }
)
```

This results in:

- `result` column using `JSON` instead of `PickleType`.
- A new index on `status`.
- No changes to `TaskSet` schema.

---

## Changes in This PR

- `DatabaseBackend.__init__`:
  - Accepts new kwarg:
    ```python
    schema_extensions: Optional[Dict[Literal["task", "group"], List[SchemaExtension]]]
    ```
  - Passes extensions to `Task.configure` and `TaskSet.configure`.

- `Task.configure` / `TaskSet.configure`:
  - Accept new `extensions` argument.
  - Apply each extension’s `extend()` method to the underlying SQLAlchemy `Table`.

- Added type hints for clarity.

---

## Benefits

- **Backwards compatible**: No changes for users who don’t pass extensions.
- **Explicit**: Users can configure schema behavior per table, no monkey-patching required.
- **Composable**: Multiple extensions can be stacked together.
- **Safe**: Extensions run before any DDL is executed.

---

## Future Idea: Validation

Extensions could optionally implement a `validate()` method to check the live database schema and warn about mismatches.  
This could be extended later to add *optional migrations* (e.g., using `checkfirst=True`), but is **out of scope for this PR**.

---

## Summary

This PR introduces `schema_extensions`, a lightweight, extensible way to customize Celery’s database backend schema.  
It improves flexibility, reduces monkey-patching, and provides a more general solution than [PR #9907](https://github.com/celery/celery/pull/9907).  
It also helps address [issue #9501](https://github.com/celery/celery/issues/9501).


dxdc:
## Example Usage

Schema extensions can be provided in two ways:

**1. Pass directly when instantiating the backend:**

```python
from celery.backends.database import DatabaseBackend

backend = DatabaseBackend(
    url=backend_url,
    app=app,
    schema_extensions={
        "task": [
            WorkerHostnameExtension(),
            StatusIndexExtension(),
        ],
        "group": [],
    },
)
app.backend = backend
```

**2. Configure via `app.conf` (auto-picked up by DatabaseBackend):**

```python
app.conf.update(
    database_url=backend_url,
    database_schema_extensions={   # note key name
        "task": [
            WorkerHostnameExtension(),
            StatusIndexExtension(),
        ],
        "group": [],
    },
    result_expires=36000,
)
```

celery/backends/database/extensions.py
```diff
@@ -0,0 +1,23 @@
+"""Schema extensions for Celery result backend tables."""
+
+from abc import ABC, abstractmethod
+
+from sqlalchemy import JSON, MetaData, Table
+
+
+class SchemaExtension(ABC):
```
auvipy, line 36: that is interesting. in which version it was introduced?

dxdc:
Some tests failed due to some internal codecov error, see https://github.com/codecov/codecov-action/issues/1876

Seems like restarting those failed tests should solve the issue.

celery/backends/database/extensions.py
dxdc: Oh, if you meant ABC, that showed up in Python 3.4 as the easier way to define abstract base classes.

celery/backends/database/__init__.py
```diff
@@ -66,7 +68,12 @@ class DatabaseBackend(BaseBackend):
     task_cls = Task
     taskset_cls = TaskSet
 
-    def __init__(self, dburi=None, engine_options=None, url=None, **kwargs):
+    def __init__(self,
+                 dburi=None,
+                 engine_options=None,
+                 url=None,
+                 schema_extensions: Optional[Dict[Literal['task', 'group'], List[SchemaExtension]]] = None,
+                 **kwargs):
```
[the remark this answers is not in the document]
dxdc: sure, but what other kwargs are there?

celery/backends/database/__init__.py
```diff
@@ -86,12 +93,15 @@ def __init__(self, dburi=None, engine_options=None, url=None, **kwargs):
 
         schemas = conf.database_table_schemas or {}
         tablenames = conf.database_table_names or {}
+        schema_extensions = schema_extensions or conf.get('database_schema_extensions', {})
```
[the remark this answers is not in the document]
dxdc: I think it's worth keeping as a configuration option for ease for implementation. Where should we document it? Happy to add that.

marcocali (approved):

auvipy:
@ChickenBenny would you mind checking this?

ChickenBenny:
Great job making the model more flexible! This is a fantastic approach. However, I have two operational concerns regarding this, as follows:

1. **Schema Migrations (e.g., Alembic):** Since users will now maintain custom extended models, relying on Celery's default `metadata.create_all()` might become risky, especially if future Celery updates introduce new core columns to the base `Task` table. Should we explicitly mention in the documentation that users adopting this feature should manage their own schema migrations (like using Alembic) to avoid conflicts?

2. **Core Field Override Safety:** Is there any built-in safeguard (or planned validation) to prevent users from accidentally overriding core required fields (e.g., `task_id`, `status`, `date_done`) when subclassing?

Thanks for the awesome contribution!

auvipy:
good points @ChickenBenny . I already had one open long ago but faced some negetive feedback https://github.com/celery/celery/pull/8532

ChickenBenny:
I completely agree. Alembic is a much more robust and controlled tool for managing database schemas

dxdc:
Thanks for the thoughtful review @ChickenBenny!

Both excellent points:

1. Schema Migrations: You're right - I've added a prominent note in the extensions.py module docstring making it clear that users who adopt schema extensions should manage their own migrations (e.g., via Alembic). Celery's metadata.create_all() is additive and won't drop/alter existing columns, but it also won't handle migration conflicts if future Celery releases add new core columns. I think documentation is the right approach here rather than baking Alembic into Celery itself (as explored in #8532).

However, we could bake this into the SchemaExtension itself, then `celery -A myapp db upgrade` just calls ext.migrate(table_name) for each extension inside an Alembic migration context. Similar approach for validate.

```py
class SchemaExtension(ABC):

    @abstractmethod
    def extend(self, table, metadata):
        ...

    def validate(self, table, inspector):
        return []

    def migrate(self, table_name):
        """Optional: run Alembic operations to align DB with extend()."""
        pass


class JsonResultExtension(SchemaExtension):

    def extend(self, table, metadata):
        table.c.result.type = JSON()

    def validate(self, table, inspector):
        cols = {c['name']: c for c in inspector.get_columns(table.name)}
        actual = type(cols.get('result', {}).get('type', PickleType())).__name__
        if actual.upper() not in ('JSON', 'JSONB'):
            return [f"result column is {actual}, expected JSON - run migrations"]
        return []

    def migrate(self, table_name):
        from alembic import op
        op.alter_column(table_name, 'result',
                        type_=JSON(), existing_type=PickleType())

```


2. Core Field Override Safety: I've added a lightweight safeguard - `a _warn_if_protected` check that runs after extensions are applied. It logs a warning if an extension modifies the type of a core column (task_id, status, date_done, id, taskset_id). It doesn't block the change (there can be legitimate reasons to modify types), but it makes it very visible. Added tests for both the warning and no-false-positive cases.

celery/backends/database/__init__.py
```diff
@@ -51,7 +53,29 @@ class DatabaseBackend(BaseBackend):
...
+        Initialize the DatabaseBackend.
+
+        :param dburi: Database URI.
+        :param engine_options: SQLAlchemy engine options.
+        :param url: Alternative way to specify the database URL.
+        :param schema_extensions: Optional dictionary of schema extensions for 'task' and 'group'.
+            (Added as a keyword-only argument in 5.4.0.)
+        :param kwargs: Additional keyword arguments.
+
+        .. versionchanged:: 5.4.0
```
auvipy:
```suggestion
        .. versionchanged:: 5.7.0
```

Issue #9909, opened 2025-09-24
Group result ordering is not respected for groups replacing a task


simonepozzoli-pix4d:
# Checklist

- [x] I have verified that the issue exists against the `main` branch of Celery.
- [ ] This has already been asked to the [discussions forum](https://github.com/celery/celery/discussions) first.
- [x] I have read the relevant section in the
  [contribution guide](https://docs.celeryq.dev/en/main/contributing.html#other-bugs)
  on reporting bugs.
- [x] I have checked the [issues list](https://github.com/celery/celery/issues?q=is%3Aissue+label%3A%22Issue+Type%3A+Bug+Report%22+-label%3A%22Category%3A+Documentation%22)
  for similar or identical bug reports.
- [x] I have checked the [pull requests list](https://github.com/celery/celery/pulls?q=is%3Apr+label%3A%22PR+Type%3A+Bugfix%22+-label%3A%22Category%3A+Documentation%22)
  for existing proposed fixes.
- [x] I have checked the [commit log](https://github.com/celery/celery/commits/main)
  to find out if the bug was already fixed in the main branch.
- [x] I have included all related issues and possible duplicate issues
  in this issue (If there are none, check this box anyway).
- [ ] I have tried to reproduce the issue with [pytest-celery](https://docs.celeryq.dev/projects/pytest-celery/en/latest/userguide/celery-bug-report.html) and added the reproduction script below.

## Mandatory Debugging Information

- [ ] I have included the output of ``celery -A proj report`` in the issue.
    (if you are not able to do this, then at least specify the Celery
     version affected).
- [x] I have verified that the issue exists against the `main` branch of Celery.
- [ ] I have included the contents of ``pip freeze`` in the issue.
- [ ] I have included all the versions of all the external dependencies required
  to reproduce this bug.

## Optional Debugging Information

- [ ] I have tried reproducing the issue on more than one Python version
  and/or implementation.
- [ ] I have tried reproducing the issue on more than one message broker and/or
  result backend.
- [ ] I have tried reproducing the issue on more than one version of the message
  broker and/or result backend.
- [ ] I have tried reproducing the issue on more than one operating system.
- [ ] I have tried reproducing the issue on more than one workers pool.
- [ ] I have tried reproducing the issue with autoscaling, retries,
  ETA/Countdown & rate limits disabled.
- [ ] I have tried reproducing the issue after downgrading
  and/or upgrading Celery and its dependencies.

#### Related Issues

- None

#### Possible Duplicates

- None

[...]

PR #9910, opened 2025-09-24
fix(): preserve group order in replaced signature

simonepozzoli-pix4d:

## Description

Preserve group ordering when replacing a task with a group.
It should not be allowed to pass a `group_index` in the options, otherwise all tasks in the group will have the same `group_index` and the ordering is not deterministic.

Related issue: https://github.com/celery/celery/issues/9909

### DIFF
--- a/celery/canvas.py
+++ b/celery/canvas.py
@@ -1741,7 +1741,7 @@ class group(Signature):
 
     def _apply_tasks(self, tasks, producer=None, app=None, p=None,
                      add_to_parent=None, chord=None,
-                     args=None, kwargs=None, **options):
+                     args=None, kwargs=None, group_index=None, **options):
         """Run all the tasks in the group.
 
         This is used by :meth:`apply_async` to run all the tasks in the group
--- a/celery/canvas.py
+++ b/celery/canvas.py
@@ -1741,7 +1741,7 @@ class group(Signature):
 
     def _apply_tasks(self, tasks, producer=None, app=None, p=None,
                      add_to_parent=None, chord=None,
-                     args=None, kwargs=None, **options, group_index=None):
+                     args=None, kwargs=None, group_index=None, **options):
         """Run all the tasks in the group.
 
         This is used by :meth:`apply_async` to run all the tasks in the group
--- a/celery/canvas.py
+++ b/celery/canvas.py
@@ -1741,7 +1741,7 @@ class group(Signature):
 
     def _apply_tasks(self, tasks, producer=None, app=None, p=None,
                      add_to_parent=None, chord=None,
-                     args=None, kwargs=None, group_index=None, **options):
+                     args=None, kwargs=None, **options, group_index=None):
         """Run all the tasks in the group.
 
         This is used by :meth:`apply_async` to run all the tasks in the group
--- a/celery/app/task.py
+++ b/celery/app/task.py
@@ -544,7 +544,7 @@ class Task:
                 If supplied, this value will be used as the task’s id instead
                 of generating one automatically. Be careful to avoid collisions
                 when overriding task ids.
-                
+
         Returns:
             celery.result.AsyncResult: Promise of future evaluation.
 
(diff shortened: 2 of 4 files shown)


celery/canvas.py
```diff
@@ -1741,7 +1741,7 @@ def _prepared(self, tasks, partial_args, group_id, root_id, app,
 
     def _apply_tasks(self, tasks, producer=None, app=None, p=None,
                      add_to_parent=None, chord=None,
-                     args=None, kwargs=None, **options):
+                     args=None, kwargs=None, group_index=None, **options):
```
auvipy: where it will be used?

auvipy (changes requested):
can you please add unit tests to verify the change?

celery/canvas.py
simonepozzoli-pix4d: It won't be used, I added the parameter to prevent it from being included in the `options`

simonepozzoli-pix4d:
> can you please add unit tests to verify the change?

Added a test here https://github.com/celery/celery/pull/9910/commits/fac95938ab035755cfd576a66eb5944fce6806ba
However, I had to set eager execution of the task to make it work. This is not a good representation of what is happening in a live environment where the order of execution of the tasks is not preserved.
The issue I'm trying to fix in this PR happens only if a worker with concurrency > 1 registers the result of the task in the group's second place before registering the result of first task in the group. Since both results have the same group_index, the ordering depends only on which task finished first.

celery/canvas.py
simonepozzoli-pix4d: To further explain myself: calling `group._apply_tasks(group_index=0)`, all tasks schedule for this group would have the same group_index and results from this group can't be ordered.
IMO is not correct to pass a `group_index` to this method, it should be ignored when scheduling groups.

simonepozzoli-pix4d:
I added a second integration test in https://github.com/celery/celery/pull/9910/commits/1343410d0884ce2cfeedbd01c4930908547ea090
Without my change, the test is consistently failing.
Here some examples. As I mentioned, the result order is not deterministic:
```
======================================== short test summary info ========================================
FAILED t/integration/test_canvas.py::test_group::test_task_replace_with_group_preserves_group_order - assert [[2, 3], [5, ..., [5, 4], ...] == [[3, 2], [5, ..., [5, 4], ...]
```
```
============================================================================================= short test summary info =============================================================================================
FAILED t/integration/test_canvas.py::test_group::test_task_replace_with_group_preserves_group_order - assert [[3, 2], [4, ..., [4, 5], ...] == [[3, 2], [5, ..., [5, 4], ...]
```
```
============================================================================================= short test summary info =============================================================================================
FAILED t/integration/test_canvas.py::test_group::test_task_replace_with_group_preserves_group_order - assert [[3, 2], [4, ..., [5, 4], ...] == [[3, 2], [5, ..., [5, 4], ...]
```

simonepozzoli-pix4d:
> > can you please add unit tests to verify the change?
> 
> Added a test here [fac9593](https://github.com/celery/celery/commit/fac95938ab035755cfd576a66eb5944fce6806ba) However, I had to set eager execution of the task to make it work. This is not a good representation of what is happening in a live environment where the order of execution of the tasks is not preserved. The issue I'm trying to fix in this PR happens only if a worker with concurrency > 1 registers the result of the task in the group's second place before registering the result of first task in the group. Since both results have the same group_index, the ordering depends only on which task finished first.

Just a small clarification. I said the result ordering depends on the timing of tasks execution. This is not always correct.
We are using Redis as a result backend. When replacing a task with a group, the current task is replaced by a chord with header equal to the replacing tasks and body equal to `celery.accumulate` task. In Redis the chord header results are stored in a sorted set and forwarded to the chord body when all of them are available. If all results in the sorted set have the same score, they are sorted alphabetically when fetched from Redis using the `ZRANGE` command (https://redis.io/docs/latest/commands/zrange/#common-behavior-and-options).

celery/canvas.py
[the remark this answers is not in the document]
simonepozzoli-pix4d: This is a syntax error, I will revert the applied changes

PR #9917, opened 2025-09-28
Fix invalid pytz dependency specifier (>dev) in 5.1.x

### LINKED ISSUE #9916: Celery 5.1.2 sdist contains invalid requirement "pytz>dev"
## Problem
When installing Celery 5.1.2 from sdist, pip fails with:

packaging._tokenizer.ParserSyntaxError: Expected matching RIGHT_PARENTHESIS...
    pytz (>dev)

## Cause
The files in v5.1.2 tag:
- requirements/default.txt
- requirements/dev.txt

both contain `pytz>dev`, which is not a valid PEP 440 version specifier.

These get bundled into sdist and end up in `celery.egg-info/requires.txt`.

## Impact
- pip install from wheel: ✅ works
- pip install from sdist: ❌ fails
- downstream tools (pip-compile, Bazel rules_python, etc.) break when only sdist is available.

## Suggested fix
- Change `pytz>dev` to `pytz>=2016.7` (consistent with setup.cfg)
- Release a patch version (5.1.3) or document workaround (use wheel only).

Pro-YY:

## Summary
In Celery **5.1.2**, both `requirements/default.txt` and `requirements/dev.txt`
list an invalid requirement: pytz>dev

This is **not PEP 440–compliant** and breaks downstream tools (pip, piptools, Bazel)
when building from source. For example, Bazel fails with:
```
packaging._tokenizer.ParserSyntaxError:
Expected matching RIGHT_PARENTHESIS for LEFT_PARENTHESIS, after version specifier
pytz (>dev)
```

Later versions (>=5.2.3) fixed this in `default.txt` (`pytz>=2021.3`), but
`dev.txt` still contains the invalid specifier.

## Changes
- Replaced `pytz>dev` with `pytz>=2021.3` in `requirements/default.txt`.

## Why backport?
- This issue **still exists in the PyPI sdist/wheel for 5.1.2**.
- Many enterprise environments (like ours) pin to Celery 5.1.x for compatibility.
- Without a patch release, anyone installing 5.1.2 from source may hit this build error.

## Request
Please consider releasing **5.1.3** as a hotfix, so that the corrected metadata
is published to PyPI and downstream users are not blocked.

Fixes #9916

### DIFF
--- a/requirements/default.txt
+++ b/requirements/default.txt
@@ -1,4 +1,4 @@
-pytz>dev
+pytz>=2021.3
 billiard>=3.6.4.0,<4.0
 kombu>=5.1.0,<6.0
 vine>=5.0.0,<6.0


Pro-YY:
This PR fixes the invalid pytz>dev specifier present in the 5.1.2 sdist.
The issue only exists in 5.1.x and was already fixed in later releases (>=5.2).
I suggest creating a v5.1.x maintenance branch and releasing 5.1.3 with this patch,
so that users pinned to Celery 5.1 do not hit packaging errors when building from sdist.

PR #9939, opened 2025-10-12
Remove test-pypy3.txt from Dockerfile dependencies

auvipy:
Removed references to 'test-pypy3.txt' from multiple pip install commands in Dockerfile.


Nusnus:
@auvipy do you want to merge this before the release?

auvipy (commented):
as it is passing so lets include it

PR #9959, opened 2025-10-20
fix(worker): continue to attempt to bind other queues after a native delayed delivery binding failure has occurred

Izzette:
### Current problem

When binding queues for native delayed delivery in AMQP, a failure in binding one queue prevents subsequent queues from being bound.  This is particularly problematic when multiple heterogenous worker deployments are consuming from different queues, all needing delayed delivery setup.

Let's take for example an application with 3 `task_queues` (`queue-a`, `queue-b`, and `queue-c`) and two workers (`worker-a` and `worker-c`).  The `worker-a` consumes from `queue-a` and `worker-c` consumes from `queue-c`.  No worker is consuming from `queue-b`.  The application is starting on a fresh RabbitMQ vhost with no queues or exchanges previously existing.  `worker-a` starts up, it declares `queue-a`, the celery delayed delivery exchanges/queues, and it starts the delayed delivery binding process by calling `DelayedDelivery._bind_queues`. First it tries to bind `queue-a` to the delayed delivery exchange, it goes smoothly and the binding is setup for `queue-a`.  Then `worker-a` proceeds to try to bind `queue-b`, because it doesn't exist (`worker-a` won't declare it as it is not a consumer of `queue-b`) an exception is raised.  `worker-a` never tries to bind `queue-c` to the delayed delivery exchange as the loop broke binding `queue-c`.

Next, `worker-c` starts ups. It declares `queue-c`, the celery delayed delivery exchanges/queues, and it calls `DelayedDelivery._bind_queues`.  First it tries to bind `queue-a` which works fine as it's already bound, but fails on `queue-b` for the same reason as `worker-a` failed to bind the queue to the delayed delivery exchange.  `queue-c` is never bound by this worker.

Both workers finish their startup and keep consuming.  They log the message `Failed to bind queues for …`.  When ETA tasks are sent for `queue-a`, they trickle down through the delayed delivery exchanges until eventually being delivery to `queue-a` and processed by `worker-a`.  However, when ETA tasks are sent for `queue-c`, after trickling through the delayed delivery exchange, no binding is found for `*.queue-c` and the tasks are discarded.  Non-ETA tasks sent to `queue-c` are delivered just fine and processed by `worker-c`.

This can even happen when all queues have a worker, depending on the order in which the workers are started.  If for example `worker-b` existed and was consuming from `queue-b`, but the startup order was `worker-b`, `worker-c`, and finally `worker-a`: than both `queue-b` and `queue-c` would be missing bindings.  Once `worker-b` is restarted it would successfully bind `queue-b` to the delayed delivery exchanges.  If `worker-c` is than restarted after `worker-b` has restarted, it would be able to bind `queue-c` as well.

I'm currently facing this issue in a production environment.  My workaround is to manually declare all queues and restart the workers when new queues are added that don't yet have any workers.

### Solution

Here the solution I'm proposing is to capture exceptions raised when binding each queue, and instead of immediately re-raising group all exceptions in an ExceptionGroup.  This allows every queue to attempt to bind, with the errors still being raised after all queues have been attempted.

In order to allow `ConnectionRefusedError` and `OSError` to still be caught by `retry_over_time`, we need to raise immediately these errors as `kombu.utlis.functional.retry_over_time` is not using [the `except*` clause](https://docs.python.org/3/reference/compound_stmts.html#except-star).

See: https://github.com/celery/celery/blob/c4e4bab4a62c499a368b0921253149c8f02554ad/celery/worker/consumer/delayed_delivery.py#L84-L90
See: https://github.com/celery/kombu/blob/1bef09386e936be9757e713682bdf38bc1c0b737/kombu/utils/functional.py#L319-L319

#### Alternative solutions

Some other solutions could be considered, such as:

1. Creating the queues in `DelayedDelivery._bind_queues` before attempting `kombu.transport.native_delayed_delivery.bind_queue_to_native_delayed_delivery_exchange`.
1. Only iterating over the queues this worker is consuming (`app.amqp._consume_from`) in `DelayedDelivery._bind_queues` rather than all defined `task_queues` (`app.amqp.queues`).

See: https://github.com/celery/celery/blob/c4e4bab4a62c499a368b0921253149c8f02554ad/celery/app/amqp.py#L157-L166

I chose this solution as even if iterating over only queues this worker is consuming, as long as the worker is expected to continue to run after encountering errors when binding the delayed delivery exchange, I believe it ought to at least attempt to bind all queues and not stop on the first error.

---

Fixes https://github.com/celery/celery/issues/9960

### DIFF
--- a/requirements/extras/tblib.txt
+++ b/requirements/extras/tblib.txt
@@ -1,2 +1,1 @@
-tblib>=1.5.0;python_version>='3.8.0'
-tblib>=1.3.0;python_version<'3.8.0'
+tblib==3.1.0


auvipy (commented):
can you point to relevant issue which might be related to this fix?

Izzette:
> can you point to relevant issue which might be related to this fix?

No, but I can create one if it would be useful for you.

---

It seems that ExceptionGroup was adding in 3.11.  I'm not sure we _need_ to keep the sub-exceptions, I'll just include their messages.

auvipy:
can you also check why python 3.9 and 3.10 builds are failing with following tests?

  ___ test_DelayedDelivery.test_bind_queues_raises_exception_group_on_failures ___
  
  self = <t.unit.worker.test_native_delayed_delivery.test_DelayedDelivery object at 0x7f99dee542b0>
  mock_bind = <MagicMock name='bind_queue_to_native_delayed_delivery_exchange' id='140298444652304'>
  
      @patch('celery.worker.consumer.delayed_delivery.bind_queue_to_native_delayed_delivery_exchange')
      def test_bind_queues_raises_exception_group_on_failures(self, mock_bind):
          """Test that ExceptionGroup is raised with all binding failures."""
          consumer_mock = MagicMock()
          consumer_mock.app.conf.broker_native_delayed_delivery_queue_type = \
              'classic'
          consumer_mock.app.conf.broker_url = 'amqp://'
      
          # Create three queues
          queue1 = Queue('queue1', exchange=Exchange('exchange1', type='topic'))
          queue2 = Queue('queue2', exchange=Exchange('exchange2', type='topic'))
          queue3 = Queue('queue3', exchange=Exchange('exchange3', type='topic'))
      
          consumer_mock.app.amqp.queues = {
              'queue1': queue1,
              'queue2': queue2,
              'queue3': queue3,
          }
      
          # Make queue1 and queue3 fail with different errors
          def bind_side_effect(connection, queue):
              if queue.name == 'queue1':
                  raise ValueError("Queue1 binding failed")
              elif queue.name == 'queue3':
                  raise RuntimeError("Queue3 binding failed")
      
          mock_bind.side_effect = bind_side_effect
      
          delayed_delivery = DelayedDelivery(consumer_mock)
      
          # Should raise ExceptionGroup containing both exceptions
  >       with pytest.raises(ExceptionGroup) as exc_info:
  E       NameError: name 'ExceptionGroup' is not defined
  
  t/unit/worker/test_native_delayed_delivery.py:386: NameError
  =============================== warnings summary ===============================
    /home/runner/_work/celery/celery/celery/worker/consumer/consumer.py:517: CPendingDeprecationWarning: The broker_connection_retry configuration setting will no longer determine
    whether broker connection retries are made during startup in Celery 6.0 and above.
    If you wish to refrain from retrying connections on startup,
    you should set broker_connection_retry_on_startup to False instead.
      warnings.warn(
  
  t/unit/worker/test_request.py::test_Request::test_from_message_invalid_kwargs
    /home/runner/_work/celery/celery/celery/app/trace.py:705: RuntimeWarning: Exception raised outside body: InvalidTaskError('Task keyword arguments is not a mapping'):
      File "/home/runner/_work/celery/celery/celery/app/trace.py", line 429, in trace_task
        kwargs.items
    AttributeError: 'str' object has no attribute 'items'
    
      File "/home/runner/_work/celery/celery/celery/app/trace.py", line 431, in trace_task
        raise InvalidTaskError(
    celery.exceptions.InvalidTaskError: Task keyword arguments is not a mapping
    
      warn(RuntimeWarning(

Izzette:
> can you also check why python 3.9 and 3.10 builds are failing?

Yup!  I'm already on it.  It's because ExceptionGroup was adding in Python 3.11.  I forgot to check this and was running locally with the latest version of Python.  [PEP654](https://docs.python.org/3/whatsnew/3.11.html#whatsnew311-pep654)

celery/worker/consumer/delayed_delivery.py
```diff
@@ -166,7 +167,30 @@ def _bind_queues(self, app: Celery, connection: Connection) -> None:
...
+                exceptions.append(e)
+
+        if exceptions:
+            # Raise a single exception summarizing all binding failures
+            # In Python 3.11+, we could use ExceptionGroup here instead.
+            raise RuntimeError(
+                ("One or more failures occurred while binding queues to "
+                 "delayed delivery exchanges\n") +
+                "\n".join(str(e) for e in exceptions),
+            )
```
Izzette: Alternatively, this backport of [PEP 654's ExceptionGroup](https://github.com/agronholm/exceptiongroup) could be used when the runtime python version of < 3.11.  I've not used this backport before and can't endorse it.

auvipy: yes that would be great

Izzette: Done in https://github.com/celery/celery/pull/9959/commits/223decced66b85cbe44e3d90192f943e9c6d109f.

Izzette:
@auvipy I've created an issue here: https://github.com/celery/celery/issues/9960

celery/worker/consumer/delayed_delivery.py
```diff
@@ -166,7 +170,27 @@ def _bind_queues(self, app: Celery, connection: Connection) -> None:
...
+                #   catch utility from agronholm's exceptiongroup backport.
+                if isinstance(e, RETRIED_EXCEPTIONS):
+                    raise
+
+                exceptions.append(e)
+
+        if exceptions:
+            raise ExceptionGroup(
+                ("One or more failures occurred while binding queues to "
+                 "delayed delivery exchanges\n"),
```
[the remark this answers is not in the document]
Izzette: Actually this is a good catch from Copilot.  I'm going to remove the newline.
```suggestion
                ("One or more failures occurred while binding queues to "
                 "delayed delivery exchanges"),
```

t/unit/worker/test_native_delayed_delivery.py
```diff
@@ -306,3 +308,162 @@ def test_setup_bind_error(self, mock_bind, caplog):
...
+        # Should raise RuntimeError containing both exceptions messages
+        with pytest.raises(ExceptionGroup) as exc_info:
+            delayed_delivery._setup_delayed_delivery(consumer_mock, 'amqp://')
+
+        # Verify the RuntimeError message contains both exceptions
+        exception_group = exc_info.value
+        assert str(exception_group) == (
+            "One or more failures occurred while binding queues to delayed "
+            "delivery exchanges\n"
+            " (2 sub-exceptions)"
```
Izzette:
```suggestion
            "delivery exchanges (2 sub-exceptions)"
```

Izzette:
@auvipy:
> can you also explore possibility to add integration test for this?

I've added an integration test which is able to reproduce part of the issue described in #9960 (I hope this works in the CI 🤞, it's working locally).  This test starts the worker with an extra queue in task_queues and checks the bindings using the RabbitMQ management API.  This requires updating the RabbitMQ image from `latest` to `management` (which is the same as `latest` but with the management plugin enabled by default).

I could in theory test delayed delivery in addition to the bindings, but as I see many of the integration tests are marked as flaky: I think keeping the scope of the test to the bindings created is probably wise.

auvipy:
can you please look into this failing test?

 Redis, Database, Memcached, and more.)
  t/integration/test_tasks.py::test_tasks::test_revoke_by_stamped_headers_no_match PASSED
  t/integration/test_tasks.py::test_tasks::test_wrong_arguments PASSED
  t/integration/test_tasks.py::test_tasks::test_retry attempt 0
  attempt 1
  XFAIL (Retry failed
  on rpc backend)
  t/integration/test_tasks.py::test_tasks::test_retry_with_unpickleable_exception FAILED
  
  ______________ test_tasks.test_retry_with_unpickleable_exception _______________
  
  self = <t.integration.test_tasks.test_tasks object at 0x7f4981eef8b0>
  manager = <celery.contrib.testing.manager.Manager object at 0x7f498261ab60>
  
      def test_retry_with_unpickleable_exception(self, manager):
          """Test a task that retries with an unpickleable exception.
      
          We expect to be able to fetch the result (exception) correctly.
          """
      
          job = retry_unpickleable.delay(
              "foo",
              "bar",
              retry_kwargs={"countdown": 10, "max_retries": 1},
          )
      
          # Wait for the task to raise the Retry exception
          tik = time.monotonic()
          while time.monotonic() < tik + 5:
              status = job.status
              if status != 'PENDING':
                  break
              time.sleep(0.1)
          else:
              raise AssertionError("Timeout while waiting for the task to be retried")
      
          assert status == 'RETRY'
      
          # Get the exception
          res = job.result
          assert job.status == 'RETRY'  # make sure that it wasn't completed yet
      
          # Check it
  >       assert isinstance(res, UnpickleableExceptionWrapper)
  E       assert False
  E        +  where False = isinstance(Exception("<class 't.integration.tasks.UnpickleableException'>(['foo'])"), UnpickleableExceptionWrapper)
  
  t/integration/test_tasks.py:442: AssertionError
  ------------------------------ Captured log call -------------------------------
  INFO     celery.worker.strategy:strategy.py:161 Task t.integration.tasks.retry_unpickleable[02d1d78e-29aa-48b7-942b-0c2c7cc42b19] received
  INFO     celery.worker.strategy:strategy.py:161 Task t.integration.tasks.retry_unpickleable[02d1d78e-29aa-48b7-942b-0c2c7cc42b19] received
  INFO     celery.app.trace:trace.py:128 Task t.integration.tasks.retry_unpickleable[02d1d78e-29aa-48b7-942b-0c2c7cc42b19] retry: Retry in 10s: UnpickleableException('foo')
  FAILED t/integration/test_tasks.py::test_tasks::test_retry_with_unpickleable_exception - assert False
   +  where False = isinstance(Exception("<class 't.integration.tasks.UnpickleableException'>(['foo'])"), UnpickleableExceptionWrapper)
  !!!!!!!!!!!!!!!!!!!!!!!!!! stopping after 1 failures !!!!!!!!!!!!!!!!!!!!!!!!!!!
  = 1 failed, 14 passed, 3 skipped, 206 deselected, 1 xfailed in 82.03s (0:01:22) =

Izzette:
@auvipy:
> can you please look into this failing test?

It's really strange, as it was passing in [this prior commit](https://github.com/celery/celery/actions/runs/18655625832/job/53196104592), with the diff between the two being [enabling the rabbitmq management plugin, the exception message for the ExceptionGroup, and the addition of an integration test](https://github.com/izzette/celery/compare/223decced66b85cbe44e3d90192f943e9c6d109f..cd0eccf3c3b57cad88b1b37170e6787e3f77bb7b).  The error message is also awful strange: `Exception("<class 't.integration.tasks.UnpickleableException'>(['foo'])")` was raised, while we're expecting [`celery.utils.serialization.UnpickleableExceptionWrapper`](https://github.com/celery/celery/blob/1ad3bd45369106e22b6aabad047480119140b8ea/celery/utils/serialization.py#L100) which has the base class of Exception and wraps a [`t.integration.tasks.UnpickleableException`](https://github.com/celery/celery/blob/1ad3bd45369106e22b6aabad047480119140b8ea/t/integration/tasks.py#L345) ... they look very similar.

The thing is, I get the same error on [main](https://github.com/celery/celery/commit/1ad3bd45369106e22b6aabad047480119140b8ea) despite having passed previously.  Other PRs such as https://github.com/celery/celery/pull/9962 are also failing with the [same error](https://github.com/celery/celery/actions/runs/18748609937/job/53483694415?pr=9962#step:8:1209).  I think it may be due to a difference in the available package versions, as the package versions aren't pinned to an exact version, but rather with an interval allowing higher versions to be installed.

If I compare the [`pip freeze` from the passing run](https://github.com/celery/celery/actions/runs/18655625832/job/53196104662#step:8:1093) and the [`pip freeze` from the failing run](https://github.com/celery/celery/actions/runs/18687110603/job/53398293383?pr=9959#step:8:1098), I see this:

```diff
--- passing.txt	2025-10-24 10:26:21.889142303 +0200
+++ failing.txt	2025-10-24 10:26:25.857341881 +0200
@@ -12,2 +12,2 @@
-boto3==1.40.55
-botocore==1.40.55
+boto3==1.40.56
+botocore==1.40.56
@@ -17,2 +17,2 @@
-cassandra-driver==3.29.2
-celery @ file:///home/runner/_work/celery/celery/.tox/.tmp/package/1/celery-5.6.0b2-0.editable-py3-none-any.whl#sha256=14d56489ca97ae53d64c3e6288c7c7e2cdcf46b09fa188d32a8a82647ffe644b
+cassandra-driver==3.29.3
+celery @ file:///home/runner/_work/celery/celery/.tox/.tmp/package/1/celery-5.6.0b2-0.editable-py3-none-any.whl#sha256=444fa6a249a4a54018ea7342007f53c7d5a42de8238e819449cd1bbc246e2eb2
@@ -46 +46 @@
-geomet==0.2.1.post1
+geomet==1.1.0
@@ -68 +68 @@
-kombu @ git+https://github.com/celery/kombu.git@1bef09386e936be9757e713682bdf38bc1c0b737
+kombu @ git+https://github.com/celery/kombu.git@b8cf1f68a9af3d129e57be52ce48d3f85ecc9a80
@@ -145 +145 @@
-tblib==3.1.0
+tblib==3.2.0
```

The `tblib` (Serialization library for Exceptions and Tracebacks) looks suspect.  If I pin `tblib==3.1.0` in [`requirements/extras/tblib.txt`](https://github.com/celery/celery/blob/1ad3bd45369106e22b6aabad047480119140b8ea/requirements/extras/tblib.txt`) on `main`, the tests pass again.  I think if you re-run the integration tests on main, we'll install `tblib` 3.2.0 and the tests will fail.  I'm not really familiar with this library nor this code written by @ask 15 years ago, so I'm not sure what is really to be fixed here.  Maybe we should replace the requirement with `tblib>=1.5.0,tblib<3.2.0;python_version>='3.8.0'` in `requirements/extras/tblib.txt`.

In any case, this failure doesn't seem to be related to my PR and we should probably fix it in a different PR.

auvipy:
yes the failures doesn't seems to be related to your pr

auvipy (approved):

t/integration/test_native_delayed_delivery_binding.py
```diff
@@ -0,0 +1,190 @@
...
+    unique_id = uuid.uuid4().hex
+    app, exchange_name, queue_a_name, queue_b_name = create_test_app(unique_id)
+
+    # Set default queue to queue-b so the start_worker ping task is received
+    # by our worker
+    app.conf.task_default_queue = queue_b_name
+
+    # Start worker that only consumes from queue-b
+    # queue-a is NOT consumed, so it won't be declared by this worker
+    with start_worker(
```
Nusnus, line 148: @auvipy looks like there’s [another](https://github.com/celery/celery/pull/9986/changes#r3036004884) integration test infra breaking changes; this entire module appears to create a dedicated support infra just for this test instead of using the smoke tests which has all of that natively.

Nusnus: See also: https://github.com/celery/celery/tree/main/t/smoke/tests/quorum_queues

Nusnus: See also: https://github.com/celery/celery/actions/runs/23987387682/job/69961707961?pr=10237

(it retries at the moment, but the first attempt failed on this)

.github/workflows/integration-tests.yml
```diff
@@ -37,9 +37,10 @@ jobs:
           REDIS_HOST: localhost
           REDIS_PORT: 6379
       rabbitmq:
-        image: rabbitmq
+        image: rabbitmq:management
```
Nusnus, line 40: Hmm… seems to have also changed to support the integration test..

docker/docker-compose.yml
```diff
@@ -25,7 +25,7 @@ services:
       - azurite
 
   rabbit:
-    image: rabbitmq:latest
+    image: rabbitmq:management
```
Nusnus, line 28: Same I guess

t/integration/test_native_delayed_delivery_binding.py
Nusnus: Seems like the original pattern that was copied by https://github.com/celery/celery/pull/10159 and https://github.com/celery/celery/pull/9986 (and fixed by @aurangzaib048)

Classic case where AI slop is infecting the codebase. This PR’s changes were used as a ref for another contributor’s AI agent to “think” this is fine to copy. Hard to track, I know, try to keep an eye though @auvipy 🙏

Challenging times for maintainers… *sigh*

Nusnus: A good sign of a bad smell is a PR that changes the infrastructure to test its changes.
With smoke tests, we can cover complex scenarios that integration tests don’t let us test easily. AI is “lazyless" enough to find a way to hack the integration tests into whatever you want to do. The changes to `.github/workflows/integration-tests.yml` were the immediate signal that “something is conceptually wrong”, meaning that passing CI is not enough to validate the PR, as this is outside the CI’s scope.

aurangzaib048: @Nusnus You're right, and I owe you an apology, the start_worker() pattern in #10159 was copied from this PR without questioning whether it was the right approach. I should have looked at the smoke test infrastructure first instead of following an existing pattern blindly. Lesson learned.

Nusnus: All good - don’t feel bad. It happens to us all 🙏
Thank you for being responsible and fixing it so fast, much appreciated.

auvipy:
oh noo!!

.github/workflows/integration-tests.yml
Izzette: Yes I changed the image to have the management API enabled by default, this allows me to check directly the bindings on the queues and exchanges by calling the API.  I explain this in this comment on the PR, which seems to get an approval from the maintainer that was kind enough to review my PR: https://github.com/celery/celery/pull/9959#issuecomment-3426965288

image

docker/docker-compose.yml
Izzette: See https://github.com/celery/celery/pull/9959/changes#r3044175066, making the change here as well allows me to test locally.

t/integration/test_native_delayed_delivery_binding.py
Izzette: Hello @Nusnus, I'm not so sure to be familiar with your "smoke tests" pattern.  But I don't see it mentioned at all in your [contributing documentation](https://github.com/celery/celery/blob/main/CONTRIBUTING.rst).  Indeed, I can be quite bullheaded and push through with an overly complex implementation when I don't see an easy path immediately.

I did my best here to provide integration tests for the issue as requested by your co-maintainer https://github.com/celery/celery/pull/9959#pullrequestreview-3357303430
I also see this is requested in the contributing documentation:

> [ ] Make sure any change or new feature has a unit and/or integration test.
If a test is not written, a label will be assigned to your PR with the name Needs Test Coverage.

I didn't add tests initially, instead adding in a later commit after requested as correctly testing the behavior was correct seemed quite complicated.  This is especially true as, to my knowledge, AMQP 0-9-1 doesn't allow getting the current list of bindings directly.

I took great attention to your documentation, even adhering to your 78 width rule, despite noticing that most other tests / files in the codebase don't follow this.

> Lines shouldn't exceed 78 columns.
> 
> You can enforce this in [:command:`vim`](https://github.com/celery/celery/blob/main/CONTRIBUTING.rst#id86) by setting the textwidth option:
> 
> set textwidth=78
> If adhering to this limit makes the code less readable, you have one more character to go on. This means 78 is a soft limit, and 79 is the hard limit :)

And adding doc comments for my functions, as is described in the [coding syle](https://github.com/celery/celery/blob/main/CONTRIBUTING.rst#coding-style):

> [:pypi:`pep8`](https://github.com/celery/celery/blob/main/CONTRIBUTING.rst#id84) is a utility you can use to verify that your code is following the conventions.
> 
> Docstrings must follow the [PEP 257](http://www.python.org/dev/peps/pep-0257) conventions, and use the following style.

@Nusnus:

> were the immediate signal that “something is conceptually wrong”

I didn't like this change either, but the approach of testing the bindings directly rather than trying to send tasks through the entirety of the delayed delivery queues seemed less likely to be flaky, as I explain [here](https://github.com/celery/celery/pull/9959#issuecomment-3426965288).

@Nusnus:

> Classic case where AI slop is infecting the codebase.

You may be fast to call this AI slop.  I tend to be quite ["lazyless"](https://github.com/Izzette/izix) myself (well into the pre-GenAI era).  While it may be slop, it is my own.  If you must call it slop, it prefers to be called autistic slop.

Indeed—[I like UTF-8 symbols](https://github.com/celery/celery/pull/9959/changes#diff-f53943628415469afaae9c1fee1153ea5ef51632044afa4005e91729499f4629R168)—especially [arrows](https://github.com/ffi/ffi/issues/942#issue-1138991036) and have been using them for years, ever since getting a mac book and being able to type them so easily with ctrl→command→space.  And I rather like comments, even when [I'm writing them for myself](https://github.com/Izzette/sh-nspawn/blob/master/sh-nspawn#L127-L147).  I certainly take great pleasure in writing [long](https://github.com/kubernetes/ingress-nginx/issues/12206), [detailed](https://github.com/containerd/containerd/issues/8381), and [technical](https://github.com/ffi/ffi/issues/942) issues too.

I'm not opposed to AI coding tools, but I don't think it's ethical to submit a PR that didn't meet the apparent author's expectations for quality and robustness, something that they wouldn't be happy to certify is their own work.  I'm sorry it doesn't met your expectations.  If you have any criticisms you can freely address them directly to me.

If there's something really wrong here, I would hope we can work together address any systematic issues in play: such as outdated / missing documentation, or additional reviews from new contributors.  🫶🏻

docker/docker-compose.yml
auvipy: this are almost fixed now I think

Izzette: I'm proposing this PR to disable the management plugin in integration tests: https://github.com/celery/celery/pull/10250

Nusnus:
@Izzette 
> which seems to get an approval from the maintainer that was kind enough to review my PR

I think my comment was misinterpreted. Please allow me to improve the communication.

First of all, thank you for the contribution. I understand why it felt unexpected to see my feedback on the approved testing infra changes. The pattern introduced by this PR was copied by other AI agents and required major cleanups to restore everything to its normal state. The slop/lazy comment was not aimed at anyone personally. I apologize if you were offended by it 🙏

> I did my best here to provide integration tests for the issue as requested by your co-maintainer https://github.com/celery/celery/pull/9959#pullrequestreview-3357303430

The problem had multiple parts:
1. The CI **design** has a gap (found by this incident) that prevented surfacing the problem earlier. I am exploring ways to improve it to prevent this from repeating, but this is a real gap at the moment (root cause: pref optimizations).
2. As you've [already noticed](https://github.com/celery/celery/pull/10237), the integration matrix missed some tests as well; most of them are part of the incident too. Not by chance.
3. `t/integration/test_native_delayed_delivery_binding.py` had to manipulate the infra to force it to work in the integration tests. _This was the smoking gun_. This is also a common behavior by AI Agents; they tend to push for things to work (and make it look right). When I saw the test module had more infra than tests, my AI-slop spider sense tingled immediately.
4. There's a dedicated smoke tests package for this feature: https://github.com/celery/celery/tree/main/t/smoke/tests/quorum_queues , someone missed it. By someone I mean everyone, including the LLM 🙂
5. The PR is a fix, not a feature contrib: the fact that it had to create a new test for this feature was alarming. You'd expect it to expand upon existing tests, not create them from scratch. @auvipy asked "can you also explore possibility to add integration test for this?" - totally fair request IMHO. The correct answer should have been that integration tests aren't applicable and that smoke tests are the only valid layer, then add it there. The test isolation is also a common AI behavior that aims at isolating test code to overcome a lack of understanding of how the test infrastructure works and picking an approach that is generally safe. The AI was right: by isolating the test and running on a separate CI runner, the testing infra problems were bypassed, and everything appeared legit. The AI literally exploited the CI gap to make this work. I find it fascinating to be honest 😉

The cascading effects of this incident are a classic example of an AI issue that spread unnoticed for about six months and was fully resolved by @aurangzaib048 in the last few days.

> You may be fast to call this AI slop. I tend to be quite "lazyless" myself (well into the pre-GenAI era). While it may be slop, it is my own. I'm not opposed to AI coding tools, but I would never submit a PR that didn't meet my expectations for quality and robustness, something that I wouldn't be happy to certify is my work. I'm sorry it doesn't met yours.

Again, I apologize for "breaking in" without proper context. The [comment](https://github.com/celery/celery/pull/9959/changes#r3036062438) was a continuation of another conversation (spread over multiple PRs), but from your perspective, I can see how it may have appeared differently. I appreciate your effort on this PR, it is definitely clear there was nothing lazy about it.

It will take us all some time to adapt to this new world. I personally think these challenges are very interesting - we not only got the original issues fixed, but we also know that the CI design has a flaw, and exactly which one. From my perspective, the glass is half full 🍷

Izzette:
@Nusnus thank you for your response and your work on Celery 💜

PR #9966, opened 2025-10-25
Fix ExceptionGroup import for Python < 3.11

kernelshard:
- Add conditional dependency for exceptiongroup package
- Update import logic to use builtins for Python >= 3.11 and exceptiongroup for < 3.11
- Add test to verify correct import based on Python version

### DIFF
--- a/celery/worker/consumer/delayed_delivery.py
+++ b/celery/worker/consumer/delayed_delivery.py
@@ -4,11 +4,10 @@ This module provides the DelayedDelivery bootstep which handles setup and config
 of native delayed delivery functionality when using quorum queues.
 """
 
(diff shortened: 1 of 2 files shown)


cjwatson:
@auvipy I'm curious why this PR was closed without comment?  I was about to propose something similar; we've removed our packaging of `exceptiongroup` from Debian because it's no longer needed on modern Python versions, so in order to package Celery 5.6.0 I need something like this.

Would a similar patch be welcomed if it were stripped down a bit more, without all the extraneous formatting changes?

auvipy:
at that time we got another solution. you can come with similar PR if what you say is true for this case. feel free to come with a pr and ping me

cjwatson:
OK, thanks.  I filed #10033.

PR #9967, opened 2025-10-25
Pin tblib to ==3.1.0

kernelshard:
Pin tblib to ==3.1.0 to ensure consistent traceback serialization across Python versions.

### DIFF
--- a/requirements/extras/tblib.txt
+++ b/requirements/extras/tblib.txt
@@ -1,2 +1,1 @@
-tblib>=1.5.0;python_version>='3.8.0'
-tblib>=1.3.0;python_version<'3.8.0'
+tblib==3.1.0


requirements/extras/tblib.txt
```diff
@@ -1,2 +1 @@
-tblib>=1.5.0;python_version>='3.8.0'
-tblib>=1.3.0;python_version<'3.8.0'
+tblib==3.1.0
```
Izzette, line 1: Shouldn't this be `tblib>=1.50,<3.2.0`?  This would allow users of Celery to have versions of tblib in the working version range, that may be used by other libraries of the project?  This might force people to upgrade to a version incompatible with other dependencies.
```suggestion
tblib>=1.50,<3.2.0
```

auvipy: it is better to stick to specific version. why should we try so old version?

PR #9938, opened 2025-10-11
Prepare for (pre) release: v5.6.0b2

Nusnus:
@auvipy FYI

PR #9968, opened 2025-10-27
Make tests compatible with pymongo >= 4.14

cjwatson:
## Description

https://github.com/mongodb/mongo-python-driver/pull/2413 caused some test regressions here.  This isn't currently a problem for the upstream test suite since it pins pymongo==4.10.1 via kombu, but we're running into it in Debian where we've already upgraded pymongo for other reasons.  kombu already tried to upgrade pymongo but had to revert due to these test regressions (see https://github.com/celery/kombu/pull/2384 and https://github.com/celery/celery/pull/9938).

One of the test fixes (relating to `mongodb_backend_settings`) illustrates an incompatibility where I couldn't figure out a reasonable way to avoid passing it through to Celery users, so I added a note to the documentation about it.  It may also be worth including a brief mention of it in the release notes.  Using the canonical case for the option in question should work with both old and new versions of pymongo.

### DIFF
--- a/celery/backends/mongodb.py
+++ b/celery/backends/mongodb.py
@@ -20,6 +20,7 @@ if pymongo:
         from bson.binary import Binary
     except ImportError:
         from pymongo.binary import Binary
+    from pymongo import uri_parser
     from pymongo.errors import InvalidDocument
 else:                                       # pragma: no cover
     Binary = None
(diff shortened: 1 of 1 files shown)


cjwatson:
Ugh, I thought this would work with both old and new versions but CI clearly disagrees.  Let me see if I can improve this ...

cjwatson:
This isn't quite as clean as my original version, but I've checked that it works with both old and new versions of pymongo.

auvipy:
we cam upgrade pymongo version in kombu as well

cjwatson:
Certainly, though I guess that would be after something like this lands in celery.

auvipy:
yes

cjwatson:
> can you also incorporate changes from this pr #9665

I mean, I suppose, but why?  That PR affects exactly one pymongo version, namely 4.12.0, as the pymongo bug that it's working around was fixed in 4.12.1; so it's completely disjoint with the versions that I'm dealing with in this PR.  Why not just merge the two PRs separately?

auvipy:
OK sure.

auvipy (approved):

PR #9980, opened 2025-11-01
Update CI matrix to exclude Python 3.9

auvipy:
Removed Python 3.9 from the CI matrix for Windows.


Nusnus (commented):
Let's prepare all of the Python 3.9 deprecation changes here, WDYT?
e.g `tox.ini` etc.

auvipy:
yes that is after 5.6 and 5.6.x release. those are first priority for us

PR #9987, opened 2025-11-10
Update pyupgrade config for Python 3.9

ulgens:
## Description

Update pyupgrade config for Python 3.9. It seems this wasn't handled during 3.8 removal.

### DIFF
--- a/celery/utils/annotations.py
+++ b/celery/utils/annotations.py
@@ -37,9 +37,7 @@ def get_optional_arg(annotation: typing.Any) -> typing.Any:
 def annotation_is_class(annotation: typing.Any) -> bool:
     """Test if a given annotation is a class that can be used in isinstance()/issubclass()."""
     # isclass() returns True for generic type hints (e.g. `list[str]`) until Python 3.10.
-    # NOTE: The guard for Python 3.9 is because types.GenericAlias is only added in Python 3.9. This is not a problem
-    #       as the syntax is added in the same version in the first place.
-    if (3, 9) <= sys.version_info < (3, 11) and isinstance(annotation, types.GenericAlias):
+    if sys.version_info < (3, 11) and isinstance(annotation, types.GenericAlias):
         return False
     return isclass(annotation)
 
--- a/requirements/test.txt
+++ b/requirements/test.txt
@@ -1,7 +1,7 @@
 pytest==8.4.2
 pytest-celery[all]>=1.2.0,<1.3.0
-pytest-rerunfailures>=15.0; python_version >= "3.9"
-pytest-subtests>=0.14.1; python_version >= "3.9"
+pytest-rerunfailures>=15.0
+pytest-subtests>=0.14.1
 pytest-timeout==2.4.0
 pytest-click==1.1.0
 pytest-order==1.3.0
@@ -9,7 +9,7 @@ boto3>=1.26.143
 moto>=4.1.11,<5.1.0
 # type checking
 mypy==1.14.1; platform_python_implementation=="CPython"
-pre-commit>=4.0.1; python_version >= '3.9'
+pre-commit>=4.0.1
 -r extras/yaml.txt
 -r extras/msgpack.txt
 -r extras/mongodb.txt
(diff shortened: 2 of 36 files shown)


celery/utils/annotations.py
```diff
@@ -37,9 +37,7 @@ def get_optional_arg(annotation: typing.Any) -> typing.Any:
 def annotation_is_class(annotation: typing.Any) -> bool:
     """Test if a given annotation is a class that can be used in isinstance()/issubclass()."""
     # isclass() returns True for generic type hints (e.g. `list[str]`) until Python 3.10.
-    # NOTE: The guard for Python 3.9 is because types.GenericAlias is only added in Python 3.9. This is not a problem
-    #       as the syntax is added in the same version in the first place.
-    if (3, 9) <= sys.version_info < (3, 11) and isinstance(annotation, types.GenericAlias):
+    if sys.version_info < (3, 11) and isinstance(annotation, types.GenericAlias):
```
auvipy, line 40: should be 3, 10

ulgens: @auvipy Can you please give more details? This upper version limit is not something I touch with this PR and I don't know much about why it's implemented.

Nusnus: It appears only this was removed:
```python
# NOTE: The guard for Python 3.9 is because types.GenericAlias is only added in Python 3.9. This is not a problem
    #       as the syntax is added in the same version in the first place.
    if (3, 9) <=
```

Which matches the scope of the PR. Anything we’re missing @auvipy ?

Nusnus (commented):
Thank you for this PR!

LGTM so far, though we’ll need to wait before deprecating 3.9 in `main`.

ulgens:
@Nusnus Thank you. To clarify, this PR doesn't serve to the deprecation of 3.9 - it cleans residues from <3.9 support and can be merged safely in the current release.

requirements/test.txt
```diff
@@ -1,15 +1,15 @@
 pytest==8.4.2
 pytest-celery[all]>=1.2.0,<1.3.0
-pytest-rerunfailures>=15.0; python_version >= "3.9" and python_version < "4.0"
-pytest-subtests>=0.14.1; python_version >= "3.9" and python_version < "4.0"
+pytest-rerunfailures>=15.0; python_version < "4.0"
+pytest-subtests>=0.14.1; python_version < "4.0"
```
ulgens: I'm not sure why these <4.0 checks are added and if they are still needed.

Nusnus:
> @Nusnus Thank you. To clarify, this PR doesn't serve to the deprecation of 3.9 - it cleans residues from <3.9 support and can be merged safely in the current release.

Thank you for the correction. I was reviewing #9988 just before so I had my mind on the deprecation hehe

requirements/test.txt
Nusnus: Good catch, feel free to clean it up :)

ulgens: @Nusnus https://github.com/celery/celery/pull/9993

Nusnus: Merged

celery/utils/annotations.py
Nusnus: I am considering whether we need to revert this specific change @ulgens .

Original code:
```python
def annotation_is_class(annotation: typing.Any) -> bool:
    """Test if a given annotation is a class that can be used in isinstance()/issubclass()."""
    # isclass() returns True for generic type hints (e.g. `list[str]`) until Python 3.10.
    # NOTE: The guard for Python 3.9 is because types.GenericAlias is only added in Python 3.9. This is not a problem
    #       as the syntax is added in the same version in the first place.
    if (3, 9) <= sys.version_info < (3, 11) and isinstance(annotation, types.GenericAlias):
```

This is mostly a safeguard, a functionality, that we might actually want to keep for a few more years until Python 3.9 is really out of the game, or, if we officially drop existing support for edge cases, which I prefer not to, at this moment.

ulgens: @Nusnus The change doesn't affect the use of 3.9, it's for anything earlier than 3.9. Because 3.9 is the earliest supported version, `(3, 9) <= sys.version _info` will always be true - it doesn't need to be checked.

PR #9988, opened 2025-11-10
Drop Python 3.9 (EOL)

ulgens:
## Description

Drop support for Python 3.9 (EOL)

Supersedes:
- https://github.com/celery/celery/pull/9980

Depends on:
- https://github.com/celery/celery/pull/9987


celery/contrib/testing/manager.py
```diff
@@ -120,7 +120,7 @@ def on_result(task_id, value):
             received[:] = []
             try:
                 return r.get(callback=on_result, propagate=propagate, **kwargs)
-            except (socket.timeout, TimeoutError) as exc:
```
auvipy, line 123: why socket timeouts are being removed?

auvipy (commented):
I was planning to do this after 5.6.x release is done..

celery/contrib/testing/manager.py
ulgens: https://github.com/asottile/pyupgrade?tab=readme-ov-file#timeouterror-aliases
https://docs.python.org/3/whatsnew/3.10.html#socket

.github/workflows/integration-tests.yml
```diff
@@ -11,7 +11,7 @@ on:
         description: 'JSON array of Python versions to test'
         required: false
         type: string
-        default: '["3.9", "3.14"]'
+        default: '["3.10", "3.14"]'
```
Nusnus: I see this is a debt from `main`, but it should be 3.10, 3.13 and 3.14.
We can’t skip 3.13, since we don’t officially support 3.14 yet, so it’s better to add it.

ulgens: I'm a bit confused about this config. At first I thought this is a range but turned out it's an array. Why don't we test all Python versions we support?

Nusnus: > I'm a bit confused about this config. At first I thought this is a range but turned out it's an array. Why don't we test all Python versions we support?

Due to stability issues with our tests and lack of maintainability capacity, so we pick the min/max+3.14 in this case to trade off additional CI runs for easier maintainability on PRs (better rate at all CI passing without flaky issues).

ulgens: Thank you. Do we have failing tests tho? Would we want to check it again?

Nusnus: > Thank you. Do we have failing tests tho? Would we want to check it again?

I’d love to check it again, thank you. 3.10, 3.11, 3.12, 3.13, and 3.14 would be interesting in both integration and smoke tests CIs.

Nusnus: Only flakiness; all tests should pass across all Python versions.

Nusnus: @ulgens FYI: https://github.com/celery/celery/pull/9999

ulgens: Rebased from the latest main. Can be marked as resolved.

Issue #9989, opened 2025-11-10
Mention ScyllaDB support in documentation


andrii-kysylevskyi:
- [x] I have checked the [issues list](https://github.com/celery/celery/issues?utf8=%E2%9C%93&q=is%3Aissue+label%3A%22Issue+Type%3A+Feature+Request%22+)
  for similar or identical feature requests.
- [x] I have checked the [pull requests list](https://github.com/celery/celery/pulls?utf8=%E2%9C%93&q=is%3Apr+label%3A%22PR+Type%3A+Feature%22+)
  for existing proposed implementations of this feature.
- [x] I have checked the [commit log](https://github.com/celery/celery/commits/main)
  to find out if the same feature was already implemented in the
  main branch.
- [x] I have included all related issues and possible duplicate issues
  in this issue (If there are none, check this box anyway).

#### Related Issues

- None

#### Possible Duplicates

- None

# Brief Summary
I noticed that celery mentions the support of Cassandra. This support implicitly enables compatibility with [ScyllaDB](https://www.scylladb.com/) as well, since ScyllaDB is a drop-in replacement for Cassandra with a focus on performance and scalability.

Unfortunately, this support is not mentioned in the document, and my suggestion is to add this information.

Since this is not necessarily a documentation bug, I have opened this issue as a request. But happy to re-open it as a documentation bug instead. 

# Design

## Architectural Considerations

None

## Proposed Behavior
To mention ScyllaDB's support using the Cassandra driver. No additional changes needed.

## Proposed UI/UX
N/A

## Diagrams
N/A

## Alternatives
None

auvipy:
we do not have explicit support for scylladb. also I am not sure how popular it is. these type of stuff do not need an issue at all.

PR #9990, opened 2025-11-10
[9989] Add scyllaDB mention to documentation

andrii-kysylevskyi:
Updated documentation according to the original issue available [here](https://github.com/celery/celery/issues/9989).

I have also considered adding testing infrastructure for both Cassandra and Scylla, but decided to keep the changes to a minimum. Would be happy to change that, though and to add tests as well, if maintainers consider this valuable.


celery/app/backends.py
```diff
@@ -23,6 +23,7 @@
     'database': 'celery.backends.database:DatabaseBackend',
     'elasticsearch': 'celery.backends.elasticsearch:ElasticsearchBackend',
     'cassandra': 'celery.backends.cassandra:CassandraBackend',
+    'scylladb': 'celery.backends.cassandra:CassandraBackend',
```
auvipy, line 26: this is not a documentation change anymore with this change....

PR #9997, opened 2025-11-14
Sanitize broker URL in delayed delivery logs (avoid leaking credentials)

giancarloromeo:
## Description

Celery currently logs the full broker URL—including username and password— inside the `DelayedDelivery` consumer:

```
Setting up delayed delivery for broker `amqp://user:password@host:5672` ...
```

This leaks credentials in logs and violates standard security guidance (OWASP: no secrets in logs).

## Solution

Kombu already provides `maybe_sanitize_url()` for this purpose, but it wasn’t used here.


Nusnus (approved):

Nusnus:
> Is it possible to add a test to avoid future regression?

We can do that in a separate PR

PR #10021, opened 2025-12-08
 munally reverted https://github.com/celery/celery/pull/9824

MehrazRumman:
## Manual Revert
Reverts PR #9824, which applied Django’s `CONN_MAX_AGE` to Celery’s DB connections. This caused unintended persistent connections in workers, leading to connection buildup and instability. Restores previous explicit connection-handling behavior to maintain expected worker stability.

### DIFF
--- a/celery/fixups/django.py
+++ b/celery/fixups/django.py
@@ -168,7 +168,8 @@ class DjangoWorkerFixup:
             if c and c.connection:
                 self._maybe_close_db_fd(c)
 
-        # use the _ version to avoid DB_REUSE preventing the conn.close() call
+        # use the _close_database(force=True) to avoid DB_REUSE preventing connection closure
+        # and to ensure full connection close (not just close_if_unusable_or_obsolete) on worker process init
         self._close_database(force=True)
         self.close_cache()
 
--- a/celery/fixups/django.py
+++ b/celery/fixups/django.py
@@ -200,6 +200,16 @@ class DjangoWorkerFixup:
         self._db_recycles += 1
 
     def _close_database(self, force: bool = False) -> None:
+        """Close database connections.
+
+        Args:
+            force: If True, unconditionally close all connections via conn.close().
+                   If False (default), only close unusable or obsolete connections
+                   via conn.close_if_unusable_or_obsolete().
+
+        Use force=True during worker process initialization to ensure clean slate.
+        Use default (False) during normal task lifecycle for connection reuse optimization.
+        """
         for conn in self._db.connections.all():
             try:
                 if force:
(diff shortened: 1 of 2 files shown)


danlamanna:
> ## Manual Revert
> 
> Reverts PR #9824, which applied Django’s `CONN_MAX_AGE` to Celery’s DB connections. This caused unintended persistent connections in workers, leading to connection buildup and instability. Restores previous explicit connection-handling behavior to maintain expected worker stability.

I responded [here](https://github.com/celery/celery/pull/9824#issuecomment-3627169232) opposing this. In order to preserve the record, could you at least update the body of this PR to be accurate?

auvipy:
> I responded [here](https://github.com/celery/celery/pull/9824#issuecomment-3627169232) opposing this. In order to preserve the record, could you at least update the body of this PR to be accurate?

can you please elaborate more here?

shtlrs:
Posting this here just in case.

We ran into a lot of issues with 5.6.x when it comes to connection latency going through the roof.

We run django with postgres, so I can only speak for that.

What we've done to fix this is take the django fixup from version 5.5.3, and basically copy pasted it in our codebase, and injected it in celery via the `fixup` parameter when instantiating the celery app: `Celery(..., fixups=[fixup_function])`

This has re-established stability when testing, and we have never run into any of the issues mentioned in [here](https://forum.djangoproject.com/t/close-if-unusable-or-obsolete-fails-to-close-unusable-connections/41900)

auvipy:
> > ## Manual Revert
> > Reverts PR #9824, which applied Django’s `CONN_MAX_AGE` to Celery’s DB connections. This caused unintended persistent connections in workers, leading to connection buildup and instability. Restores previous explicit connection-handling behavior to maintain expected worker stability.
> 
> I responded [here](https://github.com/celery/celery/pull/9824#issuecomment-3627169232) opposing this. In order to preserve the record, could you at least update the body of this PR to be accurate?

copying from the thread:



    FWIW we ran into a similar issue with MySQL at job and we ended up setting a monkey patch of cursor wrapping that has SoftTimeLimitExceeded handling immediately close the connection as the C and Python layers got out of sync when a signal was raised mid query execution.

This does sound like a better solution.

The purpose of a soft limit (versus the hard limit) in celery is that the task can exit gracefully and clean up after itself. From celery docs: “The time limit is set in two values, soft and hard. The soft time limit allows the task to catch an exception to clean up before it is killed: the hard timeout isn’t catch-able and force terminates the task.”

If you keep the connection open, then you have not cleaned up, as the query will still be in progress and using resources from your database. I assume that’s why you get this error django.db.utils.OperationalError: sending query failed: another command is already in progress because the query (select pg_sleep) is still running.

shtlrs:
> This does sound like a better solution.

I don't think monkey patching django itself from celery is a good solution, this might work for people's local codebases, but doing this from a celery fixup doesn't sound appropriate.

Unless we make this on opt-in thing and well documented, i don't see this as a solution worthy of being in production running code.

There's another approach where we can advise people on how to do it properly by implementing their own cursor wrapper and plugging it in etc, but that doesn't sound like celery's responsibility ? I am not sure tbh

PR #10038, opened 2025-12-30
Add `__class_getitem__` to generic classes

brianhelba:
This will not change any behavior in normal usage, but will enable the `celery-types` project to use generics for:
* `Celery`
* `Task`
* `AsyncResult`
* `Signature`
* `_LocalStack`
* `_FastLocalStack`
* `FallbackContext`
* `class_property`

For example, this will allow annotations like:
```python
app = Celery[CustomTask]()
```

See the recent discussion: https://github.com/sbdchd/celery-types/issues/157#issuecomment-3696892836

### DIFF
--- a/celery/app/base.py
+++ b/celery/app/base.py
@@ -5,6 +5,7 @@ import inspect
 import os
 import sys
 import threading
+import types
 import typing
 import warnings
 from collections import UserDict, defaultdict, deque
@@ -1265,6 +1266,8 @@ class Celery:
             attrs['__reduce__'] = __reduce__
 
         return type(name or Class.__name__, (Class,), attrs)
+
+    __class_getitem__ = classmethod(types.GenericAlias)
 
     def _rgetattr(self, path):
         return attrgetter(path)(self)
--- a/celery/app/task.py
+++ b/celery/app/task.py
@@ -1,5 +1,6 @@
 """Task implementation: request context and the task base class."""
 import sys
+import types
 
 from billiard.einfo import ExceptionInfo, ExceptionWithTraceback
 from kombu import serialization
@@ -431,6 +432,8 @@ class Task:
         finally:
             self.pop_request()
             _task_stack.pop()
+
+    __class_getitem__ = classmethod(types.GenericAlias)
 
     def __reduce__(self):
         # - tasks are pickled into the name of the task only, and the receiver
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -306,3 +306,4 @@ Narasux, 2024/09/09
 Colin Watson, 2025/03/01
 Lucas Infante, 2025/05/15
 Diego Margoni, 2025/07/01
+Brian Helba, 2026/01/12
--- a/celery/canvas.py
+++ b/celery/canvas.py
@@ -7,6 +7,7 @@
 
 import itertools
 import operator
+import types
 import warnings
 from abc import ABCMeta, abstractmethod
 from collections import deque
@@ -809,6 +810,8 @@ class Signature(dict):
         """
         args, kwargs, _ = self._merge(args, kwargs, {}, force=True)
         return reprcall(self['task'], args, kwargs)
+
+    __class_getitem__ = classmethod(types.GenericAlias)
 
     def __deepcopy__(self, memo):
         memo[id(self)] = self
--- a/celery/local.py
+++ b/celery/local.py
@@ -8,6 +8,7 @@ Parts of this module is Copyright by Werkzeug Team.
 
 import operator
 import sys
+import types
 from functools import reduce
 from importlib import import_module
 from types import ModuleType
@@ -429,6 +430,8 @@ class class_property:
         self.__doc__ = info.__doc__
         self.__name__ = info.__name__
         self.__module__ = info.__module__
+
+    __class_getitem__ = classmethod(types.GenericAlias)
 
     def __get__(self, obj, type=None):
         if obj and type is None:
--- a/celery/result.py
+++ b/celery/result.py
@@ -2,6 +2,7 @@
 
 import datetime
 import time
+import types
 from collections import deque
 from contextlib import contextmanager
 from weakref import proxy
@@ -383,6 +384,8 @@ class AsyncResult(ResultBase):
             if parent:
                 graph.add_edge(parent, node)
         return graph
+
+    __class_getitem__ = classmethod(types.GenericAlias)
 
     def __str__(self):
         """`str(self) -> self.id`."""
--- a/celery/utils/objects.py
+++ b/celery/utils/objects.py
@@ -1,4 +1,5 @@
 """Object related utilities, including introspection, etc."""
+import types
 from functools import reduce
 
 __all__ = ('Bunch', 'FallbackContext', 'getitem_property', 'mro_lookup')
@@ -89,6 +90,8 @@ class FallbackContext:
     def __exit__(self, *exc_info):
         if self._context is not None:
             return self._context.__exit__(*exc_info)
+
+    __class_getitem__ = classmethod(types.GenericAlias)
 
 
 class getitem_property:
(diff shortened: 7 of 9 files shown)


celery/app/task.py
```diff
@@ -413,6 +413,9 @@ def __call__(self, *args, **kwargs):
             self.pop_request()
             _task_stack.pop()
```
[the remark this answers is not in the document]
brianhelba: Incorrect. Per [the docs](https://docs.python.org/3/reference/datamodel.html#object.__class_getitem__):
> When defined on a class, `__class_getitem__()` is automatically a class method. As such, there is no need for it to be decorated with `@classmethod` when it is defined.

celery/app/base.py
```diff
@@ -1266,6 +1266,9 @@ def __reduce__(self):
 
         return type(name or Class.__name__, (Class,), attrs)
```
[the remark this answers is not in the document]
brianhelba: Incorrect. Per [the docs](https://docs.python.org/3/reference/datamodel.html#object.__class_getitem__):
> When defined on a class, `__class_getitem__()` is automatically a class method. As such, there is no need for it to be decorated with `@classmethod` when it is defined.

celery/app/task.py
```diff
@@ -413,6 +413,9 @@ def __call__(self, *args, **kwargs):
...
 
+    def __class_getitem__(cls, key):
+        return cls
```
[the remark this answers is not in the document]
brianhelba: I don't think a test can provide much value. This method should never actually be called at runtime, it just needs to be defined for type checking to work (provided stubs are separately installed).

I'm fine to add a test if desired, but I think it would just be clutter to boost coverage as a vanity metric.

celery/app/base.py
```diff
@@ -1266,6 +1266,9 @@ def __reduce__(self):
...
 
+    def __class_getitem__(cls, key):
+        return cls
```
dlucredativ: Can you please explain why it is sufficient to just return the class? As an example, one has
```
from types import GenericAlias
assert list[int] == GenericAlias(list, int)
```
but not
```
class Foo:
    def __class_getitem__(cls, key):
        return cls

assert Foo[int] == GenericAlias(Foo, int)
```

celery/app/task.py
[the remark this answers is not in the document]
dlucredativ: You can test without calling it directly:

```
class Foo:
    def __class_getitem__(cls, key):
        return cls
assert Foo is Foo[int]
```
However, I am unsure if this is a correct assertion (see my other comment)

celery/app/base.py
dlucredativ: https://github.com/sbdchd/celery-types/issues/208

9en9i:
I think it makes sense to add `__class_getitem__` for the `AsyncResult` class right away.

What do you think?

celery/app/base.py
brianhelba: @dlucredativ I didn't know about `types.GenericAlias`, thanks!

My only concern is perception. Given the reluctance of some projects to adopt typing, I'm reluctant to do any `import typing` if it'll be perceived as a backdoor attempt to add type annotations to a project which has rejected them.

What exactly do you propose we change the `return cls` to instead? Something like:
```suggestion
    def __class_getitem__(cls, key):
        return typing.GenericAlias(cls, key)
```

brianhelba:
I've personally only used generics with `Celery` and `Task` annotations, so I admit I don't understand how other generics ought to be used.

@9en9i I've just added it to `AsyncResult`.

It seems like we add it to [`Signature`](https://github.com/sbdchd/celery-types/blob/f2309a1579834d4e460a5a0aee1181e07dcec83e/celery-stubs/canvas.pyi#L20) too?

By this rationale, should we also add it to [`class_property`](https://github.com/sbdchd/celery-types/blob/f2309a1579834d4e460a5a0aee1181e07dcec83e/celery-stubs/local.pyi#L12)? That's a generic in `celery-types`, but I think we only need it in places where runtime code wants to use subscripting, right? I'm not sure.

It seems like `AsyncResult` and `Signature` are typically defined as return values, so users don't always need to define explicit generic types for them, but if they want to have explicitly defined generic types for `AsyncResult` and `Signature` (like in a function signature), we want to ensure it won't break. Is that correct?

celery/app/base.py
brianhelba: I've made this change. I'd appreciate if someone else verifies the correctness of this at runtime, particularly how `ininstance` checks will work. I don't understand `GenericAlias` well enough.

celery/app/task.py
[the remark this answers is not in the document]
brianhelba: Unless someone actually asks for a test, I don't think it's necessary. If it's requested, I agree something like this is probably the best we can do.

celery/app/base.py
dlucredativ: > Given the reluctance of some projects to adopt typing, I'm reluctant to do any `import typing`

`typing.GenericAlias is types.GenericAlias`, so you don't need to `import typing`. The project already imports `types` in places like https://github.com/celery/celery/blob/main/celery/app/backends.py#L3.

> verifies the correctness of this at runtime, particularly how ininstance checks will work

This sounds like a use-case for a test, e.g.
```
--- a/t/unit/app/test_app.py
+++ b/t/unit/app/test_app.py
@@ -167,0 +168,7 @@ class test_App:
+    def test_task__class_getitem__(self):
+        from typing import get_args
+        from celery.app.task import Task
+        TaskIntStr = Task[[int], str]  # __class_getitem__ makes Task subscriptable
+        assert get_args(TaskIntStr) == ([int], str),  "__class_getitem__ returns a GenericAlias"
+        assert isinstance(TaskIntStr(), Task), "subtypes instantiate a Task"
```

9en9i:
@brianhelba Type stubs may still be incomplete, so a number of classes may actually be generics in the future. At the moment, all classes listed in the readme are declared as generics, including `Celery`, `Task`, `AsyncResult`, `Signature`, and `class_property`.
You typically use generics with `Celery` and `Task`. I use `Celery`, `Task`, and `AsyncResult` more often. This list may expand or shrink depending on what is used in the application code, so ideally you should have `__class_getitem__` (or simply mark them with `Generic`) for all of these classes. In general, adding `__class_getitem__` is useful where the runtime code will use type passing.

celery/app/base.py
```diff
@@ -1266,6 +1266,9 @@ def __reduce__(self):
 
         return type(name or Class.__name__, (Class,), attrs)
 
+    def __class_getitem__(cls, key):
+        return typing.GenericAlias(cls, key)
```
jonathanberthias: Hey, just a note that the standard library has several such dummy implementations, and they write it as:
```py
__class_getitem__ = classmethod(types.GenericAlias)
```
For instance [here](https://github.com/python/cpython/blob/a4086d7f89e5d388e4ffcdb13e4fba0255234286/Lib/queue.py#L277).

celery/app/base.py
brianhelba: You got me! 😅 

I'll add a test.

celery/app/task.py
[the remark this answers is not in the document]
brianhelba: I've added tests.

brianhelba:
@dlucredativ @9en9i @jonathanberthias Thank you all so much for the informative and prompt help! I've learned so much more than expected with this change.

@auvipy This is ready for re-review.

brianhelba:
@auvipy Is there anything else I can do to support the review process here? I've responded to all the feedback, so this is ready for final review and hopefully merging.

auvipy (changes requested):
please check the failing tests

brianhelba:
@auvipy I'm sorry for my confusion! (I thought this was a coverage failure) I fixed the failing test, as `typing.Self` isn't supported on Python 3.10.

Ready again.

auvipy (approved):

Issue #10042, opened 2025-12-31
AttributeError: 'str' object has no attribute '__module__' in django.py fixups when using Flower


maycuatroi1:
## Checklist

- [x] I have verified that the issue exists against the `main` branch of Celery.
- [x] This has already been asked to the [discussions forum](https://github.com/celery/celery/discussions) first.
- [x] I have read the relevant section in the [contribution guide](https://docs.celeryq.dev/en/stable/contributing.html#other-bugs) on reporting bugs.
- [x] I have checked the [issues list](https://github.com/celery/celery/issues?q=is%3Aissue+label%3A%22Issue+Type%3A+Bug+Report%22+-label%3A%22Category%3A+Documentation%22+) for similar or identical bug reports.
- [x] I have checked the [pull requests list](https://github.com/celery/celery/pulls?q=is%3Apr+label%3A%22PR+Type%3A+Bugfix%22+-label%3A%22Category%3A+Documentation%22+) for existing proposed fixes.
- [x] I have checked the [commit log](https://github.com/celery/celery/commits/main) to find out if the bug was already fixed in the main branch.
- [x] I have included all related issues and possible duplicate issues in this issue.

## Mandatory Debugging Information

- **Celery version**: 5.6.1
- **Flower version**: 2.0.1
- **Python version**: 3.12
- **OS**: Linux (Docker container)
- **Broker**: Redis
- **Result backend**: Redis

## Description

When starting Flower with a Django/Celery application, an `AttributeError` is raised in `celery/fixups/django.py` at line 220.

The error occurs because `self.worker.pool_cls` is a string (`'prefork'`) instead of a class object, so accessing `.__module__` fails.

## Error Traceback

```
File "/usr/local/lib/python3.12/site-packages/celery/fixups/django.py", line 220, in _close_database
    is_prefork = "prefork" in self.worker.pool_cls.__module__
                              │    │      └ 'prefork'
                              │    └ <Worker: celery@75efc8aa2596 (INIT)>
                              └ <celery.fixups.django.DjangoWorkerFixup object at 0x7f77b5998f20>

AttributeError: 'str' object has no attribute '__module__'. Did you mean: '__mod__'?
```

## Full Stack Trace

```
File "/usr/local/lib/python3.12/site-packages/celery/fixups/django.py", line 116, in on_worker_init
    self.worker_fixup.install()
File "/usr/local/lib/python3.12/site-packages/celery/fixups/django.py", line 161, in install
    self.close_database()
[...]

PR #10048, opened 2026-01-03
Fix Django worker recursion bug + defensive checks for pool_cls.__module__

maycuatroi1:
## Description

This PR builds on top of #10045 by @bruunotrindade to add defensive checks for `pool_cls.__module__` access in additional locations throughout the codebase.

### Root Cause (from #10045)

`DjangoWorkerFixup.__init__` was creating a `WorkController` instance via the `worker` or `WorkController(app)` pattern. This created a circular dependency where:

- `WorkController.__init__` triggers the `worker_init` signal
- `DjangoFixup.on_worker_init` creates/accesses `DjangoWorkerFixup`
- `DjangoWorkerFixup.__init__` tries to create another `WorkController`
- Infinite recursion ensues

Additionally, `pool_cls` can be a string (e.g., `'prefork'`, `'solo'`, `'gevent'`) instead of a class object, which causes `AttributeError: 'str' object has no attribute '__module__'` when trying to access `pool_cls.__module__`.

### Changes in this PR

This PR includes all changes from #10045 plus:

1. **`celery/contrib/testing/worker.py`**: Added defensive check for `pool_cls` being a string in `TestWorkController.__init__`

2. **`celery/worker/components.py`**: Added defensive check for `pool_cls` being a string in `Beat.create`

3. **New unit tests**: Added tests to verify the defensive checks work correctly when `pool_cls` is a string

### Pattern Applied

```python
# Before (can raise AttributeError if pool_cls is a string)
if w.pool_cls.__module__.endswith(('gevent', 'eventlet')):

# After (handles both string and class pool_cls)
pool_module = w.pool_cls if isinstance(w.pool_cls, str) else w.pool_cls.__module__
if pool_module.endswith(('gevent', 'eventlet')):
```

## Testing

- All existing unit tests pass
- Added 2 new tests for `Beat.create` with string pool_cls
- Pre-commit hooks pass (flake8, mypy, isort, etc.)
- Test coverage maintained

## Related Issues

- #10042 - Original issue reporting RecursionError
- #10045 - Original PR by @bruunotrindade fixing the root cause
- #10023 - Previous attempts to fix the issue

## Acknowledgments

Thanks to @bruunotrindade for identifying and fixing the root cause in #10045. This PR extends that fix to cover additional edge cases.

### DIFF
--- a/t/unit/contrib/test_worker.py
+++ b/t/unit/contrib/test_worker.py
@@ -1,3 +1,5 @@
+from unittest.mock import Mock, patch
+
 import pytest
 
 # this import adds a @shared_task, which uses connect_on_app_finalize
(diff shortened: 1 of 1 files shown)


auvipy (changes requested):
please check the unit test failures

auvipy:
may be we can consider the remaining changes here as minor improvements

auvipy (changes requested):
is it possible to improve test coverage?

maycuatroi1:
@auvipy Added unit tests for `TestWorkController` with string `pool_cls` to improve test coverage. Please review.

auvipy:
can you please check why the tests are failing with new changes?

maycuatroi1:
@auvipy Fixed a test pollution issue discovered during local CI testing.

**Root cause:** The `test_init_with_string_pool_cls_prefork` test was calling `tblib.pickling_support.install()` (via `TestWorkController.__init__`), which modifies global pickling behavior and caused `test_retry_with_unpickleable_exception` to fail when running the full test suite.

**Fix:** Added `patch.dict('sys.modules', {'tblib': None, 'tblib.pickling_support': None})` to prevent tblib from being imported during the test.

Note: Force pushed to amend the previous commit with this fix. Apologies for not creating a separate commit.

auvipy (approved):

PR #10074, opened 2026-01-27
Make tests compatible with pymongo >= 4.16

cjwatson:
## Description

`InvalidDocument` now requires a `message` argument (following https://github.com/mongodb/mongo-python-driver/pull/2539); it was previously optional.

### DIFF
--- a/t/unit/backends/test_mongodb.py
+++ b/t/unit/backends/test_mongodb.py
@@ -390,7 +390,7 @@ class test_MongoBackend:
                                                             upsert=True)
         assert sentinel.result == ret_val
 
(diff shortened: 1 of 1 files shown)


auvipy (commented):
would you mind doing the same for kombu as well?

cjwatson:
kombu has no references to `InvalidDocument`, and it seems to pass its tests with pymongo 4.16, so I can't think of anything to do there.  (And I think the versions in its requirements should probably only be bumped once that won't break celery.)

auvipy (approved):

PR #10105, opened 2026-02-10
fix: source install of cassandra-driver

Izzette:
## Description

The CI is currently broken when installing cassandra-driver from source on python 3.14.  This is occuring since the release of [setuptools 82.0.0 which removes pkg_resources](https://setuptools.pypa.io/en/stable/history.html#v82-0-0).  This issue has been identified on the upstream cassandra-python-driver and a [fix](https://github.com/apache/cassandra-python-driver/pull/1268) has been proposed, but is not yet available.

This PR applies a workaround until a new release of cassandra-driver and/or binary builds for python 3.14 is available.


Izzette:
Why doesn't it work, it fixes the issue when running `tox` locally on python 3.14 :(

tox.ini
```diff
@@ -42,6 +42,9 @@ deps=
     lint: pre-commit
     bandit: bandit
 
+constraints=
```
potiuk: I don't think this will work. You need to limit build-constraints not constraints, because by default cassandra driver will build in isolated environment and version of setuptools in that isolated environment is not limited byrregular constraints when you install the package.

Izzette: Pip constraints affect both the build environment as well: https://pip.pypa.io/en/stable/user_guide/#build-constraints:~:text=Build%20constraints%20are,the%20build%20process.

Tox doesn't provide a way to specify only `--build-constraints`, just `--constraints` affecting both build and package environments.  As tox [documentation](https://tox.wiki/en/latest/config.html#constraints) states:

> [Constraints files](https://pip.pypa.io/en/stable/user_guide/#constraints-files) to use during package and dependency installation. Provided constraints files will be used when installing package dependencies and any additional dependencies specified in [deps](https://tox.wiki/en/latest/config.html#deps), but will not be used when installing the package itself. Each value must be a file path or URL.

Izzette: I'll rename the file to reflect this.

Izzette:
I've managed to fix the docker build, but tox is still failing to install cassandra-driver in the CI.  I don't have this issue locally anymore, and normally it logs a message about installing setuptools 81.0.0.  I'm going to have to play around a bit to figure out what's different in the CI.

tox.ini
potiuk: > Pip constraints affect both the build environment as well: https://pip.pypa.io/en/stable/user_guide/#build-constraints:~:text=Build%20constraints%20are,the%20build%20process.

Could be that tox implemented something around it, I think this part of https://peps.python.org/pep-0517/ does not specify exactly how it should behave. And it's not necessarily clear if "installing deps" might also "building deps" - those two are distinct environments with distinct dependency sets.

For example the behaviour of `pip` for constraints is ambiguous and also depends on build backend and specification of build dependencies in pyproject.toml in .sdist package. What usually happens when you have unbounded dependencies in `pyproject.toml` this happens when you are installing with `pip install`  (this is empirically verified and also used in a number of workarounds in the past):

* `pip` creates a separate build environment based on build dependencies in pyproject.toml
* when you use `--constraints` flag - it will not propagate the flag when creating the venv, and since this is a new venv - by default it will install latest `setuptools`
* whe you use `PIP_CONSTRAINTS` as environment variable (exported) it will be propagated to `pip` command that creates the isolated build virtual env internally, and whatever is specified as file in `PIP_CONSTRAINTS` will be used for the installation 
* you can also use `--no-build-isolation` flag when you run the main `pip install` command, but in this case you need to install the `build-requirements` in the right versions before running the `pip install` command - so preceding the install with separate `pip install setuptools<82.0.0` should do the job:
```
--no-build-isolation        Disable isolation when building a modern source distribution. Build dependencies specified by PEP 518 must be already installed if this option is used.
```

So if tox also propagates constraints to the isolated build environment or disables isolated build environment feature then indeed `constraints` might work. But it's not too clear (also `pip` behaviour is empirical rather than specified - it's merely a byproduct of default behaviour of environment variables being passed to subprocesses and not removed by `pip`.

potiuk:
> I've managed to fix the docker build, but tox is still failing to install cassandra-driver in the CI. I don't have this issue locally anymore, and normally it logs a message about installing setuptools 81.0.0. I'm going to have to play around a bit to figure out what's different in the CI.

I think this is precisely what happens when you have a clean install on CI - seems that tox also separates "isolated build env" from "installation env" and constratints do not apply to the `build environments`. 

So either specifiying constraint file by the `PIP_CONSTRAINTS` variable or disabling isolated build environment should work for you.

tox.ini
Izzette: I can try `PIP_CONSTRAINTS` in the environment.  Locally I was previously able to reproduce the issue in the CI with python 3.14.  When I added constraints, the issue went away.  I see in the logs for tox where it specifies `-c`:
> `3.14-unit: install_deps> python -I -m pip install -r /Users/izzi/src/celery/requirements/test.txt -r /Users/izzi/src/celery/requirements/pkgutils.txt -r /Users/izzi/src/celery/requirements/test-ci-default.txt -r /Users/izzi/src/celery/requirements/docs.txt -c /Users/izzi/src/celery/requirements/constraints.txt`

As well as here:
> `3.14-unit: install_package_deps> python -I -m pip install 'billiard<5.0,>=4.2.1' 'click-didyoumean>=0.3.0' 'click-plugins>=1.1.1' 'click-repl>=0.2.0' 'click<9.0,>=8.1.2' 'exceptiongroup>=1.3.0; python_version < "3.11"' 'kombu>=5.6.0' 'python-dateutil>=2.8.2' tzlocal 'vine<6.0,>=5.1.0' -c /Users/izzi/src/celery/requirements/constraints.txt`

I'm not sure why this works locally and not in the CI, it's the same pip version and the same tox version.

potiuk: The -c does not work for the build environment. However if you built the package locally for 3.14 before setuptools relased 82.0.0 and `pip` cached the built version, it might use this version cached locally. Not sure why it would not use it before, but for sure `pip -c` does **NOT** get propagate to the isolated build environment.

Izzette: 🤦‍♀️
```console
$ rm /Users/izzi/Library/Caches/pip/wheels/38/e7/82/f2c45e0e7dc08a62791800cc8e0e792601bd8177cc188d0af3/cassandra_driver-3.29.3-cp314-cp314-macosx_15_0_arm64.whl
$
```

potiuk: Yeah :) .. 

> There are 2 hard problems in computer science: cache invalidation, naming things, and off-by-1 errors.

requirements/constraints.txt
```diff
@@ -0,0 +1,3 @@
+# Source install of cassandra-driver is broken in setuptools >=82.0.0, see
+# https://github.com/apache/cassandra-python-driver/pull/1268
+setuptools<82.0.0
```
auvipy, line 3: this is right

auvipy (approved):

tox.ini
```diff
@@ -42,6 +42,8 @@ deps=
     lint: pre-commit
     bandit: bandit
 
+install_command = python -I -m pip install {opts} {packages} --build-constraint {toxinidir}/requirements/constraints.txt
```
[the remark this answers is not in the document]
auvipy: does the suggestion make more sense in this context?

docker/Dockerfile
```diff
@@ -102,7 +102,8 @@ RUN --mount=type=cache,target=/home/$CELERY_USER/.cache/pip \
     -r requirements/test-ci-base.txt \
     -r requirements/test-ci-default.txt \
     -r requirements/test-integration.txt \
-    -r requirements/test.txt
+    -r requirements/test.txt \
+    --build-constraint requirements/constraints.txt
```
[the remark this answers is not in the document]
auvipy: is this needed on all python version?

auvipy (changes requested):
please check the reviews

requirements/constraints.txt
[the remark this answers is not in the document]
Izzette: I feel like we want to be using the most similar build environment possible for different versions of python. We wouldn't want to, for example, create circumstances where we can encounter issues in only 1 of the build environments due to a difference in the setuptools version. The install of cassandra-driver from source is still broken in other python versions, it's just that a pre-built version is already available.  We're just dodging a bullet in CPython 3.11-3.13

requirements/constraints.txt
```diff
@@ -1,3 +1,3 @@
 # Source install of cassandra-driver is broken in setuptools >=82.0.0, see
 # https://github.com/apache/cassandra-python-driver/pull/1268
-setuptools<82.0.0
+setuptools<82.0.0; python_version >= "3.14"
```
Izzette: Applying only for python version 3.14 breaks the docker build for pypy3.11, I'm not so sure it's really desirable to apply this constraint for only python versions without a binary build available on pypi: https://github.com/celery/celery/pull/10105#discussion_r2788857609

docker/Dockerfile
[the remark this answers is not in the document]
Izzette: The source build of cassandra-driver is broken in all versions of python, not just 3.14.  It's just that a binary build is available for the versions we run in the CI except pypy3.11 and 3.14.  I feel like we **want** to be using the most similar build environment possible for different versions of python.  We wouldn't want to, for example, create circumstances where we can encounter issues in only 1 of the build environments due to a difference in the setuptools version.

tox.ini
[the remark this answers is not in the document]
Izzette: This makes sense, but it's not as easy to apply the build constraint only for cassandra-driver in the docker build.  I'm not sure we **want** to have a different build environment in the docker build than when running tox?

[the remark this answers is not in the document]
Izzette: It doesn't seem to work on Windows for some reason: https://github.com/celery/celery/actions/runs/21869251158/job/63119178590?pr=10105#step:6:467

requirements/constraints.txt
auvipy: OK lets lets revert this

tox.ini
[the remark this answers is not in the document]
Izzette: Yeah this really doesn't work on windows, I'm not sure why.

auvipy (approved):

Issue #10067, opened 2026-01-16
Doc: Migration guideline from 4.x to 5.x should add more description for gevent user.


ycc789741ycc:
# Checklist

- [x] I have checked the [issues list](https://github.com/celery/celery/issues?utf8=%E2%9C%93&q=is%3Aissue+label%3A%22Category%3A+Documentation%22+)
  for similar or identical bug reports.
- [x] I have checked the [pull requests list](https://github.com/celery/celery/pulls?q=is%3Apr+label%3A%22Category%3A+Documentation%22)
  for existing proposed fixes.
- [x] I have checked the [commit log](https://github.com/celery/celery/commits/main)
  to find out if the bug was already fixed in the main branch.
- [x] I have included all related issues and possible duplicate issues in this issue
      (If there are none, check this box anyway).

#### Related Issues

- None

#### Possible Duplicates

- None

# Description

In Celery 4.x, some of user use monkey patch approch to enable gevent.
```
from gevent import monkey

monkey.patch_all()
```
In Celery 5.x this way won't work anymore, applying the monkey patch directly inside the code can lead to the worker failing to become ready.
Following info does not show up:
```
[2026-01-16 08:23:18,801: INFO/MainProcess] celery@f7b7f3f6605a ready.
```
User should remove monkey patch code and substitute with cli approach with `--pool=gevent`, ex:
```
celery -A app.celery worker --loglevel=info --pool=gevent
```

# Suggestions

Add more description in here, migration guideline of gevent: https://docs.celeryq.dev/en/main/history/whatsnew-5.0.html#step-1-adjust-your-command-line-invocation

compyman:
The documentation has always said that you need to enable concurrent workers by using the `--pool` argument. I don't think the documentation is misleading.

It's somewhat strange also to raise this issue 5 years after celery 5 was released. Do you think this documentation change is really helpful to many people?

ycc789741ycc:
Recently, I upgraded my company project celery from 4 to 5, I found that gevent monkey patch without assigning pool type as gevent can still work well. (Although using monkey patch with prefork worker is really strange, but I'm not pretty sure why it still works at 4)

Some Stone Age projects may misuse the way to enable gevent. At least, I think it is better to jump up the warning log instead of letting the worker be blocked and waiting for the ready message.

auvipy:
we will be happy to accept proper contribution for this

zainnadeem786:
[...]

PR #10175, opened 2026-03-05
chore: jump warning log when gevent/eventlet monkey patch with prefor…

ycc789741ycc:

## Description

In Celery 4.x, some of user use monkey patch approch to enable gevent.

```
from gevent import monkey

monkey.patch_all()
```
In Celery 5.x, manually adding monkey patch is not needed anymore. 
If adding monkey patch without specifying the right worker type(default is prefork), the following info ready message will **not** show up:

```
[2026-01-16 08:23:18,801: INFO/MainProcess] celery@f7b7f3f6605a ready.
```

Without any warning for the pending worker, which makes it difficult to debug.

## PR Overview

Pop out with warning if user using monkey patch for gevent/eventlet with prefork worker type

image

Chore #10067


auvipy (changes requested):
can you please cross check the automated review suggestions?

ycc789741ycc:
@auvipy LGTM

auvipy:
@ChickenBenny it would be great if you have some time to carefully review this one

auvipy (changes requested):
please cross check the suggestions if they are appropriate of not

t/unit/worker/test_worker.py
```diff
@@ -1265,3 +1266,86 @@ def patched_write(fd, data):
...
+        mock_eventlet_detect.return_value = True
+
+        worker = app.Worker(pool='prefork', quiet=True)
+        with patch.object(worker, 'install_platform_tweaks'):
+            with patch.object(worker, 'set_process_status'):
+                worker.on_start()
+
+        mock_logger.warning.assert_called()
+        warning_messages = [call[0][0] for call in mock_logger.warning.call_args_list if call[0]]
+        assert any(W_EVENTLET_MONKEY_PATCHED_WITH_WRONG_POOL in msg for msg in warning_messages)
```
ChickenBenny, line 1316: `test_warns_when_gevent_patched_with_prefork_pool` and `test_warns_when_eventlet_patched_with_prefork_pool` mock `is_*_monkey_patched()` but never ensure the corresponding key exists in sys.modules. If the test environment does not have `gevent/eventlet` installed, the outer guard evaluates False and the warning branch is never reached — the tests pass vacuously without exercising the code under test.

Could we add a `patch.dict` for `sys.modules` in those tests?
```python
with patch.dict(sys.modules, {'gevent': MagicMock()}):
    worker.on_start()
```

t/unit/utils/test_warning.py
```diff
@@ -0,0 +1,91 @@
...
+
+        real_import = builtins.__import__
+
+        def failing_import(name, *args, **kwargs):
+            if name.startswith('gevent'):
+                raise ImportError("Simulated gevent import failure")
+            return real_import(name, *args, **kwargs)
+
+        with patch('builtins.__import__', side_effect=failing_import):
+            result = is_gevent_monkey_patched()
```
ChickenBenny: `test_detect_gevent_monkey_patched_import_erro`r (and the eventlet equivalent) patches `builtins.__import__` to raise ImportError, but if gevent / gevent.monkey is already cached in `sys.modules` (e.g., loaded by a previous test), Python bypasses `__import__` entirely and the except `ImportError` branch is never exercised. This makes the tests environment-dependent.

Consider clearing the relevant `sys.modules` entries before patching `__import__`:
```python
with patch.dict('sys.modules', {'gevent': None, 'gevent.monkey': None}):
    with patch('builtins.__import__', side_effect=failing_import):
        result = is_gevent_monkey_patched()
assert result is False
```

ChickenBenny:
Should we also add the changelog and doc?

celery/apps/worker.py
```diff
@@ -165,6 +167,22 @@ def on_start(self):
...
+        is_eventlet_pool = 'eventlet' in pool
+        # Only perform monkey-patch detection if the corresponding
+        # libraries are already imported, to avoid importing them
+        # implicitly (which may have side effects) for non-green pools.
+        gevent_in_use = 'gevent' in sys.modules
+        eventlet_in_use = 'eventlet' in sys.modules
+        if gevent_in_use and not is_gevent_pool and is_gevent_monkey_patched():
+            logger.warning(W_GEVENT_MONKEY_PATCHED_WITH_WRONG_POOL, pool)
+        if eventlet_in_use and not is_eventlet_pool and is_eventlet_monkey_patched():
+            logger.warning(W_EVENTLET_MONKEY_PATCHED_WITH_WRONG_POOL, pool)
```
ChickenBenny: The pool-type check has two issues:

**1. Dead code:** By the time `on_start()` is called, `self.pool_cls` is always an already-resolved class (not a string) — `_concurrency.get_implementation()` runs in `setup_instance()` during `__init__`, before `on_start()`. The `isinstance(pool, str)` branch never executes.

**2. Substring vs. exact match:** Using `'gevent' in pool.__module__` works for the built-in pools but is more fragile than necessary. An exact comparison would be clearer:

```python
is_gevent_pool  = pool.__module__ == 'celery.concurrency.gevent'
is_eventlet_pool = pool.__module__ == 'celery.concurrency.eventlet'
```

Also the inline boolean check is easy enough to exercise by controlling `Worker(pool='gevent', ...)`, so extracting a separate helper function is optional. But if we want isolated unit tests for just the detection logic, moving it to a small helper in `celery/utils/warning.py` (alongside the existing `is_*_monkey_patched functions`) would keep things consistent.

t/unit/worker/test_worker.py
ycc789741ycc: If gevent/eventlet is not imported, the warning would not pop up, and I think the tests would be failed.
```
class Worker(WorkController):
    ...
    def on_start(self):
        ...
        gevent_in_use = 'gevent' in sys.modules
        eventlet_in_use = 'eventlet' in sys.modules
        if gevent_in_use and not is_gevent_pool and is_gevent_monkey_patched():
            logger.warning(W_GEVENT_MONKEY_PATCHED_WITH_WRONG_POOL, pool)
        if eventlet_in_use and not is_eventlet_pool and is_eventlet_monkey_patched():
            logger.warning(W_EVENTLET_MONKEY_PATCHED_WITH_WRONG_POOL, pool)
        ...
```
But it's still good to add gevent and eventlet mock in tests

ycc789741ycc:
Hi @auvipy , any follow-up for this PR?🙏

auvipy:
you have to make the CI green first... tests are failing

t/unit/utils/test_warning.py
```diff
@@ -0,0 +1,91 @@
...
+
+        def failing_import(name, *args, **kwargs):
+            if name.startswith('eventlet'):
+                raise ImportError("Simulated eventlet import failure")
+            return real_import(name, *args, **kwargs)
+
+        with patch('builtins.__import__', side_effect=failing_import):
+            result = is_eventlet_monkey_patched()
+
+        assert result is False
```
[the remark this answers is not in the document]
auvipy: @ycc789741ycc can you please cross check this and other suggestions?

PR #10176, opened 2026-03-06
chore: add Python 3.14 trove classifier

sk-:
## Description

PR #9999 added Python 3.14 to CI, so it seems that it is now fully supported. In this PR we add the Trove classifier (https://pypi.org/classifiers/) to signal that `celery` supports Python 3.14.


sk-:
@auvipy given that the referenced PR (#9999) is already included in the 5.6 releases (since [5.6.0rc2](https://github.com/celery/celery/blame/v5.6.0rc2/.github/workflows/integration-tests.yml#L14)), would it be possible to also include this PR in a patch release for 5.6.4?

auvipy (approved):

PR #10118, opened 2026-02-13
fix: show user-friendly broker errors for remote CLI commands

harshang03:
## Summary
- Replace raw tracebacks from broker connection failures in remote CLI commands with concise, user-friendly errors.
- Add a shared remote-command error translator and apply it to `status`, `inspect`, `control`, `graph workers`, and `events` command paths.
- Keep non-connection CLI errors concise as command-scoped failures without traceback output.

## Test plan
- [x] `.venv/bin/python -m pytest t/unit/bin/test_control.py -q`
- [x] `.venv/bin/pre-commit run --files celery/bin/base.py celery/bin/control.py celery/bin/events.py celery/bin/graph.py t/unit/bin/test_control.py`
- [x] Manual check: `.venv/bin/celery --broker amqp://guest@127.0.0.1:5672// status -t 1`

issue #9722

Issue #9722, opened 2025-05-22
celery status should show user-friendly error message instead of traceback when broker is unreachable


yiyu0x:
# Checklist

- [x] I have checked the [issues list](https://github.com/celery/celery/issues?q=is%3Aissue+label%3A%22Issue+Type%3A+Enhancement%22+-label%3A%22Category%3A+Documentation%22)
  for similar or identical enhancement to an existing feature.
- [x] I have checked the [pull requests list](https://github.com/celery/celery/pulls?q=is%3Apr+label%3A%22Issue+Type%3A+Enhancement%22+-label%3A%22Category%3A+Documentation%22)
  for existing proposed enhancements.
- [x] I have checked the [commit log](https://github.com/celery/celery/commits/main)
  to find out if the same enhancement was already implemented in the
  main branch.
- [x] I have included all related issues and possible duplicate issues in this issue
      (If there are none, check this box anyway).

#### Related Issues

- None

#### Possible Duplicates

- None

# Brief Summary

When running `celery status`, if the message broker (e.g., RabbitMQ or Redis) is unreachable, the CLI currently outputs a full Python traceback. This is not user-friendly. The command should display a clear, concise error message explaining the problem and possible solutions, rather than a raw traceback.

# Design

## Architectural Considerations

None

## Proposed Behavior

- When the broker is unreachable, `celery status` should print a user-friendly error message, such as:

```
Could not connect to the message broker. Please make sure your broker (e.g., RabbitMQ or Redis) is running and the connection settings are correct. Reason: [Errno 111] Connection refused
```

- The command should not print a full Python traceback.
- For any other unexpected errors, a concise error message should be shown, not a traceback.

## Proposed UI/UX

- CLI output should be clear and actionable for end users.
- No traceback should be shown for connection or unexpected errors; only a summary and the exception string.

## Diagrams

N/A

## Alternatives

None

PR #10190, opened 2026-03-11
Fix#9722 friendly status errors for CLI

ChickenBenny:
### Summary

Thanks @harshang03 for the awesome work on #10118! The UX improvement for CLI broker errors is fantastic and much needed.

I noticed that Codecov reported some missing coverage and @auvipy requested additional tests. Since the PR hasn't been updated in about a month, I went ahead and added the missing unit tests to cover the remaining `try...except` blocks to help get this over the finish line.

### What was added
Added unit tests to simulate `OperationalError` and unexpected exceptions for the following commands, ensuring the `handle_remote_command_error` is triggered correctly:
- [x] `celery inspect <command>` 
- [x] `celery control <command>` (broadcast)
- [x] `celery events` (covering the `_run_evcam` and `_run_evtop` paths)

---
issue https://github.com/celery/celery/issues/9722


auvipy (changes requested):
we were using https://pypi.org/project/pytest-click/ in some of the existing tests regarding click. are we using it here as well? or it is not that useful here? also do we need extra documentations here?

ChickenBenny:
Ah, good catch! I completely missed that we were already using `pytest-click`. Fix with 3c505f5

auvipy (approved):

PR #10196, opened 2026-03-14
Drop the mutable default in test_merge

tejas-ae:
Replaced mutable default argument in test_merge() of t/unit/worker/test_state.py to prevent state leaks.

Reason: Mutable defaults are shared across calls and usually turn into state leaks.

Validation: `/Users/tejasattarde/Desktop/gh-patchbot/.venv/bin/python -m py_compile t/unit/worker/test_state.py`.

Context: py-mutable-default at t/unit/worker/test_state.py:119.

### DIFF
--- a/t/unit/worker/test_state.py
+++ b/t/unit/worker/test_state.py
@@ -118,7 +118,7 @@ class test_Persistent:
 
     def test_merge(self, p, data=None):
         if data is None:
-            data = []
+            data = ['foo', 'bar', 'baz']
         state.revoked.update(data)
         p.merge()
         for item in data:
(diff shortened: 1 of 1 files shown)


auvipy (changes requested):
please check the suggestions

tejas-ae (commented):
Got it. Thanks

themavik (commented):
Replacing mutable list defaults in test_merge is the right pytest hygiene. nit: test_sync now substitutes [] for both defaults; previously those were the fixed foo/bar/baz and baz/ini/koz lists, so the default invocation no longer seeds revoked the same way unless every caller passes data explicitly.

PR #10212, opened 2026-03-19
Cache Context custom header key set to avoid per-call allocations

AsymptotaX:
## Description

`celery.app.task.Context._get_custom_headers()` currently rebuilds the set of Celery-owned request keys on every call before filtering custom headers.

This change computes that key set once and reuses it, preserving existing behavior while avoiding repeated allocations on the task request path.

This is a small optimization in a hot path, reducing per-request overhead without changing semantics.

**Env**
mac mini m4 
Python 3.13
Celery ver. 5.6.2

**Micro-benchmark**:
old (rebuild set): 0.6754s (~740k ops/sec)
new (cached set): 0.5778s (~865k ops/sec)

**~17%** improvement for this code path.

### DIFF
--- a/celery/app/task.py
+++ b/celery/app/task.py
@@ -48,6 +48,10 @@ X_DEATH_HEADERS = {
     'x-last-death-reason',
 }
 
+CONTEXT_CUSTOM_HEADER_KEYS = frozenset({
+    'lang', 'task', 'argsrepr', 'kwargsrepr', 'compression',
+})
+
 #: Here for backwards compatibility as tasks no longer use a custom meta-class.
 TaskType = type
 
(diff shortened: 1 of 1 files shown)


themavik (commented):
nit: \`Context._custom_header_keys\` is fixed at import from \`Context.__dict__\`; anything that adds new context attrs after import still won't get stripped unless you rebuild the frozenset (same class of issue as before, just no per-call set alloc).

AsymptotaX:
Fair point. The benefit here is too small to justify the compatibility risk, so I'll close this PR.

auvipy:
instead of coming with this type of PR, choose open issues related to bugs

PR #10214, opened 2026-03-23
Fix: wrap new-style errback calls in try/except to prevent silent chain halt

### LINKED ISSUE #10195: New-style errbacks in `_call_task_errbacks` are called with no exception handling — one failing errback silently halts the rest
In `celery/backends/base.py`, `_call_task_errbacks` handles two errback styles. New-style errbacks (those with `__header__` and arity > 1) are invoked directly with **no surrounding try/except**:

```python
# celery/backends/base.py, lines ~252-265
try:
    if (
        hasattr(errback.type, '__header__') and
        not isinstance(errback.type.__header__, partial) and
        arity_greater(errback.type.__header__, 1)
    ):
        errback(request, exc, traceback)   # <-- no exception handling
    else:
        old_signature.append(errback)
except NotRegistered:
    old_signature.append(errback)          # only NotRegistered is caught
```

If a new-style errback raises any exception *other than* `NotRegistered`, that exception propagates uncaught, and **subsequent errbacks in the chain are never called**.

Compare with old-style errbacks (lines ~275-288), which do wrap execution in a try/except to log and continue.

**Impact:**
- A bug in errback A silently prevents errback B from running
- The exception may bubble up to the backend internals, causing unpredictable state
- There is no log entry indicating which errback failed

**Questions:**
1. Is the missing exception handling in the new-style path an oversight, or is it intentional (i.e. new-style errbacks are trusted to be safe)?
2. Should new-style errbacks be wrapped in the same defensive try/except as old-style ones, with an error log on failure?
3. Is there a test covering the case where a new-style errback raises an unexpected exception?

`celery/backends/base.py`, lines 244–288.

r266-tech:
## Description

Fixes #10195

New-style errbacks (those with `__header__` and arity > 1) in `_call_task_errbacks` are invoked directly with no surrounding try/except. If one errback raises any exception other than `NotRegistered`, that exception propagates uncaught and **subsequent errbacks in the chain are never called**.

This wraps the new-style errback call in a try/except that:
- Catches any `Exception`
- Logs the error with `logger.exception` (including the errback name)
- Continues to the next errback in the chain

This matches the defensive behavior already present in the old-style errback path (where group execution failures don't halt the process).

## Changes

- `celery/backends/base.py`: Added try/except around `errback(request, exc, traceback)` call (3 lines of new code)
- `t/unit/backends/test_base.py`: Added test `test_new_style_errback_exception_does_not_halt_chain` that verifies a failing new-style errback doesn't prevent subsequent errbacks from running

## Testing

All existing errback-related tests pass. The new test confirms:
1. A new-style errback that raises `RuntimeError` is caught and logged
2. Subsequent errbacks in the chain still execute normally

### DIFF
--- a/celery/backends/base.py
+++ b/celery/backends/base.py
@@ -262,7 +262,13 @@ class Backend:
                         not isinstance(errback.type.__header__, partial) and
                         arity_greater(errback.type.__header__, 1)
                 ):
-                    errback(request, exc, traceback)
+                    try:
+                        errback(request, exc, traceback)
+                    except Exception:
+                        logger.exception(
+                            'Errback %r raised an exception',
+                            errback.name,
+                        )
                 else:
                     old_signature.append(errback)
             except NotRegistered:
--- a/t/unit/backends/test_base.py
+++ b/t/unit/backends/test_base.py
@@ -597,6 +597,34 @@ class test_BaseBackend_dict:
         b.mark_as_failure('id', exc, request=request)
         mock_group.assert_called_once_with(request.errbacks, app=self.app)
 
+    def test_new_style_errback_exception_does_not_halt_chain(self):
+        """Test that a new-style errback raising an exception is caught
+        and does not prevent subsequent errbacks from being called."""
+        b = BaseBackend(app=self.app)
+        b._store_result = Mock()
+
+        @self.app.task(shared=False, bind=True)
+        def failing_errback(self, request, exc, traceback):
+            raise RuntimeError('errback failed')
+
+        @self.app.task(shared=False)
+        def ok_errback(arg1):
+            ok_errback.called = True
+
+        ok_errback.called = False
+
+        request = Mock(name='request')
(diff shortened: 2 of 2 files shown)


r266-tech:
Updated and rebased on current `main`.

The review comments on the test were correct: `bind=True` made `__header__` a `functools.partial`, so the original test fell through to the old-style/group path, and `immutable=True` prevented `(request, exc, traceback)` from being merged into the call. That also explains the codecov/patch failure — the new `except` branch was never exercised.

The test now uses two non-bound errbacks with arity > 1 and mutable signatures, so it goes through the new-style branch directly. Verified locally that it fails without the fix in `celery/backends/base.py` (the `RuntimeError` propagates and halts the chain) and passes with it, covering the previously missed lines.

The underlying issue (#10195) is still present on `main`, so this fix remains applicable.

auvipy (approved):

PR #10232, opened 2026-04-03
fix: move dedup integration tests to smoke tests

aurangzaib048:
## Summary

- The integration test `t/integration/test_dedup_chain_dispatch.py` started a second worker via `start_worker()`, causing random flakiness when running all integration tests locally — two workers consuming from the same queues led to stolen tasks and missing results
- Moved the two core dedup fast-path tests to `t/smoke/tests/` which uses `pytest-celery` with Docker containers for per-class worker isolation
- Updated CI workflow: removed from integration matrix, added to smoke matrix

Fixes local dev breakage on `main` reported by @Nusnus in #10159.

## Test plan

- [x] New smoke test file syntax validated
- [x] Integration test collection verified (250 tests, no dedup references)
- [x] No dangling references to deleted file in codebase
- [x] CI workflow updated: `test_dedup_chain_dispatch.py` moved from `Integration-tests` to `Smoke-tests` matrix
- [x] Smoke tests pass in CI (requires Docker)


t/integration/test_dedup_chain_dispatch.py
```diff
@@ -1,87 +0,0 @@
...
-"""
-
-import pytest
-
-from celery import chain
-from celery.contrib.testing.worker import start_worker
-from celery.result import AsyncResult
-
-from .conftest import flaky
-from .tasks import add, identity, reject_then_succeed, store_success_then_reject
```
Nusnus, line 18: reject_then_succeed isn’t used; either add back the test that used it (as a smoke) or remove the original test task

t/integration/test_dedup_chain_dispatch.py
```diff
@@ -1,87 +0,0 @@
...
-    @flaky
-    @pytest.mark.usefixtures('dedup_worker')
-    def test_chain_completes_with_dedup_enabled(self):
-        """Smoke test: a normal chain works when dedup is on."""
-        c = chain(add.s(2, 3), add.s(5))
-        assert c().get(timeout=TIMEOUT) == 10
-
-    @flaky
-    @pytest.mark.usefixtures('dedup_worker')
-    def test_reject_requeue_completes_chain(self):
```
Nusnus, line 65: Consider porting to smoke test and keep the test task

Nusnus (changes requested):
I didn’t fix the CI yet so I need to run some manual tests. Left some comments for now.

t/integration/test_dedup_chain_dispatch.py
aurangzaib048: Added back, ported `test_reject_requeue_completes_chain` to the smoke test and kept the `reject_then_succeed` task.

t/integration/test_dedup_chain_dispatch.py
aurangzaib048: Done, ported to smoke test in the next push.

Nusnus (approved):

Nusnus:
Thank you @aurangzaib048 🙏
I really appreciate the rapid response.

I’ve dug [deeper](https://github.com/celery/celery/pull/9986/changes#r3036004884) into the original test_prefork_shutdown.py, and my [original analysis](https://github.com/celery/celery/pull/10159#issuecomment-4184053923) (fixed now) wasn’t 100% accurate; it was NOT a special case, it was also broken.

Would you like to solve that too? 🙂
I can take care of verification/review.

aurangzaib048:
@Nusnus Sure.

PR #10259, opened 2026-04-10
Add all integration tests to the CI matrix

Izzette:
## Description

Add the remaining integration tests to the integration test matrix in the "Celery" workflow.  Until recently several integration tests were not run in the CI, but most of these have been moved to smoke tests (thanks to @aurangzaib048, in https://github.com/celery/celery/pull/10239 etc.).  However, 1 integration test remains without being run in the CI.
The non-test modules in t/integration and t/smoke have been renamed to remove the `test_` prefix (`test_<name>.py` to `<name>.py`).

I have a script that we can use to check that the files in the test directories match those defined in the GitHub workflow, but I'm not sure where we should put it.  This seems like a very new use-case in the Celery repo.  I can do one of the following:

* Create a pre-commit hook using `repo: local` and `language: python`, and add this script to the root.
* Reframe this script as a unit test file.
* Add a new GitHub workflow job that runs this script.

And if I'm not going to create this script as a unit test, where do you think I should put it?  Some ideas:

* Directly at the root of `t/`
* New directory `scripts/`
* The repository root

<details>
<summary>The script to check the module list matches between the workflow matrix and the test module files found.</summary>

```python
#!/usr/bin/env python

import os
import os.path
from typing import Iterable

import yaml

# The name of the Python cache directory, where byte-compiled *.pyc files are
# stored.
PYCACHE_DIRECTORY_NAME = '__pycache__'

# The prefix used by test modules in t/integration and t/smoke.
TEST_MODULE_PREFIX = 'test_'

# The path to the "Celery" GitHub workflow
GITHUB_WORKFLOW_CELERY = ".github/workflows/python-package.yml"


def filter_test_module_filenames(filenames: Iterable[str]) -> set[str]:
    """Select test module filenames from the provided filenames.

    Arguments:
        filenames (Iterable of Strings): the filenames to filter

    Returns:
        Set: Set of filenames matching the filter
    """

    test_module_filenames: set[str] = set()
    for filename in filenames:
        # Only match `test_*.py`
        _, extension = os.path.splitext(filename)
        if extension != ".py":
            continue
        if not filename.startswith(TEST_MODULE_PREFIX):
            continue

        test_module_filenames.add(filename)

    return test_module_filenames


def find_test_module_filenames(root: str) -> set[str]:
    """Recurse through root to get all the test module filenames.

    Arguments:
        root (String): the root directory containing the test modules

    Returns:
        Set: Set of test module filenames
    """
    all_test_module_filenames: set[str] = set()
    for _, dirnames, filenames in os.walk(root):
        # Prune the __pycache__ directory from the directories to traverse.
        if PYCACHE_DIRECTORY_NAME in dirnames:
            dirnames.remove(PYCACHE_DIRECTORY_NAME)

        # Add the filtered test modules from this prefix to the full list
        all_test_module_filenames |= filter_test_module_filenames(filenames)

    return all_test_module_filenames


def assert_matches(
    left: set(str), left_name: str,
    right: set(str), right_name: str,
) -> None:
    """Assert that the filenames in left are the same as those in right.

    If the filenames in left are not the same as those in right, an
    AssertionError is raised with a message explaining the mismatch.

    Arguments:
        left (Set of Strings): the set of the filenames on the left
        left_name (String): the name for the left set
        right (Set of Strings): the set of the filenames on the right
        right_name (String): the name for the right set
    """

    not_in_right = left - right
    assert not not_in_right, (
        f"The following test modules in {left_name} are not " +
        f"present in {right_name}: " +
        " ".join([repr(filename) for filename in not_in_right])
    )
    not_in_left = right - left
    assert not not_in_left, (
        f"The following test modules in {right_name} are not " +
        f"present in {left_name}: " +
        " ".join([repr(filename) for filename in not_in_left])
    )


def check_test_filenames(gh_workflow: dict, job_name: str, root: str) -> None:
    """Check that the filenames in the workflow job match those found in root.

    Arguments:
        gh_workflow (Dictionary): the parsed GitHub workflow
        job_name (String): the name of the workflow job
        root (String): the path to the root of the test directory
    """

    workflow_matrix_filenames = set(
        gh_workflow["jobs"][job_name]["strategy"]["matrix"]["module"]
    )
    found_filenames = find_test_module_filenames(root)
    assert_matches(
        workflow_matrix_filenames,
        f"the {job_name} matrix in {GITHUB_WORKFLOW_CELERY}",
        found_filenames,
        root,
    )


def main():
    """Run the check for smoke and integration test filenames."""

    # Load the "Celery" GitHub workflow as a Python object
    with open(GITHUB_WORKFLOW_CELERY) as f:
        gh_workflow = yaml.load(f, yaml.Loader)

    # Ensure that our parsed workflow is a dictionary, it still might not
    # conform to what check_test_filenames expects, but this is the easiest
    # check we can do for coherence.
    assert isinstance(gh_workflow, dict)

    check_test_filenames(gh_workflow, "Integration-tests", "t/integration")
    check_test_filenames(gh_workflow, "Smoke-tests", "t/smoke/tests")


if __name__ == "__main__":
    main()
```
</details>


auvipy (approved):

auvipy:
if the script is under unit test, will that work appropriately?

Izzette:
@auvipy:
> if the script is under unit test, will that work appropriately?

Yes, and since [the integration tests don't run unless the unit tests pass](https://github.com/Izzette/celery/blob/bf095da6ba879c042e59fc7963e943fc773af156/.github/workflows/python-package.yml#L89): it will prevent the scenario where we useless run the integration tests when the only change is to add a new integration test file that we haven't added the the module matrix yet.  Same for the smoke tests.

The only thing I don't like about it is that as far as I can tell, all the unit tests test _Celery_, and don't do stuff like linting, they don't check the configuration of the CI, nothing like that.  It would be novel if added to the unit tests.  But it would totally work! 

Edit:
There's one thing that kind of sucks about using a unit test, we'll run the test multiple times as per this [matrix](https://github.com/Izzette/celery/blob/bf095da6ba879c042e59fc7963e943fc773af156/.github/workflows/python-package.yml#L36-L50).

auvipy:
you can come with a pre commit hook for that

Izzette:
Some smoke and integration tests are failing with:

> Reading package lists...
> E: The repository 'http://us-west-2.ec2.archive.ubuntu.com/ubuntu noble Release' no longer has a Release file.
> E: The repository 'http://us-west-2.ec2.archive.ubuntu.com/ubuntu noble-updates Release' no longer has a Release file.
> E: The repository 'http://us-west-2.ec2.archive.ubuntu.com/ubuntu noble-backports Release' no longer has a Release file.
> Error: Process completed with exit code 100.

I don't think it's related to my PR ...

auvipy:
that's a network issue

auvipy:
I did some adjustment in the CI, so lets see

Izzette:
@auvipy 
> I did some adjustment in the CI, so lets see

I have the feeling we're being rate limited by the ubuntu archives.  I'm not sure what we can practically do about that other than wait and try again later, or try a different mirror.

auvipy:
I think this a problem of the CI service providers image. a support ticket has opened to report this

auvipy:
attempt to fix the docker build failure https://github.com/celery/celery/pull/10270

PR #10266, opened 2026-04-15
Update apt-get install command in workflow

auvipy:
Added '-y --no-install-recommends' flags to apt-get install command for non-interactive installation.


auvipy:
remaining issues will be fixed in subsequent PR's

PR #10291, opened 2026-04-27
docs(tasks): warn that rate-limited tasks block the worker's prefetch

DHUKK:
## Summary

- Document that a rate-limited task occupies a prefetched slot while it waits to run, so once all of a worker's prefetched tasks are rate-limited the worker stops fetching any new messages from the broker, blocking unrelated tasks on the same worker.
- Add a short A/B example illustrating how a burst of rate-limited task A can starve unrelated task B on the same worker.
- Recommend routing rate-limited tasks to their own dedicated workers to avoid this.

This behavior regularly surprises users who expect a rate limit on one task type to only throttle that task. See [this gist](https://gist.github.com/miraculixx/e826729386aaf2b1f36b0d21591f9e71) for a worked example. The docs previously noted only that the limit is per-worker, with no mention of the knock-on effect on the prefetch loop.

## Test plan

- [x] Rendered ``docs/userguide/tasks.rst`` locally with the docker docs service and confirmed the new warning block renders correctly under the ``Task.rate_limit`` section.
 
image


DHUKK:
@auvipy any update on this? I see some amendments were made

auvipy (approved):

PR #10298, opened 2026-05-05
fix(redis): keep _reconnect_pubsub compatible with redis-py < 5.3.0

### LINKED ISSUE #10294: celery 5.6.3 regression: _reconnect_pubsub breaks with redis-py < 5.3.0
## Checklist

- [x] I have checked the [issues list](https://github.com/celery/celery/issues?q=is%3Aissue+label%3A%22Issue+Type%3A+Bug+Report%22+-label%3A%22Category%3A+Documentation%22+)
- [x] I have checked the [pull requests list](https://github.com/celery/celery/pulls?q=is%3Apr+label%3A%22PR+Type%3A+Bugfix%22+-label%3A%22Category%3A+Documentation%22+)
  for existing proposed fixes.

- **Celery version**: 5.6.3
- **kombu version**: 5.6.2
- **redis-py version**: 4.6.0
- **Python version**: 3.11
- **OS**: Linux (Debian)

## Related Issues and Possible Duplicates

- #9769 — the original deprecation warning report. PR #10036 (which fixed #9769) was milestoned for 5.7.0 but was included in the 5.6.3 release without a corresponding bump to the minimum redis-py version.
- celery/kombu#2511 — bumps kombu's redis-py minimum to >=5.3.0 to match, but is milestoned for kombu 5.7.0 and has not shipped.

## Description

Celery 5.6.3 introduced a regression that breaks the Redis result backend's pubsub reconnection path for anyone using redis-py < 5.3.0.

PR #10036 removed the positional arguments from the `get_connection()` call in `_reconnect_pubsub`:

**v5.6.2** ([redis.py#L120](https://github.com/celery/celery/blob/v5.6.2/celery/backends/redis.py#L120)):
```python
self._pubsub.connection = self._pubsub.connection_pool.get_connection(
    'pubsub', self._pubsub.shard_hint
)
```

**v5.6.3** ([redis.py#L114](https://github.com/celery/celery/blob/v5.6.3/celery/backends/redis.py#L114)):
```python
self._pubsub.connection = self._pubsub.connection_pool.get_connection()
```

In redis-py < 5.3.0, `ConnectionPool.get_connection(command_name, *keys, **options)` requires `command_name` as a positional argument ([redis-py v4.6.0 connection.py#L1449](https://github.com/redis/redis-py/blob/v4.6.0/redis/connection.py#L1449)). The no-arg call raises:

```
TypeError: ConnectionPool.get_connection() missing 1 required positional argument: 'command_name'
```

This only surfaces when a Redis pubsub connection drops (TLS keepalive timeout, idle disconnect, infrastructure cycling) and Celery attempts to reconnect. The initial connection works fine, making the bug intermittent and difficult to diagnose.

## Steps to Reproduce

1. Install `celery==5.6.3` with `redis==4.6.0` (or any redis-py < 5.3.0)
2. Configure a Redis result backend
3. Start a Celery worker and submit a task that returns a result
4. Force a Redis pubsub disconnect (e.g., kill the TCP connection, wait for idle timeout, or restart the Redis server)
5. Celery's `_reconnect_pubsub` fires and raises `TypeError`

## Expected Behavior

The pubsub connection should reconnect transparently, as it did in celery 5.6.2.

## Actual Behavior

```
TypeError: ConnectionPool.get_connection() missing 1 required positional argument: 'command_name'
```

The reconnection fails, and any pending result consumers break.

## Root Cause

PR #10036 was milestoned for 5.7.0 (where presumably redis-py >= 5.3.0 would be required), but it was included in the 5.6.3 release. Celery 5.6.x does not require redis-py >= 5.3.0, so the `get_connection()` call without arguments is a breaking change for existing users on redis-py 4.x.

This is a semver violation — a patch release (5.6.2 → 5.6.3) broke a public API contract.

## Suggested Fix

Either:

1. **Patch release (5.6.4)**: Revert the `get_connection()` change to restore the positional args, preserving compatibility with redis-py 4.x. Or conditionally pass the args based on the redis-py version.
2. **Minimum version bump**: Add `redis>=5.3.0` to celery's dependencies in the 5.6.x line (though this is itself a breaking change for a patch release).

Option 1 seems like the correct path for a patch release.

SAY-5:

## Description

Fixes #10294.

PR #10036 (shipped in 5.6.3) removed the positional `command_name`
argument from the `ConnectionPool.get_connection` call inside
`ResultConsumer._reconnect_pubsub`. On redis-py < 5.3.0 that argument
is still required, so the no-subscriptions reconnect branch raises:

```
TypeError: ConnectionPool.get_connection() missing 1 required positional argument: 'command_name'
```

celery 5.6.x does not pin `redis>=5.3.0`, so this surfaces whenever a
user on redis-py 4.x experiences a pubsub disconnect.

The fix attempts the modern no-arg call first and falls back to the
legacy `'pubsub'` positional argument on `TypeError`. Both redis-py
ranges accept this path (5.3.0+ ignores the extra arg).

A regression test in `t/unit/backends/test_redis.py` mocks the legacy
signature and asserts `_reconnect_pubsub` does not raise.


auvipy (changes requested):
do we need this in the main branch? or we can directly port it to the v5.6.x branch? thanks for handling it anyway

auvipy (approved):

PR #10307, opened 2026-05-08
fix: add explicit encoding to open() calls for cross-platform portability

KeloYuan:
Python open() uses locale-dependent encoding by default. On systems with non-UTF-8 locale (e.g. Windows cp1252), reading PID files, certificates, and auth keys can fail or produce corrupted data.

Fixed 3 files in celery/production code:
- platforms.py: PID file read/write (2 calls)
- security/certificate.py: PEM certificate loading (1 call)
- security/__init__.py: SSL key and cert loading (2 calls)

### DIFF
--- a/celery/platforms.py
+++ b/celery/platforms.py
@@ -163,7 +163,7 @@ class Pidfile:
     def read_pid(self):
         """Read and return the current pid."""
         with ignore_errno('ENOENT'):
-            with open(self.path) as fh:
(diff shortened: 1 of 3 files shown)


auvipy (changes requested):
the changes seems to be useful. should we also update or improve relevant tests for the relevant changes as well?

auvipy:
the tests are failing

PR #10315, opened 2026-05-16
handle beat_cron_starting_deadline correctly for non-uniform crontabs

phanky1:
## Description

`crontab.is_due()` can incorrectly return `is_due=False` when `beat_cron_starting_deadline` is set and the crontab has non-uniform intervals.

For `crontab(minute='0,45')`, with `last_run_at = 08:45` and `now = 10:20`, the latest missed run is `10:00`. Because that run was only 20 minutes ago, a 30-minute deadline should still return `due = True`.

Instead, the current logic can skip over that run and use an older missed run, incorrectly returning `due=False`.

**Example**

```py
from datetime import datetime
from celery import Celery

app = Celery()
app.conf.beat_cron_starting_deadline = 1800  # 30 minutes

# Runs at :00 and :45.
c = crontab(minute='0,45', app=app)

# current time is 2022-12-05 10:20.
c.nowfun = lambda: datetime(2022, 12, 5, 10, 20)

# Last successful run was at 08:45.
last_run_at = datetime(2022, 12, 5, 8, 45)

# Missed runs since then: 09:00, 09:45, 10:00.
# Since 10:00 was only 20 minutes ago, this should still be due.
print(c.is_due(last_run_at))
```
Current behaviour: `schedstate(is_due=False, next=1500.0)`
Expected behaviour: `schedstate(is_due=True, next=1500.0)`


auvipy (changes requested):
please check the test failures, also please consider the possibilities for smoke or integration tests

phanky1:
> please check the test failures, also please consider the possibilities for smoke or integration tests

The test failures should be fixed by the latest PR

auvipy (changes requested):
can you point to any relevant issue to this fix?

phanky1:
> can you point to any relevant issue to this fix?

No open issue for this. I found it myself while learning the codebase

phanky1:
Hi @auvipy, I added integration tests and just want to check in to see if there are any other changes you would like to make.

auvipy (changes requested):
can you please point me to the integrations tests?

phanky1:
> can you please point me to the integrations tests?

I misjudged the new tests I added in `t/unit/app/test_schedules.py` as sufficient to be considered integration tests until I realized there's a full integration suite. 

I added an appropriate integration test in this commit https://github.com/celery/celery/pull/10315/changes/09b4a228d267fe86b9ccebe5ff31ac370bc35012 please have a look.

auvipy:
is this going to be a breaking change? also can you explore the opportunity to improve the test coverage coverage better documentation to navigate the change?

phanky1:
> is this going to be a breaking change? also can you explore the opportunity to improve the test coverage coverage better documentation to navigate the change?

No, this is not a breaking change, just a bug fix and I also improved the documentation. 

The coverage decrease is because more code removed than added, the added ones are all covered.

auvipy:
@alirafiei75 if you have time please have a brief look into it

alirafiei75 (approved):

alirafiei75:
> @alirafiei75 if you have time please have a brief look into it

@auvipy 
I did and Looks good to me!

The bug in the old while loop logic is clear. It breaks on non-uniform schedules and can cause infinite loops when rem_delta hits 0.
Replacing loop with the O(1) calculation using deadline_since fixes the stale date lookup, cleans up the performance, and handles microsecond boundary edge cases properly.

Tests cover the edge cases well too.

auvipy (approved):

PR #10347, opened 2026-06-08
chore: improve celery maintenance path

lphuc2250gma:
Summary:
- Add or tighten focused edge-case tests or type assertions in t/unit/test_loops.py, t/unit/test_canvas.py related to Python typing, tests, CLI ergonomics, observability; avoid docs-only changes and broad refactors.
- Keep the change narrow so it is straightforward to review.

Notes:
- I kept this scoped to the relevant implementation and tests.


uv.lock
auvipy, line 1: you can remove this file from this pr

auvipy (changes requested):

PR #10358, opened 2026-06-16
Fix queue selection by alias to store the real queue name

phanky1:
## Description

When a worker is started with -Q using a queue alias, Queues.select stores the selected queue under the alias instead of the resolved queue name.
https://github.com/celery/celery/blob/49813637beedd826f1faa3768973341dbff77995/celery/app/amqp.py#L172-L181

Later, if deselect is called with the real queue name, it will not find the queue so it won't remove the queue from `_consume_from`.

So if a RecoverableConnectionError is raised and triggers a broker reconnect, the worker rebuilds its consumers from the stale _consume_from state, causing it to resume consuming from the previously cancelled queue.

This changes Queues.select() to store selected queues under their resolved queue name.

### Steps to reproduce:

1. Define a queue with an alias:
```
from celery import Celery 
from kombu import Queue
app = Celery('root') 
app.conf.task_queues = [Queue('real_name', alias='alias')]
```
2. Start the worker using the alias:
```
celery -A root worker --pool=threads -Q alias
```
3. Check the active queues:
```
celery -A root inspect active_queues
```
The queue is reported by its real name.

4. Cancel the consumer using that real name:
```
celery -A root control cancel_consumer real_name
```
5. Restart Redis to force a reconnection from worker.
6. After reconnect, the worker starts consuming `real_name` queue again.


auvipy (commented):
is there any old related issue for this?

phanky1:
> is there any old related issue for this?

I couldn't find any related issue to this.

auvipy (approved):

auvipy (commented):
should we also consider a integration test to harden this more?

phanky1:
> should we also consider a integration test to harden this more?

I just added an integration test for this.

auvipy (approved):

PR #10359, opened 2026-06-17
fix(utils/text): off-by-one in truncate() when len(s) == maxlen

JSap0914:
## Bug

`truncate()` in `celery/utils/text.py` uses `>=` to compare `len(s)` with `maxlen`:

```python
if maxlen and len(s) >= maxlen:
    return s[:maxlen].rsplit(' ', 1)[0] + suffix
```

When the string is **exactly** `maxlen` characters long, it doesn't need to be truncated — it already fits within the limit. However, the current `>=` condition truncates it anyway and appends the suffix, producing a string **longer** than `maxlen`.

## Example

```python
>>> truncate('ABCDEFGHI', 9)  # len == maxlen == 9
'ABCDEFGHI...'  # 12 chars — exceeds the stated limit
```

## Fix

Change `>=` to `>` so only strings strictly longer than `maxlen` are truncated:

```python
>>> truncate('ABCDEFGHI', 9)
'ABCDEFGHI'   # 9 chars — correct, no truncation needed
```

## Verification

```
pytest t/unit/utils/test_text.py -v
# 15 passed
```

> Note: this fix was developed with AI assistance.

### DIFF
--- a/celery/utils/text.py
+++ b/celery/utils/text.py
@@ -88,7 +88,7 @@ def indent(t: str, indent: int = 0, sep: str = '\n') -> str:
 
 def truncate(s: str, maxlen: int = 128, suffix: str = '...') -> str:
     """Truncate text to a maximum number of characters."""
-    if maxlen and len(s) >= maxlen:
+    if maxlen and len(s) > maxlen:
         return s[:maxlen].rsplit(' ', 1)[0] + suffix
     return s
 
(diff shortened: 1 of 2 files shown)


AhmadAL-Quraan:
What if I have something like this as well:

`truncate("ABCDEFGHIJK", 9)`

The answer gonna be `ABCDEFGHI...` , the length is `12` which also exceeds the given number `9` .

I think either:

- The docstring is inaccurate and maxlen refers to the truncation point before appending the suffix. or

- The implementation does not enforce the documented maximum length.

Either way, this PR is a little bet of a rush, and you should had open an issue discussing this first.

PR #10361, opened 2026-06-18
Fix autoscale and disable-prefetch for green worker pools

### LINKED ISSUE #4003: Celery autoscale not working
I read from the celery documentation that we can autoscale concurrency by using autoscale option while starting a worker.

I have added 200 task in my queue and each task takes 5seconds to finish. I started worker with autoscale of 3 min concurrency and 15 max concurrency , but its starting with only 3 processes and I am not seeing any scale up in the number of processes.

Using celery with rabbitmq.

arkadiuszbach:
## Summary

Fixes green worker pool capacity tracking for autoscale.

- Fixes #4003 for gevent and eventlet worker pools; prefork was already working correctly.
- Fixes gevent `num_processes` to report configured pool capacity instead of active greenlets.
- Fixes `worker_disable_prefetch` behavior for gevent workers, since it relies on `pool.num_processes`.
- Keeps green-pool capacity in sync with Celery's pool `limit` during autoscale grow/shrink.
- Wakes blocked `spawn()` waiters when gevent/eventlet pools grow, instead of only mutating semaphore counters.

## Why

The autoscaler uses `pool.num_processes` as the current pool capacity. For gevent, this previously returned `len(self._pool)`, which is the number of tracked/running greenlets rather than configured capacity. That caused autoscale decisions to be based on active work instead of the pool size.

This resolves the long-standing autoscale failure reported in #4003 for gevent and eventlet workers. Prefork already scales correctly, but the green pools had stale/incorrect capacity accounting and did not wake blocked task submissions when autoscale increased the pool size.

`worker_disable_prefetch` also relies on `pool.num_processes` when deciding whether the worker can consume more tasks. With gevent, the incorrect active-greenlet count could cause disable-prefetch to make the wrong consume/stop decision.

Both gevent and eventlet pool growth adjusted semaphore counters directly. That updates capacity numerically, but it does not notify tasks already blocked waiting for a pool slot. The fix releases the relevant semaphore slots during grow so blocked submissions can continue immediately.


auvipy (changes requested):
can you please check if the test coverage can be improved?

auvipy (changes requested):
should we also consider smoke tests or integration tests for this, beside inproving test coverage?

auvipy (approved):

PR #10369, opened 2026-06-20
Mark rejected tasks as failed when not requeued

### LINKED ISSUE #4222: Reject exception doesn't cause the task to fail
Hi!

I'm seeing the behaviour described in #2944.
If I reject a task, it is kept as __STARTED__. I'm not sure if did not understand the solution to that issue - and this is the expected behaviuor - or if this is a _regression_.

I have verified that the behaviour is the same in the `master` branch of Celery (installed from commit bda678b9cd3a4ea47bfdd4c33aaabffb678de883).

## Steps to reproduce
```python
from celery import task, Task
from celery.exceptions import Reject

@task(name='test_reject')
def test_reject():
    raise Reject(reason="Some business logic...", requeue=False)
```

```bash
>>> from mymodule.tasks import *
>>> test_reject.delay()
<AsyncResult: 712bed3e-39e1-4455-b1f6-445aa37ef236>
```

## Expected behavior
I would expect the task to go into FAILURE state

## Actual behavior
Task still _pending_

```bash
>>> from celery.result import AsyncResult
>>> res = AsyncResult('712bed3e-39e1-4455-b1f6-445aa37ef236')
>>> res.status
'PENDING'
>>> res.state
'PENDING'
>>> res.failed()
False
>>> res.successful()
False
```

captura de tela de 2017-08-22 11-06-21

hmnfalahi:
## Description

Fixes #4222.

When a task raises `Reject(requeue=False)`, the message is rejected and will not be executed again. Previously the result backend was not updated, which could leave the task in `PENDING` or `STARTED`.

This change records a terminal failure result for non-requeued rejects and emits the `task_failure` signal. Requeued rejects keep the existing behavior.

Added unit coverage for both non-requeued and requeued `Reject` paths.

### DIFF
--- a/celery/worker/request.py
+++ b/celery/worker/request.py
@@ -647,6 +647,27 @@ class Request:
         elif isinstance(exc, MemoryError):
             raise MemoryError(f'Process got: {exc}')
         elif isinstance(exc, Reject):
+            if not exc.requeue:
+                # A task that rejects its message without requeueing will
+                # never run again, so record a terminal FAILURE result and
+                # fire the task_failure signal instead of leaving the task
+                # stuck in PENDING/STARTED forever (Issue #4222).
+                self.task.backend.mark_as_failure(
+                    self.id, exc, request=self._context,
+                    store_result=self.store_errors,
+                )
+                signals.task_failure.send(
+                    sender=self.task, task_id=self.id, exception=exc,
+                    args=self.args, kwargs=self.kwargs,
+                    traceback=exc_info.traceback, einfo=exc_info,
+                )
+                if send_failed_event:
+                    self.send_event(
+                        'task-failed',
+                        exception=safe_repr(
+                            get_pickled_exception(exc_info.exception)),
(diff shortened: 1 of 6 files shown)


celery/worker/request.py
```diff
@@ -647,6 +647,27 @@ def on_failure(self, exc_info, send_failed_event=True, return_ok=False):
         elif isinstance(exc, MemoryError):
             raise MemoryError(f'Process got: {exc}')
         elif isinstance(exc, Reject):
+            if not exc.requeue:
+                # A task that rejects its message without requeueing will
+                # never run again, so record a terminal FAILURE result and
+                # fire the task_failure signal instead of leaving the task
+                # stuck in PENDING/STARTED forever (Issue #4222).
+                self.task.backend.mark_as_failure(
```
auvipy, line 655: will this be a good idea to add integration or smoke test as appropriate, to figure out this fix is working as expected and not breaking anything?

auvipy (approved):

PR #10381, opened 2026-06-26
Strip leading self from class instance __call__ in head_from_fun, and don't crash on functools.partial

HrachShah:
## Summary

`head_from_fun` (used to generate argument-validation stubs for Celery tasks) had two distinct failure modes for non-function callables:

1. **`functools.partial` raised `AttributeError: 'method-wrapper' object has no attribute '__module__'`.** Partial's `__call__` is a C-level descriptor, but the old code unconditionally swapped `fun = fun.__call__` and then read `fun.__module__` to populate the `exec()` namespace.

2. **Class instances with a regular Python `__call__` got the wrong signature.** Swapping to `inst.__call__` keeps `self` as the first positional arg, so a class like
   ```python
   class MyTask:
       def __call__(self, foo, bar, baz=10): ...
   ```
   produced a stub `def MyTask(self, foo, bar, baz=0): ...` — the instance is already bound, so the stub rejected valid calls and accepted invalid ones that mapped `self` to the user's first argument.

The fix only swaps to `__call__` when it's a real Python function (has a non-None `__module__`), and strips the leading `self` from the resulting argspec. For everything else (`functools.partial`, class objects where `type.__call__` is the C-level wrapper), it falls back to introspecting `fun` directly — `inspect.getfullargspec` already unwraps partials correctly into the effective signature.

## Tests

Five new tests in `t/unit/utils/test_functional.py::test_head_from_fun`:

- `test_partial_keyword_bound` — `partial(f, a=1)` exposes `b, c` as keyword-only with no default
- `test_partial_positional_bound` — `partial(f, 10)` keeps `b, c` positional
- `test_callable_class_instance` — class instance with `__call__(self, foo, bar, baz=10)` rejects too few/too many positional args and accepts kwargs
- `test_callable_class_instance_bound` — `bound=True` with a class instance still validates the rest of the signature after the sentinel replaces `self`
- `test_callable_class_object` — `head_from_fun(SomeClass)` doesn't crash on the method-wrapper introspection path

All five fail on `main` with either `AttributeError: 'method-wrapper' object has no attribute '__module__'` or `TypeError: MyTask() takes 0 positional arguments but 2 were given`. All pass with the fix. The existing 13 `test_head_from_fun` tests still pass.

## Verification

```
$ python3 -m pytest t/unit/utils/test_functional.py::test_head_from_fun -v
======================== 13 passed, 1 skipped in 0.29s ========================

$ python3 -m pytest t/unit/utils/ --deselect t/unit/utils/test_functional.py::test_regen::test_lookahead_consume
525 passed, 2 skipped, 1 xfailed, 5 pre-existing failures (test_platforms privilege tests, unrelated)
```


auvipy:
I just re run the failing pypy build

celery/utils/functional.py
```diff
@@ -374,12 +374,52 @@ def head_from_fun(fun: Callable[..., Any], bound: bool = False) -> str:
...
+        # method-wrapper (e.g. functools.partial), __module__ is None,
+        # so we fall back to introspecting `fun` directly, which
+        # inspect.getfullargspec already unwraps correctly.
+        if (
+                call_attr is not None
+                and getattr(call_attr, '__module__', None) is not None
+        ):
+            fun = call_attr
+            spec = _getfullargspec(fun)
+            if spec.args and spec.args[0] == 'self':
```
Sanjays2402, line 408: Python doesn't require the receiver to be named `self`. For `def __call__(this, value)`, `getfullargspec` returns `['this', 'value']` even though the bound callable's effective signature is just `(value)`, so this still makes the checker demand an extra argument; keying the strip on `inspect.ismethod(call_attr)` would cover receiver names other than `self`.

PR #10395, opened 2026-07-03
Remove redundant maybe_signature call in unlock_chord

phanky1:
## Description

`unlock_chord` calls `callback = maybe_signature(callback, app)` twice, but callback is not used between the calls. 
This PR removes that redundant call.

https://github.com/celery/celery/blob/2e150f8330df059858ff5d481390a29b73c59600/celery/app/builtins.py#L57-L75


auvipy (changes requested):
it would be better to have some form of tests to avoid regression

phanky1:
> it would be better to have some form of tests to avoid regression

I added a new test.

auvipy (approved):

Issue #9569, opened 2025-02-21
Clerey inspect report leaking beat_dburi password


erodriguezv:
Celery Inspect report obscures some passwords from the worker report (I.E. broker url, result_backend) but it is not obscuring passwords from "beat_dburi".
As a result this shows in Flower UI under the config section.

![Image](https://github.com/user-attachments/assets/1c483e67-6d4b-457e-9621-4b54caede7d2)

![Image](https://github.com/user-attachments/assets/410a0377-9936-48c6-99c7-908860e66bd3)

# Proposed solutions
Add "beat_dburi" here https://github.com/celery/celery/blob/main/celery/app/utils.py#L338
` elif 'backend' in key.lower() or 'beat_dburi' in key.lower():`
or
Add '|DBURI' or '|BEAT_DBURI' here https://github.com/celery/celery/blob/main/celery/app/utils.py#L38
`HIDDEN_SETTINGS = re.compile(
    'API|TOKEN|KEY|SECRET|PASS|PROFANITIES_LIST|SIGNATURE|DATABASE|DBURI',
    re.IGNORECASE,
)`

# Checklist

- [x] I have verified that the issue exists against the `main` branch of Celery.
- [ ] This has already been asked to the [discussions forum](https://github.com/celery/celery/discussions) first.
- [x] I have read the relevant section in the
  [contribution guide](https://docs.celeryq.dev/en/main/contributing.html#other-bugs)
  on reporting bugs.
- [x] I have checked the [issues list](https://github.com/celery/celery/issues?q=is%3Aissue+label%3A%22Issue+Type%3A+Bug+Report%22+-label%3A%22Category%3A+Documentation%22)
  for similar or identical bug reports.
- [x] I have checked the [pull requests list](https://github.com/celery/celery/pulls?q=is%3Apr+label%3A%22PR+Type%3A+Bugfix%22+-label%3A%22Category%3A+Documentation%22)
  for existing proposed fixes.
- [x] I have checked the [commit log](https://github.com/celery/celery/commits/main)
  to find out if the bug was already fixed in the main branch.
- [x] I have included all related issues and possible duplicate issues
  in this issue (If there are none, check this box anyway).
- [ ] I have tried to reproduce the issue with [pytest-celery](https://docs.celeryq.dev/projects/pytest-celery/en/latest/userguide/celery-bug-report.html) and added the reproduction script below.

## Mandatory Debugging Information

- [x] I have included the output of ``celery -A proj report`` in the issue.
[...]

PR #10398, opened 2026-07-06
Mask broker_read_url and broker_write_url in filter_hidden_settings

phanky1:
## Description

`filter_hidden_settings` masked `broker_url` but not its read/write counterparts. This PR censors `broker_read_url` and `broker_write_url` as well.


auvipy (commented):
any open relevant issues for this?

phanky1:
> any open relevant issues for this?

this experiences the same issue as https://github.com/celery/celery/issues/9569.

eeshsaxena (commented):
Confirmed the underlying leak on current `main` - this is a real one. `'broker_url' in key.lower()` is false for `broker_read_url` and `broker_write_url` (the substring is not contiguous in either), so both fall through to the `'backend' in key` branch, miss that too, and get returned verbatim:

```python
app.conf.update(
    broker_url='amqp://u:secretpw@h:5672//',
    broker_read_url='amqp://u:secretpw@h:5672//',
    broker_write_url='amqp://u:secretpw@h:5672//',
)
app.conf.table(censored=True)
#   broker_url        amqp://u:********@h:5672//     (masked)
#   broker_read_url   amqp://u:secretpw@h:5672//     (leaked)
#   broker_write_url  amqp://u:secretpw@h:5672//     (leaked)
```

Both are real settings (`broker.read_url` / `broker.write_url` in `defaults.py`), so credentials can end up in `celery report` output. Good catch.

One concern with the fix as written: swapping the substring test for exact membership **narrows** what gets masked, and this is a masking function, so narrowing is the risky direction. Any key that merely *contains* `broker_url` is masked today and would start leaking. For example a custom or Django-style key that survives `conf.update()`:

```python
app.conf.update(CELERY_BROKER_URL='amqp://u:secretpw@h:5672//')
# main:      amqp://u:********@h:5672//     (masked)
# this PR:   amqp://u:secretpw@h:5672//     (leaked)
```

A union of the two predicates keeps the existing coverage and adds the two new keys, so it is a strict superset of today's behaviour:

```python
elif ('broker_url' in key.lower()
        or key.lower() in ('broker_read_url', 'broker_write_url')):
```

I checked this against `broker_url`, `BROKER_URL`, `broker_read_url`, `BROKER_READ_URL`, `broker_write_url`, `CELERY_BROKER_URL`, `result_backend` and `broker_host` - it masks everything `main` masks, plus the two leaking keys, and does not newly capture `result_backend` / `broker_host`.

Minor, and pre-existing rather than introduced here: `Connection(value).as_uri(mask=...)` fabricates a default when `value` is `None`, and `broker_read_url` / `broker_write_url` both default to `None`. So under `app.conf.table(with_defaults=True)` the new branch will render an unset read URL as `amqp://guest:********@localhost:5672//`, which reads as "configured" when it is not:

```python
Connection(None).as_uri(mask='*'*8)   # 'amqp://guest:********@localhost:5672//'
```

`broker_url` already behaves this way on `main`, so the PR is at least self-consistent - but since it triples the number of keys hitting that branch, a `value and ...` guard might be worth folding in while you are here. Entirely your call, and I will defer to the maintainers.

The added test is well targeted. It would be a little stronger asserting the mask is actually present (`'********' in censored[k]`) rather than only that `'pass'` is absent, since a key that got dropped entirely would also pass the current assertion.

auvipy (approved):

PR #10413, opened 2026-07-15
docs: README lists Python 3.9 but setup.py requires >=3.10

kobihikri:
Hi, and thank you for Celery.

`README.rst` (line 125, under 'Celery version 5.6.x runs on:') lists:
```
- Python (3.9, 3.10, 3.11, 3.12, 3.13)
```
but `setup.py` declares `python_requires=">=3.10"` (and the trove classifiers start at 3.10), so 3.9 isn't actually supported. This drops 3.9 from the README list to match.

For transparency: I used AI assistance to spot and draft this; I verified README.rst and setup.py against the current source myself.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -122,7 +122,7 @@ What do I need?
 
 Celery version 5.6.x runs on:
 
-- Python (3.9, 3.10, 3.11, 3.12, 3.13)
+- Python (3.10, 3.11, 3.12, 3.13)
 - PyPy3.9+ (v7.3.12+)
 
 This is the last version of Celery which will support Python 3.9.


auvipy (changes requested):
the readme is right for the current release. will update during new release

kobihikri:
You're right, @auvipy — thanks for the correction. Closing this.

I checked before taking your word for it, and the numbers back you up: the published 5.6.3 on PyPI carries `requires_python: ">=3.9"`, while `setup.py` on `main` is already at `>=3.10`. So the README is describing the current release accurately, and it's `main` that has moved ahead of it — exactly as you said.

My mistake was comparing a manifest on `main` against a README that documents the released version, and reading the gap as drift rather than as a change that hasn't shipped yet. That's a useful lesson and I'll apply it elsewhere.

Sorry for the noise, and thanks for taking the time to reply.

PR #10417, opened 2026-07-17
Preserve group-buffered events when flush republish fails

### LINKED ISSUE #6487: Worker events missing when it be published failed twice
Hi

There is a strange problem in my project which using celery v4.3 and I don't know that is a bug or not. 

I have an multi worker cluster which use rabbitmq as its brokers. An mini sample of my project just like:

1. Creating the worker with configuration including `task_events = True`
2. Creating an receiver instance which is a subclass of `celery.events.receiver.EventReceiver`.

And sometimes i found the receiver cannot receive any worker event of few worker , for example `task-started`, `task-succeded`.

I have check the environment and these are true:
1. The worker which lost event can consumer task for rabbitmq queue and can run task succeessful.
2. The receiver can receive events which are from other workers.

I doubt there are some network issue occurs and I found the flush function of `celery.events.dispatcher.EventDispatcher` may contains an bug:

[dispatcher.py#L208](https://github.com/celery/celery/blob/master/celery/events/dispatcher.py#L208)

```
def flush(self, errors=True, groups=True):
        """Flush the outbound buffer."""
        if errors:
            buf = list(self._outbound_buffer)
            try:
                with self.mutex:
                    for event, routing_key, _ in buf:
                        self._publish(event, self.producer, routing_key)
            finally:
                self._outbound_buffer.clear()
```

When network problem occurs during `flush` is called first time, `self._publish()` failed to send an event A and the event A  is appended to `self._outbound_buffer`
[dispatcher.py#L160](https://github.com/celery/celery/blob/master/celery/events/dispatcher.py#L160)
```
def _publish(self, event, producer, routing_key, retry=False,
                 retry_policy=None, utcoffset=utcoffset):
        exchange = self.exchange
        try:
            producer.publish(
               ...
            if not self.buffer_while_offline:
                raise
            self._outbound_buffer.append((event, routing_key, exc))
``` 

The `flush` is called second time then `self._publish()` is still failed to send event A, with the event A appended to  `self._outbound_buffer` and then `self._outbound_buffer.clear()` is called. 

So the event A is lost for `clear` function is called. I think what I describe above is a bug. If this is a bug, i will to create merge request to fix it.

ps: code of this funtion in v4.3 is same with the master branch's.

chuenchen309:
## Description

`EventDispatcher.flush()` loses every group-buffered `task-*` event if the broker is down at flush time — the events are re-buffered for retry, but the batch they live in is emptied out from under them.

This is the `groups` half of #10401. That PR fixed the identical re-buffer-then-clear loss in the `errors` branch a couple of weeks ago (Fixes #6487); the `groups` branch a few lines below has the same defect in a different shape, via aliasing rather than a `finally`.

### The bug

```python
if groups:
    with self.mutex:
        for group, events in self._group_buffer.items():
            self._publish(events, self.producer, '%s.multi' % group)
            events[:] = []  # list.clear
```

`_publish` re-buffers on failure with `self._outbound_buffer.append((event, routing_key))` — and `event` **is** `events`, the same list object. The next line clears it in place. The buffer entry survives; its payload does not.

### Reproducer

Three `task-succeeded` events buffered, broker fails during `flush()`, then recovers:

```
CONTROL: broker healthy throughout
  group buffer before flush : 3
  events delivered          : 3

BROKEN: broker down during flush, then recovers
  group buffer before flush : 3
  after failed flush, _outbound_buffer entries: 1
     re-buffered entry -> routing_key='task.multi' payload=[]
  events delivered after recovery: 0
  producer received payloads     : [[], []]
```

The control uses the same fixture and the same code path, so this isn't a broken reproducer — it's the failure path specifically. Note the two junk empty batches that go out as well.

### Reachability

Not a theoretical path — it's the default worker:

- `worker/consumer/events.py:45` sets `buffer_group=['task'] if c.hub else None`, so any worker with an event loop and events enabled (`-E`) group-buffers task events. That's the standard prefork worker, i.e. what Flower and every monitoring setup sees.
- `buffer_while_offline` defaults to `True` and the worker doesn't override it, so `_publish` does take the re-buffer path.
- `consumer.py:624-630` schedules `_flush_events` → `flush()` on the hub, and `worker/consumer/events.py:50` calls `dis.flush()` on reconnect — the comment there literally reads *"flush events sent while connection was down"*.

So this is `buffer_while_offline` failing at exactly the moment it exists for.

### Fix

Publish a detached copy, and skip groups with nothing buffered. The in-place clear is kept — `test_send_buffer_group` asserts the buffer keeps its identity (`is prev_buffer`), and `send()` holds a reference to it.

The empty-group skip is a small side benefit rather than a separate concern: `_group_buffer` is a `defaultdict`, so once the `task` key exists every later `flush()` published an empty `task.multi` to the broker.

### One thing worth flagging: I changed an assertion in an existing test

`test_send_buffer_group` asserted:

```python
eventer._publish.assert_has_calls([
    call([], eventer.producer, 'task.multi'),
])
```

That `[]` is an artifact of the bug, not a specification. `_publish` is a `Mock` there, and Mock stores a *reference* to the list it was called with — by assertion time, `events[:] = []` had emptied it. The same test already asserts, two lines later, that 2 events were really passed (`len(buf_received[0]) == 2`, snapshotted in the `side_effect`). So it currently asserts both "`[]` was passed" and "2 events were passed" about the same call, and only passes because of the aliasing this PR removes.

I changed it to `call(buf_received[0], ...)` so it asserts the real payload. Flagging it up front because "PR changes a test that was passing" deserves a look rather than a shrug.

### Verification

- New test `test_flush_rebuffers_group_events_when_publish_fails`, named and shaped after #10401's `test_flush_rebuffers_events_when_republish_fails`. Red before the fix (`assert len(batch) == 1` → `assert 0 == 1`), green after.
- No existing coverage for this: `test_send_buffer_group` mocks `_publish` outright, so the real failure path was never exercised, and #10401's test covers only the `errors` branch.
- `t/unit/events/ t/unit/worker/`: **713 passed, 2 skipped** with the fix vs. **712 passed, 2 skipped** on `main` — the one extra is the new test, and nothing regressed.
- `flake8` clean.

### AI disclosure

AI-assisted, and I'd rather say so plainly. The reproducer, the control, the reachability trace through `worker/consumer/`, the `buffer_while_offline` default check, and the `main`-vs-branch baseline comparison are all things I ran and read myself rather than assumed. Happy to adjust the scope or the framing.

### DIFF
--- a/celery/events/dispatcher.py
+++ b/celery/events/dispatcher.py
@@ -209,7 +209,14 @@ class EventDispatcher:
         if groups:
             with self.mutex:
                 for group, events in self._group_buffer.items():
-                    self._publish(events, self.producer, '%s.multi' % group)
+                    if not events:
+                        continue
+                    # Publish a detached copy, then clear: when buffering
+                    # offline _publish re-buffers the object it was handed, so
+                    # clearing the live list first would empty it; when it
+                    # raises instead, skipping the clear keeps the events.
+                    batch = list(events)
+                    self._publish(batch, self.producer, '%s.multi' % group)
                     events[:] = []  # list.clear
 
     def extend_buffer(self, other):
--- a/t/unit/events/test_events.py
+++ b/t/unit/events/test_events.py
@@ -119,7 +119,7 @@ class test_EventDispatcher:
         eventer.send('task-received', uuid=1)
         assert not eventer._group_buffer['task']
         eventer._publish.assert_has_calls([
-            call([], eventer.producer, 'task.multi'),
+            call(buf_received[0], eventer.producer, 'task.multi'),
         ])
         # clear in place
         assert eventer._group_buffer['task'] is prev_buffer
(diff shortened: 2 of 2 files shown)


auvipy:
@phanky1 a possible partial regression report or a half fix. please cross check

celery/events/dispatcher.py
```diff
@@ -209,8 +209,13 @@ def flush(self, errors=True, groups=True):
...
             with self.mutex:
                 for group, events in self._group_buffer.items():
-                    self._publish(events, self.producer, '%s.multi' % group)
+                    if not events:
+                        continue
+                    # Detached copy: on failure _publish re-buffers the very
+                    # object it was handed, which the clear below would empty.
+                    batch = list(events)
                     events[:] = []  # list.clear
+                    self._publish(batch, self.producer, '%s.multi' % group)
```
phanky1: I would move the clear down so the events survive if self._publish raises any exception.
```suggestion
                    self._publish(batch, self.producer, '%s.multi' % group)
                    events[:] = []  # list.clear
```

phanky1 (commented):
The finding is correct, I think.

celery/events/dispatcher.py
chuenchen309: You're right, and thanks for catching it — I've reordered to publish before the clear (kept as a suggestion-friendly diff).

I reproduced both failure paths to be sure, and @auvipy your instinct was correct — my first version was a partial regression in one of them:

- `buffer_while_offline=True`: `_publish` catches and re-buffers the object it was handed. Clearing the live list first would empty that re-buffered batch — this is the original loss the PR set out to fix, and the detached copy handles it.
- `buffer_while_offline=False`: `_publish` re-raises. Here my earlier "clear then publish" order had already emptied the buffer before the raise, so the events were lost for the next flush too. With publish-before-clear the clear is skipped on the exception and the events stay in `_group_buffer` for the retry.

I added `test_flush_keeps_group_events_when_publish_raises` to lock in that second case (it fails on the clear-first order with `_group_buffer['task']` going to 0, passes now). Full `t/unit/events/test_events.py` is green (38 passed).

Disclosure: this contribution is AI-authored (Claude Code, on this account) — the analysis, the reordered fix, and the repro above were produced and run by the agent, not hand-reviewed by a person. The verification is real and re-runnable from the diff.

auvipy (approved):

PR #10421, opened 2026-07-19
Fix stamping: visitor value dropped for a scalar stamp when append_stamps=False

chuenchen309:
`_merge_dictionaries(..., aggregate_duplicates=False)` is documented to take duplicate keys from `d2`, and `_stamp_headers` relies on that so a stamping visitor overrides a header of the same name (the default path, since `append_stamps` defaults to `False`). But for a scalar duplicate the code re-assigns `d1`'s own value instead of `d2`'s:

```python
d1[key] = [value] if aggregate_duplicates else value   # keeps d1, should take d2
```

So the visitor's value is silently dropped. This contradicts both the docstring and the worked example in `_stamp_headers`:

```python
>>> from celery.canvas import _merge_dictionaries
>>> h = {'foo': 'bar1'}
>>> _merge_dictionaries(h, {'foo': 'bar2'}, aggregate_duplicates=False)
>>> h['foo']
'bar1'          # the inline example in _stamp_headers says this should be 'bar2'
```

The fix takes `d2[key]` in the scalar branch, matching the documented behavior. The `aggregate_duplicates=True` path (list aggregation) is unchanged. Added a regression test; `test_stamping.py` + `test_canvas.py` pass (651).

Scope note: I left the list-valued duplicate branch as-is to keep the change surgical — only the scalar case, which is what the docstring example covers. Happy to extend if you'd prefer.

---

Disclosure: this PR was authored by an AI coding agent (Claude Code) running on this account — it found the mismatch against the docstring and worked example, reproduced it, wrote the fix and the test, and wrote this description. The account holder reviews every change and is accountable for it, and the verification above is re-runnable from the diff. Happy to close it if it isn't the kind of contribution you want.


auvipy:
any related open issues for this you can refer to?

chuenchen309:
No existing issue that I could find — I searched open/closed issues for `stamping` / `append_stamps` / `visitor` and didn't see one covering this. It came out of the code itself: `_merge_dictionaries(..., aggregate_duplicates=False)` documents that duplicate keys are taken from `d2`, and `_stamp_headers` depends on that so a visitor's stamp overrides an existing header of the same name (the default `append_stamps=False` path). For a **scalar** duplicate, though, the branch re-assigns `d1`'s own value instead of `d2`'s, so the visitor's value is silently dropped — the regression test in the PR reproduces exactly that.

Happy to open a tracking issue for it if you'd prefer to have one to reference.

auvipy:
@Nusnus need your approval here

t/unit/tasks/test_canvas.py
Nusnus, line 1: Potential bug - check if these pass:
```python
def test_scalar_vs_list_no_self_extend_no_d2_mutation():
    d1, d2 = {"foo": "bar"}, {"foo": [1, 2]}
    _merge_dictionaries(d1, d2, aggregate_duplicates=False)
    assert d1["foo"] == [1, 2]
    assert d2 == {"foo": [1, 2]}  # d2 must not be mutated

def test_restamp_replaces_previous_stamp():
    sig = dummy.si()
    sig.stamp(VisitorReturning({"mystamp": "first"}))
    sig.stamp(VisitorReturning({"mystamp": "second"}))
    assert sig.options["mystamp"] == "second"  # works on main, breaks with this PR
```

For me:
```console
...
t/unit/tasks/test_pr10421_pins.py::test_scalar_vs_list_no_self_extend_no_d2_mutation FAILED [ 66%]
t/unit/tasks/test_pr10421_pins.py::test_restamp_replaces_previous_stamp FAILED [100%]
 
______________ test_scalar_vs_list_no_self_extend_no_d2_mutation _______________
>       assert d1["foo"] == [1, 2]
E       AssertionError: assert [1, 2, 1, 2] == [1, 2]
 
_____________________ test_restamp_replaces_previous_stamp _____________________
>       assert sig.options["mystamp"] == "second"  # works on main, breaks with this PR
E       AssertionError: assert 'first' == 'second'
E         - second
E         + first
 
FAILED t/unit/tasks/test_pr10421_pins.py::test_scalar_vs_list_no_self_extend_no_d2_mutation
FAILED t/unit/tasks/test_pr10421_pins.py::test_restamp_replaces_previous_stamp
========================= 2 failed, 1 passed in 0.14s ==========================
```

Nusnus (changes requested):

t/unit/tasks/test_canvas.py
chuenchen309: You're right on both, and sorry for the slow reply. I reproduced them and they're worse than "potential" — the second one means this PR trades one bug for another, so it shouldn't go in as written.

**Your first case.** On this branch:

```
d1['foo'] = [1, 2, 1, 2]
d2        = {'foo': [1, 2, 1, 2]}     # mutated
d1['foo'] is d2['foo'] -> True
```

`d1[key] = d2[key]` binds the *same list object*, and the following `if isinstance(d2[key], list) and isinstance(d1[key], list)` is a plain `if`, not an `elif` — so it immediately extends that list with itself. On `main` the same input gives `d1['foo'] == 'bar'` and `d2` untouched.

**Your second case.** Confirmed, and it's the important one:

```
this branch:  after 1st stamp -> 'first'   after 2nd -> 'first'
main:         after 1st stamp -> 'first'   after 2nd -> 'second'
```

The cause is that `_merge_dictionaries(..., aggregate_duplicates=False)` is called from `_stamp_headers` twice with **opposite** precedence intent:

- `_merge_dictionaries(headers, visitor_headers, aggregate_duplicates=append_stamps)` — here `d2` (the visitor) should win. That is exactly what the comment above it says: *"Use append_stamps=False to prioritize visitor_headers over headers"*.
- `_merge_dictionaries(headers, redacted_options, aggregate_duplicates=append_stamps)` — here `d1` (the headers that already carry the visitor's value) should win over `d2` (the pre-existing `self.options`).

`main` resolves both as "d1 wins", which is right for the second call and wrong for the first — that's the drop I was reporting. My change resolves both as "d2 wins", which fixes the first and breaks the second, which is your re-stamp case. One flag can't mean both things, so patching the shared helper in either direction is wrong.

**Where I think this should go.** Two options, and I'd rather you pick than guess:

1. Leave `_merge_dictionaries` alone (d1 keeps winning) and fix the precedence at the first call site only — the visitor's headers are supposed to take priority, so merge them in a direction that expresses that, rather than encoding it in a shared flag.
2. Give the helper an explicit winner instead of overloading `aggregate_duplicates` (which currently means both "aggregate into a list" and "who wins on conflict"). More invasive, but it removes the ambiguity that produced this whole thread.

Either way the aliasing needs fixing too — whatever ends up assigned should be copied (`list(...)`) and the following `if` should be an `elif`, otherwise a caller's dict can be mutated through the merge.

If you'd prefer this closed rather than reworked, I'm happy to close it — the underlying report (the visitor's value being dropped for a scalar duplicate key) would still be worth an issue, and I can open one with the repro above.

AI-assisted with agentic coding tools; all changes were reviewed and tested by me.

celery/canvas.py
```diff
@@ -103,7 +103,7 @@ def _merge_dictionaries(d1, d2, aggregate_duplicates=True):
                 _merge_dictionaries(d1[key], d2[key])
             else:
                 if isinstance(value, (int, float, str)):
-                    d1[key] = [value] if aggregate_duplicates else value
+                    d1[key] = [value] if aggregate_duplicates else d2[key]
                 if isinstance(d2[key], list) and isinstance(d1[key], list):
                     d1[key].extend(d2[key])
```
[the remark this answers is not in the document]
auvipy: @chuenchen309 can you cross check this as well?

Issue #9506, opened 2025-01-21
setup_logging silently discards exceptions


drienkop:
# Checklist

- [x] I have verified that the issue exists against the `main` branch of Celery.
- [ ] This has already been asked to the [discussions forum](https://github.com/celery/celery/discussions) first.
- [x] I have read the relevant section in the
  [contribution guide](https://docs.celeryq.dev/en/main/contributing.html#other-bugs)
  on reporting bugs.
- [x] I have checked the [issues list](https://github.com/celery/celery/issues?q=is%3Aissue+label%3A%22Issue+Type%3A+Bug+Report%22+-label%3A%22Category%3A+Documentation%22)
  for similar or identical bug reports.
- [x] I have checked the [pull requests list](https://github.com/celery/celery/pulls?q=is%3Apr+label%3A%22PR+Type%3A+Bugfix%22+-label%3A%22Category%3A+Documentation%22)
  for existing proposed fixes.
- [x] I have checked the [commit log](https://github.com/celery/celery/commits/main)
  to find out if the bug was already fixed in the main branch.
- [x] I have included all related issues and possible duplicate issues
  in this issue (If there are none, check this box anyway).
- [ ] I have tried to reproduce the issue with [pytest-celery](https://docs.celeryq.dev/projects/pytest-celery/en/latest/userguide/celery-bug-report.html) and added the reproduction script below.

## Mandatory Debugging Information

- [x] I have included the output of ``celery -A proj report`` in the issue.
    (if you are not able to do this, then at least specify the Celery
     version affected).
- [x] I have verified that the issue exists against the `main` branch of Celery.
- [x] I have included the contents of ``pip freeze`` in the issue.
- [x] I have included all the versions of all the external dependencies required
  to reproduce this bug.

## Optional Debugging Information

- [x] I have tried reproducing the issue on more than one Python version
  and/or implementation.
- [x] I have tried reproducing the issue on more than one message broker and/or
  result backend.
- [] I have tried reproducing the issue on more than one version of the message
  broker and/or result backend.
- [ ] I have tried reproducing the issue on more than one operating system.
- [ ] I have tried reproducing the issue on more than one workers pool.
- [ ] I have tried reproducing the issue with autoscaling, retries,
  ETA/Countdown & rate limits disabled.
- [ ] I have tried reproducing the issue after downgrading
  and/or upgrading Celery and its dependencies.

#### Related Issues

[...]

Issue #6596, opened 2021-01-13
Documentation for creating JSON logs for Kubernetes deployment


ShahBinoy:
# Checklist

- [x] I have checked the [issues list](https://github.com/celery/celery/issues?utf8=%E2%9C%93&q=is%3Aissue+label%3A%22Category%3A+Documentation%22+)
  for similar or identical bug reports.
- [x] I have checked the [pull requests list](https://github.com/celery/celery/pulls?q=is%3Apr+label%3A%22Category%3A+Documentation%22)
  for existing proposed fixes.
- [x] I have checked the [commit log](https://github.com/celery/celery/commits/master)
  to find out if the bug was already fixed in the master branch.
- [x] I have included all related issues and possible duplicate issues in this issue
      (If there are none, check this box anyway).

#### Related Issues

- None

#### Possible Duplicates

- None

# Description

Celery is really light on documentation for customizing Logging.
Internet and S/O questions have remained un-answered.
. What is celery's recommended way to write logs? 
. How to customize and format logs to JSON format?
. How to add custom fields to the logs, does it respect the `extra` field ?

# Suggestions

There should be a detailed section on what and how to put correct logging enhancements

auvipy:
do you have any practical exp/suggestion to share?

ShahBinoy:
> do you have any practical exp/suggestion to share?

@auvipy 
I am deploying celery workers on Kubernetes as `Deployments` and need to send logs to Kibana. Currently the logs dont render nicely because the output is not formatted in JSON. I need to override or re-format the Log output to JSON some additional `extra` fields as `dict`

Is that helpful

auvipy:
> > do you have any practical exp/suggestion to share?
> 
> @auvipy
> I am deploying celery workers on Kubernetes as `Deployments` and need to send logs to Kibana. Currently the logs dont get parsed/indexed correctly because the output is not formatted in JSON. I need to override or re-format the Log output to JSON some additional `extra` fields as `dict`
> 
> Is that helpful

yes, can you come with a PR with some of your suggested improvements? we can then share our views & might give you a helping hand

thedrow:
@brainbreaker Feel free to issue a PR at any time you'd like.
If you can do this in the following weeks, we'll get this merged by the time 5.1 will land GA.

brainbreaker:
To share a bit about how we do logging for our Python apps running on K8s -> 
[...]

PR #10424, opened 2026-07-21
Add worker_skip_logging_setup setting to disable logging configuration

### LINKED ISSUE #1867: Feature request: Create a config option for Celery to not touch logging config anywhere ever
Working with Celery in a more complex environment, like ours, is very frustrating because it seems to remove handlers and then seemingly sets up arbitrary loggers (like a `kombu.log.NullHandler` for the 'celery' logger). This is done even when we set `CELERYD_HIJACK_ROOT_LOGGER` and `CELERY_REDIRECT_STDOUTS` to `False`.

At my company we have a default config for the `logging` module which we want all services to use all the time with no exceptions (because these messages get picked up and reformatted by logstash/logstash-forwarder and put into elasticsearch). The only way to use this with Celery currently seems to bring Celery up then iterate through all loggers and remove them forcefully and then set up a sane logging configuration. Which means we set logging up twice in our environment, with the alternative being to patch Celery or monkey-patch Celery.

So my feature request is this: I'd like to set `CELERY_I_HUMBLY_REQUEST_YOU_STOP_EFFING_WITH_MY_LOGGING = True` and then Celery should just not do anything to `logging` _at all_ aside from writing log messages.

I apologize if this is already possible and I just missed the adequate settings in the docs.

razchiriac:
## Description

Fixes #1867 (open since 2014).

Adds a `worker_skip_logging_setup` setting (bool, default `False`). When
enabled, Celery performs no logging configuration at all: no handlers are
added or removed, no logger levels are changed, the root logger is left
untouched, and stdout/stderr are not redirected
(`worker_redirect_stdouts` is ignored).

This is byte-for-byte equivalent to the long-standing community
workaround of connecting an empty receiver to the `setup_logging` signal
(the guard returns a truthy value from `setup_logging_subsystem`, the
same contract as a non-empty receiver list, so `setup()` also skips
stdout redirection). Applications that configure Python logging
themselves — structlog, dictConfig, Django LOGGING, JSON logging for
Kubernetes (#6596) — no longer need to define a do-nothing receiver,
which also side-steps the receiver-exception pitfalls tracked in #9506.

Existing behavior is completely unchanged when the setting is disabled
(the default). `worker_hijack_root_logger` remains the lighter option
that only prevents removal of existing root handlers.

## Changes

- `celery/app/defaults.py`: register the new worker option.
- `celery/app/log.py`: guard in `Logging.setup_logging_subsystem`.
- `docs/userguide/configuration.rst`: new setting entry
  (`versionadded:: 5.7`) plus a note under `worker_hijack_root_logger`.
- `docs/userguide/signals.rst`: note under the `setup_logging` signal.
- `t/unit/app/test_log.py`: tests for setting-on (root logger and
  celery loggers untouched without any signal receiver; no stdout
  redirection) and setting-off (current behavior preserved).


DomHudson:
@ionelmc  What do you think of merging this? It would be good to close #1867 once and for all :)

ionelmc:
> @ionelmc What do you think of merging this? It would be good to close #1867 once and for all :)

Did you want to ping @auvipy instead? I made some comments on that issue, yes, but 12 years ago... today, I am basically drawing blanks.

auvipy:
> @ionelmc What do you think of merging this? It would be good to close #1867 once and for all :)

that might be great for sure

auvipy (changes requested):
can you please fix the merge conflicts and address the open suggestion?

MehrazRumman (commented):
Reviewed at `d7ddb90` against current `main`. The change itself is correct and I would approve it once the three mechanical items below are done. Notes for @razchiriac to unblock @auvipy's request:

**1. The merge conflict is a single hunk in `celery/app/log.py`.** `main` added a `datefmt=None` parameter to `setup_logging_subsystem()` (the `worker_log_datefmt` feature) on the same signature lines this PR touches. Resolution is to keep `main`'s signature and put the new guard directly under it:

```python
    def setup_logging_subsystem(self, loglevel=None, logfile=None, format=None,
                                colorize=None, hostname=None, datefmt=None,
                                **kwargs):
        if self.app.conf.worker_skip_logging_setup:
            ...
            self.already_setup = True
            return True
        if self.already_setup:
            return
```

The other four files auto-merge cleanly (`git merge-tree` confirms). The pre-commit.ci failure is only the "error during mergeable check" caused by the conflict, not a lint problem.

**2. Copilot's first finding is already fixed, the second is still open.** The autofix commit moved the setting check above `already_setup`, so repeated `setup()` calls now correctly return `True` and never redirect. The suggested regression test for that is not in the branch yet: `test_setup_skip_logging_setup_no_stdout_redirect` calls `setup(..., redirect_stdouts=True)` once. Call it a second time and re-assert `sys.stdout`/`sys.stderr` are unchanged; that locks in the property the reorder was made for.

**3. One doc inaccuracy.** The configuration entry says this is "equivalent to connecting an empty receiver" and that `after_setup_logger`/`after_setup_task_logger` won't be sent. That is true, but the early return also means the **`setup_logging` signal itself is never sent** when the setting is on. So an application (or a library it imports) that connects a real `setup_logging` receiver and also enables this setting silently loses its hook. Worth one sentence in both `configuration.rst` and `signals.rst` so nobody has to discover it by debugging. For completeness, the early return also skips `reset_multiprocessing_logger()`, `set_default_encoding_file()`, and the `_MP_FORK_*` env vars, which the empty-receiver path still runs; I think skipping them is the right reading of "no logging configuration at all", but "byte-for-byte equivalent" in the description overstates it.

Things I checked that are fine: every caller of `setup_logging_subsystem()`/`log.setup()` on `main` (worker, beat, prefork initializer, `events.snapshot`, `bin/worker.detach`, testing worker) only tests the return for truthiness, so returning `True` instead of a receivers list is safe, and `apps/worker.py` gates its own `redirect_stdouts()` on that value. `versionadded:: 5.7` matches the other new entries on `main`. The `worker_` prefix while also affecting beat follows the `worker_hijack_root_logger` precedent.

Issue #10434, opened 2026-07-26
Buffered task events are silently dropped when the worker uses the threads pool


oliverhaas:
<details>
<summary><b>Checklist, related issues and environment</b></summary>

# Checklist

- [x] I have verified that the issue exists against the `main` branch of Celery.
- [ ] This has already been asked to the [discussions forum](https://github.com/celery/celery/discussions) first.
- [x] I have read the relevant section in the
  [contribution guide](https://docs.celeryq.dev/en/main/contributing.html#other-bugs)
  on reporting bugs.
- [x] I have checked the [issues list](https://github.com/celery/celery/issues?q=is%3Aissue+label%3A%22Issue+Type%3A+Bug+Report%22+-label%3A%22Category%3A+Documentation%22)
  for similar or identical bug reports.
- [x] I have checked the [pull requests list](https://github.com/celery/celery/pulls?q=is%3Apr+label%3A%22PR+Type%3A+Bugfix%22+-label%3A%22Category%3A+Documentation%22)
  for existing proposed fixes.
- [x] I have checked the [commit log](https://github.com/celery/celery/commits/main)
  to find out if the bug was already fixed in the main branch.
- [x] I have included all related issues and possible duplicate issues
  in this issue (If there are none, check this box anyway).
- [ ] I have tried to reproduce the issue with [pytest-celery](https://docs.celeryq.dev/projects/pytest-celery/en/latest/userguide/celery-bug-report.html) and added the reproduction script below. -> Reproduction of such a threading issue in tests is unfortunately super complicated. I have reproduced the bug in multiple ways, but not in a perfectly clean-but-complete test; see below.

The bug is inside `EventDispatcher` itself, so it reproduces with no worker, no broker and no
pytest-celery fixtures. A unit test and a script are attached instead.

## Mandatory Debugging Information

- [x] I have included the output of ``celery -A proj report`` in the issue.
    (if you are not able to do this, then at least specify the Celery
     version affected).
- [x] I have verified that the issue exists against the `main` branch of Celery.
- [x] I have included the contents of ``pip freeze`` in the issue.
- [x] I have included all the versions of all the external dependencies required
  to reproduce this bug.

## Optional Debugging Information

- [x] I have tried reproducing the issue on more than one Python version
  and/or implementation.
- [x] I have tried reproducing the issue on more than one message broker and/or
  result backend.
- [x] I have tried reproducing the issue on more than one version of the message
  broker and/or result backend.
[...]

PR #10435, opened 2026-07-26
fix: don't drop buffered events appended during publish

oliverhaas:
See #10434

Note that it would have also been possible to put the append of events under the mutex, but the approach I've chosen seemed simpler given the constraints and a tiny bit more efficient as well.


oliverhaas:
The actual logic change is just one line.

Let me know if the tests are too artificial or not. If this was my own project (and not a public package), I would probably not have these kind of very specific regression tests, but I don't have an alternative to get coverage for this.

auvipy (approved):

PR #10436, opened 2026-07-28
fix: route custom headers through stamping mechanism to prevent AMQP field-table corruption (#10412) [FaaFyfxR9WAQrL7FcAgEHJvztd8cVMxvjHRS55rw1nwH]

### LINKED ISSUE #10412: AMQP header table corrupted when custom headers (e.g. APM tracing) merge with built-in protocol headers on task publish
### Summary
Publishing a task via `Task.apply_async(headers=...)` with a custom header key present (e.g. Elastic APM's celery integration, which sets `headers["elasticapm"]["trace_parent_string"]` from a `before_task_publish` signal handler) produces an AMQP header field-table that RabbitMQ cannot decode. The broker's channel process crashes (`function_clause` in `rabbit_binary_parser:parse_table`), and after repeated crashes RabbitMQ closes the whole AMQP connection — client sees `amqp.exceptions.InternalError: (0, 0): (541) INTERNAL_ERROR`.

### Environment
- celery 5.6.1 (checked 5.6.2/5.6.3 changelogs — no related fix)
- kombu 5.6.2, amqp (py-amqp) 5.3.1
- elastic-apm 6.25.0 (celery integration via before_task_publish -> set_celery_headers)
- RabbitMQ broker, AMQPS/TLS
- Python 3.13

### Evidence
549 broker-side crashes captured over a 14h production window, all on tasks published from inside an active elastic-apm HTTP transaction. Confirmed identical corruption across 9+ different task names — not payload-specific, it's the shared header-construction path (anything merging a custom `headers=` dict with celery's built-in protocol headers).

Reimplemented RabbitMQ's actual rabbit_binary_parser:parse_table byte-for-byte (from rabbitmq/rabbitmq-server) and ran it against all 549 captured crash dumps. Result: 549/549 fail at the identical offset — offset 0, invalid AMQP type tag 0x6b ('k') immediately after a corrupted zero-length field name, right where the literal bytes spell "kwargsrepr" (celery's hardcoded protocol header key, which needs a 0x0A length-prefix, not 0x00).

Sample decoded bytes (internal hostname redacted):
```
b'\x00kwargsreprS\x00\x00\x00\x02{}\x06originS\x00\x00\x00,gen2418@<redacted-host>\rignore_resultt\x00\x15replaced_task_nestingI\x00\x00\x00\x00\x0fstamped_headersV\x06stampsF\x00\x00\x00\x00\x00elasticapmF\x00\x00\x00P\x13trace_parent_stringS\x00\x00\x00700-a6bf418165f95e74425b39378f2f1dad-ed201c37b0d03203-00'
```

Only fires when stamped_headers/stamps/elasticapm/trace_parent_string are present — i.e. only when a custom header key merges in alongside celery's auto-generated protocol headers.

### Suspected cause
Looks like it lands in celery's header-stamping/merge logic (celery/app/amqp.py) that folds a custom `headers=` dict into celery's built-in protocol headers (kwargsrepr, origin, etc.) and wraps "extra" keys into stamped_headers/stamps. Related older report: elastic/apm-agent-python#1370 ("There is a Celery bug concerning its headers field", referencing celery/celery#4875) — this may be a resurgence of that class of bug, now producing on-wire corruption instead of just header loss.

### Ask
Is there a known issue tracking the header-merge logic mishandling custom headers combined with built-in protocol headers? Happy to share the full 549-sample dataset and the parse simulator script.

waterWang:
## Description

Fixes #10412 — AMQP header field-table corruption when custom headers (e.g. Elastic APM's `elasticapm` key) are merged directly into the protocol headers dict via `dict.update()`.

### Root Cause

In `send_task_message`, user-provided `headers=` dict was merged directly into the built-in protocol headers via `headers2.update(headers)`. When integrations (such as Elastic APM) inject nested dict values (e.g. `headers["elasticapm"]["trace_parent_string"]`), these end up as top-level AMQP field-table entries. The resulting wire-format can contain a zero-length field-name that RabbitMQ's `rabbit_binary_parser:parse_table` rejects with `function_clause`, causing connection-wide crashes.

### Fix

- **Before**: `if headers: headers2.update(headers)` — every user-supplied key becomes a top-level AMQP field.
- **After**: Known protocol header keys (`lang`, `task`, `id`, `kwargsrepr`, etc.) are still allowed to override via `update()`. All other keys are routed through the `stamped_headers` / `stamps` mechanism that Celery already has for custom header propagation.

This preserves backward compatibility for protocol header overrides while preventing custom integration headers from corrupting the AMQP wire format.

### Testing

The existing test suite covers the stamping mechanism path. Verified that:
- Known protocol header overrides still work (e.g. `headers={'lang': 'custom'}`)
- Custom headers are correctly added to `stamped_headers` / `stamps`
- Empty/None custom headers are handled gracefully
- No regression for the normal (no custom headers) path

### DIFF
--- a/celery/app/amqp.py
+++ b/celery/app/amqp.py
@@ -23,6 +23,22 @@ __all__ = ('AMQP', 'Queues', 'task_message')
 #: earliest date supported by time.mktime.
 INT_MIN = -2147483648
 
+#: Celery protocol header keys that must not be polluted by
+#: arbitrary user-supplied header keys.  Custom / integration
+#: headers (e.g. Elastic APM's ``elasticapm``) are routed
+#: through the ``stamped_headers`` / ``stamps`` mechanism
+#: instead of being merged as top-level AMQP field-table entries,
+#: which would otherwise risk wire-level corruption (RFC-like
+#: zero-length field-name) when the resulting dict is serialised
+#: by py-amqp / kombu.
+_PROTOCOL_HEADER_KEYS = frozenset({
+    'lang', 'task', 'id', 'shadow', 'eta', 'expires',
+    'group', 'group_index', 'retries', 'timelimit',
+    'root_id', 'parent_id', 'argsrepr', 'kwargsrepr',
+    'origin', 'ignore_result', 'replaced_task_nesting',
+    'stamped_headers', 'stamps',
+})
+
 #: Human readable queue declaration.
 QUEUE_FORMAT = """
 .> {0.name:<16} exchange={0.exchange.name}({0.exchange.type}) \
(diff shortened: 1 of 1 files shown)


auvipy (changes requested):
this changes also need appropriate unit tests to be considered

Issue #10388, opened 2026-07-01
Malformed numeric worker environment variables can fail with import-time ValueError


MicroMilo:
### Summary

Several worker-related numeric environment variables are parsed with direct `int()` / `float()` calls at module import time. If one of these variables is malformed, Celery can fail before worker initialization with a raw Python traceback instead of a Celery configuration error.

### Reproduction

```bash
CELERY_WORKER_REVOKES_MAX=abc python - <<'PY'
import celery.worker.state
PY
```

### Actual behavior

The import fails with a traceback ending in:

```text
ValueError: invalid literal for int() with base 10: 'abc'
```

### Expected behavior

Invalid numeric environment variables should ideally be reported as configuration errors that name the variable and expected type, for example:

```text
Invalid value for CELERY_WORKER_REVOKES_MAX: expected integer, got 'abc'
```

If the intended product semantics are that these low-level environment variables may raise raw Python conversion errors, it would be useful to document that explicitly. Otherwise this looks like a config parsing / feedback mismatch.

### Source of the mismatch

`celery.worker.state` defines numeric worker settings from environment variables at import time:

```python
REVOKES_MAX = int(os.environ.get('CELERY_WORKER_REVOKES_MAX', 50000))
REVOKE_EXPIRES = float(os.environ.get('CELERY_WORKER_REVOKE_EXPIRES', 10800))

SUCCESSFUL_MAX = int(os.environ.get('CELERY_WORKER_SUCCESSFUL_MAX', 1000))
SUCCESSFUL_EXPIRES = float(os.environ.get('CELERY_WORKER_SUCCESSFUL_EXPIRES', 10800))
```

These values are later consumed by worker state structures, for example:

```python
successful_requests = LimitedSet(maxlen=SUCCESSFUL_MAX, expires=SUCCESSFUL_EXPIRES)
```

Because the conversion happens directly during import, malformed values do not pass through a Celery configuration/error-reporting layer.

A similar direct numeric environment parse also exists for RDB port configuration:

```python
CELERY_RDB_PORT = int(os.environ.get('CELERY_RDB_PORT') or DEFAULT_PORT)
```

### Why this matters

In container, systemd, or Kubernetes deployments, malformed environment variables are a common configuration failure mode. A clear Celery-facing error that includes the variable name and expected type would make startup failures easier to diagnose than an import-time Python traceback.

### Possible fix

Centralize numeric environment parsing behind helpers that include the variable name in any error message, for example:

```python
[...]

PR #10438, opened 2026-07-28
fix(worker): raise ImproperlyConfigured for malformed worker numeric environment variables

amir-rezaei:
## Description
This PR resolves issue #10388 by wrapping numeric worker environment variable parsing in `_int_env` and `_float_env` helpers in `celery/worker/state.py`.

## Details
- Raises a clear `ImproperlyConfigured` exception naming the environment variable and invalid value instead of a raw `ValueError` traceback during import.


auvipy:
@Namraa310806 may be you can take this with proper tests.

Namraa310806:
Hi @auvipy , Just wanted to clarify the scope before I start — would you prefer me to keep the implementation approach from #10438 and focus on adding the proper regression tests, or should I also improve the fix itself based on the review feedback from that PR and then add the tests?

auvipy:
also improve the fix as well...

PR #10451, opened 2026-08-01
perf(utils): fast path saferepr for flat, short values

dexhunter:
`saferepr` builds its output by pushing an iterator onto a `deque`, walking it
in `reprstream` through a chain of `isinstance` checks, wrapping every container
in a generator tacked onto the rest of the work with `itertools.chain`, and
joining the token stream. That machinery is what makes truncation and depth
limiting work, and it earns its cost on a large or deeply nested object.

Most of what celery hands it is neither. `add.delay(1, 2)` renders `(1, 2)` and
`{}`, and the streaming path costs about 4.4 microseconds to do it.

## Where it runs

Two calls per published task, at `celery/app/amqp.py:363` and `:365`, inside
`as_task_v2`, which every `apply_async` and `send_task` goes through. One more
per successfully executed task, at `celery/app/trace.py:633`, on the return
value. That is three calls per task round trip at stock settings.

`saferepr` is 80% of the cost of building a task message. Over 20,000
`as_task_v2` calls with `args=(1, 2), kwargs={}`, cProfile attributes 0.719 s of
the 0.903 s total to `saferepr` — 180,000 `reprstream` yields and 1,180,000
`isinstance` calls.

## The change

A fast path for a leaf, or a flat container of leaves, that returns only when
the result is short enough not to need truncating. Everything else goes to the
existing code untouched: nested containers, subclasses of `str`/`dict`/`list`,
`bytes`, sets, anything with a `seen` set passed in, and any rendering that
would exceed `maxlen`.

The point of that restriction is that neither of the two hard behaviours has to
be reimplemented. Only depth-1 structures reach the fast path, so no value of
`maxlevels` can cap them. The result is returned only when it fits, so the
`maxlen` budget can never engage.

## Measurements

Wall time, median of 9 runs, on an idle machine, at stock defaults —
`argsrepr_maxsize`, `kwargsrepr_maxsize` and `resultrepr_maxsize` are all 1024
and `maxlevels` is 3. No setting was changed to produce these.

`as_task_v2`, the whole per-publish message build:

| call | before | after | |
| --- | --- | --- | --- |
| `args=(1, 2), kwargs={}` | 11.168 us | 4.370 us | 2.56x |
| `args=("user@example.com", 42), kwargs={"notify": True}` | 15.486 us | 4.805 us | 3.22x |

`saferepr` on its own, over payloads shaped like real task arguments and return
values:

| payload | before | after | |
| --- | --- | --- | --- |
| `()` | 2.711 us | 0.253 us | 10.73x |
| `(12345,)` | 3.499 us | 0.406 us | 8.61x |
| `(1, 2)` | 4.441 us | 0.541 us | 8.21x |
| `("user@example.com", 42)` | 4.885 us | 0.617 us | 7.92x |
| `("3f2504e0-4f89-11d3-9a0c-0305e82c3301",)` | 3.920 us | 0.471 us | 8.33x |
| `{}` | 2.839 us | 0.271 us | 10.49x |
| `{"user_id": 123, "notify": True}` | 10.491 us | 1.017 us | 10.31x |
| `{"path": "/var/log/app.log", "n": 10, "dry": False}` | 14.776 us | 1.388 us | 10.64x |
| `(["alpha", "beta", "gamma"],)` | 8.722 us | 8.760 us | 1.00x |
| `({"outer": {"inner": {"leaf": 1}}},)` | 13.790 us | 9.978 us | 1.38x |
| `(list(range(20)),)` | 23.261 us | 22.867 us | 1.02x |
| return `None` | 1.808 us | 0.256 us | 7.06x |
| return `{"status": "ok", "count": 7}` | 10.718 us | 1.049 us | 10.22x |
| return `"x" * 900` | 2.388 us | 0.326 us | 7.32x |
| mean | 7.732 us | 3.443 us | 2.25x |

The three rows near 1.00x are the ones that fall back. A tuple holding a list is
two levels deep, so the fast path declines it and the original code runs
unchanged. **Nested arguments get no improvement from this patch**, and the
`(list(range(20)),)` row at 1.02x is inside measurement noise rather than a real
gain. The win is on flat arguments, which is what most tasks take.

Instruction counts, measured with Callgrind because the machine was shared, over
a fixed 29-payload corpus: **5.5068 M down to 2.4505 M per pass**. Three runs of
the patched code gave 2.4505 / 2.4482 / 2.4511, a spread of 0.12%.

## Behaviour

Unchanged, and the parts that are easy to get wrong are the reason the fast path
is as narrow as it is:

- `saferepr` is not `repr`. It always single-quotes, escapes only `'`, and
  leaves backslashes, newlines and tabs raw. The fast path does the same.
- Only exact types are handled. A `str` or `dict` subclass renders as
  `ClassName(...)` and goes to the original path.
- A one-element tuple still renders `(x,)`.
- `maxlen=None` and `maxlevels=None` both still mean unlimited.
- Truncation and depth capping are untouched, since the fast path returns only
  when neither can apply.

`t/unit/utils/test_saferepr.py` passes, all 70 cases. I also diffed the patched
function against the original across the fixed cases above plus randomly
generated payloads, at seven `(maxlen, maxlevels)` combinations including
`(1024, 3)`, `(None, None)`, `(64, 3)`, `(16, 2)` and `(8, 3)`, checking for
byte-identical output.

## Test plan

- [x] `t/unit/utils/test_saferepr.py` — 70 passed, before and after
- [x] Differential check against the unpatched function on a fixed adversarial
      battery and randomly generated payloads, at the seven `(maxlen,
      maxlevels)` settings listed above, requiring byte-identical output
- [x] `flake8`, `isort`, `codespell` and `pyupgrade --py38-plus` clean on the
      changed file, at the versions pinned in `.pre-commit-config.yaml`

---

The shape of the fast path came out of an optimization run; the trajectory is at
https://dashboard.weco.ai/share/-9x3Ti2y-uc7gDXgoddzNtwLT5ht8pY- if it is of
interest. Ten of its nineteen candidates were rejected for getting truncation or
depth wrong, which is most of why this patch declines so much. Its best
candidate went further by unrolling the type dispatch four times for 2.00 M
instructions against this patch's 2.45 M; I did not think 18% was worth a
hundred lines of repetition, so what is here is the readable version.

Note `celery/utils/saferepr.py` is also touched by #9987, which updates type
comments repo-wide. The two do not overlap in content and I am happy to rebase
whichever lands second.


auvipy (changes requested):
I am not fully sure about the changes. may be you can also add some regression tests to verify this

dexhunter:
@auvipy Added regression coverage in c7d8479f2. It exercises the public saferepr API and verifies that exact-type flat scalars and containers use the optimized path, including the exact maxlen boundary, while nested containers, list subclasses, and over-budget flat containers use the real streaming path and preserve its output. Validation: 79 focused tests passed, and all project pre-commit hooks passed on the changed files. Applied to the current base (74c46f8e3), the six fast-path cases fail, so the coverage specifically guards this optimization.

PR #10452, opened 2026-08-02
Count buffered items when initializing BufferMap.total

uttam12331:
### Summary

`BufferMap` subclasses `OrderedDict`, and its buffers are stored as the dict **values** (`_get_or_create_buffer` returns `self[key]`, a `Messagebuffer`). When a `BufferMap` is constructed with an `iterable`, `__init__` seeds the running total:

```python
self.total = sum(len(buf) for buf in self.items())
```

`items()` yields `(key, value)` tuples, so `len(buf)` is `len((key, value)) == 2` for every entry — never the buffer's length. The total ends up as `2 * number_of_keys` instead of the number of buffered items.

### Reproduce

```python
from celery.utils.collections import BufferMap, Messagebuffer

bm = BufferMap(100, iterable={"a": Messagebuffer(100, [1, 2]),
                              "b": Messagebuffer(100, [3, 4, 5])})
bm.total   # -> 4  (2 keys × 2), expected 5 (2 + 3 buffered items)
```

`self.total` backs `_evictcount()` and `__repr__`, so a `BufferMap` seeded from existing data reports a corrupt size.

### Fix

Iterate `self.values()`, so the total reflects the real buffered-item count — matching how `put()` (`self.total += 1`) and `extend()` (`self.total += len(it)`) maintain it:

```python
self.total = sum(len(buf) for buf in self.values())
```

### Tests

Added `test_total_from_iterable`, which seeds a `BufferMap` from two buffers and asserts `total == 5`. It fails on `main` (`4`) and passes with the fix. The existing `BufferMap` tests all construct empty maps, so they're unaffected (`39 passed`).

### Note

This is a separate bug from the adjacent `self.bufmaxsize = 1000` (ignoring the `bufmaxsize` argument) addressed in #10138 — this PR touches only the `self.total` line and leaves that one alone.


auvipy:
what issue it is fixing?

PR #10471, opened 2026-08-10
Utilizing Identity instead of Sequence for id column to avoid RDS Proxy from pinning the session connection

### LINKED ISSUE #10472: RDS Proxy pinning the database connection due to Sequence Column Feature
# Checklist

- [x ] I have verified that the issue exists against the `main` branch of Celery.
- [ x] This has already been asked to the [discussions forum](https://github.com/celery/celery/discussions) first.
- [ x] I have read the relevant section in the
  on reporting bugs.
- [ x] I have checked the [issues list](https://github.com/celery/celery/issues?q=is%3Aissue+label%3A%22Issue+Type%3A+Bug+Report%22+-label%3A%22Category%3A+Documentation%22)
- [ x] I have checked the [pull requests list](https://github.com/celery/celery/pulls?q=is%3Apr+label%3A%22PR+Type%3A+Bugfix%22+-label%3A%22Category%3A+Documentation%22)
  for existing proposed fixes.
- [ x] I have checked the [commit log](https://github.com/celery/celery/commits/main)
- [ x] I have included all related issues and possible duplicate issues

     version affected).
- [ ] I have verified that the issue exists against the `main` branch of Celery.
  to reproduce this bug.

  and/or implementation.
  result backend.
  broker and/or result backend.

#### Related Issues

- None

#### Possible Duplicates

- None

## Environment & Settings

**Celery version**:

<details>
<p>

```
```

</p>
</details>

# Steps to Reproduce

## Required Dependencies

- **Minimal Python Version**: 3.10
- **Minimal Celery Version**: 5.6.3
- **Minimal Broker Version**: RabbitMQ 3.13.7
- **Minimal Result Backend Version**: Postgres 17.7
- **Minimal OS and/or Kernel Version**: Ubuntu 24.04.4 LTS

### Python Packages

<details>
<p>

```
```

</p>
</details>

### Other Dependencies

<details>
<p>
N/A
</p>
</details>

<details>
<p>

```python
```

</p>
</details>

# Expected Behavior

When using Celery with RDS PostgreSQL and RDS Proxy, we expect RDS Proxy to use its connection pool to reuse database connections. This should result in lower CPU utilization and help avoid hitting the `max_connections` limit, allowing us to run more worker pods and execute tasks faster.

# Actual Behavior

According to the AWS CloudWatch metrics, when comparing Client Connections, Currently Session Pinned, and Currently Borrowed, we can see that all three metrics have the same value. This is a HUGE red flag in the RDS Proxy world.

This indicates that every database connection currently in use is pinned to a specific client and cannot be reused by other application requests until that specific client session ends.
 
Image

According to the RDS Proxy logs, we also see a large number of pinned connection messages. Here is an example of an actual log message:
`The client session was pinned to the database connection [dbConnection=3658691037] for the remainder of the session. The proxy can't reuse this connection until the session ends. Reason: SQL changed session settings that the proxy doesn't track. Consider moving session configuration to the proxy's initialization query. Digest: "insert into celery_taskmeta (id, task_id, status, result, date_done, traceback, name, args, kwargs, worker, retries, queue) values (nextval($1), $2, $3, $4, $5:: timestamp, $6, $7, $8, $9, $10, $11, $12) returning celery_taskmeta. id".`

When we increase the worker pod count beyond the database's max_connections limit, it results in OperationTimeout errors.

We want to add AWS RDS Proxy in front of the AWS RDS database to reduce CPU utilization and avoid database failover events. However, due to this limitation, RDS Proxy is unable to effectively reuse database connections, which is causing production issues, primarily slow task completion and database failover events.

https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/rds-proxy-pinning.html#:~:text=Manipulating%20sequences%20using%20functions%20such%20as%20nextval%20and%20setval.

# Proposal
Utilize modern Postgres feature (Identity feature) https://www.postgresql.org/docs/current/ddl-identity-columns.html 
Identity columns are handled entirely on the database server. Therefore, RDS Proxy should not pin the database connection because the driver no longer needs to send an SQL statement containing the nextval keyword.

rajpatel322:

## Description

backends/database/models.py uses the Sequence feature to generate incremental integer value for the ID column which serves as the primary key for the Task and TaskSet tables. Using the Sequence feature results in inserting `nextval` keyword into the SQL insert statement resulting in database session being pinned according to the AWS documentation (https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/rds-proxy-pinning.html#:~:text=functions%20such%20as-,nextval%20and%20setval,-)

To avoid RDS Proxy from pinning the session, we want to utilize the modern Postgres approach which is to utilize Identity feature (https://www.postgresql.org/docs/current/ddl-identity-columns.html)   

NOTE: All patches should be made against main, not a maintenance branch like
3.1, 2.5, etc.  That is unless the bug is already fixed in main, but not in
that version series.

Fixes #10472
-->

### DIFF
--- a/celery/backends/database/models.py
+++ b/celery/backends/database/models.py
@@ -56,7 +56,8 @@ class Task(ResultModelBase):
     @classmethod
     def configure(cls, schema=None, name=None):
         cls.__table__.schema = schema
-        cls.id.default.schema = schema
+        if isinstance(cls.id.default, sa.Sequence):
+            cls.id.default.schema = schema
         cls.__table__.name = name or cls.__tablename__
 
 
@@ -115,5 +116,6 @@ class TaskSet(ResultModelBase):
     @classmethod
     def configure(cls, schema=None, name=None):
         cls.__table__.schema = schema
-        cls.id.default.schema = schema
+        if isinstance(cls.id.default, sa.Sequence):
+            cls.id.default.schema = schema
         cls.__table__.name = name or cls.__tablename__
--- a/celery/backends/database/models.py
+++ b/celery/backends/database/models.py
@@ -30,9 +30,7 @@ class Task(ResultModelBase):
     __tablename__ = 'celery_taskmeta'
     __table_args__ = {'sqlite_autoincrement': True}
 
-
-    id = sa.Column(DialectSpecificInteger, sa.Identity(cache=20),
-                primary_key=True)
+    id = sa.Column(DialectSpecificInteger, sa.Identity(cache=20), primary_key=True)
     task_id = sa.Column(sa.String(155), unique=True)
     status = sa.Column(sa.String(50), default=states.PENDING)
     result = sa.Column(PickleType, nullable=True)
@@ -94,8 +92,7 @@ class TaskSet(ResultModelBase):
     __tablename__ = 'celery_tasksetmeta'
     __table_args__ = {'sqlite_autoincrement': True}
 
-    id = sa.Column(DialectSpecificInteger, sa.Identity(cache=20),
-                primary_key=True)
+    id = sa.Column(DialectSpecificInteger, sa.Identity(cache=20), primary_key=True)
     taskset_id = sa.Column(sa.String(155), unique=True)
     result = sa.Column(PickleType, nullable=True)
     date_done = sa.Column(sa.DateTime, default=_get_utc_now,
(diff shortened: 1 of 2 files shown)


auvipy (changes requested):
it is going to be a breaking change, so we need some way to keep backward compatibility here

rajpatel322:
> it is going to be a breaking change, so we need some way to keep backward compatibility here

I believe someone from the community created a new PR: https://github.com/celery/celery/pull/10474

auvipy:
you can incorporate changes from that PR as well. that contributor seems like a spammer

PR #10479, opened 2026-08-12
fix: make ResultConsumer.cancel_for idempotent to prevent GC-time deadlock in Redis backend

### LINKED ISSUE #10477: Redis backend: AsyncResult.__del__ deadlocks apply_async() by re-entering PubSub._lock
# Checklist

  on reporting bugs.
  for existing proposed fixes.

> Notes on the two unchecked boxes, for transparency: this was not raised in Discussions first because there is a deterministic reproduction script and a production stack trace, so there was no open question to discuss. The reproduction is a standalone script rather than a `pytest-celery` case because it needs to control garbage-collection timing precisely; it requires only a Redis server and no worker.
>
> "Verified against `main`" means: reproduced on 5.6.3 (latest release), and the three implicated functions — `AsyncResult.__del__`, `ResultConsumer.cancel_for`, `RedisBackend.on_task_call` — are byte-identical on `main` as of 2026-08-12.

  to reproduce this bug.

  and/or implementation.
  result backend.
- [x] I have tried reproducing the issue on more than one version of the message
  broker and/or result backend.
- [x] I have tried reproducing the issue after downgrading

#### Related Issues

- #6969 — same class of deadlock (GC finalizer vs. `threading.Lock` in the Redis backend), closed unmerged as a redis-py issue. Different trigger: `PubSub.__del__` during *disconnection*, not `AsyncResult.__del__` during an in-flight `SUBSCRIBE`.
- #6963 — the Redis result consumer is not drained, so unsubscribes accumulate. Same machinery, resource-leak symptom rather than deadlock.
- #4707 — no unsubscribe from `celery-task-meta-*` channels when `ignore_result=True`.

#### Possible Duplicates

- None

## Environment & Settings

**Celery version**: 5.6.3 (also reproduced on 5.6.2)

<details>
<p>

```
software -> celery:5.6.3 (recovery) kombu:5.6.2 py:3.11.15
            billiard:4.2.4 redis:6.3.0
            kernel version:25.5.0 imp:CPython
settings -> transport:redis results:redis://localhost:6379/0

broker_url: 'redis://localhost:6379/0'
result_backend: 'redis://localhost:6379/0'
deprecated_settings: None
```

</p>
</details>

# Steps to Reproduce

## Required Dependencies

- **Minimal Python Version**: 3.11
- **Minimal Celery Version**: reproduced on 5.6.2 and 5.6.3; implicated code is unchanged on `main`
- **Minimal Kombu Version**: 5.6.2
- **Minimal Broker Version**: any Redis (broker is irrelevant — the deadlock is in the result backend)
- **Minimal Result Backend Version**: any Redis
- **Minimal OS and/or Kernel Version**: N/A (reproduced on macOS; originally observed on Linux/aarch64)
- **Minimal Broker Client Version**: N/A
- **Minimal Result Backend Client Version**: **redis-py <= 6.3.0** (see below — 6.4.0 is unaffected)

### Python Packages

<details>
<p>

```
amqp==5.3.1
billiard==4.2.4
celery==5.6.3
click==8.4.2
click-didyoumean==0.3.1
click-plugins==1.1.1.2
click-repl==0.3.0
kombu==5.6.2
packaging==26.3
prompt_toolkit==3.0.53
PyJWT==2.13.0
python-dateutil==2.9.0.post0
redis==6.3.0
six==1.17.0
tzdata==2026.3
tzlocal==5.4.4
vine==5.1.0
wcwidth==0.8.2
```

</p>
</details>

### Other Dependencies

<details>
<p>

A Redis server. No worker is required to reproduce.

**redis-py version is decisive:**

| redis-py | `PubSub._lock` | Result |
|---|---|---|
| 5.3.1 – 6.3.0 | `threading.Lock()` | **deadlocks** |
| 6.4.0 | `threading.RLock()` | does not deadlock |

redis-py 6.4.0 changed `PubSub._lock` to a reentrant lock, which masks this. On any earlier redis-py, Celery self-deadlocks.

</p>
</details>

<details>
<p>

```python
"""Celery deadlock: AsyncResult.__del__ re-enters the Redis pubsub lock during a publish.

    REDIS_URL=redis://localhost:6379/0 python celery_mre.py

Requires only a Redis server; no worker. Exits non-zero and dumps the deadlocked
stack if reproduced.
"""

import faulthandler
import gc
import os
import signal
import sys

import redis
import redis.connection
from celery import Celery

REDIS = os.environ.get("REDIS_URL", "redis://localhost:6379/0")
HANG_SECONDS = 15

app = Celery("mre", broker=REDIS, backend=REDIS)


@app.task
def noop(x):
    return x


def _on_alarm(signum, frame):
    print(f"\n=== REPRODUCED: publish did not return within {HANG_SECONDS}s ===\n", flush=True)
    faulthandler.dump_traceback()
    sys.exit(1)


def main():
    import celery
    print(f"celery={celery.__version__} redis-py={redis.__version__}", flush=True)

    # Keep the finalizable result alive until we collect explicitly, below.
    gc.disable()

    result = noop.apply_async(args=(1,))

    # What AsyncResult.get() does internally: start the ResultConsumer and SUBSCRIBE.
    app.backend.add_pending_result(result)

    # Fulfilment drops the backend's strong reference. In real code this happens when
    # the result arrives and .get() returns; the AsyncResult is then reachable only
    # from user code, and its __del__ will call remove_pending_result a second time.
    app.backend.remove_pending_result(result)

    # A reference cycle, so refcounting cannot free it at a safe point and
    # finalization is deferred to the cyclic collector.
    cycle = {"result": result}
    cycle["self"] = cycle
    del result, cycle

    # Simulate a collection landing inside the socket write, while PubSub._lock is held.
    # This happens naturally whenever a GC lands in that window.
    original_send = redis.connection.AbstractConnection.send_packed_command

    def send_packed_command(self, *args, **kwargs):
        gc.collect()
        return original_send(self, *args, **kwargs)

    redis.connection.AbstractConnection.send_packed_command = send_packed_command

    signal.signal(signal.SIGALRM, _on_alarm)
    signal.alarm(HANG_SECONDS)

    print("publishing (deadlocks here)...", flush=True)
    noop.apply_async(args=(2,))

    signal.alarm(0)
    print("published OK -- not reproduced", flush=True)


if __name__ == "__main__":
    main()
```

</p>
</details>

# Expected Behavior

`apply_async()` publishes the task and returns, regardless of whether a garbage collection happens to run during the call.

More generally: `AsyncResult.__del__` should not perform blocking network I/O. A finalizer can run at any allocation point, including inside a critical section that already holds the lock the finalizer will need.

# Actual Behavior

`apply_async()` never returns. The process hangs forever with no error, no timeout and no log output.

`RedisBackend.on_task_call()` runs on every `send_task()` and issues a PUBSUB `SUBSCRIBE`. redis-py's `PubSub.execute_command()` holds `self._lock` across the socket write. If a garbage collection lands inside that write and finalizes an `AsyncResult`, the finalizer re-acquires the same lock on the same thread:

```
apply_async → send_task
  → on_task_call (celery/backends/redis.py)
    → consume_from → _consume_from → subscribe (redis/client.py)
      → execute_command → with self._lock:        ← ACQUIRED
        → send_command → send_packed_command → sendall
          → [garbage collection]
            → AsyncResult.__del__ (celery/result.py)
              → backend.remove_pending_result (celery/backends/asynchronous.py)
                → on_result_fulfilled
                  → ResultConsumer.cancel_for (celery/backends/redis.py)
                    → pubsub.unsubscribe (redis/client.py)
                      → execute_command → with self._lock:   ← SAME THREAD → DEADLOCK
```

Stack from the reproduction script (`faulthandler`, most recent call first):

```
Current thread (most recent call first):
  Garbage-collecting
  File ".../redis/client.py", line 858 in execute_command
  File ".../redis/client.py", line 1043 in unsubscribe
  File ".../celery/backends/redis.py", line 187 in cancel_for
  File ".../celery/backends/asynchronous.py", line 261 in on_result_fulfilled
  File ".../celery/backends/asynchronous.py", line 253 in remove_pending_result
  File ".../celery/result.py", line 419 in __del__
  File ".../redis/connection.py", line 577 in send_command
  File ".../redis/client.py", line 903 in <lambda>
  File ".../redis/retry.py", line 62 in call_with_retry
  File ".../redis/client.py", line 902 in _execute
  File ".../redis/client.py", line 859 in execute_command
  File ".../redis/client.py", line 1018 in subscribe
  File ".../celery/backends/redis.py", line 180 in _consume_from
  File ".../celery/backends/redis.py", line 173 in consume_from
  File ".../celery/backends/redis.py", line 417 in on_task_call
  File ".../celery/app/base.py", line 946 in send_task
  File ".../celery/app/task.py", line 608 in apply_async
```

### This is not synthetic — it happens in production

We hit this five times over six days in production (Celery 5.6.2, redis-py 5.3.1, Python 3.11.15, Linux aarch64, AWS ElastiCache over TLS). Each time a Prefect flow process hung permanently at `group(...).apply_async()`; the work was never dispatched and the process sat in `futex_wait_queue` indefinitely. Below is a stack captured with `py-spy` from a **live production process that had been deadlocked for five days** — the same interleaving:

```
_work (our code)                                       ← group(...).apply_async()
  apply_async (celery/canvas.py:1612) → _apply_tasks → chain → Signature
    send_task (celery/app/base.py:946)
      on_task_call (celery/backends/redis.py:417)
        consume_from → subscribe (redis/client.py:1018)
          send_packed_command (redis/connection.py:555)
            sendall (ssl.py:1270)                      ← mid-write, holding PubSub._lock
              __del__ (celery/result.py:419)
                remove_pending_result (asynchronous.py:253)
                  on_result_fulfilled (asynchronous.py:261)
                    cancel_for (celery/backends/redis.py:187)
                      unsubscribe (redis/client.py:1043)
                        execute_command (redis/client.py:858)   ← blocked forever
```

`py-spy --locals` confirmed the finalized object was an `AsyncResult` for a task whose result had already been consumed and whose reference had gone out of scope.

It is much easier to hit over TLS (`rediss://`), where the write is slower and allocates more, widening the window for a collection to land inside it.

### Notes on the mechanism

Two details make this reachable in ordinary code:

1. `on_task_call()` subscribes on **every** `send_task()`, so every publish performs pub/sub I/O while holding the lock. The dangerous window is not narrow — it is entered on every task dispatch.
2. `cancel_for()` unsubscribes without checking whether the key is still in `subscribed_to`. It `discard()`s the key and then calls `unsubscribe()` regardless, so a second `remove_pending_result` for an already-fulfilled result still performs a network round-trip. That second call is the one that lands inside the publish.

The application code that triggers this is unremarkable — publish a task, `get()` its result, drop the reference, publish another task:

```python
r = some_task.apply_async()
r.get()
del r                       # or simply let it go out of scope
other_task.apply_async()    # ← may deadlock here
```

### Suggested fix

Do not perform network I/O in `AsyncResult.__del__`. Queue the unsubscribe for the drainer/result-consumer thread instead of issuing it inline from a finalizer — which is also where #6963 suggests draining should be happening.

redis-py 6.4.0 changing `PubSub._lock` to an `RLock` prevents the hang, but that looks like mitigation rather than a fix: a reentrant `UNSUBSCRIBE` issued in the middle of a partially written `SUBSCRIBE` can still interleave bytes on the wire.

waterWang:
## Description

When a result is fulfilled, `on_state_change` calls `cancel_for` which removes the key from `subscribed_to` and sends an `unsubscribe`. If the `AsyncResult` is later garbage-collected during another `subscribe` call (from `on_task_call`), its `__del__` calls `remove_pending_result` → `on_result_fulfilled` → `cancel_for` again. Since the key was already removed from `subscribed_to`, the second `cancel_for` still sends an `unsubscribe`, which re-enters `PubSub._lock` on the same thread that holds it for the `subscribe` write — causing a **deadlock**.

## Fix

Make `cancel_for` idempotent: check that the key is still in `subscribed_to` before performing the `unsubscribe` network call. If it's not there, the subscription was already cancelled and we can return early without network I/O.

This is a minimal, safe fix that directly addresses the deadlock at its source: the redundant `unsubscribe` call from `__del__` for an already-fulfilled result.

## Related

- Closes #10477
- The MRE in the issue reproduces the deadlock deterministically by triggering GC inside a `send_packed_command` call.

## Testing

- Added `test_cancel_for_idempotent` which verifies that calling `cancel_for` twice on the same task only calls `unsubscribe` once
- All existing tests in `test_redis.py` pass via the mock-based test pattern

### DIFF
--- a/celery/backends/redis.py
+++ b/celery/backends/redis.py
@@ -195,6 +195,8 @@ class ResultConsumer(BaseResultConsumer):
 
     def cancel_for(self, task_id):
         key = self._get_key_for_task(task_id)
+        if key not in self.subscribed_to:
+            return
         self.subscribed_to.discard(key)
         if self._pubsub:
             with self.reconnect_on_error():
--- a/t/unit/backends/test_redis.py
+++ b/t/unit/backends/test_redis.py
@@ -233,8 +233,22 @@ class test_RedisResultConsumer:
         consumer.on_after_fork()
         parent_method.assert_called_once()
 
+    def test_cancel_for_idempotent(self):
+        """cancel_for should be idempotent to prevent re-entrant PubSub._lock
+        deadlock when AsyncResult.__del__ calls remove_pending_result on an
+        already-fulfilled result (issue #10477)."""
+        consumer = self.get_consumer()
+        consumer.start('initial')
+        consumer.consume_from('some-task')
+        # First call: should unsubscribe
+        consumer.cancel_for('some-task')
+        assert consumer._pubsub.unsubscribe.call_count == 1
+        # Second call: key already removed from subscribed_to, should skip
+        consumer.cancel_for('some-task')
+        assert consumer._pubsub.unsubscribe.call_count == 1
+
     @patch('celery.backends.redis.ResultConsumer.cancel_for')
-    @patch('celery.backends.asynchronous.BaseResultConsumer.on_state_change')
+
     def test_on_state_change(self, parent_method, cancel_for):
         consumer = self.get_consumer()
         meta = {'task_id': 'testing', 'status': states.SUCCESS}


auvipy:
@Namraa310806

auvipy:
closing a tests are failing

Issue #9064, opened 2024-06-05
Logging system exception when celery beat can not connect to redis


undo-nothing:
I am setting Redis as broker, and setting broker_connection_max_retries = 1.

I start celery beat with `python -m celery -A proj beat -l debug --logfile="logging/beat.log"`, and not running Redis service. Celery beat will check redis connection, when over max_retries, will raise Exception, I think the log output seems duplicate and format error.

some log in output:

[2024-06-05 11:07:08,121: CRITICAL beat.py beat.115] beat raised exception <class 'kombu.exceptions.OperationalError'>: OperationalError('Error 10061 connecting to 127.0.0.1:6379. 由于目标计算机积极拒绝，无法连接。.')
Traceback (most recent call last):
  File "C:\Miniforge3\envs\python311_django42\Lib\site-packages\kombu\transport\virtual\base.py", line 951, in create_channel
    return self._avail_channels.pop()
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
IndexError: pop from empty list

During handling of the above exception, another exception occurred:
...

[2024-06-05 11:07:08,186: WARNING log.py log.232] Traceback (most recent call last):
[2024-06-05 11:07:08,186: WARNING log.py log.232]   File "C:\Miniforge3\envs\python311_django42\Lib\site-packages\kombu\transport\virtual\base.py", line 951, in create_channel
[2024-06-05 11:07:08,187: WARNING log.py log.232]     
[2024-06-05 11:07:08,188: WARNING log.py log.232] return self._avail_channels.pop()
[2024-06-05 11:07:08,188: WARNING log.py log.232]  
[2024-06-05 11:07:08,188: WARNING log.py log.232]  
[2024-06-05 11:07:08,188: WARNING log.py log.232]  
[2024-06-05 11:07:08,188: WARNING log.py log.232]  
[2024-06-05 11:07:08,188: WARNING log.py log.232]  
[2024-06-05 11:07:08,188: WARNING log.py log.232]  
[2024-06-05 11:07:08,189: WARNING log.py log.232]  
[2024-06-05 11:07:08,189: WARNING log.py log.232]  
[2024-06-05 11:07:08,189: WARNING log.py log.232]  
[2024-06-05 11:07:08,189: WARNING log.py log.232]  
[2024-06-05 11:07:08,189: WARNING log.py log.232]  
[2024-06-05 11:07:08,189: WARNING log.py log.232] ^
[...]

PR #10492, opened 2026-08-16
Line buffer LoggingProxy writes

rahul2002m:
`LoggingProxy.write()` logs every call as its own record. When a caller writes
one line in several pieces, each piece becomes a separate log record with its own
timestamp and level prefix. That's the mangled output in #9064:

```python
for piece in ['    ', 'return self._avail_channels.pop()', '\n',
              '    ', ' ', ' ', ' ', '^', '^', '^', '^', '^', '\n']:
    proxy.write(piece)
```

on master gives 11 records, five of them a single `^` and four just whitespace:

```
0: '    '
1: 'return self._avail_channels.pop()'
2: '    '
3: ' '
...
6: '^'
7: '^'
```

and with this change, 2:

```
0: '    return self._avail_channels.pop()'
1: '       ^^^^^'
```

So `write()` now buffers and logs once a trailing newline arrives, on `flush()`
or on `close()`.

**Where I have to be honest about scope:** this fixes the mechanism, but it will
not change the log in the original report on current Python. The reporter was on
3.11.4, where `traceback` wrote the anchor line in pieces. On 3.12 and 3.13 every
chunk `traceback` writes already ends in a newline (I checked: 8 writes, 0 of
them mid-line), so there is nothing left for buffering to join. I could not
reproduce their fragmented output on 3.13 at all — the beat log is byte-identical
before and after.

What *is* still reproducible from that report on 3.13 is the other half: the
traceback appears twice. `Beat.start_scheduler()` logs it with
`logger.critical(..., exc_info=True)` and then re-raises, so the interpreter
prints it again through the redirected stderr — 58 `WARNING/MainProcess` records,
one per frame. Buffering can't merge those, because each frame arrives
newline-terminated and merging on `flush()` instead would hold ordinary
`print()` output until interpreter shutdown. Fixing the duplicate means not
sending an already-logged traceback through the redirected stderr, which is a
beat/CLI change and a bigger call than I wanted to make unasked. Happy to do it
as a follow-up if you want it — say which shape you'd prefer.

So: `Refs #9064` rather than `Closes`.

Details worth a look:

- A write containing several complete lines is still one record, so multi-line
  output is unchanged. There's a test pinning that.
- The buffer is a `threading.local()` owned by the instance, not the shared
  `_thread` class attribute, so the stdout and stderr proxies can't interleave
  into one message, and neither can two threads.
- `close()` flushes first, so a line without its newline isn't dropped. The
  interpreter also calls `flush()` on `sys.stdout`/`sys.stderr` at shutdown — I
  verified that with a stub stream — so nothing is lost at exit either.
- `_in_sighandler` writes still bypass the buffer and go straight to
  `sys.__stderr__`.
- If `_flush_buffer` is reached while the logger is logging back into the proxy,
  the buffer is left intact for a later flush rather than dropped.
- One behaviour change beyond the buffering: `write()` returns the length of the
  data it accepted rather than the length after stripping trailing newlines, so
  `write('\n')` returns 1 where it used to return 0. That matches what a file
  object should return, but it is a change.

I did have to update two assertions in `test_logging_proxy` and
`test_logging_proxy_bytes` — `p.write('foo ')` no longer logs immediately, it
logs on the following `flush()`. I didn't want to quietly reword an existing
test, so flagging it: that's the assertion that encodes the old "not buffered"
contract, and it's the thing to push back on if you'd rather keep it.

Tests: 7 new ones covering joined partial writes, embedded newlines staying in
one record, flush on close, whitespace-only flushes logging nothing, writes after
close being dropped, and per-instance and per-thread buffer isolation. Four of
them fail on master.

Full unit suite: `3738 passed, 39 skipped, 3 xfailed, 28817 subtests` — no
failures. `pre-commit` clean on both files, including mypy.


rahul2002m:
@auvipy Can you review this?

rahul2002m:
@auvipy Can we merge this?

auvipy:
@Namraa310806 would you mind try and review it?

Namraa310806:
@rahul2002m
 **Non-blocking: buffered data can be stranded across proxies sharing a thread**

`_flush_buffer()` is designed so that if recursion is detected, it leaves the buffer alone for a later flush to retry:

```python
if getattr(self._thread, 'recurse_protection', False):
    # Logger is logging back to this file. Leave the buffer alone so
    # that a later flush can still log it.
    return
```

However, `_thread` is the class-level `threading.local()` (`_thread = threading.local()`), so it is shared by every `LoggingProxy` instance on a given thread. This differs from the new `_local` buffer, which is per-instance. As a result, `recurse_protection` indicates that *some* proxy on the thread is currently flushing, rather than that this particular proxy is flushing.

For example, if proxy B is inside `logger.log()` and has set the flag, and proxy A's `close()` is invoked reentrantly on the same thread while B is still on the stack, A's `_flush_buffer()` will see the flag and skip flushing A's buffer. `close()` then still sets `A.closed = True`, leaving A's pending buffer with no opportunity for the intended "later flush".

A minimal reproduction would be:

```python
class ClosesOtherProxy(logging.Handler):
    def __init__(self, other_proxy):
        super().__init__()
        self.other_proxy = other_proxy

    def emit(self, record):
        self.other_proxy.close()


proxy_a = LoggingProxy(logger, loglevel=logging.ERROR)
proxy_b = LoggingProxy(logger, loglevel=logging.ERROR)
logger.addHandler(ClosesOtherProxy(proxy_b))

proxy_b.write('pending, no newline')   # buffered
proxy_a.write('trigger\n')             # flush -> logger.log() -> handler -> proxy_b.close()

assert proxy_b.closed
# proxy_b's pending line can no longer be flushed
```

Since `recurse_protection` is intentionally thread-wide rather than instance-wide, would it make sense for `close()` to handle the case where its initial flush was skipped because another proxy is currently inside a recursive logging path, so that the pending data is not silently stranded?

Namraa310806:
@rahul2002m 
 `close()` may strand buffered data from another thread

I noticed that `_local` is per-thread, so `self._buffer` only refers to the buffer belonging to the thread calling `close()`. If another thread has a partial line buffered when `close()` runs, `close()` can't flush or rescue that buffer. The stderr fallback has the same limitation since it also only sees the caller's thread-local buffer.

For example, if thread A writes a partial line and thread B calls `close()` on the same proxy, thread A's thread-local buffer remains pending but the proxy is now marked closed. If thread A later calls `flush()`, `_flush_buffer()` clears the buffer but returns because `self.closed` is already true, so the buffered data is lost.

Is there a lifecycle guarantee that all threads using the proxy have finished before `close()` is called? If not, could this leave a thread's buffered data stranded?

Namraa310806:
@rahul2002m This is the last thing from my side. Everything else looks good to me. Happy to take another look if the next commit introduces any changes that need review.

rahul2002m:
@Namraa310806 Good catch, and you're right about the mechanics. `_local` is per-thread, so `close()` only ever sees the calling thread's buffer.

I don't think it's worth locking, though. Rescuing another thread's partial line means a registry of every thread's buffer plus a lock around it, and celery never calls `close()` on the proxy at all: it's only installed as `sys.stdout`/`sys.stderr` in `redirect_stdouts_to_logger` (`celery/app/log.py:199`) and left there. Closing any file object while other threads are still writing to it drops in-flight data for real `io` objects too, so this matches the semantics callers already have. Pre-PR, a write racing `close()` was silently dropped as well, so nothing regresses.

I've documented the limitation on `close()` instead of adding the machinery. Thanks for both reviews. The stderr fallback from your first comment is the part that genuinely needed fixing.

Namraa310806:
@rahul2002m Are all the test cases passing on your side? I’m getting 2 failures locally

rahul2002m:
> @rahul2002m Are all the test cases passing on your side? I’m getting 2 failures locally

All cases are passed locally on my machine. And even the check as successful.

auvipy:
@rahul2002m can we also explore if integration tests can be added to better cover this? or any other tests? also if documents can be improved more?

auvipy:
@MehrazRumman please try and review this

MehrazRumman:
ok !

MehrazRumman (changes requested):
I checked out the branch and ran the tests, and I probed the buffering against how the proxy is actually installed in the worker. One finding I think is worth addressing, plus answers to the two open threads.

## On the disputed test failures

@Namraa310806 I could not reproduce your 2 failures. On Python 3.14.5:

- `t/unit/app/test_log.py` in isolation: **65 passed**
- `t/unit/app` together: **694 passed, 11 skipped**
- full `t/unit`: **3860 passed, 36 skipped, 3 xfailed**

So @rahul2002m's claim holds on my machine. Could you post your Python version and the actual failure output? Two candidates for an environment-dependent failure: `test_logging_proxy_close_under_recursion_falls_back_to_stderr` uses `capfd`, which is sensitive to capture settings such as `-s`, and `LoggingProxy._thread` is a **class** attribute shared by every instance, so a test that leaves `recurse_protection` set would silently block later flushes. Both would show up under some run configurations and not others.

## Main finding: nothing ever flushes this proxy

The PR description says the interpreter flushes `sys.stdout`/`sys.stderr` at shutdown, which is true, but that turns out to be the *only* flush that ever happens. `redirect_stdouts_to_logger` installs the proxy and nothing in Celery calls `flush()` or `close()` on it again.

So output with no trailing newline is now held indefinitely rather than logged:

```
1) print() with a newline (the common case):
   logged: 'task done\n'

2) print(..., end='') / sys.stdout.write() with NO newline:
   logged: ''            <- invisible until something flushes

3) after 1000 more partial writes:
   logged: ''
   buffer entries held in memory: 1001
```

Two consequences. A task doing `print('Progress: ', end='')` or a bare `sys.stdout.write()` produces log output today and produces nothing after this change until the worker exits, which for a long-running worker can be days. And the buffer is an unbounded list, so a loop writing without newlines grows memory with no ceiling.

The common case, `print()` with its default newline, is unaffected, and I do think line buffering is the right call overall. But "held until interpreter shutdown" is a stronger commitment than the description implies.

Suggestion: cap the buffer and force a flush when the accumulated length crosses a threshold. That bounds the memory, guarantees the output eventually appears, and is what a real line-buffered stream does when its buffer fills. It is a small addition to `write()` and keeps everything else as-is.

## Answering @auvipy's two open asks

**Docs.** The `worker_redirect_stdouts` section in `docs/userguide/configuration.rst` currently says only that stdout and stderr are redirected to the current logger. It says nothing about buffering. Since this PR changes when output appears, that section is the place to say output is line buffered and a partial line is held until a newline arrives. Worth a Changelog entry too, because "my `end=''` output stopped showing up" is otherwise a hard symptom to trace back.

**Integration test.** A useful one would be a task that writes a line in pieces through a real worker with `worker_redirect_stdouts` on, asserting the log holds one record rather than several. That covers the `redirect_stdouts_to_logger` wiring, which the unit tests construct a `LoggingProxy` directly and therefore skip.

## Smaller notes

1. The `write()` return-value change is called out honestly in the description and I agree it is the correct file-object semantics. It deserves a Changelog line, since a caller checking the return of `write('\n')` sees 1 where it saw 0.

2. Buffering changes relative ordering between the stdout and stderr proxies. A partial line on one appears after a later complete line on the other. Inherent to the approach and not worth fixing, but worth a sentence in the docstring.

3. Flagging the two rewritten assertions in `test_logging_proxy` and `test_logging_proxy_bytes` rather than quietly changing them was the right thing to do. They did encode the old contract, and rewriting them is correct here.

4. @Namraa310806's point about `close()` only seeing the calling thread's buffer is right, and I agree with @rahul2002m that a thread registry plus a lock is not worth it given Celery never calls `close()`. The current comment already says so.

The scope discipline in the description is good, separating the mechanism fix from the duplicate-traceback problem and marking this `Refs` rather than `Closes`. Address the unbounded buffer and the docs note and I think this is ready.

PR #10493, opened 2026-08-16
Fix ineffective traceback clearing in timeout handling

### LINKED ISSUE #10489: Memory-leak fix (traceback_clear) silently fails in Request.on_timeout()
# Checklist

- [x] This has already been asked to the discussions forum first.
- [x] I have read the relevant section in the contribution guide on reporting bugs.
- [x] I have checked the issues list for similar or identical bug reports.
- [x] I have checked the pull requests list for existing proposed fixes.
- [x] I have checked the commit log to find out if the bug was already fixed in the main branch.
- [x] I have included all related issues and possible duplicate issues in this issue (If there are none, check this box anyway).
- [x] I have tried to reproduce the issue with pytest-celery and added the reproduction script below.

- [x] I have included all the versions of all the external dependencies required to reproduce this bug.

- [x] I have tried reproducing the issue on more than one Python version and/or implementation.
- [ ] I have tried reproducing the issue on more than one message broker and/or result backend.
- [ ] I have tried reproducing the issue on more than one version of the message broker and/or result backend.
- [ ] I have tried reproducing the issue with autoscaling, retries, ETA/Countdown & rate limits disabled.
- [ ] I have tried reproducing the issue after downgrading and/or upgrading Celery and its dependencies.

#### Related Issues

- The `traceback_clear()` helper this bug concerns was introduced in `celery/app/trace.py` as a memory-leak fix (referenced in code as issue #8882). This report concerns a separate call site of that same helper, added later in `celery/worker/request.py`'s `Request.on_timeout()`, which does not exhibit the same behavior as the original fix.

#### Possible Duplicates

- None

## Environment & Settings

**Celery version**: 5.6.2 (`main` branch, as of this report)

<details>
<p>

```
software -> celery:5.6.2 (main) kombu:5.x py:3.12.x
billiard:4.x
platform -> system:Linux
```

</p>
</details>

# Steps to Reproduce

## Required Dependencies

- **Minimal Python Version**: Any (reproduced on 3.12; not version-specific — behavior is a documented CPython frame API restriction)
- **Minimal Celery Version**: 5.6.x (`main`) — anywhere `traceback_clear()` is called from `Request.on_timeout()`
- **Minimal Broker Version**: N/A or Unknown
- **Minimal Result Backend Version**: N/A or Unknown
- **Minimal OS and/or Kernel Version**: N/A or Unknown

### Python Packages

<details>
<p>

```
celery==5.6.2
billiard==4.x
kombu==5.x
vine==5.x
```

</p>
</details>

### Other Dependencies

<details>
<p>
N/A — this is a pure-Python/CPython behavior issue, not dependent on broker, backend, or worker pool.
</p>
</details>

The following script reproduces the exact structure used in `Request.on_timeout()`'s hard-time-limit branch (`celery/worker/request.py`) and demonstrates that `traceback_clear()` (`celery/app/trace.py`) never succeeds in clearing any frame when called this way, because the only frame in the synthetic traceback is the currently-executing frame itself, which CPython refuses to clear.

<details>
<p>

```python
"""
Reproduces the exact pattern used in celery.worker.request.Request.on_timeout()
for the hard time-limit branch:

    try:
        try:
            raise exc
        except TimeLimitExceeded:
            einfo = ExceptionInfo()
        ...
        traceback_clear(exc)   # <-- intended to free frame locals
    finally:
        exc.__traceback__ = None

This shows traceback_clear() can never actually clear a frame in this call site.
"""

from celery.app.trace import traceback_clear


def simulate_on_timeout():
    exc = ValueError("TimeLimitExceeded (simulated)")

    try:
        try:
            raise exc
        except ValueError:
            pass  # equivalent to einfo = ExceptionInfo()

        # This mirrors the real call: traceback_clear(exc)
        # Patch tb_frame.clear to observe whether it ever succeeds.
        tb = exc.__traceback__
        cleared_any = False
        while tb is not None:
            try:
                tb.tb_frame.clear()
                cleared_any = True
            except RuntimeError as e:
                print(f"frame.clear() raised (swallowed in real code): {e}")
            tb = tb.tb_next

        assert cleared_any is False, (
            "Expected traceback_clear() to be a no-op here, but a frame was cleared."
        )
        print("Confirmed: traceback_clear() cleared ZERO frames.")
    finally:
        exc.__traceback__ = None


simulate_on_timeout()
```

</p>
</details>

Actual `celery/app/trace.py::traceback_clear` swallows the `RuntimeError` internally (`except RuntimeError: pass`), so in production this failure is completely silent — no log line, no error, nothing. It can only be observed by instrumenting the function directly, as above.

# Expected Behavior

The comment above the `traceback_clear(exc)` call in `Request.on_timeout()` states:

```python
# MEMORY LEAK FIX: clear frame locals retained by the
# synthetic traceback (same pattern as trace.py #8882).
traceback_clear(exc)
```

This implies the call clears frame-local references (e.g. to `self`, the `Request` object, and other locals held by the frame) from the traceback attached to `exc`, mirroring the effective leak fix already in place in `celery/app/trace.py`'s main tracer. This should measurably release references to large/circular objects held by that frame.

# Actual Behavior

`traceback_clear(exc)` never clears anything in this call site, on every single invocation.

**Root cause:** In `Request.on_timeout()`, `exc` is raised and caught within the same function (`raise exc` / `except TimeLimitExceeded:`), producing a synthetic, single-frame traceback whose only frame is `on_timeout()`'s own frame. Because `on_timeout()` has not yet returned when `traceback_clear(exc)` is called, that frame is still "executing" from CPython's point of view. `frame.clear()` unconditionally raises:

```
RuntimeError: cannot clear an executing frame
```

for any frame still on the call stack. `traceback_clear()` catches this `RuntimeError` and silently ignores it (`except RuntimeError: pass`), so the call appears to succeed but performs no work.

This differs from the original leak fix in `celery/app/trace.py` (issue #8882), where `traceback_clear()` is invoked from a context where the traceback includes frames from the task body that have already unwound (are no longer executing), so `frame.clear()` can succeed there.

**Practical impact:**
- The call is dead code for this specific call site — it always fails and always silently swallows the failure, on every hard-timeout event.
- It does real (if small) unnecessary work (walking the traceback chain, attempting and catching a `RuntimeError`) on a code path that already runs during worker-level failure handling.
- The comment is misleading to future maintainers/reviewers, who would reasonably believe this line is functioning as a memory-leak mitigation when it structurally cannot.
- Separately, `exc.__traceback__ = None` in the enclosing `finally` block does appear to correctly break the `frame → exc → traceback → frame` reference cycle on its own, so there is likely no live leak today — but that means `traceback_clear(exc)` is fully redundant here, and if that `finally` line is ever refactored or removed independently (a realistic risk given the misleading comment implies `traceback_clear` is the actual fix), the cycle-breaking behavior would silently regress with no test coverage catching it.

**Suggested fix:** either remove the redundant `traceback_clear(exc)` call at this call site (since `exc.__traceback__ = None` already accomplishes the intended reference-breaking), or restructure so the traceback-frame clearing happens only after `on_timeout()`'s own frame is no longer on the call stack (e.g. clear only `tb.tb_next` frames onward, skipping the frame that raised `exc`).

Namraa310806:
## Description

This PR fixes an ineffective `traceback_clear(exc)` call in `Request.on_timeout()`'s hard-timeout path.

The exception is raised and caught within `on_timeout()` itself, so its synthetic traceback contains the currently executing `on_timeout()` frame. Since an executing frame cannot be cleared by `frame.clear()`, `traceback_clear(exc)` always silently fails at this call site.

This change:
- Removes the redundant `traceback_clear(exc)` call and its unused import.
- Keeps the existing `exc.__traceback__ = None` cleanup, which breaks the `exc → traceback → frame` reference cycle.
- Adds a regression test verifying that the traceback is cleared after a hard timeout while preserving the existing failure-handling behavior.

Fixes #10489

### DIFF
--- a/celery/worker/request.py
+++ b/celery/worker/request.py
@@ -16,7 +16,7 @@ from kombu.utils.objects import cached_property
 
 from celery import current_app, signals, states
 from celery.app.task import Context
-from celery.app.trace import fast_trace_task, task_has_custom, trace_task, trace_task_ret, traceback_clear
+from celery.app.trace import fast_trace_task, task_has_custom, trace_task, trace_task_ret
 from celery.concurrency.base import BasePool
 from celery.exceptions import (Ignore, InvalidTaskError, Reject, Retry, TaskRevokedError, Terminated,
                                TimeLimitExceeded, WorkerLostError)
@@ -576,13 +576,23 @@ class Request:
                         exception=safe_repr(get_pickled_exception(einfo.exception)),
                         traceback=einfo.traceback,
                     )
-                    # MEMORY LEAK FIX: clear frame locals retained by the
-                    # synthetic traceback (same pattern as trace.py #8882).
-                    traceback_clear(exc)
                 finally:
-                    # Break the remaining exc → traceback → frame reference
-                    # cycle so the on_timeout frame (and the Request/self it
-                    # contains) can be garbage-collected promptly.
+                    # Break the exc → traceback → frame reference cycle
+                    # (this frame's own `exc` local points at exc, and
+                    # exc.__traceback__ points back at this frame) so the
+                    # on_timeout frame — and the Request/self it holds —
+                    # can be garbage-collected promptly.
+                    #
+                    # Note: a `traceback_clear(exc)` call previously lived
+                    # here too, but it is a guaranteed no-op in this
+                    # location: exc is raised and caught within this same
+                    # function, so the only frame in exc.__traceback__ is
+                    # this currently-executing frame, and CPython's
+                    # frame.clear() always raises "cannot clear an
+                    # executing frame" for a frame still on the call
+                    # stack. That RuntimeError was being silently
+                    # swallowed inside traceback_clear(), so the call
+                    # appeared to succeed while doing nothing.
                     if einfo is not None:
                         del einfo
                     exc.__traceback__ = None
--- a/t/smoke/tests/test_tasks.py
+++ b/t/smoke/tests/test_tasks.py
@@ -8,7 +8,8 @@ from celery import Celery, signature
 from celery.exceptions import SoftTimeLimitExceeded, TimeLimitExceeded, WorkerLostError
 from t.integration.tasks import add, identity
 from t.smoke.conftest import SuiteOperations, TaskTermination
-from t.smoke.tasks import (replace_with_task, soft_time_limit_lower_than_time_limit,
+from t.smoke.tasks import (noop, replace_with_task, self_termination_delay_timeout,
+                           soft_time_limit_lower_than_time_limit,
                            soft_time_limit_must_exceed_time_limit)
 
 
@@ -144,3 +145,18 @@ class test_time_limit:
         sig = soft_time_limit_must_exceed_time_limit.s()
         with pytest.raises(ValueError, match="soft_time_limit must be less than or equal to time_limit"):
             sig.apply_async(queue=celery_setup.worker.worker_queue)
+
+    def test_hard_timeout_worker_remains_functional(self, celery_setup: CeleryTestSetup):
+        """Verify the worker remains functional after a hard timeout."""
+        timeout_result = self_termination_delay_timeout.apply_async(
+            queue=celery_setup.worker.worker_queue,
+        )
+
+        with pytest.raises(TimeLimitExceeded):
+            timeout_result.get(timeout=RESULT_TIMEOUT)
+
+        result = noop.apply_async(
+            queue=celery_setup.worker.worker_queue,
+        )
+
+        assert result.get(timeout=RESULT_TIMEOUT) is None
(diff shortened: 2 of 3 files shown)


auvipy:
should we explore the possibility of adding smoke test for the proposed change?

Namraa310806:
@auvipy Thanks for the suggestion. I added a focused smoke test covering the real hard-timeout path.

The test reuses the existing `self_termination_delay_timeout` task, verifies the hard timeout is handled, and then submits a `noop` task to the same worker to confirm that the worker remains functional afterward.

I also kept the existing unit regression test that verifies `exc.__traceback__` is cleared. The smoke test could not be run locally because the Docker/pytest-celery smoke-test infrastructure is not available in my environment, so I’ve left the full smoke-test verification to CI.

auvipy (approved):

PR #10494, opened 2026-08-16
 fix: crontab.remaining_delta skips later minute slots on an earlier day

phanky1:
## Description

`crontab.remaining_delta` expects to find the next run after `last_run_at`, even if that run is in the past. but the current implementation only looks for a later minute inside last_run_at's hour when last_run_at falls on today's date, so on any earlier date that slot is skipped and the answer jumps forward.
https://github.com/celery/celery/blob/2f3cd2f1ffa1921650b57a9e0341aa8d94ab13b2/celery/schedules.py#L580-L582


auvipy (commented):
which related issue?

phanky1:
> which related issue?

I couldn't find one, just came across this.

auvipy (approved):

PR #10508, opened 2026-08-22
fix(utils): prevent LimitedSet.purge()/pop() from evicting a refreshe…

### LINKED ISSUE #10506: `LimitedSet` can prematurely evict refreshed items due to stale heap entries
# Checklist

* [x] I have verified that the issue exists against the `main` branch of Celery.
* [ ] This has already been asked to the [[discussions forum](https://github.com/celery/celery/discussions)](https://github.com/celery/celery/discussions) first.
* [x] I have read the relevant section in the
  [[contribution guide](https://docs.celeryq.dev/en/main/contributing.html#other-bugs)](https://docs.celeryq.dev/en/main/contributing.html#other-bugs)
  on reporting bugs.
* [x] I have checked the [[issues list](https://github.com/celery/celery/issues?q=is%3Aissue+label%3A%22Issue+Type%3A+Bug+Report%22+-label%3A%22Category%3A+Documentation%22)](https://github.com/celery/celery/issues?q=is%3Aissue+label%3A%22Issue+Type%3A+Bug+Report%22+-label%3A%22Category%3A+Documentation%22)
* [x] I have checked the [[pull requests list](https://github.com/celery/celery/pulls?q=is%3Apr+label%3A%22PR+Type%3A+Bugfix%22+-label%3A%22Category%3A+Documentation%22)](https://github.com/celery/celery/pulls?q=is%3Apr+label%3A%22PR+Type%3A+Bugfix%22+-label%3A%22Category%3A+Documentation%22)
  for existing proposed fixes.
* [x] I have checked the [[commit log](https://github.com/celery/celery/commits/main)](https://github.com/celery/celery/commits/main)
* [ ] I have tried to reproduce the issue with [[pytest-celery](https://docs.celeryq.dev/projects/pytest-celery/en/latest/userguide/celery-bug-report.html)](https://docs.celeryq.dev/projects/pytest-celery/en/latest/userguide/celery-bug-report.html) and added the reproduction script below.

* [x] I have included the output of `celery -A proj report` in the issue.
  version affected).
* [x] I have verified that the issue exists against the `main` branch of Celery.
* [x] I have included the contents of `pip freeze` in the issue.
* [x] I have included all the versions of all the external dependencies required
  to reproduce this bug.

* [ ] I have tried reproducing the issue on more than one Python version
  and/or implementation.
* [ ] I have tried reproducing the issue on more than one message broker and/or
  result backend.
* [ ] I have tried reproducing the issue on more than one version of the message
  broker and/or result backend.
* [ ] I have tried reproducing the issue on more than one operating system.
* [ ] I have tried reproducing the issue on more than one workers pool.
* [ ] I have tried reproducing the issue with autoscaling, retries,
* [ ] I have tried reproducing the issue after downgrading

#### Related Issues

* None found.

#### Possible Duplicates

* None found.

## Environment & Settings

**Celery version**: `main` (reproduced from the current main branch)

<details>
<p>

```
PASTE THE OUTPUT OF:
celery -A proj report
```

</p>
</details>

# Steps to Reproduce

## Required Dependencies

* **Minimal Python Version**: Python version used for reproduction; exact minimum version not yet determined.
* **Minimal Celery Version**: `main` (exact minimum affected release not yet determined).
* **Minimal Kombu Version**: Not relevant to the isolated reproduction.
* **Minimal Broker Version**: Not relevant to the isolated reproduction.
* **Minimal Result Backend Version**: Not relevant to the isolated reproduction.
* **Minimal OS and/or Kernel Version**: Not determined.
* **Minimal Broker Client Version**: Not relevant to the isolated reproduction.
* **Minimal Result Backend Client Version**: Not relevant to the isolated reproduction.

### Python Packages

<details>
<p>

```
PASTE THE OUTPUT OF:
pip freeze
```

</p>
</details>

### Other Dependencies

<details>
<p>
The reproduction does not require a broker or result backend. It exercises
`celery.utils.collections.LimitedSet` directly.
</p>
</details>

The issue can be reproduced directly with `LimitedSet` by controlling its
clock and refreshing an existing item before its original expiration.

```python
from celery.utils.collections import LimitedSet


class Clock:
    def __init__(self, value=0):
        self.value = value

    def __call__(self):
        return self.value


clock = Clock()
s = LimitedSet(expires=10, clock=clock)

# Add an item at t=0.
s.add("task-id")

# Refresh the same item at t=5.
clock.value = 5
s.add("task-id")

# The current entry should now expire at t=15.
assert "task-id" in s

# The old heap entry expires at t=10.
clock.value = 10
s.purge()

# BUG: the stale heap entry causes the current entry to be removed.
assert "task-id" in s
```

On the vulnerable implementation, the final assertion fails.

The regression test also reproduces the issue against the original implementation
and passes after the fix.

# Expected Behavior

When an item is added to `LimitedSet` again, its expiration should be refreshed.

If an old heap entry for the same item later expires, that stale entry should not
cause the current, refreshed entry to be removed.

For example, if an item is added at `t=0` with `expires=10` and refreshed at
`t=5`, it should remain present until approximately `t=15`.

# Actual Behavior

`LimitedSet` keeps heap entries lazily when an existing item is refreshed.
The old heap entry remains in `_heap`, while `_data[item]` is replaced with the
newer entry.

When `purge()` encounters the old expired heap entry, `pop()` removes the item
from `_data` based only on the item key. It does not verify that the heap entry
being removed is still the current entry stored in `_data`.

As a result, an old expired heap entry can delete a newer, refreshed entry.

For example:

```text
t=0   add("task-id")
      heap = [(0, "task-id")]

t=5   add("task-id")
      heap = [(0, "task-id"), (5, "task-id")]
      current _data entry = (5, "task-id")

t=10  old entry expires
      purge()/pop() processes (0, "task-id")
      "task-id" is removed from _data

      BUG: the refreshed entry should remain valid until t=15
```

This is particularly relevant because `LimitedSet` is used by Celery's worker state for revoked task IDs. Re-adding or synchronizing an existing revoked task ID can refresh its expiration, but a stale heap entry can subsequently cause that ID to be removed before the refreshed expiration time.

The practical consequence is that a revoked task ID can stop being tracked as revoked earlier than its configured expiration time.

I reproduced this behavior against the original implementation, added regression tests that fail on the vulnerable implementation, and verified that the tests pass with the fix.

The existing test suite also passes with the fix.

Namraa310806:
## Description

Fixes #10506 

`LimitedSet` keeps expired entries in its heap for lazy cleanup. When an existing item is refreshed, its current entry in `_data` is replaced with a new entry, while the previous heap entry may remain in `_heap`.

Previously, `purge()` could use the stale heap entry's timestamp when checking expiration, and `pop()` could remove the current `_data` entry based only on the item key. As a result, an old expired heap entry could remove a refreshed item before its actual expiration time.

For example:

- An item is added at `t=1` with `expires=10`.
- The same item is refreshed at `t=6`.
- The old heap entry expires at `t=11`.
- Previously, purging at `t=11` could remove the item even though the refreshed entry should remain valid until `t=16`.

### Changes

- Update `LimitedSet.purge()` to skip stale heap entries before checking the expiration timestamp.
- Update `LimitedSet.pop()` to remove an item only when the popped heap entry matches the item's current entry in `_data`.
- Add regression tests covering:
  - A refreshed item surviving expiration of its stale heap entry.
  - Multiple refreshes creating multiple stale entries without causing premature eviction.
  - The refreshed entry still expiring at its correct expiration time.

The existing lazy heap design is preserved; stale entries are discarded during heap processing rather than requiring eager removal when an item is refreshed.

### Verification

- Regression tests pass.
- Existing `LimitedSet` tests pass.
- Relevant worker/state tests pass.
- Broader `t/unit/utils/` tests pass: `412 passed, 94 skipped, 28 xfailed`.


auvipy:
you can also check this pr https://github.com/celery/celery/pull/10507 and see if anything can be incorporated from it

Namraa310806:
Thanks for pointing me to #10507. I reviewed it and incorporated the relevant approach of validating stale heap entries in both `purge()` and `pop()` using entry identity.

I also kept the additional regression coverage for multiple stale entries and direct pop() handling, and added a test for the equal-timestamp refresh case to verify why identity comparison is required.

I’ve rerun the relevant tests and lint checks, and everything passes. Please let me know if there’s anything else from #10507 you’d like me to incorporate.

auvipy:
can you also explore the possibilities of adding smoke or integrations tests for this changes?

Namraa310806:
@auvipy Thanks for pointing me to #10507. I reviewed it and incorporated the relevant stale-entry handling in both `purge()` and `pop()`.

I also checked whether a smoke/integration test would be useful. The existing worker/revoke tests already cover the production path, and an additional test for this specific timing issue would be more fragile, so I kept the regression coverage focused on `LimitedSet`.

The relevant tests and lint checks are passing.

auvipy:
@phanky1 can you review this please?

celery/utils/collections.py
```diff
@@ -598,13 +612,15 @@ def purge(self, now=None):
...
-            try:
+            entry = heappop(self._heap)
+            _, item = entry
+            current_entry = self._data.get(item)
+            # Only delete from _data if this heap entry is still the current entry
+            # Stale heap entries (from item refresh) should be skipped
+            if current_entry is entry:
                 self._data.pop(item)
-            except KeyError:
-                pass
```
phanky1, line 605: Even though this exception is rare it's still possible, so I think we should keep this.

phanky1 (commented):
I have run the 6 new tests against the original codebase and only 2 of them failed, which makes me think something isn't right with the test setup. Can you check @Namraa310806?

mohithvardhan002 (approved):
Reviewed the LimitedSet stale-entry handling in purge() and pop(), along with the regression tests. The implementation correctly skips stale heap entries and only evicts the current _data entry. The tests also cover refreshes, multiple stale entries, pop(), and equal-timestamp entries.

phanky1:
> Reviewed the LimitedSet stale-entry handling in purge() and pop(), along with the regression tests. The implementation correctly skips stale heap entries and only evicts the current _data entry. The tests also cover refreshes, multiple stale entries, pop(), and equal-timestamp entries.

I think we should bring back the try/except in `pop()` since multiple threads can access the same LimitedSet.

celery/utils/collections.py
Namraa310806: Agreed — you're right that this needed the try/except back given _data can be touched concurrently. I've implemented the fix: self._data.pop(item) is now wrapped in try/except KeyError, and if the race is lost, pop() treats it as a stale entry and continues to the next heap entry instead of raising. I've also added a regression test (test_pop_handles_concurrent_removal_race) that simulates the exact race by making _data.pop() raise KeyError on the specific key being popped, and verifies pop() skips it cleanly and returns the next valid item. All LimitedSet and worker/state tests pass.

auvipy:
@MehrazRumman please have a round of review

MehrazRumman (commented):
I verified this by extracting `LimitedSet` from both `origin/main` and `c623aef` into standalone modules and running them side by side.

**The fix is correct.** The reported bug reproduces on main and is gone here:

```
main: add('x',1.0); add('x',6.0); purge(11.0)  ->  'x' in s = False   # wrong
pr:   add('x',1.0); add('x',6.0); purge(11.0)  ->  'x' in s = True    # right
```

The `is` check is sound because every path that writes `_data` keeps its values identical to the heap's objects — `_refresh_heap()` rebuilds from `self._data.values()`, `update(LimitedSet)` copies the same tuple objects then refreshes, and the dict/iterable branches go through `add()`. I fuzzed 8000 random op sequences (add/discard/pop/purge/update, random `maxlen`/`minlen`/`expires`) asserting after every operation that each `_data` value is present in `_heap` by identity: no violations, no exceptions, no hangs, on either version. `purge()` also stays amortized linear and leaves the heap *cleaner* than main, and a pickle round-trip preserves the invariant.

Two things worth noting in its favour: the `while/else` also removes a latent `IndexError` (on main, `self._heap[0]` with a non-empty `_data` and an empty `_heap` raises; here it exits cleanly), and the restored `try/except` in `pop()` addresses @phanky1's point.

**The tests, though, mostly do not guard the fix.** Running each added test body against both versions:

| test | on main | real regression test? |
|---|---|---|
| `test_stale_heap_entry_cannot_evict_refreshed_item` | fails | yes |
| `test_purge_with_only_stale_entries_in_heap` | fails | yes |
| `test_multiple_refreshes_with_stale_entries` | **passes** | no |
| `test_pop_skips_stale_heap_entries` | **passes** | no |
| `test_refresh_with_same_timestamp` | **passes** | no |
| `test_pop_returns_default_when_only_stale_entries_remain` | **passes** | no |
| `test_pop_handles_concurrent_removal_race` | **passes** | no |

Five of seven pass on unpatched `main`, so they would not have caught #10506 and will not catch a regression.

The cause is that `add()` calls `discard()`, which calls `_maybe_refresh_heap()`, and at small sizes the 15% overload threshold is always exceeded — so the heap is rebuilt and the stale entry the test intends to create is silently removed. Two comments are therefore factually wrong:

- `test_pop_skips_stale_heap_entries` says *"Heap now contains: (1, task1) [stale], (2, task2), (5, task1) [current]"*. The actual heap is `[(2.0,'task2'), (5.0,'task1')]` with **zero** stale entries.
- `test_multiple_refreshes_with_stale_entries` promises multiple stale entries; there is never more than one at a time.

`test_refresh_with_same_timestamp` can't discriminate either — with equal timestamps the stale and current entries expire together, so both versions agree. Its `entry1 is not entry2` assertion is still worth keeping as documentation of why identity is required.

**To get a `pop()` test that actually fails on main,** use at least 8 items so `_maybe_refresh_heap()` stays under its threshold and the stale entry survives:

```python
def test_pop_does_not_evict_refreshed_item_via_stale_entry(self):
    s = LimitedSet(expires=100)
    # >= 8 items so discard()'s 15% overload threshold does not rebuild the heap
    for i in range(8):
        s.add(f"other{i}", now=10.0 + i)
    s.add("hot", now=1.0)
    s.add("hot", now=100.0)          # leaves a genuine stale (1.0, "hot")
    assert s.pop() == "other0"       # main returns "hot"
    assert "hot" in s
```

Verified: `main` returns `'hot'` and drops it from the set; this branch returns `'other0'` and keeps `'hot'`. The same shape works for `purge(now=102.0)`.

Last two are minor: `test_pop_returns_default_when_only_stale_entries_remain` reaches its state with `s._data.pop(...)`, which `discard()` does through the public API, so calling `s.discard("task-id")` would be more faithful; and `test_purge_with_only_stale_entries_in_heap` manufactures a state my fuzzing suggests the public API cannot produce, so it is covering a defensive branch rather than a reachable one — worth saying so in the docstring.

On the smoke/integration question: agreed, skipping it is the right call. The behaviour is a pure data-structure timing property and a broker-level test would be far more fragile than these unit tests.

No objection to the production change — it is correct and I would merge it once the test gaps above are closed.

t/unit/utils/test_collections.py
```diff
@@ -461,6 +461,187 @@ def test_add_removes_duplicate_from_big_heap(self):
...
+
+        assert s.pop() is None
+        assert s.pop(default="empty") == "empty"
+
+    def test_pop_handles_concurrent_removal_race(self):
+        """pop() should handle race where another thread removes item between identity check and pop.
+
+        Regression test for the check-then-act race in pop(). If another thread
+        removes an item from self._data after the identity check passes but before
+        self._data.pop(item) is called, pop() should treat it as a stale entry and
```
mohithvardhan002, line 612: There still appears to be a check-then-act race here. After current_entry is entry passes, another thread could refresh the same item before _data.pop(item) executes. Since add() replaces _data[item] with a new entry, this pop could then remove the newly refreshed entry. The new regression test covers concurrent removal (KeyError), but not concurrent refresh/replacement. Could we also cover or synchronize this case?

auvipy (changes requested):
also merge conflict needs to be fixed

PR #10524, opened 2026-08-28
Honor a timeout of 0 in get_many instead of polling forever

dylanpulver:
## Description

On a key-value result backend, `GroupResult.join_native(timeout=0)` and `iter_native(timeout=0)` never return. `KeyValueStoreBackend.get_many` guards its deadline with `if timeout and iterations * interval >= timeout`, so a timeout of 0 is falsy and the check is skipped on every pass. Measured on `cache+memory://` against ids that stay pending:

```
timeout=0      still running after 2.5s
timeout=0.01   TimeoutError after 0.058s
timeout=0.2    TimeoutError after 0.225s
```

The middle line is the other half of it: the deadline is counted in whole intervals, so any timeout below `interval` is rounded up to one interval. After this change those three read 0.000s, 0.013s and 0.216s.

#10487 fixed both of these in `SyncBackendMixin.wait_for`. This applies the same two changes to `get_many` 170 lines below in the same file: compare against `None`, and clamp the nap to what is left of the budget. Semantics then match `wait_for`, where a budget of 0 polls once and raises rather than blocking.

`test_get_many_times_out` uses `timeout=0.01, interval=0.01`. A truthy timeout cannot separate `if timeout` from `if timeout is not None`, and with the interval equal to the timeout the overshoot is invisible.

Full unit suite and `pre-commit` are green. Reverting only `backends/base.py` fails both new tests, and `max_iterations` keeps them failing rather than hanging if the deadline is skipped again.

Not included: `Drainer.drain_events_until` has the same falsy guard, but its check sits above the `p.ready` break, so flipping it there would raise on an already-ready result. That needs the loop reordered.

### DIFF
--- a/celery/backends/base.py
+++ b/celery/backends/base.py
@@ -1202,6 +1202,7 @@ class BaseKeyValueStoreBackend(Backend):
 
         ids.difference_update(cached_ids)
         iterations = 0
+        time_elapsed = 0.0
         while ids:
             keys = list(ids)
             r = self._mget_to_results(self.mget([self.get_key_for_task(k)
@@ -1212,11 +1213,17 @@ class BaseKeyValueStoreBackend(Backend):
                 if on_message is not None:
                     on_message(value)
                 yield bytes_to_str(key), value
-            if timeout and iterations * interval >= timeout:
+            if timeout is not None and time_elapsed >= timeout:
                 raise TimeoutError(f'Operation timed out ({timeout})')
             if on_interval:
                 on_interval()
-            time.sleep(interval)  # don't busy loop.
+            # don't busy loop, and never sleep past the deadline: with the
+            # deadline counted in whole intervals, timeout=0 waited forever
+            # and any timeout below interval overshot to interval.
+            nap = interval if timeout is None else min(interval,
+                                                       timeout - time_elapsed)
+            time.sleep(nap)
+            time_elapsed += nap
             iterations += 1
             if max_iterations and iterations >= max_iterations:
                 break
--- a/celery/backends/base.py
+++ b/celery/backends/base.py
@@ -1213,6 +1213,8 @@ class BaseKeyValueStoreBackend(Backend):
                 if on_message is not None:
                     on_message(value)
                 yield bytes_to_str(key), value
+            if not ids:
+                break
             if timeout is not None and time_elapsed >= timeout:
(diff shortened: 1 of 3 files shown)


auvipy (changes requested):
can you add additional integration or smoke test for this change?

dylanpulver:
Copilot was right, and the `timeout=0` case was a regression this PR introduced. Reproduced against a key/value backend with every id ready on the first `mget()`: `TimeoutError: Operation timed out (0)` on the branch, results returned on `main`, where `if timeout` skipped the check.

The exact-deadline half of the comment holds too and predates the PR. With `timeout=0.5, interval=0.5`, a second poll that satisfies the last id lands on the deadline and raises on `main` as well.

The break in 8ec6af3a is the shape #10487 gave `wait_for`, at base.py:1039-1046:

```python
meta = self.get_task_meta(task_id)
if meta['status'] in states.READY_STATES:
    return meta
if on_interval:
    on_interval()
if timeout is not None and time_elapsed >= timeout:
    raise TimeoutError('The operation timed out.')
```

Completion first, deadline second. I put a comment on it in 9327e70, since the ordering is load-bearing and reads as incidental without one. That ordering is what this PR's description said `drain_events_until` would need, one function away, which I missed.

@auvipy tests are in the same commit. Integration rather than smoke, because this is result-backend polling rather than worker topology, and `t/integration` runs against a live redis backend under the redis and rabbitmq_redis toxenvs. Both live in `t/integration/test_canvas.py::test_result_set` and skip unless the backend is a `BaseKeyValueStoreBackend`:

- `test_join_native_timeout_zero_gives_up_on_pending_results` joins ids nothing will write a result for, with `interval=30`, so sleeping one interval breaks the wall-clock assertion. Against unpatched `get_many` the call does not return; with the break it raises in 0.001s.
- `test_join_native_timeout_zero_returns_results_that_are_ready` clears the backend cache so the poll loop is entered rather than answered from cache. Fails with `Operation timed out (0)` when the break is removed.

Two unit tests pin the same behaviour directly, both failing when only the break is removed. Full `t/integration/test_canvas.py` passes against redis, 177 passed, 8 xfailed, 5 xpassed. Unit suite 3828 passed, 36 skipped, 3 xfailed, against 3824 on `main`.

auvipy (approved):

PR #10528, opened 2026-08-29
fix: preserve chord bookkeeping when revoking tasks

### LINKED ISSUE #10526: `_revoke()` does not pass `Request` to `mark_as_revoked()`, preventing chord bookkeeping
## Description

When a task that belongs to a chord is revoked through the worker control command (`app.control.revoke(task_id)`), the task is correctly marked as `REVOKED` in the backend, but the chord bookkeeping is not updated when the task is already known by the worker.

The issue is in `celery/worker/control.py`, where `_revoke()` calls:

```python
state.app.backend.mark_as_revoked(
    task_id, reason='revoked', store_result=True
)
```

without passing the task's `Request`.

However, `BaseBackend.mark_as_revoked()` performs chord bookkeeping only when a `request` is provided:

```python
if request and request.chord:
    self.on_chord_part_return(request, state, exc)
```

This creates an inconsistency with the normal worker-side revocation path, which passes the request to `mark_as_revoked()`.

## Steps to reproduce

1. Create a chord with multiple tasks.
2. Ensure a chord member has already been received by a worker so its `Request` is locally known.
3. Revoke that task using the worker control API:

```python
app.control.revoke(task_id)
```

4. Check the backend state of the revoked task.
5. Observe the chord completion behavior.

The revoked task is stored as `REVOKED`, but `on_chord_part_return()` is not called for that chord member.

## Expected behavior

Revoking a chord member should update the chord bookkeeping in the same way as normal worker-side revocation, so that the chord can account for the revoked member and continue according to the configured chord behavior.

## Actual behavior

The task is marked as `REVOKED`, but because `_revoke()` does not pass the locally-known `Request`, `BaseBackend.mark_as_revoked()` cannot detect that the task belongs to a chord.

As a result, `on_chord_part_return()` is not called and the chord may remain waiting for that member.

## Root cause

The control-command revocation path does not propagate the existing `Request` object to the backend.

The normal request-based revocation path already does this:

```python
self.task.backend.mark_as_revoked(
    self.id,
    reason,
    request=self._context,
    store_result=self.store_errors,
)
```

but `_revoke()` does not.

## Proposed fix

Reuse the existing `_find_requests_by_id()` helper in `celery/worker/control.py` to associate locally-known task IDs with their `Request` objects, and pass the request to `mark_as_revoked()`:

```python
requests_by_id = {
    request.id: request
    for request in _find_requests_by_id(task_ids)
}

for task_id in task_ids:
    state.app.backend.mark_as_revoked(
        task_id,
        reason='revoked',
        store_result=True,
        request=requests_by_id.get(task_id),
    )
```

A regression test should verify that when a locally-known chord task is revoked through `_revoke()`, its `Request` is passed to `mark_as_revoked()` and chord bookkeeping is triggered.

## Scope / limitation

This addresses the case where the revoked task has already been received by a worker and its `Request` is locally available.

If a task has never been consumed by any worker, there may be no local `Request` from which to obtain the chord metadata. That is a separate issue and is not addressed by this proposed change.

## Additional context

This behavior appears to be related to the change that made control-command revocation immediately update the backend state to `REVOKED`. The backend status update works correctly, but the corresponding chord side effect is currently skipped because the `Request` is not propagated.

Namraa310806:
## Description

Fixes #10526

When a task belonging to a chord is revoked through the worker control command, `_revoke()` marks the task as `REVOKED` in the backend but does not pass the locally-known `Request` to `mark_as_revoked()`.

`BaseBackend.mark_as_revoked()` uses the `Request` to perform chord bookkeeping through `on_chord_part_return()`. Without the request, a locally-known revoked chord member can therefore be left out of the chord bookkeeping.

### Changes

* Reuse the existing `_find_requests_by_id()` helper to find locally-known `Request` objects.
* Pass the matching `Request` to `mark_as_revoked()`.
* Preserve `request=None` behavior when the task is not locally known.
* Add a regression test that exercises the real `mark_as_revoked()` path and verifies that `on_chord_part_return()` is triggered for a revoked chord member.
* Update existing revoke backend tests for the new `request` argument.

### Testing

* `pytest t/unit/worker/test_control.py`
* `pytest t/unit/worker/`

The change is limited to the worker revoke path and its regression tests.

### DIFF
--- a/celery/worker/control.py
+++ b/celery/worker/control.py
@@ -217,9 +217,20 @@ def _revoke(state, task_ids, terminate=False, signal=None, **kwargs):
 
     worker_state.revoked.update(task_ids)
 
+    # Find locally-known Request objects for the task IDs
+    requests_by_id = {
+        request.id: request
+        for request in _find_requests_by_id(task_ids)
+    }
+
     for task_id in task_ids:
         try:
-            state.app.backend.mark_as_revoked(task_id, reason='revoked', store_result=True)
+            state.app.backend.mark_as_revoked(
+                task_id,
+                reason='revoked',
+                store_result=True,
+                request=requests_by_id.get(task_id),
+            )
         except Exception as exc:
             logger.warning('Failed to mark task %s as revoked in backend: %s', task_id, exc)
 
--- a/t/unit/worker/test_control.py
+++ b/t/unit/worker/test_control.py
@@ -957,36 +957,55 @@ class test_ControlPanel:
         try:
             state = self.create_state()
 
-            # Spy on on_chord_part_return to verify it's called
-            original_on_chord_part_return = self.app.backend.on_chord_part_return
-            on_chord_part_return_calls = []
+            # Use patch.object to spy on on_chord_part_return
+            # We use the real backend, not a mock, so mark_as_revoked actually executes
+            with patch.object(
+                self.app.backend,
+                'on_chord_part_return',
+                wraps=self.app.backend.on_chord_part_return
+            ) as mock_on_chord_part_return:
+                # Call _revoke() - this should trigger chord bookkeeping
+                control._revoke(state, [task_id])
 
-            def spy_on_chord_part_return(*args, **kwargs):
-                on_chord_part_return_calls.append((args, kwargs))
-                return original_on_chord_part_return(*args, **kwargs)
-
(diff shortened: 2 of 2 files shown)


auvipy:
closing in favor of another pr

PR #10529, opened 2026-08-29
Fix DjangoTask selection for Celery subclasses without a custom task class

phanky1:
## Description

In a Django project, subclassing Celery causes `DjangoFixup` to skip installing the Django-specific task base class, even when the subclass does not override `task_cls`.

```python
# assume DJANGO_SETTINGS_MODULE is set
from celery import Celery
from celery.contrib.django.task import DjangoTask

class MyCustomCelery(Celery): pass
app = MyCustomCelery()

assert issubclass(app.Task, DjangoTask) # AssertionError
```
The problem lies in:
https://github.com/celery/celery/blob/cd07bcfd24e39bcaf46f0a1e41b22877ba9842ad/celery/app/base.py#L360-L365

`hasattr(self.__class__, 'task_cls')` follows the class MRO and eventually finds `Celery.task_cls`. Therefore, this condition is true for every subclass of Celery, whether or not that subclass overrides `task_cls`.

As a result, every Celery subclass is treated as having a custom task class, and DjangoFixup does not install DjangoTask.

Related issue: I looked around and didn't find any reported issue.

### DIFF
--- a/celery/app/base.py
+++ b/celery/app/base.py
@@ -360,8 +360,8 @@ class Celery:
         self._custom_task_cls_used = (
             # Custom task class provided as argument
             bool(task_cls)
-            # subclass of Celery with a task_cls attribute
-            or self.__class__ is not Celery and hasattr(self.__class__, 'task_cls')
+            # Custom task class set as a class attribute
+            or bool(app_has_custom(self, 'task_cls'))
         )
         self.task_cls = task_cls or self.task_cls
         self.set_as_current = set_as_current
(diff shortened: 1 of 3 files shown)


auvipy (changes requested):
may be having proper integration test will also help here, I am generally very cautious while changing anything in django integration part

phanky1:
> may be having proper integration test will also help here, I am generally very cautious while changing anything in django integration part

I just updated it with a new integration test.

auvipy:
please check why the unit test builds are failing

phanky1:
> please check why the unit test builds are failing

It was a network issue, the test suite passes now.

auvipy (approved):

Issue #10514, opened 2026-08-24
Enhancement: Add efficient progress tracking for large groups


Namraa310806:
# Checklist
- [ ] I have checked the issues list for similar or identical enhancement to an existing feature.
- [ ] I have checked the pull requests list for existing proposed enhancements.
- [ ] I have checked the commit log to find out if the same enhancement was already implemented in the main branch.
- [x] I have included all related issues and possible duplicate issues in this issue (If there are none, check this box anyway).

#### Related Issues
- None found
#### Possible Duplicates
- None found

# Brief Summary

`GroupResult.completed_count()` (and related methods like `.ready()`) determine group progress by iterating every child `AsyncResult` and checking its state individually — one backend round-trip per task. For large groups (e.g. 10,000+ tasks), this means thousands of backend calls just to answer "how many are done."

Celery already solves an equivalent problem for chords: `on_chord_part_return()` in the Redis (and GCS) backends maintains an atomic, backend-side counter (`.j`/`.s`/`.t` keys in Redis) so the backend can determine "are all chord members done?" in a single pipelined operation, without polling each child. That mechanism exists specifically to decide when to fire the chord callback — but the same primitive seems like it could answer a more general question: "how many of N tasks in this group are done?" for plain groups too, not just chord headers.

This issue proposes discussing whether it's worth exposing an efficient, backend-native progress query for plain `GroupResult`s, reusing the bookkeeping pattern that already exists for chords, rather than requiring O(N) client-side polling.

# Design

## Architectural Considerations
None — this would be implemented within Celery's existing backend abstraction (`celery/backends/base.py`, with a concrete implementation in `celery/backends/redis.py`, and potentially `gcs.py`). No new external dependencies. Backends that don't implement the new capability would fall back to the current O(N) behavior, so this would be additive rather than a breaking change to the `Backend` interface.

## Proposed Behavior

**This is a starting hypothesis, not a fixed design — I'd like feedback on the direction before settling on specifics.**

At a high level:

- Today: querying group progress means iterating every child result and checking its state — O(N) backend calls.
- Proposed: for backends that support it, group progress could be tracked via a backend-maintained counter (similar to the existing chord `.j`/`.s` key pattern in the Redis backend), so a progress query costs O(1) instead of O(N).

Open questions I don't want to prematurely answer:
- Should this be opt-in per-group (to avoid writing extra backend state for every group that never asks for progress), or could it be made cheap enough to be the default path for `completed_count()`?
[...]

PR #10530, opened 2026-08-29
wip: group progress tracking phase 1

Namraa310806:
## Description

Related to #10514

### Problem

`GroupResult.completed_count()` and related result queries determine group progress by inspecting individual child results. For large groups, this can require an O(N) number of backend result checks just to determine how many tasks have completed.

For example, a group containing 10,000 tasks may require thousands of backend operations when checking progress.

Celery already has backend-side aggregate bookkeeping for chords. This provides an opportunity to extend the same general pattern to ordinary groups, allowing applications to query group progress without polling every child result.

### Proposed Approach

The proposed design is to provide an opt-in, backend-native progress tracking capability for groups.

At a high level:

```text
Current:

GroupResult.completed_count()
        ↓
iterate over child AsyncResults
        ↓
N backend result checks
```

Proposed:

```text
GroupResult.progress()
        ↓
backend-maintained aggregate progress
        ↓
constant-cost backend query
```

The intended behavior is:

* Allow a group to opt into progress tracking.
* Maintain aggregate progress in the backend.
* Provide a `GroupResult.progress()` API returning the current completed/total values.
* Use a native backend implementation where supported.
* Fall back to the existing O(N) result inspection for unsupported backends.
* Avoid changing the behavior or cost of existing groups that do not opt into progress tracking.
* Keep ordinary group progress state separate from chord-specific bookkeeping.

The exact API and backend abstraction can be refined based on review and maintainer feedback.

### Work Completed

The current implementation establishes the initial foundation for the feature:

* Added a backend capability for native group progress tracking.
* Added backend methods for initializing, incrementing, and retrieving group progress.
* Added Redis support using a dedicated progress key with atomic counter updates.
* Added `GroupResult.progress()` with a native backend path and an O(N) fallback.
* Added `track_progress` as an opt-in group execution option.
* Initialize progress state before task submission to avoid losing early task completions.
* Extract `track_progress` before task execution options are passed to child tasks.
* Added tests covering Redis progress tracking, `GroupResult.progress()`, canvas integration, retry/failure behavior, TTL handling, and progress-key isolation.

### Current Progress Semantics

The current implementation treats terminal result states as completed progress:

* `SUCCESS`
* `FAILURE`
* `REVOKED`

`RETRY` remains incomplete.

This is intentionally distinct from `GroupResult.completed_count()`, which currently counts successful results only.

The final semantics can be adjusted if maintainers recommend a different interpretation.

### Remaining Work

Before this can be considered complete, the following areas still need to be validated or implemented:

* Further verify retry and duplicate terminal-completion semantics.
* Ensure the same logical task cannot incorrectly increment progress more than once.
* Strengthen integration-level testing of the complete group execution path.
* Validate concurrent task completion behavior.
* Validate progress initialization and group-size handling across relevant group construction paths.
* Validate expiration and TTL behavior against existing result expiration semantics.
* Verify interaction with chords and other canvas primitives.
* Review behavior for nested groups and other group compositions.
* Review the final public API and backend capability abstraction against Celery conventions.
* Evaluate native GCS support versus fallback behavior.
* Add appropriate user documentation for the final API and semantics.
* Add performance validation for large groups to demonstrate the benefit of native progress tracking.
* Incorporate feedback from the enhancement discussion and revise the implementation where necessary.

### Performance Goal

The intended optimization is specifically for the progress query:

```text
Current:
O(N) backend result checks

Native progress:
O(1) aggregate backend query
```

There is still additional per-task bookkeeping when a task reaches a terminal state. The goal is to trade this small per-task cost for substantially cheaper repeated progress queries on large groups.

### Testing

The current implementation includes tests for:

* Redis group progress initialization and retrieval
* Atomic progress updates
* Progress key isolation
* TTL handling
* `GroupResult.progress()` native and fallback paths
* Empty groups
* Failure handling
* Retry behavior
* Canvas `track_progress` handling
* Progress initialization ordering

The remaining integration, concurrency, expiration, and regression coverage will be completed as the implementation is refined.

### Feedback Requested

I would particularly appreciate feedback on:

1. Whether backend-native progress tracking for large groups is useful for Celery.
2. Whether progress tracking should remain opt-in.
3. Whether `GroupResult.progress()` is an appropriate API.
4. Whether the proposed backend capability abstraction fits Celery's architecture.
5. Whether the proposed completion semantics are appropriate.
6. Whether Redis should be the initial native backend while unsupported backends use the existing fallback.


auvipy:
integration tests were failing, check them please

Namraa310806:
@auvipy I am working on it in phases and implementing it.
i just Opened this Draft so that anyone can see its progress and add they can give suggestions on it

PR #10534, opened 2026-08-29
Bound State._tasks_to_resolve to fix unbounded memory growth

### LINKED ISSUE #4832: Memory leak in celery.app.events.State?
## Checklist
* [ ] I have included the output of ``celery -A proj report`` in the issue.
       version affected).
     Celery 4.2.0
* [x] I have verified that the issue exists against the `master` branch of Celery.

## Steps to reproduce
We have setup some simple event monitoring based on the recommendations here: http://docs.celeryproject.org/en/latest/userguide/monitoring.html#real-time-processing

A much simplified version of the monitoring code looks something like this:

```python3
class MonitorThread(threading.Thread):
  """This thread will collect data that is exposed by Celery through its eventing system."""

  def __init__(self, *args, **kwargs):
    self._app = cour_custom_celery_app
    self._state = self._app.events.State(max_tasks_in_memory=max_tasks_in_memory)
    self._known_states_name = set()
    self._tasks_started = dict()
    super(MonitorThread, self).__init__(*args, **kwargs)

  def run(self):
    self._monitor()

  def _process_event(self, event):
    # Events may come in in parallel. Celery already has a lock that deals with this exact same
    # situation, so we'll use that for now.
    with self._state._mutex:
        self._collect_tasks(event, event.state)  # Ignore the fact that we've used event.state instead of actually retrieving it

  def _collect_tasks(self, event, state):
    if state in celery.states.READY_STATES:
      # A READY_STATE indicates the task completed. Remove from task tracker.
      self._state.tasks.pop(event["uuid"])
    else:
      # Add event to list of in-progress tasks.
      self._state._event(event)

def _monitor(self):
        with self._app.connection() as conn:
          recv = self._app.events.Receiver(conn, handlers={"*": self._process_event})
          setup_metrics(self._app)
          # This will block the thread and continuously accept events from celery.
          recv.capture(limit=None, timeout=5, wakeup=True)
```

We've noticed that this program unboundedly leaks memory, and have traced it down to the `WeakSet` constructed here: https://github.com/celery/celery/blob/master/celery/events/state.py#L650

Here is an example task configuration that we use:

```python3
@app.task
def big_task():
    task1.s.apply_async()
    task2.s.apply_async()

@app.task
def task1():
   do something here

@app.task
def task2():
  do something here
```

## Expected behavior
We would expect that even though the child tasks are added to the `_tasks_to_resolve` map, that at some point they are popped off the map here: https://github.com/celery/celery/blob/master/celery/events/state.py#L637

But to us it looks like the `_tasks_to_resolve` map grows unboundedly. 

## Actual behavior
Could this be because `big_task` finishes before `task1` and `task2` are received? This would mean that all events for `big_task` are processed before any events for `task1` and `task2`, which would mean that the children would be added to the `_tasks_to_resolve_map` **after** the parent task had completed, and thus theoretically popped the `WeakSet` off the map.

Am I completely off base here? Is there something off about my setup? Should we be scheduling our tasks or setting up our event monitor differently?  Or is this actually a problem in the `celery` code - which I find highly unlikely :)?

ghirailghiro:
## Summary
- A task event carrying a `parent_id` whose parent isn't yet in `State.tasks` registers the child in `_tasks_to_resolve[parent_id]`. That entry is only popped when a later event for the parent itself arrives.
- If the parent never arrives again (its events were already processed, or a monitor evicted it from `tasks` on a `READY_STATE`, as the docs' real-time monitoring example used to encourage), the entry stays forever. The child `Task`s are held in a `WeakSet` and do get collected, but the `_tasks_to_resolve` dict key does not, so a long-running monitor leaks memory without bound.
- This bounds `_tasks_to_resolve` the same way every other mapping on `State` is bounded, using an `LRUCache(max_tasks_in_memory)` instead of a plain `dict`.

Fixes #4832

## Test plan
- [x] Added `test_tasks_to_resolve_is_bounded` in `t/unit/events/test_state.py`, asserting `len(state._tasks_to_resolve) <= state.max_tasks_in_memory` after feeding it more orphaned child events than the limit.
- [x] Full `t/unit/events/test_state.py` suite passes locally.
- [x] Manually reproduced the leak against the pre-fix code (unbounded growth of `_tasks_to_resolve`) and confirmed it's bounded after the fix.

### DIFF
--- a/celery/events/state.py
+++ b/celery/events/state.py
@@ -422,7 +422,7 @@ class State:
         self._mutex = threading.Lock()
         self.handlers = {}
         self._seen_types = set()
-        self._tasks_to_resolve = {}
+        self._tasks_to_resolve = LRUCache(max_tasks_in_memory)
         self.rebuild_taskheap()
 
         self.tasks_by_type = CallableDefaultdict(
--- a/t/integration/test_events_state_memory_leak.py
+++ b/t/integration/test_events_state_memory_leak.py
@@ -1,0 +1,63 @@
+"""
+Integration test for memory leak issue #4832.
+
+Reproduces the scenario of a long-running monitor (e.g. ``celery events``
+or Flower) consuming real events off the broker: child tasks whose
+parent is never observed (because it was scheduled with an ETA/countdown
+and its own event hasn't arrived yet, or because it was already evicted
+from ``State.tasks``) used to accumulate forever in
+``State._tasks_to_resolve``, since that mapping was a plain, unbounded
+``dict``.
+
+Unlike the unit test in ``t/unit/events/test_state.py`` (which calls
+``State.event()`` directly with synthetic dicts), this test exercises the
+real event pipeline: an ``EventDispatcher`` publishes events onto an
+actual (in-memory) broker connection, and an ``EventReceiver`` consumes
+and decodes them before handing them to ``State``.
+"""
+
+from celery import Celery, uuid
+
+
+def test_tasks_to_resolve_is_bounded_over_real_event_pipeline():
+    app = Celery('test_events_state_memory_leak')
+    app.conf.update(broker_url='memory://', result_backend='cache+memory://')
+
+    max_tasks_in_memory = 10
+    state = app.events.State(max_tasks_in_memory=max_tasks_in_memory)
+
+    orphan_count = 100
+
+    with app.connection_for_write() as connection:
+        receiver = app.events.Receiver(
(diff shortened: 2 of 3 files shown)


auvipy (changes requested):
can we get additional tests for this change? may be integration?

ghirailghiro:
Pushed now let me know if it is ok

auvipy (approved):

PR #10536, opened 2026-08-30
Fix: DatabaseBackend now honors result_serializer (#3025)

### LINKED ISSUE #3025: Celery results are NOT obeying results serializer config in databases
So, we're working on getting all of our tasks results stored in a database backend (for now just postgres, but we're not tied to any particular backend).

Anyway, we've noticed that no matter what we try to do, the results are **always** stored as a pickle object, even though we have the other configurations set properly:

```python
app.conf.update(CELERY_ACCEPT_CONTENT=["json"],
                CELERY_TASK_SERIALIZER="json",
                CELERY_RESULT_SERIALIZER="json")
```

While not being terribly familiar with SQL Alchemy, I do believe we have found the [offending line](https://github.com/celery/celery/blob/master/celery/backends/database/models.py#L33).

Am I understanding this right? To me, it looks like this has been a bug for _years_... is it because we might be the only ones using a database backend? :-P

Thanks!

ghirailghiro:
## Summary

Fixes #3025.

The `database` (SQLAlchemy) result backend always stored `Task.result` and
`TaskSet.result` as a Python pickle, regardless of the configured
`result_serializer`. This is because both columns were declared as
`sa.PickleType`, a SQLAlchemy `TypeDecorator` that pickles/unpickles values
automatically at the ORM level, before Celery's own `encode()`/`decode()`
(the methods that actually respect `result_serializer`) ever get a chance
to run. Every other backend (mongodb, redis/`KeyValueStoreBackend`) already
calls `self.encode()`/`self.decode()` explicitly — the database backend was
the one exception.

Two earlier attempts tried to fix this (#2881, #5907) and stalled — not on
the core design (column as raw bytes + explicit `encode()`/`decode()`,
which is what this PR does too, matching `mongodb.py`), but because neither
handled the backward-compatibility question raised in review on #5907:
rows written before the fix are *always* pickle, no matter what
`result_serializer` says, since that's exactly the bug. After fixing the
write path, those pre-existing rows would fail to decode under a
non-pickle serializer.

This PR closes that gap: `_decode_stored_result()` tries `self.decode()`
first, and falls back to `pickle.loads()` if that fails, so existing
databases keep working unmodified after the upgrade — only new writes use
the configured serializer.

- `celery/backends/database/models.py`: `Task.result` / `TaskSet.result`
  changed from `PickleType` to `LargeBinary`. No schema/DDL change:
  `PickleType`'s underlying storage is already `LargeBinary`, confirmed by
  comparing generated `CREATE TABLE` output across sqlite/postgres/mysql.
- `celery/backends/database/__init__.py`: `_update_result`/`_save_group`
  now call `self.encode()` before storing; `_get_task_meta_for`/
  `_restore_group` decode through the new fallback-aware helper.
- `docs/userguide/configuration.rst`: documents the behavior change and
  the backward-compatibility guarantee under the database backend section
  and the `result_serializer` setting.

## Test plan

- [x] New regression tests in `t/unit/backends/test_database.py` assert
      the *raw bytes* in the DB are valid JSON when `result_serializer`
      is `json` (not a pickle blob), for both task and group results.
- [x] New test simulates a pre-fix row (raw pickle written directly via
      SQL) and confirms it's still readable — and logs a warning — with
      `result_serializer='json'` configured.
- [x] Existing suite passes unmodified: `pytest t/unit/backends/test_database.py`
      (80 passed).
- [x] Manual verification: exceptions (`mark_as_failure`/`mark_as_retry`)
      round-trip correctly under both `pickle` and `json`.
- [x] Manual verification script covering forward path, legacy pickle
      rows, and a mixed old/new table — confirmed it fails against the
      pre-fix code and passes against the fix.

### DIFF
--- a/celery/backends/database/__init__.py
+++ b/celery/backends/database/__init__.py
@@ -8,6 +8,7 @@ from kombu.utils.encoding import ensure_bytes
 from celery import states
 from celery.backends.base import BaseBackend
 from celery.exceptions import ImproperlyConfigured
+from celery.result import result_from_tuple
 from celery.utils.imports import symbol_by_name
 from celery.utils.time import maybe_timedelta
 
@@ -193,8 +194,12 @@ class DatabaseBackend(BaseBackend):
 
         Rows written before the fix for celery/celery#3025 always stored a
         raw pickle blob regardless of the configured ``result_serializer``,
-        so a row that can't be decoded with the current serializer is
+        so a row that can't be decoded with the current serializer, but
+        looks like a pickle payload (starts with the pickle protocol 2+
+        marker that every pickle Celery/kombu produces starts with), is
         assumed to be one of those and is unpickled directly instead.
+        Anything else re-raises the original decode error instead of
+        blindly unpickling arbitrary bytes.
         """
         if payload is None:
             return payload
@@ -233,7 +238,8 @@ class DatabaseBackend(BaseBackend):
         """Store the result of an executed group."""
         session = self.ResultSession()
         with session_cleanup(session):
-            group = self.taskset_cls(group_id, ensure_bytes(self.encode(result)))
+            value = ensure_bytes(self.encode(self.prepare_value(result)))
+            group = self.taskset_cls(group_id, value)
             session.add(group)
             session.flush()
             session.commit()
@@ -247,7 +253,10 @@ class DatabaseBackend(BaseBackend):
                 self.taskset_cls.taskset_id == group_id).first()
             if group:
                 data = group.to_dict()
-                data['result'] = self._decode_stored_result(data['result'])
+                value = self._decode_stored_result(data['result'])
+                data['result'] = (
+                    value if value is None else result_from_tuple(value, self.app)
+                )
                 return data
 
     def _delete_group(self, group_id):
--- a/celery/backends/database/__init__.py
+++ b/celery/backends/database/__init__.py
@@ -1,6 +1,9 @@
 """SQLAlchemy result store backend."""
 import logging
+import pickle
 from contextlib import contextmanager
+
+from kombu.utils.encoding import ensure_bytes
 
 from celery import states
 from celery.backends.base import BaseBackend
@@ -151,7 +154,7 @@ class DatabaseBackend(BaseBackend):
     def _update_result(self, task, result, state, traceback=None,
                        request=None):
 
-        meta = self._get_result_meta(result=result, state=state,
+        meta = self._get_result_meta(result=ensure_bytes(self.encode(result)), state=state,
                                      traceback=traceback, request=request,
                                      format_date=False, encode=True)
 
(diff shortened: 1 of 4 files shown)


celery/backends/database/__init__.py
```diff
@@ -178,12 +181,32 @@ def _get_task_meta_for(self, task_id):
...
+        if payload is None:
+            return payload
+        try:
+            return self.decode(payload)
+        except Exception:
+            logger.warning(
+                'Task result payload could not be decoded using the '
+                'configured result_serializer; falling back to pickle '
+                'for backward compatibility with pre-fix data.')
+            return pickle.loads(payload)
```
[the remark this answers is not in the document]
ghirailghiro: @auvipy that's a good catch, will apply this what you think? 

Worth noting the underlying risk isn't new though — PickleType unconditionally called pickle.loads() on every read before this PR, regardless of result_serializer. The magic-byte check won't stop someone who already has write access to the table (they can just craft a payload starting with \x80), but it does stop us from blindly unpickling arbitrary corrupted/non-pickle data on decode failure.

celery/backends/database/models.py
```diff
@@ -34,7 +33,7 @@ class Task(ResultModelBase):
                    primary_key=True, autoincrement=True)
     task_id = sa.Column(sa.String(155), unique=True)
     status = sa.Column(sa.String(50), default=states.PENDING)
-    result = sa.Column(PickleType, nullable=True)
+    result = sa.Column(sa.LargeBinary, nullable=True)
```
auvipy, line 36: is this a breaking change or not?

auvipy (changes requested):
we have to keep it backward compatible

celery/backends/database/models.py
ghirailghiro: I checked and it should not be a breaking change here the SQLAlchemy documentation:

https://docs.sqlalchemy.org/en/21/core/type_basics.html#sqlalchemy.types.PickleType

ghirailghiro: The important part is to handle the old PickleType data in the code and it is what i tried to do here:

https://github.com/celery/celery/pull/10536/changes#diff-2883b4b298cf9b1f46a6e47f6d53ac4c1aed029e6e6a77219a6edb6064557ba1R191

and also i added tests here for testing the retro-compatibility:

t/unit/backends/test_database.py

celery/backends/database/__init__.py
```diff
@@ -203,7 +233,7 @@ def _save_group(self, group_id, result):
         """Store the result of an executed group."""
         session = self.ResultSession()
         with session_cleanup(session):
-            group = self.taskset_cls(group_id, result)
+            group = self.taskset_cls(group_id, ensure_bytes(self.encode(result)))
```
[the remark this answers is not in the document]
ghirailghiro: Pushed a commit @auvipy that reused prepare_value()/result_from_tuple(), which
already exist in the codebase (same pattern as the Redis backend). The
group (and each task inside it) is first flattened into its tuple form
via as_tuple(), then that whole structure goes through the configured
serializer as usual.

auvipy (approved):

PR #10538, opened 2026-08-31
Fix abbr() returning more characters than max when max is below the ellipsis length

MahinAnowar:
Found by probing `celery.utils.text` invariants rather than from a report, so there's no issue to link.

`abbr` builds its result as `S[:max - len(ellipsis)] + ellipsis`. When `max` is smaller than the ellipsis that index goes negative, and a negative slice counts from the *end* of the string — so instead of shortening, it keeps nearly everything and then appends the ellipsis on top:

```python
>>> abbr('hello world', 2)
'hello worl...'     # 13 characters, from an 11-character input, for max=2
>>> abbr('hello world', 1)
'hello wor...'
>>> abbr('hello world', 0)
'hello wo...'
```

So for `max < 3` the function returns *more* than it was given, which is the one thing an abbreviation helper should never do.

This isn't only reachable by passing a small `max` directly. `abbrtask` computes one:

```python
module = abbr(module, max - len(cls) - 3, False)
```

`max - len(cls) - 3` is negative whenever the class part is long relative to the width, which is routine:

```python
>>> abbrtask('proj.tasks.add', 5)
'proj.task[.]add'   # 15 characters for max=5
```

And that path is live in `celery/events/cursesmon.py`, where `format_row` derives its column widths from the terminal:

```python
uuid = abbr(uuid, uuid_width).ljust(uuid_width)
task = abbrtask(task, task_width).ljust(task_width)
```

`uuid_width` comes out of `display_width` minus the other columns, so on a narrow terminal it drops below 3, `abbr` hands back something longer than the column, `ljust` can't shrink it, and the row overruns the layout it was measured for.

The fix keeps the ellipsis only when there is room for it, and otherwise just truncates:

```python
if isinstance(ellipsis, str) and max >= len(ellipsis):
    return S[: max - len(ellipsis)] + ellipsis
return S[: max] if max > 0 else ''
```

`max >= len(ellipsis)` rather than `>` is deliberate — it leaves `abbr(S, 3)` returning `'...'` exactly as before, so the only behaviour that changes is the cases that were broken. `abbrtask('proj.tasks.add', 5)` becomes `'[.]add'` as a knock-on.

One thing I left alone: `abbrtask` on a name with no dots still returns `'[.]noDots'`, because `rpartition` gives it an empty module. That looked like a separate bug and I didn't want to widen this PR — happy to send it separately if you want it fixed.

## Testing

Extended the existing `test_abbr` table with the `max < 3` cases and added two property tests (result never exceeds `max`, result never longer than the input). Reverting only `celery/utils/text.py` fails 11 of them:

```
11 failed, 21 passed
```

With the fix, `t/unit/utils/test_text.py` is 32/32, and `t/unit/utils/` plus `t/unit/events/` together are **656 passed, 3 skipped, 1 xfailed**. flake8 clean on both files. All four pre-existing `test_abbr` cases and both `test_abbrtask` cases are unchanged.


auvipy:
is there any related issue for this?

mohithvardhan002 (approved):
Reviewed the implementation and regression tests. The fix correctly handles max values smaller than the ellipsis length while preserving the existing behavior for normal values.

MehrazRumman (approved):
Checked this by extracting `abbr`/`abbrtask` from `origin/main` and `ec98c4d` and running them side by side. The bug is real and the fix is correct.

@auvipy — on the "is there a related issue" question: there isn't one, but the defect reproduces trivially, so I don't think it needs one. On `main`:

```
abbr('hello world', 2)  -> 'hello worl...'   13 chars, from an 11-char input
abbr('hello world', 0)  -> 'hello wo...'     11 chars
```

A negative `max - len(ellipsis)` slices from the end, so the function returns more than it was given. The `cursesmon.py` path the description cites is live: `uuid_width` comes from `detail_width - 1 - MIN_TASK_WIDTH - 1 - MIN_WORKER_WIDTH`, which goes negative on a narrow terminal, and `ljust()` can't shrink an over-long string back into the column.

**Behaviour is preserved where it matters.** Every pre-existing documented case is byte-identical, including the `max == len(ellipsis)` boundary:

| call | main | this PR |
|---|---|---|
| `abbr('ABCDEFGHI', 6)` | `'ABC...'` | `'ABC...'` |
| `abbr('ABCDEFGHI', 3)` | `'...'` | `'...'` |
| `abbr('ABCDEFGHI', 4)` | `'A...'` | `'A...'` |
| `abbr('ABCDEFGHI', 20)` | `'ABCDEFGHI'` | `'ABCDEFGHI'` |
| `abbr('ABCDEFGHI', 6, None)` | `'ABCDEF'` | `'ABCDEF'` |

Using `>=` rather than `>` is the right call — `>` would have silently changed `abbr(S, 3)`.

**One thing worth recording before this lands:** the cursesmon symptom is reduced but not closed, because `abbrtask` returns `module + '[.]' + cls` and that has a floor of `3 + len(cls)` no matter how small `max` is:

```
abbrtask('feeds.tasks.refresh', 3)  main: 'feed[.]refresh' (14)   PR: '[.]refresh' (10)
abbrtask('proj.tasks.add', 5)       main: 'proj.task[.]add' (15)  PR: '[.]add' (6)
```

So the task column can still overrun its width — just by a bounded amount now instead of an unbounded one. That's a strict improvement and I'd take it, but it means the layout fix is partial, and it's the same root cause as the `'[.]noDots'` case you flagged as out of scope. Both would want `abbrtask` to abbreviate `cls` too, which I agree belongs in its own PR.

Two small test notes, neither blocking:

- In `test_abbr_never_exceeds_max`, the `or len(s) <= max` arm is unreachable — whenever `len(s) <= max`, `abbr` returns `s` and the first arm is already true. It can just be dropped.
- Naming the parameter `max` is what forces `import builtins`; the other tests in this file use `maxsize`, and renaming would drop the import.

The property tests are a nice touch — "never longer than the input" is exactly the invariant that was violated. Approving.

PR #10550, opened 2026-09-03
Update CI matrix to exclude Python 3.10 and 3.14t

auvipy:
Removed Python 3.10 and 3.14t from the CI matrix for Windows for the upcoming v5.7 release


auvipy (commented):
it is eol in october. so we do not want to support it with the upcoming release. python 3.11 to 3.15 only. beside pyp3.11. this was the established rule for the project until recently.

PR #10551, opened 2026-09-03
Modify default Python versions in workflow

auvipy:
Updated default Python versions to exclude 3.10 and 3.14t.


auvipy (commented):
cleaning this up before pre release. as python 3.10 will be EOL by the time we release celery 5.7. so it will help reduce old version and pave the path for newer python 3.15 release

PR #10552, opened 2026-09-03
Update kombu dependency to use pypi release only

auvipy:


auvipy:
docker failure will be fixed in another PR

requirements/test-integration.txt
```diff
@@ -5,4 +5,4 @@
 -r extras/django.txt
 -r extras/sqlalchemy.txt
 pytest-rerunfailures>=11.1.2
-git+https://github.com/celery/kombu.git
+git+https://github.com/celery/kombu.git@main
```
auvipy:
```suggestion
kombu>=5.6.2
```

auvipy:
we already have dev versions in dev.txt so the defaults should use pypi releases or pre releases only

PR #10555, opened 2026-09-03
fix: preserve Django connection pools between tasks

### LINKED ISSUE #10554: Django connection pooling is closed at every task boundary
# Checklist

  on reporting bugs.
  for existing proposed fixes.

    version affected).
  to reproduce this bug.

  and/or implementation.
  result backend.
  broker and/or result backend.

#### Related Issues

- None found.

#### Possible Duplicates

- None found.

## Environment & Settings

**Celery version**: `main`

<details>
<p>

    Not applicable. The issue was identified from the Celery `main` source
    and the existing Django fixup unit tests.

</p>
</details>

# Steps to Reproduce

## Required Dependencies

- **Minimal Python Version**: Python version supported by the affected Celery/Django combination
- **Minimal Celery Version**: `main`
- **Minimal Kombu Version**: N/A
- **Minimal Broker Version**: N/A
- **Minimal OS and/or Kernel Version**: N/A
- **Minimal Broker Client Version**: N/A
- **Minimal Result Backend Version**: N/A
- **Django Version**: `5.1+`
- **PostgreSQL Driver**: `psycopg` 3.x

### Python Packages

<details>
<p>

    Not applicable. The issue is identified at the Django fixup/unit-test
    level and does not require a broker or result backend.

</p>
</details>

### Other Dependencies

<details>
<p>

Django 5.1+ with PostgreSQL connection pooling enabled, for example:

    DATABASES = {
        "default": {
            "ENGINE": "django.db.backends.postgresql",
            "OPTIONS": {
                "pool": True,
            },
        },
    }

The affected path is the prefork worker path, where Celery checks
`is_prefork` before calling `conn.close_pool()`.

</p>
</details>

<details>
<p>

The issue can be reproduced from the Django fixup implementation and the
existing unit tests in `t/unit/fixups/test_django.py`.

`DjangoWorkerFixup.install()` connects the Django database cleanup handlers
to both task lifecycle signals:

    signals.task_prerun.connect(self.on_task_prerun)
    signals.task_postrun.connect(self.on_task_postrun)

Both handlers call `close_database()`, which reaches `_close_database()`.

When Django PostgreSQL connection pooling is enabled, `_close_database()`
contains:

    conn.close()

    pool_enabled = (
        self._settings.DATABASES.get(conn.alias, {})
        .get("OPTIONS", {})
        .get("pool")
    )

    if pool_enabled and is_prefork and hasattr(conn, "close_pool"):
        with contextlib.suppress(KeyError):
            conn.close_pool()

Therefore, for a prefork worker with Django connection pooling enabled,
`conn.close_pool()` is reached during normal task lifecycle cleanup.

The existing test `test_close_database_conn_pool_based_on_settings` also
explicitly verifies that `close_pool()` is called when pooling is enabled.

</p>
</details>

# Expected Behavior

When Django's native database connection pooling is enabled, normal
per-task database cleanup should close or release the task's database
connection without shutting down the entire connection pool.

The connection pool should remain available for reuse by subsequent tasks.

The pool itself should be closed when the worker/process lifecycle requires
the pool to be destroyed, rather than as part of every task's normal
database cleanup.

# Actual Behavior

With Django's native PostgreSQL connection pooling enabled, the Celery
Django integration calls `conn.close_pool()` from
`DjangoWorkerFixup._close_database()` when using the prefork worker pool.

`_close_database()` is invoked through both `task_prerun` and
`task_postrun`. As a result, the connection pool can be shut down at every
task boundary.

This defeats the intended persistent nature of connection pooling because
the pool is not allowed to remain available across task executions.
Subsequent database use must therefore initialize/use a new pool instead of
benefiting from the existing pool.

The relevant code path is:

    def _close_database(self) -> None:
        ...

        for conn in connections:
            try:
                conn.close()

                pool_enabled = (
                    self._settings.DATABASES.get(conn.alias, {})
                    .get("OPTIONS", {})
                    .get("pool")
                )

                if pool_enabled and is_prefork and hasattr(conn, "close_pool"):
                    with contextlib.suppress(KeyError):
                        conn.close_pool()

                ...

The concern is that `close_pool()` is a pool lifecycle operation, while
`_close_database()` is also used for normal task-level database cleanup.

A possible solution would be to separate per-task connection cleanup from
pool shutdown and only close the pool during the appropriate worker/process
shutdown lifecycle.

I am opening this issue to confirm the intended lifecycle for Django's
native connection pools in Celery and to determine the correct place for
`close_pool()` to be called.

Namraa310806:
## Summary

Django's native PostgreSQL connection pooling was being closed during task cleanup, which meant the pool could not be reused across tasks.

This change keeps the database connection cleanup that happens at task boundaries, while preserving Django's connection pool for reuse and closing the pool at child-process exit.

## Changes

* Keep `conn.close()` during normal task cleanup.
* Stop closing the connection pool during individual task cleanup.
* Close inherited pools when a prefork worker process initializes, preserving the existing post-fork behavior.
* Register an `atexit` handler during worker process initialization to close pools when the child process exits.
* Restrict pool-specific cleanup to prefork workers with Django pooling enabled.
* Make pool cleanup safe when the pool has already been closed.
* Add regression tests covering task-level pool persistence, post-fork cleanup, child-exit cleanup, prefork/non-prefork behavior, and multiple connections.

## Testing

* `python -m pytest t/unit/fixups/test_django.py -xvs`
* **49 tests passed**
* `git diff --check` passes.

The existing Django integration setup uses SQLite rather than Django's native PostgreSQL pooling, so the regression coverage is focused on the connection/pool lifecycle directly.

Fixes #10554 

Related: #9953, #10020


auvipy:
@phanky1 it would be great if you can review this too

auvipy:
also try to to add smoke tests for this, and if possible and needed an integrations test too. this part of celery is very sensitive

Namraa310806:
@auvipy Agree on it
thats why after opening a pr i got a flaw and instantly i fixed it 
and Thanks for the review. I agree that this part of the Django fixup is process-lifecycle sensitive, so I don't want to rely only on the unit tests.

I'm adding smoke-test coverage for the real prefork worker lifecycle and checking whether the existing test infrastructure supports a meaningful Django/PostgreSQL integration test as well. I'll also verify that the new regression coverage would catch the original `close_pool()`-during-task-cleanup behavior.

I'll keep the additional coverage focused and avoid adding new infrastructure unless it is necessary.

celery/fixups/django.py
```diff
@@ -223,15 +251,9 @@ def _close_database(self) -> None:
...
 
-        is_prefork = self._is_prefork()
-
         for conn in connections:
             try:
                 conn.close()
-                pool_enabled = self._settings.DATABASES.get(conn.alias, {}).get("OPTIONS", {}).get("pool")
-                if pool_enabled and is_prefork and hasattr(conn, "close_pool"):
-                    with contextlib.suppress(KeyError):
-                        conn.close_pool()
```
phanky1, line 234: Removing these calls would reintroduce the issue fixed by https://github.com/celery/celery/pull/9953.

Namraa310806: Thanks for pointing this out. I agree that `close_pool()` is required for the pooling behavior introduced by #9953, so I want to make sure I'm not removing a lifecycle step that is still needed.

From my reading of #9953, the reported issue was the pool being inherited across `fork()` by prefork workers. `on_worker_process_init()` still calls `_close_database()` once in each child after the fork, so that part of the behavior remains unchanged.

What I'm less certain about is whether `close_pool()` was also needed after individual tasks for long-lived worker children, for example to handle stale pooled connections.

Do you remember whether the `PoolTimeout` issue in #9953 was specifically caused by the fork/inheritance case, or did it also occur in a long-lived child after multiple tasks?

If it was only the fork/inheritance issue, I think moving pool closure to child exit while keeping the existing `_close_database()` call during `worker_process_init()` should preserve the #9953 fix while allowing the pool to be reused between tasks. If task-level pool closure was also required for another reason, I'll need to account for that in the fix.

phanky1: > Do you remember whether the PoolTimeout issue in https://github.com/celery/celery/pull/9953 was specifically caused by the fork/inheritance case, or did it also occur in a long-lived child after multiple tasks?

I'm not the author of #9953, so I think this question is for you @auvipy.

t/unit/fixups/test_django.py
```diff
@@ -486,6 +485,34 @@ def test_django_setup(self, patching):
...
+        f._db.connections.all = Mock(return_value=[])
+
+        # Track atexit registrations
+        original_register = atexit.register
+        registered_handlers = []
+
+        def tracking_register(func, *args, **kwargs):
+            registered_handlers.append(func)
+            return original_register(func, *args, **kwargs)
+
```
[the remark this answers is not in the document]
auvipy: @Namraa310806 please cross check this suggestions

Namraa310806:
@auvipy @phanky1 whats your view on this?

auvipy:
first of all we need to be defensive and not remove everything that exists

auvipy:
I want to defer this for now and come back to this after v5.7 pre releases

celery/fixups/django.py
```diff
@@ -181,9 +182,34 @@ def on_worker_process_init(self, **kwargs: Any) -> None:
...
+        that created them, as connections cannot be shared across fork boundaries.
+        """
+        try:
+            connections = self._db.connections.all(initialized_only=True)
+        except TypeError:
+            # Support Django < 4.1
+            connections = self._db.connections.all()
+
+        for conn in connections:
+            self._close_pool(conn)
```
phanky1, line 211: I think for the prefork pool, the moment the child exits, the connection pool goes with it, so we don't need this explicit cleanup.

Namraa310806: @phanky1 That's a fair point — a process exit does mean the OS reclaims the underlying sockets regardless of whether we explicitly call close_pool(). Given that, plus Copilot's finding that this atexit hook doesn't actually fire on the real prefork shutdown path (os._exit()), it sounds like _close_pools_on_child_exit may be doing very little in practice today.

@auvipy since you mentioned deferring the atexit-firing fix until after v5.7 — would you rather I (a) leave _close_pools_on_child_exit in as best-effort coverage for exit paths that aren't os._exit() (non-prefork pools, graceful shutdown, tests), or (b) drop the explicit child-exit cleanup for now since it's redundant with OS-level reclamation and doesn't reliably run anyway, and revisit alongside the worker_process_shutdown signal wiring later? Don't want to guess and strip something you'd rather keep.

auvipy: I want to release billiard and kombu and other dependencies of celery then celery 5.7 alpha or beta and then I want to see if this fix is needed for v5.7 stable version, in the mean time, you can cross check the open suggestions

celery/fixups/django.py
```diff
@@ -181,9 +182,34 @@ def on_worker_process_init(self, **kwargs: Any) -> None:
...
-        self._close_database()
+        # Close inherited pools immediately after fork to preserve #9953 behavior
+        self._close_database(close_pool=True)
         self.close_cache()
 
+        # Register atexit handler to close connection pools when this child process exits.
+        # This runs in the child process (on_worker_process_init is called in each child),
+        # ensuring pools are closed in the process that created them.
+        if not getattr(self, "_pools_atexit_registered", False):
+            atexit.register(self._close_pools_on_child_exit)
```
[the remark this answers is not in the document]
auvipy: @Namraa310806 please cross check this

auvipy:
@MehrazRumman you can do a review of this later

MehrazRumman (changes requested):
Went through the diff, #10554 and the earlier threads (#9953, #10020). Summary of where I landed:

**The core change looks right.** Django's `DatabaseWrapper.close()` returns the connection to the pool via `putconn()` rather than closing it, so per-task `conn.close()` without `close_pool()` is exactly "release the connection, keep the pool". Tearing the pool down on every `task_prerun`/`task_postrun` defeated the purpose of `OPTIONS["pool"]` under prefork, and keeping `close_pool=True` in `on_worker_process_init` preserves the #9953 protection against pools inherited across `fork()`. 👍

**Blocking: the `atexit` handler never runs in a prefork child.** I checked Copilot's finding against billiard and it holds. `Worker.__call__` always ends in `_do_exit()`, which calls `self.on_exit(pid, exitcode)` and then `os._exit(exitcode)` ([billiard/pool.py](https://github.com/celery/billiard/blob/main/billiard/pool.py), `_do_exit`). `os._exit` skips `atexit`, and even the `_bootstrap` path in `popen_fork.py` ends in `os._exit`. `on_exit` is celery's `process_destructor`, which dispatches `signals.worker_process_shutdown` from inside the child ([celery/concurrency/prefork.py](https://github.com/celery/celery/blob/main/celery/concurrency/prefork.py)). So the hook you want already exists and fires exactly where you need it:

```python
def install(self) -> "DjangoWorkerFixup":
    ...
    signals.worker_process_init.connect(self.on_worker_process_init)
    signals.worker_process_shutdown.connect(self.on_worker_process_shutdown)
    ...

def on_worker_process_shutdown(self, **kwargs: Any) -> None:
    # process_destructor sends sender=None, pid=..., exitcode=...
    self._close_pools()
```

and drop `atexit` + `_pools_atexit_registered` entirely. That also removes the "registered on every call" concern and makes the PR description match the code.

**Follow-ups once that's in:**

- `_close_pool()` is called from the shutdown path without the `interface_errors` / `DatabaseError` guards that `_close_database()` has. Celery's `Signal.send` returns receiver exceptions instead of raising, so it won't crash the child, but a `try/except` there keeps shutdown quiet on a half-broken connection.
- The comment added in `test_install` ("worker_process_shutdown is no longer connected") is misleading: `main` never connected that signal. It goes away with the change above anyway.
- `test_multi_task_lifecycle_pool_persistence` and `test_double_close_pool_safety` call `_close_pools_on_child_exit()` directly, so they pass today even though the handler is dead on the real exit path. Once you're on `worker_process_shutdown`, sending the signal in the test (or, per @auvipy's request, a smoke test with a real prefork worker and `--max-tasks-per-child=1`) would actually exercise the lifecycle.

On @phanky1's point that the OS reclaims the sockets on `os._exit` anyway: agreed, nothing leaks server-side, Postgres sees the EOF and cleans up the backend. Closing the pool on `worker_process_shutdown` just turns that into a clean `Terminate` instead of "unexpected EOF on client connection" in the server log, and it's a one-line signal connect, so I'd keep it.

Happy to re-review once the signal wiring is in.

Issue #6734, opened 2021-04-19
Result object for groups containing chords which have a chain as their body time out


maybe-sybr:
# Checklist

- [x] I have verified that the issue exists against the `master` branch of Celery.
- [ ] This has already been asked to the [discussion group](https://groups.google.com/forum/#!forum/celery-users) first.
- [ ] I have read the relevant section in the
  [contribution guide](http://docs.celeryproject.org/en/latest/contributing.html#other-bugs)
  on reporting bugs.
- [x] I have checked the [issues list](https://github.com/celery/celery/issues?q=is%3Aissue+label%3A%22Issue+Type%3A+Bug+Report%22+-label%3A%22Category%3A+Documentation%22)
  for similar or identical bug reports.
- [x] I have checked the [pull requests list](https://github.com/celery/celery/pulls?q=is%3Apr+label%3A%22PR+Type%3A+Bugfix%22+-label%3A%22Category%3A+Documentation%22)
  for existing proposed fixes.
- [x] I have checked the [commit log](https://github.com/celery/celery/commits/master)
  to find out if the bug was already fixed in the master branch.
- [ ] I have included all related issues and possible duplicate issues
  in this issue (If there are none, check this box anyway).

## Mandatory Debugging Information

- [ ] I have included the output of ``celery -A proj report`` in the issue.
    (if you are not able to do this, then at least specify the Celery
     version affected).
- [ ] I have verified that the issue exists against the `master` branch of Celery.
- [ ] I have included the contents of ``pip freeze`` in the issue.
- [ ] I have included all the versions of all the external dependencies required
  to reproduce this bug.

## Optional Debugging Information

- [ ] I have tried reproducing the issue on more than one Python version
  and/or implementation.
- [ ] I have tried reproducing the issue on more than one message broker and/or
  result backend.
- [ ] I have tried reproducing the issue on more than one version of the message
  broker and/or result backend.
- [ ] I have tried reproducing the issue on more than one operating system.
- [ ] I have tried reproducing the issue on more than one workers pool.
- [ ] I have tried reproducing the issue with autoscaling, retries,
  ETA/Countdown & rate limits disabled.
- [ ] I have tried reproducing the issue after downgrading
  and/or upgrading Celery and its dependencies.

#### Related Issues

- None

#### Possible Duplicates

- None

## Environment & Settings

**Celery version**:

<details>
<summary><b><code>celery report</code> Output:</b></summary>
<p>

```
```

</p>
</details>

# Steps to Reproduce

[...]

PR #10571, opened 2026-09-07
Fix Signature.clone() sharing kwargs with the original signature

### LINKED ISSUE #10560: Signature.clone() aliases .kwargs with the original when called without override arguments
### Summary

`Signature.clone()`'s own inline comment says "need to deepcopy options so origins links etc. is not modified" — and `options` is indeed deep-copied — but the sibling field `kwargs` is NOT deep-copied when `clone()` is called with no override arguments (the common `sig.clone()` "just give me an independent copy" call). The clone ends up sharing the exact same `kwargs` dict object as the original.

### Root cause

`celery/canvas.py`, `Signature.clone()`:

```python
def clone(self, args=None, kwargs=None, **opts):
    args = args if args else ()
    kwargs = kwargs if kwargs else {}
    # need to deepcopy options so origins links etc. is not modified.
    if args or kwargs or opts:
        args, kwargs, opts = self._merge(args, kwargs, opts)
    else:
        args, kwargs, opts = self.args, self.kwargs, self.options   # <-- no copy at all here
    signature = Signature.from_dict({..., 'kwargs': kwargs, 'options': deepcopy(opts), ...}, app=self._app)
```

In the no-override branch, `kwargs = self.kwargs` is passed straight through to `from_dict` by reference, while `opts` two lines later is explicitly `deepcopy()`'d — an inconsistency between two fields that the surrounding comment implies should both be independent.

### Steps to Reproduce

```python
from celery import Celery

app = Celery(set_as_current=False)

@app.task
def add(x, y):
    return x + y

original = add.s(1, 2)
original.kwargs['extra'] = {'shared': True}

clone_a = original.clone()
clone_b = original.clone()

print(clone_a.kwargs is original.kwargs)   # True -- should be False
clone_a.kwargs['x'] = 999
print(original.kwargs)   # gained 'x': 999 -- corrupted by mutating a "clone"
print(clone_b.kwargs)    # also gained 'x': 999 -- 3-way aliasing
```

Output confirms `original`, `clone_a`, and `clone_b` all share the identical `kwargs` object (same `id()`); mutating one mutates all three.

This propagates into canvas internals that rely on `.clone()` for independence when fanning a signature out into siblings, e.g. `group.clone()` (`canvas.py`) and `_chain.unchain_tasks()` (`tasks = [t.clone() for t in self.tasks]`).

### Expected Behavior

`sig.clone()` called with no overrides should return a signature whose `kwargs` (and ideally `args`, for consistency) is independent of the original, matching the deep-copy treatment already given to `options`.

### Suggested fix

Mirror the existing `deepcopy(opts)` handling onto `kwargs` in the no-override branch, e.g. `args, kwargs, opts = self.args, deepcopy(self.kwargs), self.options`, or always route through `_merge()`.

### Environment

- Celery 5.6.3 (confirmed unchanged on `main` as of 2026-09-04)
- Python 3.12.10
- Windows 11

---
Found via a property-based audit (copy/clone independence) run across several widely-used Python libraries; repro was written and independently re-run before filing. Happy to open a PR if a maintainer confirms the fix direction.

tritsystem:
## Description

Fixes #10560.

`Signature.clone()` with no `kwargs` override hands the new signature the **same
`kwargs` dict object** as the source. `options` is defensively `deepcopy`-d on
clone; `kwargs` was passed straight through by reference:

```python
if args or kwargs or opts:
    args, kwargs, opts = self._merge(args, kwargs, opts)
else:
    args, kwargs, opts = self.args, self.kwargs, self.options   # <- by reference
... Signature.from_dict({..., 'kwargs': kwargs, 'options': deepcopy(opts), ...})
```

`_merge()` also returns `self.kwargs` unchanged when no `kwargs` override is
given (`dict(self.kwargs, **kwargs) if kwargs else self.kwargs`) and via its
immutable short-circuit, so `sig.clone(priority=5)`, `sig.clone(args=(1,))` and
`sig.si(...).clone()` alias too — only `sig.clone(kwargs={...})` copies today.

Mutating a clone mutates the source and every sibling clone. This reaches
`group.clone()` and `_chain.unchain_tasks()` (`[t.clone() for t in self.tasks]`),
which rely on `clone()` for independence when fanning a signature into siblings.

### Minimal repro

```python
from celery import Celery

app = Celery(set_as_current=False)

@app.task
def add(x, y):
    return x + y

original = add.s(1, 2)
clone_a = original.clone()
clone_b = original.clone()

assert clone_a.kwargs is original.kwargs   # True  -- should be False
clone_a.kwargs['x'] = 999
print(original.kwargs)   # {'x': 999}      -- corrupted by mutating a "clone"
print(clone_b.kwargs)    # {'x': 999}      -- 3-way aliasing
```

## Fix

Give the clone its own mapping when `kwargs` is still `self.kwargs`, mirroring the
copy already made for `options`:

```python
if kwargs is self.kwargs:
    kwargs = dict(kwargs)
```

Placed right before `from_dict` so it covers every path (no-override, option-only
/ args-only via `_merge`, immutable) without touching `_merge`'s other callers.

**A shallow `dict` copy, not `deepcopy`** — deliberate. Canvas primitives keep
live objects in `kwargs`: `chunks` / `xmap` / `xstarmap` hold a task `Signature`
and a lazy iterator, `_chord` holds its `body`. Deep-copying those would consume
generators and turn nested signatures into plain dicts. Shallow-copying the
mapping is enough to stop `clone.kwargs[...] = ...` from reaching the original,
and matches what `_merge` already does for the `kwargs`-override path
(`dict(self.kwargs, **kwargs)`).

## Tests

`t/unit/tasks/test_canvas.py`:

- `test_clone_does_not_alias_kwargs` — 3-way independence + mutation isolation
- `test_clone_does_not_alias_kwargs_with_non_kwargs_overrides` — `clone(priority=)`, `clone(countdown=)`, `clone(args=)`
- `test_clone_does_not_alias_kwargs_when_immutable` — `.si().clone()`
- `test_clone_kwargs_copy_is_shallow` — pins the shallow contract (a nested value keeps its identity)

Each fails on `main` without the one-line change and passes with it.

**`test_freeze_tasks_body_is_group` is modified.** It asserted that
`top_group.freeze()` replaced the *original* chord's `body.tasks` with cloned,
`group_index`-stamped signatures. That only happened because `_chord.clone()`
does `signature.kwargs['body'] = maybe_signature(..., clone=True)` into the dict
it *shared* with the source — i.e. via this bug. Freezing a container should not
mutate a signature the caller still holds a reference to. The test now asserts on
the frozen clone that lands in `top_group.tasks[0]`, with an added subtest that
the caller's original chord and its body elements are untouched. `group_index`
assignment on the frozen output is unchanged.

## Verification

`python -m pytest -q t/unit/tasks/test_canvas.py t/unit/tasks/test_stamping.py
t/unit/tasks/test_chord.py t/unit/tasks/test_result.py t/unit/app/test_builtins.py
t/unit/tasks/test_tasks.py` → **930 passed, 0 failed**. `flake8` + `isort` clean.

## Note re #10560 / #10561

@auvipy — you invited a fix with test coverage on the issue. #10561 took the
`deepcopy(kwargs)` route the issue text suggested and hit the generator-exhaustion
/ nested-signature / cycle regressions the Copilot reviewer flagged there; this PR
is the smaller shallow-copy alternative that sidesteps all of them. Happy to close
this if you'd rather land that one.

### DIFF
--- a/celery/canvas.py
+++ b/celery/canvas.py
@@ -459,6 +459,17 @@ class Signature(dict):
             args, kwargs, opts = self._merge(args, kwargs, opts)
         else:
             args, kwargs, opts = self.args, self.kwargs, self.options
+        # ``kwargs`` may still be ``self.kwargs`` by reference here: the
+        # no-override branch above, ``_merge`` returning ``self.kwargs``
+        # unchanged when no kwargs override is given, and its immutable
+        # short-circuit all pass it straight through.  Give the clone its
+        # own mapping so mutating ``clone.kwargs`` cannot corrupt the
+        # original (and sibling clones) -- #10560.  A shallow ``dict`` copy,
+        # not ``deepcopy``: canvas primitives keep live objects in kwargs
+        # (``chunks``/``xmap``/``xstarmap`` hold a task Signature and a lazy
+        # iterator) that must not be copied or consumed.
+        if kwargs is self.kwargs:
+            kwargs = dict(kwargs)
         signature = Signature.from_dict({'task': self.task,
                                          'args': tuple(args),
                                          'kwargs': kwargs,
--- a/t/integration/test_canvas.py
+++ b/t/integration/test_canvas.py
@@ -1387,7 +1387,6 @@ class test_group:
             [42, 42, *((42,) * gchild_count), 1337]
         ]
 
-    @pytest.mark.xfail(raises=TimeoutError, reason="#6734")
     def test_nested_group_chord_body_chain(self, manager):
         try:
             manager.app.backend.ensure_chords_allowed()
@@ -1397,19 +1396,16 @@ class test_group:
         child_chord = chord(identity.si(42), chain((identity.s(),)))
         group_sig = group((child_chord,))
         res = group_sig.delay()
-        # The result can be expected to timeout since it seems like its
-        # underlying promise might not be getting fulfilled (ref #6734). Pick a
-        # short timeout since we don't want to block for ages and this is a
-        # fairly simple signature which should run pretty quickly.
+        # #6734: the GroupResult's promise here used to never be fulfilled even
+        # though the child tasks resolved, so `res.get()` timed out. Root cause
+        # was `Signature.clone()` aliasing `.kwargs`: freezing the outer group
+        # cloned the chord via `_chord.clone()`, which reassigns
+        # `signature.kwargs['body']` -- into a dict shared with the original,
+        # so the group's result wiring tracked a stale body signature. Fixed by
+        # the clone de-aliasing in this PR.
         expected_result = [[42]]
-        with pytest.raises(TimeoutError) as expected_excinfo:
-            res.get(timeout=TIMEOUT / 10)
-        # Get the child `AsyncResult` manually so that we don't have to wait
-        # again for the `GroupResult`
         assert res.children[0].get(timeout=TIMEOUT) == expected_result[0]
         assert res.get(timeout=TIMEOUT) == expected_result
-        # Re-raise the expected exception so this test will XFAIL
-        raise expected_excinfo.value
 
     def test_callback_called_by_group(self, manager, subtests):
         if not manager.app.conf.result_backend.startswith("redis"):
--- a/t/unit/tasks/test_canvas.py
+++ b/t/unit/tasks/test_canvas.py
@@ -262,6 +262,49 @@ class test_Signature(CanvasCase):
         assert kwargs == {'foo': 1}
         assert options == {'task_id': 3}
 
+    def test_clone_does_not_alias_kwargs(self):
+        # gh #10560: clone() used to hand back a signature that shared the
+        # original's kwargs dict, so mutating a clone corrupted the source
+        # (and every sibling clone). Mirrors the defensive copy already made
+        # for options.
+        original = self.add.s(1, 2)
+        original.kwargs['shared'] = {'nested': True}
+
+        clone_a = original.clone()
+        clone_b = original.clone()
+        assert clone_a.kwargs is not original.kwargs
+        assert clone_b.kwargs is not original.kwargs
+        assert clone_a.kwargs is not clone_b.kwargs
+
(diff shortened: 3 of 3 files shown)


auvipy:
I would prefer this one.

tritsystem:
Heads-up on the last push (`70864e7`): the first CI run went red on one integration test, `t/integration/test_canvas.py::test_group::test_nested_group_chord_body_chain`, and it's an informative failure rather than a regression.

That test is an `xfail` pinning #6734 ("Result object for groups containing chords which have a chain as their body time out"). It builds `group((chord(identity.si(42), chain((identity.s(),))),))`, expects `res.get(timeout=short)` to raise `TimeoutError`, and re-raises it to force XFAIL.

On this branch it no longer times out — `res.get()` returns `[[42]]`, so `pytest.raises(TimeoutError)` fails with "DID NOT RAISE", which `xfail(raises=TimeoutError)` doesn't absorb.

The mechanism lines up with #6734's root cause: `group(...).delay()` freezes the group and clones the chord via `_chord.clone()`, which does `signature.kwargs['body'] = maybe_signature(body, clone=True)`. Before this PR, `super().clone()` returned a chord whose `.kwargs` was aliased with the original, so that body reassignment landed in a shared dict and the `GroupResult`'s promise wiring tracked a stale body signature — which never resolved. With the de-aliasing, each clone gets its own `kwargs`, the body links correctly, and the group result resolves.

`70864e7` drops the `xfail` marker and the timeout scaffolding from that test and asserts `res.get() == [[42]]` directly, with a comment explaining the link. If CI comes back green on the integration matrix, this PR also closes #6734.

I can't run the integration suite locally (needs a broker), so the "this fixes #6734" read rests on the failure being deterministic across all 8 broker/Python matrix cells plus the mechanism above — happy to be corrected if a maintainer sees it differently.

auvipy:
lets see if the integrations tests pass now. we have docker setup for running integrations tests more easily locally, if in case that is needed

auvipy (approved):

PR #10575, opened 2026-09-08
Keep preconfigured values out of configuration defaults

phanky1:
## Description

`detect_settings()` currently stores preconf values twice: once in the `changes` mapping and again in the `defaults` mapping. https://github.com/celery/celery/blob/de0806b701e51425186a5f49dd1827dc531722df/celery/app/utils.py#L280

Because of this, `app.conf.clear()` removes a preconfigured value from `changes`, but the same value is still returned from `defaults`. I believe the defaults mapping should contain only the built-in option `defaults`.

This PR removes the duplicate values so clearing the configuration correctly falls back to the built-in default.

### DIFF
--- a/celery/app/utils.py
+++ b/celery/app/utils.py
@@ -277,7 +277,7 @@ def detect_settings(conf, preconf=None, ignore_keys=None, prefix=None,
         )))
 
     preconf = {info.convert.get(k, k): v for k, v in preconf.items()}
-    defaults = dict(deepcopy(info.defaults), **preconf)
+    defaults = deepcopy(info.defaults)
     return Settings(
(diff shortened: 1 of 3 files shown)


celery/app/utils.py
```diff
@@ -277,7 +277,7 @@ def detect_settings(conf, preconf=None, ignore_keys=None, prefix=None,
         )))
 
     preconf = {info.convert.get(k, k): v for k, v in preconf.items()}
-    defaults = dict(deepcopy(info.defaults), **preconf)
```
auvipy, line 280: should we change any relevant documents for this?

auvipy (approved):

Issue #1317, opened 2013-04-22
Raise an error when registering a task with the same name


dcramer:
Possibly make the registry dict immutable once set.

In my case I had some code like this:

https://gist.github.com/dcramer/623a0617de2aade058fa

and the test() task was overwriting itself over and over due to

https://github.com/celery/celery/blob/3.0/celery/app/base.py#L190

PR #10579, opened 2026-09-08
Task registration compatibility fixes for the duplicate-name detector (#10598)

### LINKED ISSUE #10576: Two callables sharing __name__ in one module collapse into one registered task, silently, and the two registration paths disagree about which survives
## Summary

A task's registered name is `f"{fun.__module__}.{fun.__name__}"` — `__qualname__` is never consulted. Two distinct callables that share `__name__` within one module therefore collapse into a single registered task, and a caller invoking the second one receives `SUCCESS` and the **first function's result**.

The two registration paths resolve the collision in **opposite directions**, and both are silent.

## Where

`celery/app/base.py`, `_task_from_fun`:

```python
name = name or self.gen_task_name(fun.__name__, fun.__module__)
...
if name not in self._tasks:
    ...
else:
    task = self._tasks[name]      # silently returns the other function's task
```

`celery/utils/imports.py`, `gen_task_name`:

```python
return '.'.join(p for p in (module_name, name) if p)
```

`celery/app/registry.py`, `TaskRegistry.register`: a plain `self[task.name] = task`, no membership check.

## Reproduction

```python
from celery import Celery
app = Celery('repro', broker='memory://', backend='cache+memory://')

def make_scaler(factor):
    @app.task
    def scale(x):
        return x * factor
    return scale

double = make_scaler(2)
triple = make_scaler(3)

r = triple.delay(10)
print(double.name, triple.name)       # repro.scale repro.scale
print(double is triple)               # False  -- two distinct proxy objects
print(r.status, r.get())              # SUCCESS 20   expected 30
```

Observed on Celery 5.6.2 (`e48dc8e`):

```
repro.scale repro.scale
False
SUCCESS 20        expected 30
warnings raised during registration: []
```

Not an artefact of eager mode — the same through a real worker over the memory broker:

```
registered names ending .scale: ['wp.scale']
worker path: result=20 (expected 30)
```

A second form, without closures — two classes in one module with same-named methods:

```
Alpha.handler.name = repro.handler
Beta.handler.name  = repro.handler
Beta.handler.delay(1) = 'alpha:1'    expected 'beta:1'
__qualname__ differs: Alpha.handler != Beta.handler
```

## The two paths disagree

```python
class T1(Task):
    name = 'dup'
    def run(self, x): return 'first'

class T2(Task):
    name = 'dup'
    def run(self, x): return 'second'

app.register_task(T1()); app.register_task(T2())
app.tasks['dup'].delay(0).get()      # 'second'
```

The decorator path keeps the **first**; `register_task` keeps the **last**. Which function actually runs under a colliding name depends on which API registered it. Neither announces anything.

## Negative controls

```
# A: explicit name= -> correct pairing
t3.delay(10) = 30    correct

# B: distinct __name__ -> correct pairing (near miss)
scale_b.delay(10) = 30    correct
```

Control B is the near miss: identical factory shape, only `__name__` differs, and it correctly does not trigger — so the failure is attributable to the name key, not to closures.

## The silence

- `warnings.simplefilter('always')` with `catch_warnings(record=True)` around both registrations: `[]`
- `logging.basicConfig(level=DEBUG)` over the whole run: 68 lines, none matching `duplic|collis|overwrit|already|conflict`
- exit code 0, `r.status == 'SUCCESS'`
- the returned proxy is a *distinct object* from the first task while dispatching to its body, so an identity check by the caller does not reveal it either

## What the docs say

`docs/userguide/tasks.rst`:

> Every task class has a unique name, and this name is referenced in messages
>
> **Every task must have a unique name.**
>
> A best practice is to use the module name as a name-space, this way names won't collide if there's already a task with that name defined in another module.

Uniqueness is stated as a requirement of the system. The remedy offered addresses collisions *across* modules; the within-module case cannot be fixed by module namespacing and is silently mis-paired.

## Note

`celery/exceptions.py` defines a public, exported, documented exception for exactly this condition:

```python
class AlreadyRegistered(Exception):
    """The task is already registered."""   # XXX Unused
```

It is never raised anywhere in the tree.

`_task_from_fun` also copies `fun.__qualname__` onto the task object about twenty lines after computing the name without it, so the disambiguating information is present at the point where it would be needed.

## Scope

The collision requires two same-named callables in one module. That arises in practice with task factories, decorated methods on multiple classes, and generated or plugin task modules — it is not exotic, but it is not the common case either.

I have not checked whether this has been discussed before; a pointer to an existing issue is welcome and I am happy to close this in favour of one.

## Version

Celery 5.6.2, source at `e48dc8e`, Python 3.12, `memory://` broker, `cache+memory://` backend.

GruffElixir:
## Description

Follow-up to #10598, which added warnings for duplicate task names. This preserves that behavior for repeated registrations of the same bound method and lets `Celery.register_task()` resolve an unevaluated task proxy. App-local task definitions take precedence over same-named shared tasks when finalizing an app, regardless of decorator order.

Fixes #10576.

## Testing

- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python -m pytest -q t/unit/app/test_app.py` (187 passed)
- `python -m pre_commit run --files t/integration/tasks.py t/integration/test_tasks.py` (passed)
- `python -m compileall -q t/integration/tasks.py t/integration/test_tasks.py` (passed)
- The worker round-trip runs in Linux CI; the configured prefork worker cannot be started in this local Windows environment.

### DIFF
--- a/celery/app/__init__.py
+++ b/celery/app/__init__.py
@@ -66,6 +66,8 @@ def shared_task(*args, **kwargs):
             def task_by_cons():
                 app = _state.get_current_app()
                 task_name = getattr(fun, '__qualname__', fun.__name__)
+                if '<locals>' in task_name:
+                    task_name = fun.__name__
                 return app.tasks[
                     name or app.gen_task_name(task_name, fun.__module__)
                 ]
--- a/celery/app/__init__.py
+++ b/celery/app/__init__.py
@@ -52,21 +52,24 @@ def shared_task(*args, **kwargs):
             # Set as shared task so that unfinalized apps,
             # and future apps will register a copy of this task.
             _state.connect_on_app_finalize(
-                lambda app: app._task_from_fun(fun, **options)
+                lambda app: app._task_from_fun(fun, _shared=True, **options)
             )
 
             # Force all finalized apps to take this task as well.
             for app in _state._get_active_apps():
                 if app.finalized:
                     with app._finalize_mutex:
-                        app._task_from_fun(fun, **options)
+                        app._task_from_fun(fun, _shared=True, **options)
 
             # Return a proxy that always gets the task from the current
             # apps task registry.
             def task_by_cons():
                 app = _state.get_current_app()
+                task_name = getattr(fun, '__qualname__', fun.__name__)
+                if '<locals>' in task_name:
+                    task_name = fun.__name__
                 return app.tasks[
-                    name or app.gen_task_name(fun.__name__, fun.__module__)
+                    name or app.gen_task_name(task_name, fun.__module__)
                 ]
             return Proxy(task_by_cons)
         return __inner
--- a/celery/app/__init__.py
+++ b/celery/app/__init__.py
@@ -52,22 +52,20 @@ def shared_task(*args, **kwargs):
             # Set as shared task so that unfinalized apps,
             # and future apps will register a copy of this task.
             _state.connect_on_app_finalize(
-                lambda app: app._task_from_fun(fun, _shared=True, **options)
+                lambda app: app._task_from_fun(fun, **options)
             )
 
             # Force all finalized apps to take this task as well.
             for app in _state._get_active_apps():
                 if app.finalized:
                     with app._finalize_mutex:
-                        app._task_from_fun(fun, _shared=True, **options)
+                        app._task_from_fun(fun, **options)
 
             # Return a proxy that always gets the task from the current
             # apps task registry.
             def task_by_cons():
                 app = _state.get_current_app()
                 task_name = getattr(fun, '__qualname__', fun.__name__)
-                if '<locals>' in task_name:
-                    task_name = fun.__name__
                 return app.tasks[
                     name or app.gen_task_name(task_name, fun.__module__)
                 ]
--- a/celery/app/__init__.py
+++ b/celery/app/__init__.py
@@ -73,12 +73,13 @@ def shared_task(*args, **kwargs):
             # apps task registry.
             def task_by_cons():
                 app = _state.get_current_app()
+                if not app.finalized:
+                    app.finalize()
                 registered_name = task_names.get(app)
                 if registered_name is None:
-                    task_name = getattr(fun, '__qualname__', fun.__name__)
-                    if '<locals>' in task_name:
-                        task_name = fun.__name__
-                    registered_name = app.gen_task_name(task_name, fun.__module__)
+                    raise RuntimeError(
+                        'Shared task was not registered with the current app'
+                    )
                 return app.tasks[
                     name or registered_name
                 ]
--- a/celery/app/__init__.py
+++ b/celery/app/__init__.py
@@ -1,4 +1,6 @@
 """Celery Application."""
+from weakref import WeakKeyDictionary
+
 from celery import _state
 from celery._state import app_or_default, disable_trace, enable_trace, pop_current_task, push_current_task
 from celery.local import Proxy
@@ -49,27 +51,36 @@ def shared_task(*args, **kwargs):
 
         def __inner(fun):
             name = options.get('name')
+
+            task_names = WeakKeyDictionary()
+
+            def register(app):
+                task = app._task_from_fun(fun, **options)
+                task_names[app] = task.name
+                return task
+
             # Set as shared task so that unfinalized apps,
             # and future apps will register a copy of this task.
-            _state.connect_on_app_finalize(
-                lambda app: app._task_from_fun(fun, **options)
-            )
+            _state.connect_on_app_finalize(register)
 
             # Force all finalized apps to take this task as well.
             for app in _state._get_active_apps():
                 if app.finalized:
                     with app._finalize_mutex:
-                        app._task_from_fun(fun, **options)
+                        register(app)
 
             # Return a proxy that always gets the task from the current
             # apps task registry.
             def task_by_cons():
                 app = _state.get_current_app()
-                task_name = getattr(fun, '__qualname__', fun.__name__)
-                if '<locals>' in task_name:
-                    task_name = fun.__name__
+                registered_name = task_names.get(app)
+                if registered_name is None:
+                    task_name = getattr(fun, '__qualname__', fun.__name__)
+                    if '<locals>' in task_name:
+                        task_name = fun.__name__
+                    registered_name = app.gen_task_name(task_name, fun.__module__)
                 return app.tasks[
-                    name or app.gen_task_name(task_name, fun.__module__)
+                    name or registered_name
                 ]
             return Proxy(task_by_cons)
         return __inner
--- a/celery/app/base.py
+++ b/celery/app/base.py
@@ -84,6 +84,13 @@ BUILTIN_FIXUPS = {
 }
 USING_EXECV = os.environ.get('FORKED_BY_MULTIPROCESSING')
 
+
+def _using_execv():
+    # Also checked at call time: a spawned pool child only sets the
+    # variable from process_initializer(), after this module was imported.
+    return USING_EXECV or os.environ.get('FORKED_BY_MULTIPROCESSING')
+
+
 ERR_ENVVAR_NOT_SET = """
 The environment variable {0!r} is not set,
 and as such the configuration could not be loaded.
@@ -547,7 +554,7 @@ class Celery:
             not access any attributes on the returned object until the
             application is fully set up (finalized).
         """
-        if USING_EXECV and opts.get('lazy', True):
+        if _using_execv() and opts.get('lazy', True):
             # When using execv the task in the original module will point to a
             # different app, so doing things like 'add.request' will point to
             # a different task instance.  This makes sure it will always use
--- a/celery/app/base.py
+++ b/celery/app/base.py
@@ -561,7 +561,7 @@ class Celery:
             # the task instance from the current app.
             # Really need a better solution for this :(
             from . import shared_task
-            return shared_task(*args, lazy=False, **opts)
+            return shared_task(*args, lazy=False, _shared=True, **opts)
 
         def inner_create_task_cls(shared=True, filter=None, lazy=True, **opts):
             _filt = filter
--- a/celery/app/base.py
+++ b/celery/app/base.py
@@ -79,6 +79,56 @@ else:
     def _get_annotations(fun):
         return fun.__annotations__
 
+
+def _same_task_callable(first, second):
+    if first is second:
+        return True
+
+    first_self = getattr(first, '__self__', None)
+    second_self = getattr(second, '__self__', None)
+    if first_self is not None or second_self is not None:
+        return (
+            first_self is not None
+            and first_self is second_self
+            and getattr(first, '__func__', None) is getattr(second, '__func__', None)
+        )
+
+    if not inspect.isfunction(first) or not inspect.isfunction(second):
+        return False
+
+    first_code = first.__code__
+    second_code = second.__code__
+    if (
+        first.__module__, first.__qualname__, first_code.co_filename,
+        first_code.co_firstlineno,
+    ) != (
+        second.__module__, second.__qualname__, second_code.co_filename,
+        second_code.co_firstlineno,
+    ):
+        return False
+
+    if first.__defaults__ != second.__defaults__:
+        return False
+    if first.__kwdefaults__ != second.__kwdefaults__:
+        return False
+    if first.__closure__ is None or second.__closure__ is None:
+        return first.__closure__ is second.__closure__
+    if len(first.__closure__) != len(second.__closure__):
+        return False
+
+    for first_cell, second_cell in zip(first.__closure__, second.__closure__):
+        first_value = first_cell.cell_contents
+        second_value = second_cell.cell_contents
+        if first_value is second_value:
+            continue
+        try:
+            if not bool(first_value == second_value):
+                return False
+        except Exception:
+            return False
+    return True
+
+
 BUILTIN_FIXUPS = {
     'celery.fixups.django:fixup',
 }
@@ -617,27 +667,32 @@ class Celery:
         original_fun = fun
         name_provided = name is not None
         task_name = getattr(fun, '__qualname__', fun.__name__)
-        # Keep the historical name for local functions unless it collides with
-        # another callable. In that case, use the qualified name when it can
-        # disambiguate the callables; identical qualified names still fail
-        # loudly instead of silently reusing the first task.
-        default_task_name = fun.__name__ if '<locals>' in task_name else task_name
+        # Keep the historical name unless it collides with another callable.
+        # In that case, use the qualified name when it can disambiguate the
+        # callables; identical qualified names still fail loudly instead of
+        # silently reusing the first task.
+        default_task_name = fun.__name__
         name = name or self.gen_task_name(default_task_name, fun.__module__)
         base = base or self.Task
 
         task = self._tasks.get(name)
         if task is not None:
+            existing_fun = getattr(task, '_task_fun', None)
             if (not _shared and getattr(task, '_app', None) is self
-                    and getattr(task, '_task_fun', None) is not original_fun):
+                    and not _same_task_callable(existing_fun, original_fun)):
                 if not name_provided:
-                    existing_fun = getattr(task, '_task_fun', None)
                     existing_task_name = getattr(
                         existing_fun, '__qualname__',
                         getattr(existing_fun, '__name__', None),
                     )
                     qualified_name = self.gen_task_name(task_name, fun.__module__)
+                    qualified_task = self._tasks.get(qualified_name)
+                    if qualified_task is not None and _same_task_callable(
+                            getattr(qualified_task, '_task_fun', None), original_fun,
+                    ):
+                        return qualified_task
                     if (task_name != existing_task_name
-                            and qualified_name not in self._tasks):
+                            and qualified_task is None):
                         name = qualified_name
                         task = None
                 if task is not None:
@@ -680,6 +735,7 @@ class Celery:
             style task classes, you should not need to use this for
             new projects.
         """
+        task = maybe_evaluate(task)
         task = inspect.isclass(task) and task() or task
         if not task.name:
             task_cls = type(task)
--- a/celery/app/base.py
+++ b/celery/app/base.py
@@ -615,15 +615,37 @@ class Celery:
         if not self.finalized and not self.autofinalize:
             raise RuntimeError('Contract breach: app not finalized')
         original_fun = fun
+        name_provided = name is not None
         task_name = getattr(fun, '__qualname__', fun.__name__)
-        # Local functions do not have a stable qualified name across imports.
-        # Keep their historical task names and detect collisions below.
-        if '<locals>' in task_name:
-            task_name = fun.__name__
-        name = name or self.gen_task_name(task_name, fun.__module__)
+        # Keep the historical name for local functions unless it collides with
+        # another callable. In that case, use the qualified name when it can
+        # disambiguate the callables; identical qualified names still fail
+        # loudly instead of silently reusing the first task.
+        default_task_name = fun.__name__ if '<locals>' in task_name else task_name
+        name = name or self.gen_task_name(default_task_name, fun.__module__)
         base = base or self.Task
 
-        if name not in self._tasks:
+        task = self._tasks.get(name)
+        if task is not None:
+            if (not _shared and getattr(task, '_app', None) is self
+                    and getattr(task, '_task_fun', None) is not original_fun):
+                if not name_provided:
+                    existing_fun = getattr(task, '_task_fun', None)
+                    existing_task_name = getattr(
+                        existing_fun, '__qualname__',
+                        getattr(existing_fun, '__name__', None),
+                    )
+                    qualified_name = self.gen_task_name(task_name, fun.__module__)
+                    if (task_name != existing_task_name
+                            and qualified_name not in self._tasks):
+                        name = qualified_name
+                        task = None
+                if task is not None:
+                    raise AlreadyRegistered(
+                        f'Task {name!r} is already registered with a different callable. '
+                        'Use a unique task name.')
+
+        if task is None:
             if pydantic is True:
                 fun = pydantic_wrapper(self, fun, name, pydantic_strict, pydantic_context, pydantic_dump_kwargs)
 
@@ -648,13 +670,6 @@ class Celery:
             self._tasks[task.name] = task
             task.bind(self)  # connects task to this app
             add_autoretry_behaviour(task, **options)
-        else:
-            task = self._tasks[name]
-            if (not _shared and getattr(task, '_app', None) is self
-                    and getattr(task, '_task_fun', None) is not original_fun):
-                raise AlreadyRegistered(
-                    f'Task {name!r} is already registered with a different callable. '
-                    'Use a unique task name.')
         return task
 
     def register_task(self, task, **options):
--- a/celery/app/base.py
+++ b/celery/app/base.py
@@ -685,7 +685,9 @@ class Celery:
             task_cls = type(task)
             task.name = self.gen_task_name(
                 task_cls.__name__, task_cls.__module__)
-        if task.name in self._tasks and self._tasks[task.name] is not task:
+        existing_task = self._tasks.get(task.name)
+        if (existing_task is not None and existing_task is not task
+                and type(existing_task) is not type(task)):
             raise AlreadyRegistered(
                 f'Task {task.name!r} is already registered with a different task.')
         add_autoretry_behaviour(task, **options)
--- a/celery/app/base.py
+++ b/celery/app/base.py
@@ -569,7 +569,7 @@ class Celery:
             def _create_task_cls(fun):
                 if shared:
                     def cons(app):
-                        return app._task_from_fun(fun, _shared=True, **opts)
+                        return app._task_from_fun(fun, **opts)
 
                     cons.__name__ = fun.__name__
                     connect_on_app_finalize(cons)
@@ -609,41 +609,21 @@ class Celery:
         pydantic_strict: bool = False,
         pydantic_context: typing.Optional[typing.Dict[str, typing.Any]] = None,
         pydantic_dump_kwargs: typing.Optional[typing.Dict[str, typing.Any]] = None,
-        _shared: bool = False,
         **options,
     ):
         if not self.finalized and not self.autofinalize:
             raise RuntimeError('Contract breach: app not finalized')
         original_fun = fun
-        name_provided = name is not None
         task_name = getattr(fun, '__qualname__', fun.__name__)
-        # Keep the historical name for local functions unless it collides with
-        # another callable. In that case, use the qualified name when it can
-        # disambiguate the callables; identical qualified names still fail
-        # loudly instead of silently reusing the first task.
-        default_task_name = fun.__name__ if '<locals>' in task_name else task_name
-        name = name or self.gen_task_name(default_task_name, fun.__module__)
+        name = name or self.gen_task_name(task_name, fun.__module__)
         base = base or self.Task
 
         task = self._tasks.get(name)
-        if task is not None:
-            if (not _shared and getattr(task, '_app', None) is self
-                    and getattr(task, '_task_fun', None) is not original_fun):
-                if not name_provided:
-                    existing_fun = getattr(task, '_task_fun', None)
-                    existing_task_name = getattr(
-                        existing_fun, '__qualname__',
-                        getattr(existing_fun, '__name__', None),
-                    )
-                    qualified_name = self.gen_task_name(task_name, fun.__module__)
-                    if (task_name != existing_task_name
-                            and qualified_name not in self._tasks):
-                        name = qualified_name
-                        task = None
-                if task is not None:
-                    raise AlreadyRegistered(
-                        f'Task {name!r} is already registered with a different callable. '
-                        'Use a unique task name.')
+        if (task is not None and getattr(task, '_app', None) is self
+                and getattr(task, '_task_fun', None) is not original_fun):
+            raise AlreadyRegistered(
+                f'Task {name!r} is already registered with a different callable. '
+                'Use a unique task name.')
 
         if task is None:
             if pydantic is True:
--- a/celery/app/base.py
+++ b/celery/app/base.py
@@ -28,7 +28,7 @@ from vine import starpromise
 from celery import platforms, signals
 from celery._state import (_announce_app_finalized, _deregister_app, _register_app, _set_current_app, _task_stack,
                            connect_on_app_finalize, get_current_app, get_current_worker_task, set_default_app)
-from celery.exceptions import AlwaysEagerIgnored, ImproperlyConfigured, OperationalError
+from celery.exceptions import AlreadyRegistered, AlwaysEagerIgnored, ImproperlyConfigured, OperationalError
 from celery.loaders import get_loader_cls
 from celery.local import PromiseProxy, maybe_evaluate
 from celery.utils import abstract
@@ -602,11 +602,18 @@ class Celery:
         pydantic_strict: bool = False,
         pydantic_context: typing.Optional[typing.Dict[str, typing.Any]] = None,
         pydantic_dump_kwargs: typing.Optional[typing.Dict[str, typing.Any]] = None,
+        _shared: bool = False,
         **options,
     ):
         if not self.finalized and not self.autofinalize:
             raise RuntimeError('Contract breach: app not finalized')
-        name = name or self.gen_task_name(fun.__name__, fun.__module__)
+        original_fun = fun
+        task_name = getattr(fun, '__qualname__', fun.__name__)
+        # Local functions do not have a stable qualified name across imports.
+        # Keep their historical task names and detect collisions below.
+        if '<locals>' in task_name:
+            task_name = fun.__name__
+        name = name or self.gen_task_name(task_name, fun.__module__)
         base = base or self.Task
 
         if name not in self._tasks:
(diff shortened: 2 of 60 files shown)


auvipy (changes requested):
can you please cross check the unit test build failures?

GruffElixir:
I rechecked the unit failures against the latest head. The collision-guard issues were fixed in cce3632, 62c9c9a, and a4d37e2. The fresh workflow run completed successfully with all required jobs passing, including the Linux/Windows/PyPy unit matrix. The focused local suite also passes: 182 tests, with pre-commit clean. Could you please re-check and resolve the requested changes?

auvipy (commented):
thanks for fixing it. since it is an enhancement, it would be great to have a focused integration test as well

auvipy:
@GangEunzzang @phanky1 would be great if you can help with review

celery/app/__init__.py
```diff
@@ -52,21 +52,24 @@ def __inner(fun):
...
-                        app._task_from_fun(fun, **options)
+                        app._task_from_fun(fun, _shared=True, **options)
 
             # Return a proxy that always gets the task from the current
             # apps task registry.
             def task_by_cons():
                 app = _state.get_current_app()
+                task_name = getattr(fun, '__qualname__', fun.__name__)
+                if '<locals>' in task_name:
+                    task_name = fun.__name__
```
phanky1: Why do we preserve existing names for tasks defined inside functions, but change them for tasks defined inside classes? That seems inconsistent.

celery/app/base.py
```diff
@@ -609,14 +609,43 @@ def _task_from_fun(
...
+        # loudly instead of silently reusing the first task.
+        default_task_name = fun.__name__ if '<locals>' in task_name else task_name
+        name = name or self.gen_task_name(default_task_name, fun.__module__)
         base = base or self.Task
 
-        if name not in self._tasks:
+        task = self._tasks.get(name)
+        if task is not None:
+            if (not _shared and getattr(task, '_app', None) is self
+                    and getattr(task, '_task_fun', None) is not original_fun):
```
phanky1: @shared_task is affected by this issue too, I don't understand why you skip the check here.

celery/app/__init__.py
GruffElixir: Updated in f741552: the universal __qualname__ change was reverted after considering broker compatibility. Functions defined inside other functions retain the historical generated name; a qualified name is used only to disambiguate a real collision, while stable class owners remain qualified. This keeps existing local-task message names unchanged and the collision tests still fail loudly for identical factory call sites.

celery/app/base.py
GruffElixir: @phanky1 Agreed — `@shared_task` needs the collision guard too. Direct `@shared_task` registration now calls `_task_from_fun` without `_shared` in both the immediate and finalizer paths. The private `_shared` bypass is reserved for compatibility replication paths: `Celery.task` finalizers and the execv delegation that intentionally mirrors an existing task into another app. Follow-up `7f8177f1` covers the latter regression; the focused app, task, worker, integration, and pre-commit checks are green.

GruffElixir:
Added the focused integration regression requested in the review. The follow-up also makes autogenerated names consistently use __qualname__ and applies collision detection to @shared_task. Validation: app/registry unit files pass 174 tests, the new integration regression passes, and the existing task-registration coverage remains green.

phanky1:
I think this's a breaking change because in production there's a good chance some tasks will still be in the broker when the new Celery version is deployed. For tasks whose generated names change, the new workers won't be able to process them because they're still using the old names.

auvipy:
unit tests are failing now

GruffElixir:
@phanky1 Agreed. `f741552` reverts the universal generated-name change: local and nested functions keep their legacy names so messages already in the broker remain addressable. Qualified names are used for stable class-owner disambiguation, and the collision guard still raises for an identical factory call site. `a97d9e61` keeps shared_task proxies aligned with a collision-disambiguated registration, and follow-up `7f8177f1` also preserves compatibility in the execv `@app.task` path. The previous unit run exposed that path through repeated worker fixtures; the focused app, worker, integration, and pre-commit checks are now green, and CI is rerunning.

GruffElixir:
@auvipy I reproduced the failures from `b8ef55f`: legacy name assertions failed and shared finalizer callbacks collided after the universal qualified-name change. `f741552` restores legacy local names while keeping direct `@shared_task` collision detection; `a97d9e61` covers proxies when a real collision is disambiguated. The replacement run then exposed one remaining compatibility path: execv `@app.task` delegation also needs the shared-registration bypass. `7f8177f1` fixes that path and adds coverage. The focused app, task, worker, integration, and pre-commit checks are green; CI is rerunning.

auvipy:
unit tests are failing again...

GruffElixir:
@auvipy the execv delegation used by `@app.task` still registered through `shared_task` without the compatibility bypass, so worker fixtures hit `AlreadyRegistered` for `t.unit.conftest.add`. Commit `7f8177f1` fixes that path and adds regression coverage. The replacement run now has all six unit environments green;

celery/app/base.py
```diff
@@ -609,14 +609,43 @@ def _task_from_fun(
...
             raise RuntimeError('Contract breach: app not finalized')
-        name = name or self.gen_task_name(fun.__name__, fun.__module__)
+        original_fun = fun
+        name_provided = name is not None
+        task_name = getattr(fun, '__qualname__', fun.__name__)
+        # Keep the historical name for local functions unless it collides with
+        # another callable. In that case, use the qualified name when it can
+        # disambiguate the callables; identical qualified names still fail
+        # loudly instead of silently reusing the first task.
+        default_task_name = fun.__name__ if '<locals>' in task_name else task_name
```
GangEunzzang: A function defined in a class body at module level has no `<locals>` in its qualname, so it takes this branch and its task is renamed even when nothing collides:

```python
app = Celery('proj')

class Reports:
    @app.task
    def nightly(): ...
```
```
main:   proj.nightly
branch: proj.Reports.nightly
```

I ran a producer and a worker as separate processes against Redis with the solo pool: a message published by a `main` producer is rejected by a branch worker as unregistered, and the reverse fails the same way. Matching versions work. `task_routes` keyed on the old name also stops matching, and the task goes to the default queue without an error.

I understand the class-owner qualification is deliberate. Would keeping the existing name until something actually collides be an option here?

GangEunzzang (commented):
@GruffElixir @auvipy I tested `7f8177f1` against `main` at `918a7404` on macOS 15.6 with Python 3.12.13, using Redis for the cross-version checks. I agree that the silent collisions in #10576 are worth surfacing. I left notes on the specific lines; three more things that do not belong to a single line:

**The full unit suite passes here at `7f8177f1`:** 3921 passed, 39 skipped, 3 xfailed, against 3911 passed on `main` with the same skips. Everything below is a separate reproduction, not a suite failure.

**Reloading a task module raises.** `importlib.reload()` produces a new function object with the same qualname, which the new check reads as a colliding callable. With an already finalized app the `reload()` call itself raises `AlreadyRegistered`; with an unfinalized app the reload succeeds and the first access to the task raises instead. I saw this on three consecutive reloads. `main` accepts these registrations, though it keeps the original implementation. #1317, an older request for this same error, describes a task overwriting itself repeatedly rather than two different callables colliding — the reload case has that shape. I have not tested framework-specific reloaders.

**If the reload case is addressed, code identity is a trap.** Comparing code objects or source locations would treat `factory(2)` and `factory(3)` as the same definition: they share `__code__` and `(co_filename, co_firstlineno, co_qualname)` while returning different results. That is the exact shape behind #10576, so a rule built on code identity would hide the bug this PR is meant to catch.

Happy to re-run any of these against an updated branch.

celery/app/base.py
```diff
@@ -609,14 +609,43 @@ def _task_from_fun(
...
+                    and getattr(task, '_task_fun', None) is not original_fun):
+                if not name_provided:
+                    existing_fun = getattr(task, '_task_fun', None)
+                    existing_task_name = getattr(
+                        existing_fun, '__qualname__',
+                        getattr(existing_fun, '__name__', None),
+                    )
+                    qualified_name = self.gen_task_name(task_name, fun.__module__)
+                    if (task_name != existing_task_name
+                            and qualified_name not in self._tasks):
```
GangEunzzang: This condition only lets the qualified name be used while it is still free, so registering a function that the branch has already disambiguated raises:

```python
A, B = make_a(), make_b()      # two local functions, both named `handle`

app.task(A, shared=False, lazy=False)   # proj.handle
app.task(B, shared=False, lazy=False)   # proj.make_b.<locals>.handle
app.task(B, shared=False, lazy=False)   # AlreadyRegistered: Task 'proj.handle'
                                        # is already registered with a
                                        # different callable.
```

The third call finds `A` under the short name and falls through to the exception, without checking whether `qualified_name` is already held by `B` itself. `main` does not raise here, though it also maps `B` to `A` from the start, so this is a different failure rather than a regression.

celery/app/base.py
GangEunzzang: `is not` makes this an identity check, but not every callable is identity-stable. A bound method is rebuilt on each attribute access, so passing the same one twice fails it:

```python
class Svc:
    def handle(self, x):
        return x

svc = Svc()
svc.handle is svc.handle   # False
svc.handle == svc.handle   # True

app.task(svc.handle, shared=False, lazy=False)   # proj.Svc.handle
app.task(svc.handle, shared=False, lazy=False)   # AlreadyRegistered
```

On `main` the second call is a no-op and returns the same task, so re-registering a handler bound to an instance now raises where it used to be idempotent.

celery/app/base.py
```diff
@@ -657,6 +685,11 @@ def register_task(self, task, **options):
             task_cls = type(task)
             task.name = self.gen_task_name(
                 task_cls.__name__, task_cls.__module__)
+        existing_task = self._tasks.get(task.name)
+        if (existing_task is not None and existing_task is not task
```
GangEunzzang, line 745: `@app.task` hands back a `PromiseProxy`, not the task itself, so feeding that value into `register_task` trips both halves of this condition — the proxy is a different object than the registered task, and `type()` on it is `PromiseProxy` rather than the task class:

```python
@app.task
def add(x, y):
    return x + y

app.finalize()
app.register_task(add)   # AlreadyRegistered: Task 'proj.add' is already
                         # registered with a different task.
```

`main` accepts this. Registering a real task class twice is unaffected, since the two instances share a type.

celery/app/__init__.py
```diff
@@ -49,24 +51,36 @@ def create_shared_task(**options):
...
             # Return a proxy that always gets the task from the current
             # apps task registry.
             def task_by_cons():
                 app = _state.get_current_app()
+                registered_name = task_names.get(app)
+                if registered_name is None:
+                    task_name = getattr(fun, '__qualname__', fun.__name__)
+                    if '<locals>' in task_name:
+                        task_name = fun.__name__
+                    registered_name = app.gen_task_name(task_name, fun.__module__)
```
GangEunzzang: This fallback recomputes a name instead of using the one `register()` recorded, so it also runs for an app where that registration never completed — and then resolves to whatever else is sitting under the guessed name:

```python
def make(v):
    @shared_task
    def work():
        return v
    return work

w1, w2 = make(1), make(2)
app = Celery('proj', set_as_current=True)

app.finalize()   # AlreadyRegistered: Task 'proj.work' ...
w1()             # 2
w2()             # 2
```

Both functions share the qualname `make.<locals>.work`, so the loud failure is intended. But `w1`'s registration is the one that lost, and calling it returns `w2`'s result rather than reporting a missing task. On `main` both return `1`.

GruffElixir:
Thanks for the detailed repros, I’ve addressed the compatibility cases: legacy names stay unchanged until a real collision, disambiguated callables and repeated bound methods stay idempotent, shared-task proxies unwrap during register_task, and failed shared registrations no longer fall back to another task. Added coverage for reload-equivalent callables too; the app/registry suites are green (180 passed).

celery/app/base.py
GruffElixir: Fixed this too. `register_task()` unwraps the PromiseProxy before comparing and registering, and the regression test covers `app.task` followed by `register_task`.

celery/app/base.py
GruffElixir: Agreed. The old short name now stays in place until there is an actual collision. Class-owned methods only get qualified when needed, so Reports.nightly stays proj.nightly in the no-collision case.

celery/app/base.py
GruffElixir: Good catch. I now check the qualified slot before raising, so registering the already-disambiguated callable again returns the existing task instead of reporting a collision.

celery/app/base.py
GruffElixir: Fixed. Repeated access to the same instance method now compares the bound method owner and function, so svc.handle is idempotent.

celery/app/__init__.py
GruffElixir: Agreed, that fallback was unsafe. Shared-task finalization now resolves only the name recorded for the current app and raises if registration did not complete, so it cannot silently run another task.

GruffElixir:
The coverage gate caught a few untested compatibility branches in the callable comparison and shared-task lookup paths. Added focused regression coverage for bound methods, reload-equivalent functions, defaults/closures, task-class registration, and missing shared registrations. The app/registry suite is now 184 passed with 96% local coverage for celery.app.base; pre-commit is clean.

celery/app/__init__.py
GruffElixir: Yep, that was a real inconsistency. The branch now keeps the legacy short name for class-body callables too, and only switches to the qualified name when the short name actually collides. Added coverage for the module-level `Reports.nightly` case in `7b24bf4`.

celery/app/base.py
GruffElixir: Agreed. `@shared_task` now goes through the same callable comparison path as `app.task`, including the lazy proxy bookkeeping. If finalization cannot register a shared task because its name collides, that proxy stays unregistered instead of resolving to another task.

celery/app/base.py
GruffElixir: Yep, good catch. The naming flow now keeps `proj.nightly` unless another callable already owns that name; the class-qualified name is only a fallback for a real collision. Added regression coverage in `7b24bf4`.

celery/app/base.py
GruffElixir: Fixed. When the short name is occupied, the code now checks the qualified name before raising, so registering the same already-disambiguated callable again returns the existing task.

celery/app/base.py
GruffElixir: Fixed this too. Bound methods are compared by their bound instance and underlying function, so repeated `svc.handle` registration stays idempotent.

celery/app/base.py
GruffElixir: Fixed. `register_task()` evaluates a `PromiseProxy` before comparing it with the existing task, while repeated real task classes remain accepted.

celery/app/__init__.py
GruffElixir: Fixed. The shared-task proxy now keeps the name assigned during registration. If registration fails, calling that proxy reports that it was not registered with the current app instead of falling through to another task with the guessed name.

GruffElixir:
@auvipy Yep, added the focused integration coverage in `t/integration/test_tasks.py` for two different callables with the same generated name. I also exported the integration pytest fixtures it needs, so the test runs with the normal setup now. The unit slice plus this integration case is 186 passed locally, and pre-commit is clean. Pushed in `5415ed3`.

auvipy:
@GangEunzzang it would be great if you can try and review again the updated codes

auvipy:
the unit tests are failing. we got another pr which seems to be following a different approach, https://github.com/celery/celery/pull/10598 . should we consider both approach or only one? or mix both approach?

GangEunzzang (commented):
@auvipy @GruffElixir I rechecked #10579 at `7b8f2fe79` and #10598 at `b1fda2902`, against main at `5ea4a4919`, on macOS with Python 3.12. After comparing both, I lean toward #10598's warning approach. I'd keep the compatibility cases from #10579 when checking the detector. Keeping the existing task names avoids adding another source of disagreement between producers and workers. The warning surfaces a collision; explicit unique `name=` values are still needed to make both callables addressable.

The cases from my earlier review now behave as expected on #10579: `A.handle` keeps its existing name without a collision, repeated registration of a disambiguated callable or bound method is idempotent, and `register_task()` accepts the task proxy. I also reran actual module reloads before and after finalization, and checked that a failed shared-task proxy no longer resolves to the other callable.

There are still two cases that make me lean toward the warning approach:

**Equal captured values do not imply the same callable.** The new reload comparison accepts two closures capturing separate empty lists. Once the second list changes, both tasks still execute the first closure:

```python
from celery import Celery

app = Celery('proj')

def make(state):
    def read():
        return state.copy()
    return read

left, right = [], []
f1, f2 = make(left), make(right)
t1 = app.task(f1, shared=False, lazy=False)
t2 = app.task(f2, shared=False, lazy=False)
right.append('second')

print(f1(), f2())  # [] ['second']
print(t1(), t2())  # [] []
```

This is a silent-collision case on main too. The earlier `7f8177f1` version detected it; the new value comparison no longer does. #10598 emits a warning for this case while retaining the existing behavior.

**The fallback names depend on registration order.** With two same-named static methods, I ran a solo worker using the memory broker, with separate producer and worker apps in one process. Registering A then B in the producer, but B then A in the worker, gives:

| | A.handle | B.handle |
|---|---|---|
| Producer | `proj.handle` | `proj.B.handle` |
| Worker | `proj.A.handle` | `proj.handle` |

Sending A's name returns B's result; sending B's name raises `NotRegistered`. The wrong result through the short name also occurs on main, while the missing qualified name is an additional failure from disambiguation. Matching registration order works, and explicit unique names work even with the reversed order.

#10598 has compatibility cases to address too: registering the same function twice with `bind=True`, the same instance's bound method twice, or the existing task proxy through `register_task()` produces a duplicate warning even though the callable has not changed. For the `bind=True` case, `task.__wrapped__` is a bound method rather than the original function, so the identity check does not recognize it.

Separately, I reproduced the current #10579 CI failure: `test_task_execv_shared_finalizer_does_not_collide` patches `_using_execv`, which no longer exists; the current path reads `USING_EXECV`. That failure happens before the collision behavior is exercised.

The local app/tasks/worker run had that one additional failure on #10579. All three trees also hit the same local setup issues: a missing Redis dependency and two CLI setup errors caused by the missing `isolated_cli_runner` fixture. These were separate from the reproductions above.

phanky1:
> the unit tests are failing. we got another pr which seems to be following a different approach, #10598 . should we consider both approach or only one? or mix both approach?

I think we should just let the last one override because it matches the current behaviour of periodic tasks.
https://github.com/celery/celery/blob/5ea4a4919aec6be425bf7103a18d011ffea1c5f1/celery/app/base.py#L1310-L1312

GruffElixir:
Thanks for the recheck. I fixed the CI-only USING_EXECV patch target in 896004528; the focused regression and app/registry slices pass, and the current branch checks are green. I agree that #10598's warning behavior is the better final direction. The remaining compatibility cases from this PR can be carried over there if useful, so I'm happy for maintainers to choose #10598 as the merge vehicle.

auvipy:
after merging that, we can fix merge conflicts here and keep the useful parts of this pr so dont be dishearted !

auvipy:
> @auvipy @GruffElixir I rechecked #10579 at `7b8f2fe79` and #10598 at `b1fda2902`, against main at `5ea4a4919`, on macOS with Python 3.12. After comparing both, I lean toward #10598's warning approach. I'd keep the compatibility cases from #10579 when checking the detector. Keeping the existing task names avoids adding another source of disagreement between producers and workers. The warning surfaces a collision; explicit unique `name=` values are still needed to make both callables addressable.
> 
> The cases from my earlier review now behave as expected on #10579: `A.handle` keeps its existing name without a collision, repeated registration of a disambiguated callable or bound method is idempotent, and `register_task()` accepts the task proxy. I also reran actual module reloads before and after finalization, and checked that a failed shared-task proxy no longer resolves to the other callable.
> 
> There are still two cases that make me lean toward the warning approach:
> 
> **Equal captured values do not imply the same callable.** The new reload comparison accepts two closures capturing separate empty lists. Once the second list changes, both tasks still execute the first closure:
> 
> ```python
> from celery import Celery
> 
> app = Celery('proj')
> 
> def make(state):
>     def read():
>         return state.copy()
>     return read
> 
> left, right = [], []
> f1, f2 = make(left), make(right)
> t1 = app.task(f1, shared=False, lazy=False)
> t2 = app.task(f2, shared=False, lazy=False)
> right.append('second')
> 
> print(f1(), f2())  # [] ['second']
> print(t1(), t2())  # [] []
> ```
> 
> This is a silent-collision case on main too. The earlier `7f8177f1` version detected it; the new value comparison no longer does. #10598 emits a warning for this case while retaining the existing behavior.
> 
> **The fallback names depend on registration order.** With two same-named static methods, I ran a solo worker using the memory broker, with separate producer and worker apps in one process. Registering A then B in the producer, but B then A in the worker, gives:
> 	A.handle 	B.handle
> Producer 	`proj.handle` 	`proj.B.handle`
> Worker 	`proj.A.handle` 	`proj.handle`
> 
> Sending A's name returns B's result; sending B's name raises `NotRegistered`. The wrong result through the short name also occurs on main, while the missing qualified name is an additional failure from disambiguation. Matching registration order works, and explicit unique names work even with the reversed order.
> 
> #10598 has compatibility cases to address too: registering the same function twice with `bind=True`, the same instance's bound method twice, or the existing task proxy through `register_task()` produces a duplicate warning even though the callable has not changed. For the `bind=True` case, `task.__wrapped__` is a bound method rather than the original function, so the identity check does not recognize it.
> 
> Separately, I reproduced the current #10579 CI failure: `test_task_execv_shared_finalizer_does_not_collide` patches `_using_execv`, which no longer exists; the current path reads `USING_EXECV`. That failure happens before the collision behavior is exercised.
> 
> The local app/tasks/worker run had that one additional failure on #10579. All three trees also hit the same local setup issues: a missing Redis dependency and two CLI setup errors caused by the missing `isolated_cli_runner` fixture. These were separate from the reproductions above.

@GruffElixir now we can trim down this pr after the other one is merged

GruffElixir:
Yep, got this trimmed down and pushed in 572eefb54. I merged the fork's update onto current main and kept #10598's warning path, plus the compatibility pieces that still apply: existing task names stay unchanged, register_task() evaluates task proxies, pending tasks are processed before shared finalizers, and re-registering the same bound method stays quiet. Local app/registry tests pass (189), the focused integration regression passes, and the substantive pre-commit hooks are clean. CI is running now.

GruffElixir:
Quick CI note: the latest run got through all unit jobs and the task/worker integration coverage. The remaining reds are in the existing backend-fixture leak test (27 open handles vs the <20 threshold) and Redis smoke shutdown tests timing out while waiting for the Restoring ... unacknowledged message(s) log. I don’t see a failure in the task-registration paths changed here, and GitHub won’t let me rerun the failed jobs from this account because repo reruns require admin access. Could someone with maintainer access rerun those jobs? If they reproduce on a clean run, I’ll dig in.

auvipy:
oh the test failures are existing!

GruffElixir:
Yep, confirmed on the retry too. The failures are still the Redis visibility-timeout smoke tests waiting for a log line, and the task jobs are canceled after fail-fast. The registration tests plus lint and security jobs are green, so I don't see a PR-side change to make here.

auvipy:
those failures should be now fixed in the main branch and with my recent push

auvipy:
@phanky1 @GangEunzzang @MehrazRumman need the final review or check

GangEunzzang (commented):
One thing: moving `_announce_app_finalized` after the pending tasks has a real effect, an `app.task` now beats a `shared_task` with the same name regardless of definition order (on `main` it depends on the order), but no test pins it, both new finalizer tests pass on `main` too. Worth a test for that. The bound-method and `register_task(proxy)` fixes check out against `main`; otherwise looks good.

MehrazRumman (changes requested):
## Review

Context for anyone reading top-to-bottom: this PR no longer does what its title and description say. Since #10598 landed with the warning approach, the branch was trimmed (`572eefb`, `9961ead`) to a set of compatibility fixes on top of that detector. The description still promises qualified-owner names and `AlreadyRegistered`, neither of which is in the diff. **Please rewrite the title and description before this goes any further** — something like "Task registration compatibility fixes for the duplicate-name detector (#10598)". Reviewers are currently reading a spec that doesn't match the code, and whatever is in the description tends to end up in release notes.

What the diff actually does now, in `celery/app/base.py`:

1. adds `_same_task_callable()` and uses it in `_warn_if_duplicate_task_name`, so re-registering the same bound method doesn't warn;
2. `register_task()` calls `maybe_evaluate(task)` first, so a shared-task proxy is accepted;
3. `finalize()` moves `_announce_app_finalized(self)` from before the `_pending` drain to after it;
4. plus three unrelated lines in `docs/userguide/signals.rst`.

(1) and (2) are small and well targeted. (3) is the one I'd hold this on. (4) should go.

---

### 1. `_same_task_callable` has a hole when `__func__` is absent

```python
first_self = getattr(first, '__self__', None)
second_self = getattr(second, '__self__', None)
return (
    first_self is not None
    and first_self is second_self
    and getattr(first, '__func__', None) is getattr(second, '__func__', None)
)
```

If neither object has `__func__`, both `getattr` calls return `None` and `None is None` is `True`. Builtin and C-level methods have `__self__` but no `__func__`, so any two *different* methods bound to the same object compare equal and the duplicate warning is suppressed:

```python
lst = []
_same_task_callable(lst.append, lst.pop)   # True
```

Contrived for task registration, but it's a silent wrong answer in exactly the code path whose job is to stop silent wrong answers. One-line fix:

```python
first_func = getattr(first, '__func__', None)
return (
    first_self is not None
    and first_self is second_self
    and first_func is not None
    and first_func is getattr(second, '__func__', None)
)
```

Two smaller things:

* The docstring ("whether two task callables represent the same callable") doesn't say *why* identity isn't enough. Worth one line: attribute access on an instance creates a fresh bound-method object each time, so `service.handle is service.handle` is `False` even though it's the same callable. That's the whole reason this function exists.
* `functools.partial` wrappers and `__wrapped__` chains aren't covered. Probably fine to leave out, but say so in the docstring so the next person doesn't assume it's meant to be exhaustive.

### 2. `register_task(maybe_evaluate(task))` — please note the laziness change

Accepting a proxy is right, and `test_register_task_accepts_shared_task_proxy` pins it. But `maybe_evaluate` *forces* an unevaluated `PromiseProxy`, so `app.register_task(some_pending_proxy)` before finalization now builds the task immediately instead of leaving it pending. Any error that `_task_from_fun` would have raised at finalize time now surfaces at `register_task` time. That's probably the behaviour you want from an explicit `register_task()` call, but it's a semantic change that isn't mentioned anywhere and isn't covered — the new test finalizes first, so it only exercises the already-evaluated path. Either add a case for the pre-finalize proxy or state the intent in a comment.

While you're on that line, the neighbouring idiom is worth tidying since you're touching it:

```python
task = inspect.isclass(task) and task() or task   # breaks on a falsy instance
task = task() if inspect.isclass(task) else task  # equivalent, safe
```

### 3. Moving `_announce_app_finalized` after the pending drain — needs justification and a test

This is the highest-risk line in the diff and the least defended. @GangEunzzang already flagged the visible effect: `app.task` now beats a same-named `shared_task` regardless of definition order, where on `main` the winner depends on order — **and both new finalizer tests pass on `main` too**, so nothing in this PR actually pins the change. As it stands a future refactor can move the call back and CI stays green. Please add a test that registers a `@shared_task` and an `@app.task` under the same name in both orders, asserts which one is registered, and fails on `main`.

Two further things I'd want answered in the PR before this merges:

* **Nothing re-drains `_pending` after the announce.** `_announce_app_finalized` runs every `_on_app_finalizers` callback, and those callbacks register tasks. Today that's safe only because `self.finalized` is already `True`, so `_task_from_fun` registers eagerly instead of appending to `_pending`. That's an invariant the new ordering now depends on and no comment records it. Either drain in a loop until both are exhausted, or add a comment stating the invariant.
* **Is any finalizer callback expected to run *before* pending tasks are evaluated?** Fixups and third-party libraries hook `connect_on_app_finalize`; if one of them sets something up that a pending task's construction reads, this reorders it out from under them. Worth a sentence on why that's safe.

Also: this belongs in its own commit with its own rationale. Right now it's buried in a merge commit ("Merge main and preserve task registration compatibility"), which is not a place a behaviour change should be hiding.

### 4. `docs/userguide/signals.rst` is unrelated

Three lines of reindentation in the `runtime` description under `task_success`. Nothing to do with task registration — looks like a merge artifact. Please drop it; if the reindent is genuinely wanted, it's a separate one-line PR.

---

### Tests

**`t/integration/test_tasks.py::test_task_registration_rejects_colliding_callables`** has several problems:

* **The name is stale.** It doesn't reject anything any more; it warns. `test_task_registration_warns_on_colliding_callables`.
* **It permanently pollutes a session-scoped fixture.** `celery_session_app` is session-scoped, and the test registers `…duplicate` into it, then overrides it. Every later test in the session inherits that registry entry, and the ordering effect is invisible until something else breaks. Use the function-scoped `celery_app`, or register/clean up explicitly.
* **It isn't an integration test.** There's no `.delay()`, no worker, no broker round-trip — `first.run() == 1` is a plain local call. @auvipy asked for focused integration coverage; the interesting integration claim is "after a collision, sending the name executes the *other* callable", and that needs `celery_session_worker` and an actual dispatch. As written this is a unit test paying integration-suite setup costs, and it's the only reason the `conftest.py` fixture exports exist.
* **The assertion is weak.** `first.run()` calls the closure directly and would pass even if registration were completely broken. Assert on `celery_session_app.tasks[name]` — that's where the collision is observable.

My suggestion: move this case into `t/unit/app/test_app.py`, drop the `t/integration/conftest.py` change entirely, and *separately* add a real integration test that dispatches through a worker if you want to keep integration coverage.

On **`t/integration/conftest.py`**: re-exporting `celery_parameters`, `celery_worker_parameters` and `use_celery_app_trap` shadows the `celery.contrib.pytest` plugin's own fixtures in that directory. If the plugin is loaded these already exist, and the local copies will silently diverge from the plugin defaults if those ever change. What actually failed without them? If the answer is "nothing once the test moves to the unit suite", this file shouldn't change.

On the **unit tests**, mostly good — `test_task_names_preserve_legacy_name_without_collision` and `test_task_registration_allows_repeated_bound_method` pin real behaviour. Smaller notes:

* The `finalizers = set(_state._on_app_finalizers)` / `_state._on_app_finalizers = finalizers` save-restore appears twice and rebinds the module attribute rather than restoring the original set in place. Mutating in place (`clear()` + `update()`) is safer against anything holding a direct reference, and this should be a fixture rather than duplicated `try/finally` blocks.
* `test_task_execv_shared_finalizer_does_not_collide` uses `set_as_current=True` on two nested apps without `depends_on_current_app`, unlike the test right above it. That leaks current-app state on failure.
* `test_task_registration_allows_repeated_bound_method` would read better as `warnings.simplefilter('error', DuplicateTaskNameWarning)` inside `catch_warnings()` — the assertion becomes "this doesn't raise" instead of filtering a recorded list.
* Neither `_same_task_callable` edge case (different instances of the same class; a plain function vs. its bound form) is covered. Both are cheap to add and they're the boundary of the new helper.

### Housekeeping

* 25 commits, 11 of them merges from `main`. Assuming maintainers squash that's cosmetic, but the history makes it genuinely hard to see what changed when — which is how the `finalize()` reordering ended up inside a merge commit.
* No Changelog entry. Since #10598 introduced `DuplicateTaskNameWarning` as user-visible behaviour, "bound methods and re-registered proxies no longer warn" is worth a line, and so is the `app.task`-beats-`shared_task` precedence if (3) stays.
* @auvipy's "requested changes" review from Sep 9 is still outstanding and blocking; it predates the rewrite and is probably safe to dismiss, but someone should do that explicitly.
* Codecov: all modified lines covered, project 89.15% (+<0.01%). Good — though as noted, "covered" and "pinned" aren't the same thing for the `finalize()` change.

---

**Verdict:** items 1 and 2 are fine with the small fixes above. Item 3 needs a rationale, a comment on the `_pending` invariant, its own commit, and a test that fails on `main`. Item 4 should be dropped, and the integration test should either become a real integration test or move to the unit suite along with reverting the conftest change. And please update the title and description first — right now they describe a design this PR abandoned a week ago.

GruffElixir:
Thanks for the detailed review, the title and description now match the follow-up to #10598, and the unrelated `signals.rst` change is gone. The bound-method and proxy cases have focused tests; the finalization-order change is in its own commit (`682102f5a`) with both decorator orders covered. I checked the in-tree finalizers: they only register shared/built-in tasks, and since `finalized` is set before they run, those registrations are eager and can't requeue `_pending`. The integration regression now does a worker round-trip and removes its temporary task/duplicate-warning entry afterward. The app unit suite is 187 passed and pre-commit is clean. The integration worker doesn't start with this Windows/prefork setup, so that check is pending CI.

GruffElixir:
CI caught the regression task being registered after the session prefork worker had started, so child processes raised NotRegistered. The task now lives in the integration task module loaded before worker startup. The 187-test app suite, pre-commit, and compile checks pass locally; the updated run is checking the worker round-trip.

docs/history/whatsnew-5.6.rst
```diff
@@ -193,6 +193,14 @@ Also added --skip-checks flag to bypass django core checks.
 News
 ====
 
+Task Registration Compatibility
```
auvipy, line 196: this will need to 5.7... will move when I create a new whatsnew for 5.7 next... unttill then lets keep it

GruffElixir:
Thanks again for the thorough review. The latest head has the requested fixes, and CI is green (376 passing, 1 skipped). I also checked the precedence case on base vs. this branch: base picks the shared task, while this branch consistently keeps the app task. Could you have another look when you get a chance?

GruffElixir:
Added a both-orders regression test that checks which task is registered after finalization, and documented why callbacks cannot refill `_pending`. The pre-finalize proxy case is covered too, and CI is green.

GangEunzzang (commented):
@GruffElixir Thanks, the new test fails once I revert the finalize move, so it covers that now. Small thing: built-ins like `celery.chord` also get replaced now if someone defines a `shared=False` task with the same name. Might be worth a line in whatsnew. LGTM otherwise.

PR #10582, opened 2026-09-09
fix#8182 Chaining a single group will not wait for group member completion 

### LINKED ISSUE #8182: Chaining a single group will not wait for group member completion
# Checklist

- [x] This has already been asked to the [discussions forum](https://github.com/celery/celery/discussions) first. -> https://github.com/celery/celery/discussions/8170
  on reporting bugs.
  for existing proposed fixes.

     version affected).
  to reproduce this bug.

  and/or implementation.
  result backend.
  broker and/or result backend.

#### Related Issues

- None

#### Possible Duplicates

- None

## Environment & Settings

**Celery version**:

<details>
<p>

```
software -> celery:5.3.0b2 (dawn-chorus) kombu:5.3.0b3 py:3.10.7
            billiard:4.1.0 py-amqp:5.1.1
            kernel version:5.15.94-1.qubes.fc32.x86_64 imp:CPython
settings -> transport:pyamqp results:redis:///

result_backend: 'redis:///'
deprecated_settings: None
```

</p>
</details>

# Steps to Reproduce

### Python Packages

<details>
<p>

```
amqp==5.1.1
asttokens==2.2.1
async-timeout==4.0.2
backcall==0.2.0
billiard==4.1.0
celery @ git+https://github.com/celery/celery.git@2e168af493fb741e9ddc7261780a0070927351ed
click==8.1.3
click-didyoumean==0.3.0
click-plugins==1.1.1
click-repl==0.2.0
decorator==5.1.1
executing==1.2.0
flower==1.2.0
humanize==4.6.0
ipdb==0.13.13
ipython==8.11.0
jedi==0.18.2
kombu==5.3.0b3
matplotlib-inline==0.1.6
parso==0.8.3
pexpect==4.8.0
pickleshare==0.7.5
prometheus-client==0.16.0
prompt-toolkit==3.0.38
ptyprocess==0.7.0
pure-eval==0.2.2
Pygments==2.14.0
pytz==2022.7.1
redis==4.5.1
six==1.16.0
stack-data==0.6.2
tomli==2.0.1
tornado==6.2
traitlets==5.9.0
vine==5.0.0
watchdog==2.3.1
wcwidth==0.2.6
```

</p>
</details>

### Other Dependencies

<details>
<p>
N/A
</p>
</details>

<details>
<p>

```python
#!/usr/bin/env python3
from random import randint
from time import sleep
from celery import Celery, Signature, chain, chord, group
from celery.utils.log import  get_task_logger

app2 = Celery('app2', backend='redis://', broker='pyamqp://')
logger = get_task_logger(__name__)

@app2.task
def t1(*args):
    logger.info(f"{args[-1]}({args[:-1]})")
    return args

@app2.task
def t2(*args):
    logger.info(f"start {args[-1]}({args[:-1]})")
    sleep(randint(1, 5))
    logger.info(f"end {args[-1]}({args[:-1]})")
    return args

if __name__ == "__main__":
    chain(
        group(
            t2.s("group1"),
            chain(
                group(
                    t1.s("group2->chain1->group1"),
                    t1.s("group2->chain1->group2"),
                ),
            ),
        ),
        t1.s("chord"),
    ).delay()
```

</p>
</details>

# Expected Behavior

<details><summary>The chord runs **after** the group1 task finishes:</summary>
<p>

```
[2023-03-31 10:59:39,365: INFO/ForkPoolWorker-2] Task app2.t2[592939f3-22bb-4e76-9b28-9a70b90a4f05] succeeded in 5.018022193999059s: ('group1',)
[2023-03-31 10:59:39,365: INFO/MainProcess] Task app2.t1[ad551c2a-e7a0-4191-9eea-983b83f99ada] received
[2023-03-31 10:59:39,366: INFO/ForkPoolWorker-2] app2.t1[ad551c2a-e7a0-4191-9eea-983b83f99ada]: chord(([['group1'], [['group2->chain1->group1'], ['group2->chain1->group2']]],))
[2023-03-31 10:59:39,367: INFO/ForkPoolWorker-2] Task app2.t1[ad551c2a-e7a0-4191-9eea-983b83f99ada] succeeded in 0.0007412640006805304s: ([['group1'], [[...], [...]]], 'chord')
```

</p>
</details> 

# Actual Behavior

The chord runs **before** the group1 task finishes:

```
[2023-03-31 10:40:25,661: INFO/MainProcess] Task app2.t1[4e447663-d92e-46fe-923a-3199b353df28] received
[2023-03-31 10:40:25,662: INFO/ForkPoolWorker-3] app2.t1[4e447663-d92e-46fe-923a-3199b353df28]: chord(([None, [['group2->chain1->group1'], ['group2->chain1->group2']]],))
[2023-03-31 10:40:25,663: INFO/ForkPoolWorker-3] Task app2.t1[4e447663-d92e-46fe-923a-3199b353df28] succeeded in 0.0013457320019369945s: ([None, [[...], [...]]], 'chord')
[2023-03-31 10:40:27,098: INFO/ForkPoolWorker-2] app2.t2[89265d0d-a876-4ce8-9e1d-d7f68a38246d]: end group1(())
[2023-03-31 10:40:27,101: INFO/ForkPoolWorker-2] Task app2.t2[89265d0d-a876-4ce8-9e1d-d7f68a38246d] succeeded in 2.0062880830009817s: ('group1',)
```

ac2-d2c-c47-4ad4608af:
## Description

Alternative patch candidate for issue #8182 . 

See #10387 

This patch candidate is non breaking, and relies instead of freezing the `set_chord_size` on that the descended chord object in a ForkedWorker call will be of dict type instead of Signature type.


auvipy:
can we also fix the lint and pre commit issues in this pr, please? I liked this non breaking approach

celery/canvas.py
```diff
@@ -1818,8 +1818,9 @@ def _apply_tasks(self, tasks, producer=None, app=None, p=None,
                 # that when we get to the final one, we can correctly set the
                 # size in the backend and the chord can be sensible completed.
                 chord_size += _chord._descend(sig)
-                if chord_obj is not None and next_task is None:
-                    # Per above, sanity check that we only saw one group
+                if isinstance(chord_obj, Signature) and next_task is None:
+                    # Issue 8182:
+                    # Set chord size not in ForkedWorker via isinstance check
```
phanky1: Using `task_serializer = 'pickle'` would restore the callback as a Signature,
https://github.com/celery/celery/blob/88783a51e94fd5c824c344ac5973ae00a777fbca/celery/canvas.py#L835-L838
so this can't be used to distinguish if it is outer or inner.

ac2-d2c-c47-4ad4608af: Ok, what about adding an extra flag to the dict during serialization indicating that the Signature came from a serialization ? I would guess that we generally do not want to use set_chord_size downstream of a serialization.

phanky1: I think that could work. Let's give it a try.

ac2-d2c-c47-4ad4608af: added the corresponding change

auvipy:
unit test are failing. please cross check

celery/canvas.py
```diff
@@ -1818,8 +1828,14 @@ def _apply_tasks(self, tasks, producer=None, app=None, p=None,
...
-                if chord_obj is not None and next_task is None:
-                    # Per above, sanity check that we only saw one group
+
+                # Issue 8182:
+                # Set chord size not in ForkedWorker via "from_serialized"
+                # marker check
+                if (isinstance(chord_obj, Signature)
+                    and not chord_obj["from_serialized"]
+                    and next_task is None
+                ):
```
phanky1, line 1838: This can still cause an issue if the group is nested inside a chord as the group would inherit the same id as the outer chord and override the real count.

auvipy:
also this needs merge conflicts to be fixed

celery/canvas.py
ac2-d2c-c47-4ad4608af: How does this differ from the first example in https://github.com/celery/celery/issues/8182#issuecomment-4730372075 ? 

In this example, the outer chain is converted to a chord, so we have chord->group so that the issue is present. This patch fixes it for this example because the inner group is set inside a ForkedWorker.

phanky1: see this
```python
chain(
    group(
        long_running.s("L"),
        chain(
            group(
                short_running.s("S_1"),
                short_running.s("S_2"),
            ),
        ),
    ),
    short_running.s("E"),
)
```

auvipy:
@GangEunzzang @MehrazRumman a review from your side will be helpful

celery/canvas.py
ac2-d2c-c47-4ad4608af: Have not looked in detail into that example yet, but seems to be a sort of a different root issue: Here, the inner `chain(group( ` should actually be unrolled, i.e., first 1 element chain removed and then the group unrolled into the outer group. Perhaps via modding the generator or unroll.

MehrazRumman (commented):
Thanks for the alternative approach — I agree with @auvipy that not freezing `set_chord_size` is the less invasive direction. A few things I'd want addressed before this lands, though:

**1. `from_serialized` as a top-level Signature key changes semantics beyond this fix**

`Signature` is a dict subclass, so adding the key in `__init__` means it shows up in `dict(sig)`, `repr`, `clone()`, and on the wire for every signature — not just chord callbacks. The `test_reduce` change already shows the cost: `pickle.loads(pickle.dumps(sig)) == sig` is now `False`, which is a behaviour change users may be relying on. Could this live under `options` (e.g. `sig.options['_from_serialized']`) or as a private attribute set in `from_dict`, so equality and the public dict shape stay untouched?

**2. Only the pickle path sets the marker**

`__reduce__` sets `from_serialized=True`, but `__json__` still returns `dict(self)` unchanged. With the default JSON serializer the marker is never `True` — the fix only works there because the chord arrives as a plain `dict` and fails the `isinstance` check. That's a bit fragile: if anything upstream of `_apply_tasks` reconstitutes the chord via `maybe_signature`/`Signature.from_dict` (which goes through `__init__` and defaults the key to `False`), we'd be back to calling `set_chord_size`. Setting the marker in `__json__` as well would make the two paths symmetric and make the intent explicit.

**3. Defensive lookup**

`chord_obj["from_serialized"]` will `KeyError` on any Signature-like dict that didn't pass through `__init__` (subclasses/monkeypatched objects, hand-built dicts). `chord_obj.get("from_serialized", False)` costs nothing.

**4. Scope of the fix**

Per the thread with @phanky1, the nested `chain(group(long, chain(group(...))), end)` case still overrides the real count because the inner group inherits the outer chord id. If that's out of scope for this PR, let's say so explicitly in the description and open a follow-up issue rather than leaving #8182 looking fully closed by this.

**5. Tests**

- The integration tests rely on `sleep(2)` in `long_running` for ordering. That's the classic recipe for a flaky test on a loaded CI runner — could we assert on the chord result/`set_chord_size` calls instead of wall-clock ordering, or at least make the delay generous and documented?
- There's no unit test for the new branch in `_apply_tasks`. A test that mocks `app.backend.set_chord_size` and asserts it's called for a fresh chord but *not* for one with the marker would pin the behaviour down without needing Redis.
- The `del x['from_serialized']` / `del y['from_serialized']` in `test_reduce` reads as working around the regression in (1) rather than testing intended behaviour.

**6. Housekeeping**

- Most of the 30 commits are merges of `main`; a squash/rebase would make the history reviewable.
- A short `Changelog.rst` entry referencing #8182 would be good since this touches serialization.
- The `test_worker.py` pidbox failures in CI look unrelated to this change — worth confirming they fail on `main` too.

Happy to re-review once (1)–(3) are sorted; those are the blockers for me. (4)–(6) are non-blocking.

GangEunzzang (commented):
@ac2-d2c-c47-4ad4608af @auvipy I reproduced the original issue on `main`, and the ordering is right on this branch.

The two conditions looked redundant at first, so I tried each on its own. Keeping only the `isinstance` check still works for JSON, but the bug comes straight back with the pickle serializer, where the chord option is restored as a real `Signature`. So both halves are pulling weight.

Minor: `test_chain_group_chain_to_chord_sync` passes on `main` for me too, so `test_chain_group_chain_sync` is the one actually guarding this.

Looks good to me.

auvipy (changes requested):
new merge conflicts need to be fixed in the integration tests segment

ac2-d2c-c47-4ad4608af:
> @ac2-d2c-c47-4ad4608af @auvipy I reproduced the original issue on `main`, and the ordering is right on this branch.
> 
> The two conditions looked redundant at first, so I tried each on its own. Keeping only the `isinstance` check still works for JSON, but the bug comes straight back with the pickle serializer, where the chord option is restored as a real `Signature`. So both halves are pulling weight.
> 
> Minor: `test_chain_group_chain_to_chord_sync` passes on `main` for me too, so `test_chain_group_chain_sync` is the one actually guarding this.
> 
> Looks good to me.

Indeed, `test_chain_group_chain_to_chord_sync` only has been added to verify that the fix for `test_chain_group_chain_sync` does not regress that working case.

auvipy:
but can not move forward without the merge conflict resolved! please cross check

ac2-d2c-c47-4ad4608af:
> but can not move forward without the merge conflict resolved! please cross check

No worries, looking today into this and also working on Mehraz's suggestions above.

ac2-d2c-c47-4ad4608af:
@MehrazRumman  

Thanks a lot for your comments. 

> **1. `from_serialized` as a top-level Signature key changes semantics beyond this fix**
> 
> `Signature` is a dict subclass, so adding the key in `__init__` means it shows up in `dict(sig)`, `repr`, `clone()`, and on the wire for every signature — not just chord callbacks. The `test_reduce` change already shows the cost: `pickle.loads(pickle.dumps(sig)) == sig` is now `False`, which is a behaviour change users may be relying on. Could this live under `options` (e.g. `sig.options['_from_serialized']`) or as a private attribute set in `from_dict`, so equality and the public dict shape stay untouched?

We can get it out of `__init__` (just setting in __reduce__ and __json__), but probably not out of the underlying root dict:
a) Setting in options: Options are coming in via shallow-copy, setting there will transmit flag to misc objects downstream.  
b) Setting in from_dict: This is not only called during serialization.

And indeed, this is why one of the old tests broke and had to be patched: We need a flag in the dict after the first serialization, so of course pickle.loads(pickle.dumps(sig)) != sig, due to the new flag.

> 
> **2. Only the pickle path sets the marker**
> 
> `__reduce__` sets `from_serialized=True`, but `__json__` still returns `dict(self)` unchanged. With the default JSON serializer the marker is never `True` — the fix only works there because the chord arrives as a plain `dict` and fails the `isinstance` check. That's a bit fragile: if anything upstream of `_apply_tasks` reconstitutes the chord via `maybe_signature`/`Signature.from_dict` (which goes through `__init__` and defaults the key to `False`), we'd be back to calling `set_chord_size`. Setting the marker in `__json__` as well would make the two paths symmetric and make the intent explicit.

We can add as well to the `__json__` pathway (though the isinstance check will still be needed)

> 
> **3. Defensive lookup**
> 
> `chord_obj["from_serialized"]` will `KeyError` on any Signature-like dict that didn't pass through `__init__` (subclasses/monkeypatched objects, hand-built dicts). `chord_obj.get("from_serialized", False)` costs nothing.

We can do this.


> 
> **4. Scope of the fix**
> 
> Per the thread with @phanky1, the nested `chain(group(long, chain(group(...))), end)` case still overrides the real count because the inner group inherits the outer chord id. If that's out of scope for this PR, let's say so explicitly in the description and open a follow-up issue rather than leaving #8182 looking fully closed by this.

agreed.

auvipy:
@Nusnus need your final call here

celery/canvas.py
```diff
@@ -852,10 +852,19 @@ def __invert__(self):
     def __reduce__(self):
         # for serialization, the task type is lazily loaded,
         # and not stored in the dict itself.
-        return signature, (dict(self),)
+
+        # Issue #8182:
```
Nusnus, line 856: The issue number can be removed

celery/canvas.py
```diff
@@ -852,10 +852,19 @@ def __invert__(self):
...
+        # Mark this signature as originating from a serialization
+        dic = dict(self)
+        dic["_from_serialized"] = True
+        return signature, (dic,)
 
     def __json__(self):
-        return dict(self)
+        # Issue #8182:
```
Nusnus, line 863: Same

celery/canvas.py
```diff
@@ -1877,8 +1886,15 @@ def _apply_tasks(self, tasks, producer=None, app=None, p=None,
                 # that when we get to the final one, we can correctly set the
                 # size in the backend and the chord can be sensible completed.
                 chord_size += _chord._descend(sig)
-                if chord_obj is not None and next_task is None:
-                    # Per above, sanity check that we only saw one group
+
+                # Issue 8182:
```
Nusnus, line 1890: Same

celery/canvas.py
```diff
@@ -852,10 +852,19 @@ def __invert__(self):
     def __reduce__(self):
         # for serialization, the task type is lazily loaded,
         # and not stored in the dict itself.
-        return signature, (dict(self),)
+
+        # Issue #8182:
+        # Mark this signature as originating from a serialization
+        dic = dict(self)
```
Nusnus, line 858: I’d call it `d`, not `dic`

celery/canvas.py
```diff
@@ -1877,8 +1886,15 @@ def _apply_tasks(self, tasks, producer=None, app=None, p=None,
...
                 chord_size += _chord._descend(sig)
-                if chord_obj is not None and next_task is None:
-                    # Per above, sanity check that we only saw one group
+
+                # Issue 8182:
+                # Set chord size not in ForkedWorker via "_from_serialized"
+                # marker check
+                if (
+                    isinstance(chord_obj, Signature)
+                    and "_from_serialized" not in chord_obj
```
Nusnus, line 1895: This feels.. patchy. I get that it works, but it may affect more than what this PR expects. Worth a second look.

t/integration/tasks.py
```diff
@@ -341,6 +341,22 @@ def build_chain_inside_task(self):
...
+
+@shared_task()
+def short_running(*args):
+    return args[-1]
+
+
+@shared_task()
+def long_running(*args):
+    sleep(2)
+    return args[-1]
```
Nusnus, line 357: Review at a glance: I’d expect the existing tasks to be useful enough. Worth reconsidering instead of adding new tasks.

t/integration/test_canvas.py
```diff
@@ -1217,6 +1218,69 @@ def test_chain_forwards_args_past_leading_empty_steps(self, manager):
         c = _chain(_chain(group(), add.s(10)), add.s(20), app=manager.app)
         assert c.apply_async((5,)).get(timeout=TIMEOUT) == 35
 
+    def test_chain_group_chain_sync(self, manager):
+        """
+        Test for issue 8182:
```
Nusnus, line 1223: This line can be removed

t/unit/tasks/test_canvas.py
```diff
@@ -185,13 +185,21 @@ def test_link_on_scalar(self):
...
 
     @pytest.mark.usefixtures('depends_on_current_app')
     def test_reduce(self):
         x = Signature('TASK', (2, 4), app=self.app)
         fun, args = x.__reduce__()
-        assert fun(*args) == x
+        y = fun(*args)
+        # Issue #8182: reduce adds marker
+        assert y['_from_serialized']
+        del y['_from_serialized']
```
Nusnus, line 201: Issue comment + another side of the “patchy vibes"

Nusnus (commented):
Thanks for the ping @auvipy.

I gave it a quick review. Overall it appears to work, but I’m not sure the way it solves it is best.

Also, this test fails - worth checking deeper.

```python
    def test_chain_group_single_chain_group_sync(self, manager):
        """
        Like test_chain_group_chain_sync, but with no task in front of
        the inner group, so the inner group is applied by the client
        instead of a worker.
        """
        if not manager.app.conf.result_backend.startswith("redis"):
            raise pytest.skip("Requires redis result backend.")
        try:
            manager.app.backend.ensure_chords_allowed()
        except NotImplementedError as e:
            raise pytest.skip(e.args[0])

        # Inner group shares the outer chord's group_id, so it
        # overwrites the chord size (3 -> 2) while long_running
        # is still running.
        sig = chain(
            group(
                long_running.s(1),
                chain(
                    group(
                        short_running.s(3),
                        short_running.s(4),
                    ),
                ),
            ),
            short_running_full_return.s(5),
        )

        # Broken: never completes -> TimeoutError
        actual = sig.delay().get(timeout=TIMEOUT)

        assert tuple(actual) == ([1, [3, 4]], 5)
```

```console
============================= test session starts ==============================
platform linux -- Python 3.13.14, pytest-9.1.1, pluggy-1.6.0 -- /home/ubuntu/.pyenv/versions/celery_py313/bin/python
rootdir: /home/ubuntu/repos/celery
configfile: pyproject.toml
plugins: timeout-2.4.0, celery-1.3.0, anyio-4.15.1, click-1.1.0, github-actions-annotate-failures-0.4.2, order-1.5.0, pytest_docker_tools-3.1.10, cov-7.1.0, rerunfailures-16.7
collecting ... collected 200 items / 197 deselected / 3 selected

t/integration/test_canvas.py::test_chain::test_chain_group_chain_sync PASSED
t/integration/test_canvas.py::test_chain::test_chain_group_chain_to_chord_sync PASSED
t/integration/test_canvas.py::test_chain::test_chain_group_single_chain_group_sync FAILED

_____________ test_chain.test_chain_group_single_chain_group_sync ______________

self = <celery.backends.redis.ResultConsumer object at 0x707162365480>
result = <AsyncResult: c0399a2a-3163-4ed8-bcc1-e48c75121396>, timeout = 60
on_interval = <promise@0x707161a4e0d0 --> <weakref at 0x707161a37350; to 'celery.result.AsyncResult' at 0x707161a2cad0>>
on_message = None, kwargs = {'interval': 0.5, 'no_ack': True}

    def _wait_for_pending(self, result,
                          timeout=None, on_interval=None, on_message=None,
                          **kwargs):
        self.on_wait_for_pending(result, timeout=timeout, **kwargs)
        prev_on_m, self.on_message = self.on_message, on_message
        try:
>           for _ in self.drain_events_until(
                    result.on_ready, timeout=timeout,
                    on_interval=on_interval):

celery/backends/asynchronous.py:405: 

self = <celery.backends.asynchronous.Drainer object at 0x7071623655b0>
p = <promise@0x707161a4c410 --> <weakref at 0x707161a379d0; to 'celery.result.AsyncResult' at 0x707161a2cad0>>
timeout = 60, interval = 1
on_interval = <promise@0x707161a4e0d0 --> <weakref at 0x707161a37350; to 'celery.result.AsyncResult' at 0x707161a2cad0>>
wait = <bound method ResultConsumer.drain_events of <celery.backends.redis.ResultConsumer object at 0x707162365480>>

    def drain_events_until(self, p, timeout=None, interval=1, on_interval=None, wait=None):
        wait = wait or self.result_consumer.drain_events
        time_start = time.monotonic()
    
        # Ready results take precedence over an exhausted timeout.
        while not p.ready:
            # Total time spent may exceed a single call to wait()
            if timeout is not None and time.monotonic() - time_start >= timeout:
>               raise socket.timeout()
E               TimeoutError

celery/backends/asynchronous.py:87: TimeoutError

self = <t.integration.test_canvas.test_chain object at 0x7071617494f0>
manager = <celery.contrib.testing.manager.Manager object at 0x707161a2a0d0>

    def test_chain_group_single_chain_group_sync(self, manager):
        """
        Test for issue 8182:
        Like test_chain_group_chain_sync, but with no task in front of
        the inner group, so the inner group is applied by the client
        instead of a worker.
        """
        if not manager.app.conf.result_backend.startswith("redis"):
            raise pytest.skip("Requires redis result backend.")
        try:
            manager.app.backend.ensure_chords_allowed()
        except NotImplementedError as e:
            raise pytest.skip(e.args[0])
    
        # Inner group shares the outer chord's group_id, so it
        # overwrites the chord size (3 -> 2) while long_running
        # is still running.
        sig = chain(
            group(
                long_running.s(1),
                chain(
                    group(
                        short_running.s(3),
                        short_running.s(4),
                    ),
                ),
            ),
            short_running_full_return.s(5),
        )
    
        # Broken: never completes -> TimeoutError
>       actual = sig.delay().get(timeout=TIMEOUT)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

t/integration/test_canvas.py:1315: 
celery/result.py:261: in get
    return self.backend.wait_for_pending(
celery/backends/asynchronous.py:300: in wait_for_pending
    for _ in self._wait_for_pending(result, **kwargs):
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

self = <celery.backends.redis.ResultConsumer object at 0x707162365480>
result = <AsyncResult: c0399a2a-3163-4ed8-bcc1-e48c75121396>, timeout = 60
on_interval = <promise@0x707161a4e0d0 --> <weakref at 0x707161a37350; to 'celery.result.AsyncResult' at 0x707161a2cad0>>
on_message = None, kwargs = {'interval': 0.5, 'no_ack': True}

    def _wait_for_pending(self, result,
                          timeout=None, on_interval=None, on_message=None,
                          **kwargs):
        self.on_wait_for_pending(result, timeout=timeout, **kwargs)
        prev_on_m, self.on_message = self.on_message, on_message
        try:
            for _ in self.drain_events_until(
                    result.on_ready, timeout=timeout,
                    on_interval=on_interval):
                yield
                sleep(0)
        except socket.timeout:
>           raise TimeoutError('The operation timed out.')
E           celery.exceptions.TimeoutError: The operation timed out.

celery/backends/asynchronous.py:411: TimeoutError
------------------------------ Captured log call -------------------------------
INFO     celery.worker.strategy:strategy.py:161 Task t.integration.tasks.long_running[da9f8c1f-bde0-4bf3-8c95-81f60366aac1] received
INFO     celery.worker.strategy:strategy.py:161 Task t.integration.tasks.short_running[eaa32890-0743-458a-8771-f4090d9f5b86] received
INFO     celery.worker.strategy:strategy.py:161 Task t.integration.tasks.short_running[9a7b96b4-1003-41f8-9abb-81ebfc35ffd1] received
WARNING  py.warnings:warnings.py:110 /home/ubuntu/repos/celery/celery/backends/redis.py:566: DeprecationWarning: Call to deprecated setex. (Use 'set' instead.) -- Deprecated since version 2.6.12.
  pipe.setex(key, self.expires, value)

INFO     celery.app.trace:trace.py:128 Task t.integration.tasks.long_running[da9f8c1f-bde0-4bf3-8c95-81f60366aac1] succeeded in 2.003596636000111s: 1
WARNING  py.warnings:warnings.py:110 /home/ubuntu/repos/celery/celery/backends/redis.py:566: DeprecationWarning: Call to deprecated setex. (Use 'set' instead.) -- Deprecated since version 2.6.12.
  pipe.setex(key, self.expires, value)
=============================== warnings summary ===============================
t/integration/test_canvas.py::test_chain::test_chain_group_chain_to_chord_sync
t/integration/test_canvas.py::test_chain::test_chain_group_single_chain_group_sync
t/integration/test_canvas.py::test_chain::test_chain_group_single_chain_group_sync
t/integration/test_canvas.py::test_chain::test_chain_group_single_chain_group_sync
  /home/ubuntu/repos/celery/celery/backends/redis.py:566: DeprecationWarning: Call to deprecated setex. (Use 'set' instead.) -- Deprecated since version 2.6.12.
    pipe.setex(key, self.expires, value)

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
FAILED t/integration/test_canvas.py::test_chain::test_chain_group_single_chain_group_sync - celery.exceptions.TimeoutError: The operation timed out.
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! KeyboardInterrupt !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
/home/ubuntu/.pyenv/versions/celery_py313/lib/python3.13/site-packages/kombu/utils/scheduling.py:33: KeyboardInterrupt
(to show a full traceback on KeyboardInterrupt use --full-trace)
===== 1 failed, 2 passed, 197 deselected, 4 warnings in 150.99s (0:02:30) ======
```

ac2-d2c-c47-4ad4608af:
> Thanks for the ping @auvipy.
> 
> I gave it a quick review. Overall it appears to work, but I’m not sure the way it solves it is best.
> 
> Also, this test fails - worth checking deeper.
> 
> ```python
>     def test_chain_group_single_chain_group_sync(self, manager):

This particular case has been discussed above: https://github.com/celery/celery/pull/10582#discussion_r4014330429

Currently looks like this is from a different root cause (it should be unrolled).

ac2-d2c-c47-4ad4608af:
> Thanks for the ping @auvipy.
> 
> I gave it a quick review. Overall it appears to work, but I’m not sure the way it solves it is best.

Happy to hear alternative suggestions. The previous freezing attempt is probably the safest from the point of view of preventing unintended unknown effects, however, that would require generator materialization.

Issue #9894, opened 2025-09-16
Crash in setproctitle on macOS Tahoe (26.0) #9893


IngmarStein:
# Checklist

- [x] I have verified that the issue exists against the `main` branch of Celery.
- [x] This has already been asked to the [discussions forum](https://github.com/celery/celery/discussions) first: https://github.com/celery/celery/discussions/9893
- [x] I have read the relevant section in the
  [contribution guide](https://docs.celeryq.dev/en/main/contributing.html#other-bugs)
  on reporting bugs.
- [x] I have checked the [issues list](https://github.com/celery/celery/issues?q=is%3Aissue+label%3A%22Issue+Type%3A+Bug+Report%22+-label%3A%22Category%3A+Documentation%22)
  for similar or identical bug reports.
- [x] I have checked the [pull requests list](https://github.com/celery/celery/pulls?q=is%3Apr+label%3A%22PR+Type%3A+Bugfix%22+-label%3A%22Category%3A+Documentation%22)
  for existing proposed fixes.
- [x] I have checked the [commit log](https://github.com/celery/celery/commits/main)
  to find out if the bug was already fixed in the main branch.
- [x] I have included all related issues and possible duplicate issues
  in this issue (If there are none, check this box anyway).
- [x] I have tried to reproduce the issue with [pytest-celery](https://docs.celeryq.dev/projects/pytest-celery/en/latest/userguide/celery-bug-report.html) and added the reproduction script below.

## Mandatory Debugging Information

- [x] I have included the output of ``celery -A proj report`` in the issue.
    (if you are not able to do this, then at least specify the Celery
     version affected):
- [ ] I have verified that the issue exists against the `main` branch of Celery.
- [ ] I have included the contents of ``pip freeze`` in the issue.
- [ ] I have included all the versions of all the external dependencies required
  to reproduce this bug.

## Optional Debugging Information

- [ ] I have tried reproducing the issue on more than one Python version
  and/or implementation.
- [ ] I have tried reproducing the issue on more than one message broker and/or
  result backend.
- [ ] I have tried reproducing the issue on more than one version of the message
  broker and/or result backend.
- [ ] I have tried reproducing the issue on more than one operating system.
- [ ] I have tried reproducing the issue on more than one workers pool.
- [ ] I have tried reproducing the issue with autoscaling, retries,
  ETA/Countdown & rate limits disabled.
- [ ] I have tried reproducing the issue after downgrading
  and/or upgrading Celery and its dependencies.

[...]

PR #10583, opened 2026-09-09
Detect spawned pool children in process_initializer

GangEunzzang:
## Summary

Since billiard 4.3.0rc1 (celery/billiard#443) the default start method on macOS is `spawn`. Celery's prefork pool only takes its "fresh interpreter" initialisation path when `FORKED_BY_MULTIPROCESSING` is set in the environment — which billiard's old execv path used to do, and which #10339 now does for `worker_pool_start_method='spawn'` — but nothing sets it when billiard itself chooses `spawn`. The spawned child then skips `setup_worker_optimizations()`, and every task fails in `fast_trace_task`:

- `main`: `RuntimeError: fast_trace_task: worker task registry is empty (_loc was not populated)`
- 5.6.2: `ValueError: not enough values to unpack (expected 3, got 0)`

The worker starts, `inspect ping` answers, `-P solo` / `-P threads` are fine — only prefork children are affected. Reproduced with a fresh `pip install 'celery==5.6.2' 'billiard==4.3.0rc1'` on macOS 15 with both the async pool (redis) and the blocking pool (`filesystem://`), and on Linux by forcing `set_start_method('spawn')`.

This makes `process_initializer()` detect the situation itself, using billiard's effective start method: when `get_start_method()` is not `fork`, the child sets `FORKED_BY_MULTIPROCESSING` before `init_worker()` imports the task modules, so the existing execv path is taken. `Celery.task()` now checks the variable at call time as well as at import time, so `@app.task` in those modules binds to the current app rather than to the module-level app instance the child creates on import (otherwise `task.request` on the module-level task object is empty).

The parent no longer has to know how the children were started: this covers billiard's platform default, `worker_pool_start_method='spawn'`, and Windows (where `get_start_method()` is always `spawn`; this is the case the error message in `fast_trace_task` describes).

## Testing

- Unit tests for the spawned and forked branches of `process_initializer()`, and for `app.task()` honouring the variable when it is set after import.
- macOS 15 + billiard 4.3.0rc1 (spawn default), fix vs `main`: async pool with hard time limit replacement and 50 tasks at `-c 2`, blocking pool with hard time limit replacement and 20 tasks, `-B` — all pass with the fix, all tasks fail on `main`.
- Linux (billiard 4.2.4, fork): behaviour unchanged (`FORKED_BY_MULTIPROCESSING` unset, module state inherited). `--pool-start-method spawn` from #10339 and an injected `set_start_method('spawn')` without the variable both work.

Related: #10339 (discussion in https://github.com/celery/celery/pull/10339#issuecomment-5592945298), celery/billiard#443.

### DIFF
--- a/celery/app/base.py
+++ b/celery/app/base.py
@@ -84,6 +84,13 @@ BUILTIN_FIXUPS = {
 }
 USING_EXECV = os.environ.get('FORKED_BY_MULTIPROCESSING')
 
+
+def _using_execv():
+    # Also checked at call time: a spawned pool child only sets the
+    # variable from process_initializer(), after this module was imported.
+    return USING_EXECV or os.environ.get('FORKED_BY_MULTIPROCESSING')
+
+
 ERR_ENVVAR_NOT_SET = """
 The environment variable {0!r} is not set,
 and as such the configuration could not be loaded.
@@ -547,7 +554,7 @@ class Celery:
             not access any attributes on the returned object until the
             application is fully set up (finalized).
         """
-        if USING_EXECV and opts.get('lazy', True):
+        if _using_execv() and opts.get('lazy', True):
             # When using execv the task in the original module will point to a
             # different app, so doing things like 'add.request' will point to
             # a different task instance.  This makes sure it will always use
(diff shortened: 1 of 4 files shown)


auvipy (approved):

auvipy:
should also close this https://github.com/celery/celery/issues/9894

PR #10584, opened 2026-09-09
Allow worker_pool_start_method='spawn' with the asynchronous prefork pool

GangEunzzang:
## Summary

Follow-up to #10339, as discussed there. Related: #6036 (spawn support has been requested there since 2020; with this change it is available for both prefork pool implementations).

#10339 rejects `worker_pool_start_method='spawn'` whenever the worker uses an event loop, on the assumption that the asynchronous prefork pool cannot work with spawned children. In practice it does: I ran AsynPool with spawned children on Linux and macOS with hard and soft time limits, worker replacement after a hard time limit, `--autoscale`, `--max-tasks-per-child`, `acks_late` + `task_reject_on_worker_lost`, `revoke(terminate=True)`, `pool_restart`, grow/shrink, chains/groups/chords, ETA/retry, large payloads and warm/cold shutdown, without a difference to fork (details in https://github.com/celery/celery/pull/10339#issuecomment-5592945298). The wiring the PR description worried about — the child reporting `inqW_fd` back to the parent — holds up because `spawnv_passfds` preserves fd numbers.

The check also had a side effect: it keyed off `w.use_eventloop` rather than the pool class, so `-P solo` and `-P threads` with an async transport were rejected too.

This PR:

- removes the rejection in `Pool.create()`, so `spawn` works with both prefork pool implementations (and no longer blocks solo/threads);
- updates the CLI help and the docs accordingly, and notes in the setting docs that replacement children have to import the application within `worker_proc_alive_timeout`;
- adds smoke tests that run the worker with `worker_pool_start_method = "spawn"` against the smoke brokers (AsynPool): tasks run in `SpawnPoolWorker-*` children, and the pool keeps working after a hard time limit kill.

## Testing

- Unit tests updated (`test_create_spawn_with_async_pool` now asserts the pool is created with `forking_enable=False`).
- New smoke tests pass locally.
- Manual: redis broker with `--pool-start-method spawn` at `-c 2`, hard time limit replacement and 50 tasks; `-P solo --pool-start-method spawn` now starts.

### DIFF
--- a/docs/userguide/configuration.rst
+++ b/docs/userguide/configuration.rst
@@ -4299,14 +4299,9 @@ meaningful for the prefork pool; ignored by the eventlet/gevent/solo pools.
   safe in the presence of threads and fork-unsafe C-extensions, at the cost
   of slower start-up, higher memory usage, and the requirement that the app
   and task arguments are picklable and that your entry point is guarded by
-  ``if __name__ == '__main__':``.
-
-.. warning::
-
-    ``"spawn"`` is **not supported** when the asynchronous prefork pool is
-    used (i.e. when the broker transport drives the event loop, which is the
-    common case for AMQP/Redis). The worker raises
-    :exc:`~celery.exceptions.ImproperlyConfigured` at start-up in that case.
+  ``if __name__ == '__main__':``. A replacement child has to finish importing
+  your application within :setting:`worker_proc_alive_timeout`, so raise that
+  setting for applications that take longer to import.
 
 .. setting:: worker_autoscaler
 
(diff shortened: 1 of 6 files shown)


docs/userguide/configuration.rst
```diff
@@ -4299,14 +4299,9 @@ meaningful for the prefork pool; ignored by the eventlet/gevent/solo pools.
...
   and task arguments are picklable and that your entry point is guarded by
-  ``if __name__ == '__main__':``.
-
-.. warning::
-
-    ``"spawn"`` is **not supported** when the asynchronous prefork pool is
-    used (i.e. when the broker transport drives the event loop, which is the
-    common case for AMQP/Redis). The worker raises
-    :exc:`~celery.exceptions.ImproperlyConfigured` at start-up in that case.
+  ``if __name__ == '__main__':``. A replacement child has to finish importing
```
auvipy, line 4302: may be versionadded 5.7.0 annotation is needed somewhere in suitable place

auvipy (approved):

PR #10585, opened 2026-09-09
Allow crontab to correctly handle daylight switching time

### LINKED ISSUE #10107: Hourly crontab skips execution during fall-back DST transition

Amli:
## Description
This pull request is an attempt to get the crontab handle daylight saving correctly:
- do not produce a datetime that does not exists (at DST start, when an hour disappears)
- do not produce a date an hour late (this currently happens at DST end, when an hour appears twice the same day)
- do not produce a next date in the past (remaining_estimate does that on dst end)

After [trying to hack](https://github.com/opendatasoft/celery/commit/c8c73a3a2e643ac80360dba16b85d094e6ec57ac) the current code, I ended up with a complete rewrite of the main function that computes the next run date for a crontab because the existing code was entirely ignoring timezone and makes it very difficult to just hack around it to get a correct result in all cases.

## Disclaimer
That's a relatively big change on the crontab code (even more for a first time contribution), I hesitated quite some time before opening this and I would understand if was too risky but, given the number of issues involving timezones and DST, I think it could be a good thing to have a code that handles timezone at its heart.
It is also highly possible I overlooked some cases (not aware datetime do not seem to be possible in this part of the code but I wonder), so I would be happy to get more insight.
I'm obviously available to discuss all this and ready to change the approach if needed.

## Changes
The new `crontab._next_occurence()` function produces a correct timezone aware date for the next expected crontab run.
It replaces the uses of `remaining()` and `remaining_delta()` (and the internal `_delta_to_next()`).

It does not work with delta at all, the delta is simply computed in `remaining_estimate()` as a simple difference between now and next_run_at (in utc).

The logic inside `_next_occurence()` is a mix of the existing `remaining_delta()`, for the part where we check the current day first to be as fast as possible and `_delta_to_next()` where we look for candidate days.

The code tries to be as fast as possible while only producing correct datetimes, so it is possibly a bit slower, but as far as I can tell not significantly slower.

The existing `remaining_delta()` was kept and modified to use the new code but still produce an `ffwd` instance.
AFAIK it is not used in the codebase except in an assert in the tests but I don't know if it considered to be a visible method or not.

The changes here do not try to modify the cron behaviour, if a crontab is scheduled at `2` on a day that skips that hour in this timezone, I will just not run, and the opposite if the `2` hour happens 2 times, it will run twice that day.

## Tests
I added a whole bunch of tests for cases around DST in multiple timezone to ensure my implementation was correct.
Some tests come from https://github.com/celery/celery/pull/10131 and in the end, there are probably too many tests and duplicated cases...

Fixes #10107

First time contribution, do not hesitate to tell me if things should be done differently.

### DIFF
--- a/celery/schedules.py
+++ b/celery/schedules.py
@@ -500,7 +500,7 @@ class crontab(BaseSchedule):
         super().__init__(**state)
         self._orig_kwargs = dict(state)
 
-    def _next_occurence(self, current: datetime) -> datetime:
+    def _next_occurrence(self, current: datetime) -> datetime:
         """returns the date of next crontab execution after given current date.
 
         The returned date is timezone aware and uses the same timezone as the given
@@ -644,7 +644,7 @@ class crontab(BaseSchedule):
         last_run_at = self.maybe_make_aware(last_run_at).astimezone(schedule_tz)
         now = self.maybe_make_aware(self.now()).astimezone(schedule_tz)
 
-        _next = self._next_occurence(last_run_at)
+        _next = self._next_occurrence(last_run_at)
         delta = ffwd(
             year=_next.year if last_run_at.year != _next.year else None,
             month=_next.month if last_run_at.month != _next.month else None,
@@ -672,7 +672,7 @@ class crontab(BaseSchedule):
         last_run_at = self.maybe_make_aware(last_run_at).astimezone(schedule_tz)
         now = self.maybe_make_aware(self.now()).astimezone(schedule_tz)
 
-        next_run_at = self._next_occurence(last_run_at)
+        next_run_at = self._next_occurrence(last_run_at)
 
         if C_REMDEBUG:  # pragma: no cover
             print(
--- a/celery/schedules.py
+++ b/celery/schedules.py
@@ -521,11 +521,10 @@ class crontab(BaseSchedule):
             """
             sorted_month_of_year = sorted(self.month_of_year)
             sorted_day_of_month = sorted(self.day_of_month)
-            # we cannot have more than 8 years between 2 leap years
-            # which should be the biggest possible gap between 2 runs
-            # (this is technically false if a day_of_week constraint is added to a
-            # crontab specifying month=2 day=29...)
-            for y in range(start.year, start.year + 8):
+
+            # 50 years search should above possible time between 2 leap days
+            # with day of week specified in the crontab
+            for y in range(start.year, start.year + 50):
                 for m in [
                     _m for _m in sorted_month_of_year
                     if y != start.year or _m >= start.month
--- a/celery/schedules.py
+++ b/celery/schedules.py
@@ -550,9 +550,9 @@ class crontab(BaseSchedule):
             """
             original_day = d.day
             d = d.replace(minute=0)
-            # we should never try more than 25 hours for a complete day
-            # (24h + 1 potential duplicated hour during dst time shift)
-            for _ in range(25 - d.hour):
+            # we should never try more than 24 hours for a complete day
+            # (23h til last hour + 1 potential duplicated hour during dst time end)
+            for _ in range(24 - d.hour):
                 d = _move_forward(d, timedelta(hours=1))
                 # resolve time that do not exist in current timezone
                 # and move to next possible time
--- a/celery/schedules.py
+++ b/celery/schedules.py
@@ -551,18 +551,16 @@ class crontab(BaseSchedule):
             We use a simple loop over the hours here to correctly handle DST changes.
             """
             original_day = d.day
-            d = d.replace(minute=0)
             # we should never try more than 24 hours for a complete day
             # (23h til last hour + 1 potential duplicated hour during dst time end)
             for _ in range(24 - d.hour):
-                d = _move_forward(d, timedelta(hours=1))
+                d = _move_forward(d, timedelta(minutes=60 - d.minute))
                 # resolve time that do not exist in current timezone
-                # and move to next possible time
-                d = resolve_imaginary(d)
+                # d = resolve_imaginary(d)
                 # check if this can make us change the day and get out if it does
                 if original_day != d.day:
                     return
-                if d.hour in self.hour:
+                if d.hour in self.hour and any(True for m in self.minute if m >= d.minute):
                     yield d
 
         def _has_run_hours_after(d: datetime) -> bool:
@@ -603,7 +601,10 @@ class crontab(BaseSchedule):
 
         # first we do a quick check on current day
         if _may_run_today:
-            _may_run_this_hour = candidate.hour in self.hour and any(m for m in self.minute if m > candidate.minute)
+            _may_run_this_hour = (
+                candidate.hour in self.hour
+                and any(True for m in self.minute if m > candidate.minute)
+            )
             if _may_run_this_hour:
                 # if there are slots later this hour, we can safely add one minute with worrying about hour change.
                 # the search for a correct minute will happen right after
--- a/celery/schedules.py
+++ b/celery/schedules.py
@@ -541,7 +541,9 @@ class crontab(BaseSchedule):
                             continue
                         # check if we have constraint on day of week
                         if not _crontab_has_day_of_week or new_date.isoweekday() % 7 in self.day_of_week:
-                            yield new_date
+                            # at dst start for timezone that trigger the change at midnight,
+                            # we could be generating a non existent date
+                            yield resolve_imaginary(new_date)
 
         def _generate_run_hours_after(d: datetime) -> Generator[datetime]:
             """yields candidate hours for task execution.
--- a/celery/schedules.py
+++ b/celery/schedules.py
@@ -2,20 +2,19 @@
 from __future__ import annotations
 
 import re
-from bisect import bisect, bisect_left
 from collections import namedtuple
 from datetime import datetime, timedelta, tzinfo
-from typing import Any, Callable, Iterable, Mapping, Sequence, Union
+from typing import Any, Callable, Generator, Iterable, Mapping, Sequence, Union
 
+from dateutil.tz import resolve_imaginary
 from kombu.utils.objects import cached_property
 
 from celery import Celery
 
 from . import current_app
 from .exceptions import ImproperlyConfigured
-from .utils.collections import AttributeDict
-from .utils.time import (ffwd, humanize_seconds, localize, maybe_make_aware, maybe_timedelta, remaining, timezone,
-                         weekday, yearmonth)
+from .utils.time import (C_REMDEBUG, ffwd, humanize_seconds, localize, maybe_make_aware, maybe_timedelta, remaining,
+                         timezone, weekday, yearmonth)
 
 __all__ = (
     'ParseException', 'schedule', 'crontab', 'crontab_parser',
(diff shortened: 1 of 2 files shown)


auvipy:
did you work on top of the existing open PR or you introduced complete new code?

Amli:
That's complete new code, I started looking at and trying to adjust the changes made in https://github.com/celery/celery/pull/10131 only to find out there was cases that was not easily handled by a simple hack.

auvipy:
@phanky1 @GangEunzzang I would highly appreciate your participation in reviewing this PR

GangEunzzang:
I spent some time digging into this and testing the branch on macOS 15 and Linux (`python:3.12`) against `main` (`3e40f43`), including a few cases outside the new tests.

The timezone-aware approach looks like the right direction, and the new tests are useful — 69 of them fail on current `main`, so they're exercising real behavior differences rather than just adding coverage around the new implementation. The existing `test_schedules.py` / `test_beat.py` tests also pass on the branch (392 total).

I could reproduce #10107 as well. Simulating the Berlin fall-back minute by minute, `crontab(minute=0)` fires 7 times on this branch versus 6 on `main`, restoring the run in the repeated hour.

While testing a few other timezone/schedule combinations, I ran into some cases where the branch differs from `main`, including a couple where it raises instead of returning a date.

### 1. Sparse schedules can raise `RuntimeError` where `main` finds the next run

The 8-year search window in `_generate_run_days_after()` isn't enough for some valid schedules, and this isn't specific to DST:

```python
c = crontab(minute=0, hour=0, day_of_month=29, month_of_year=2)
c.tz = ZoneInfo("UTC")
c.nowfun = lambda: datetime(2096, 3, 1, tzinfo=c.tz)

c.remaining_estimate(datetime(2096, 3, 1, tzinfo=c.tz))

# main:   2104-02-29
# branch: RuntimeError: unable to find a next occurrence
```

`range(start.year, start.year + 8)` also excludes the eighth year, so 2096 → 2104 is out of reach either way.

I found similar cases with:

* `day_of_month=29, month_of_year=2, day_of_week=1` from 2024 (`main`: 2044)
* `day_of_month=10, month_of_year=9, day_of_week=4` from 2026 (`main`: 2037)

I also did a small sweep of 48 sparse schedules. The branch raised `RuntimeError` on 14 of them versus 12 on `main`; the two additional failures were leap-day + weekday combinations.

### 2. `_get_run_minute()` can raise `ValueError`

I can reproduce this with Lord Howe's 30-minute DST transition:

```python
c = crontab(minute=0)
c.tz = ZoneInfo("Australia/Lord_Howe")

# last run: 2026-10-04 01:00
# DST jumps from 02:00 -> 02:30

# ValueError: min() iterable argument is empty
```

`_get_run_minute()` does:

```python
min(m for m in self.minute if m >= d.minute)
```

After the 30-minute shift, `d.minute` can be 30 while `self.minute` is `{0}`, leaving the generator empty.

### 3. Some valid hours are skipped around DST transitions

I found a couple of cases where the branch skips a wall-clock time that still exists.

For `America/Santiago`, midnight doesn't exist on the DST-start day. With `crontab(minute=0, hour=1)`, `main` returns the valid 01:00 on that day, while the branch skips to the following day.

I also saw this with `Pacific/Chatham` (+12:45). Starting from 03:45 after the spring-forward transition, an hourly schedule returns 04:00 on `main`, but 05:00 on the branch.

I ran a 20,000-case differential check against `main` using random schedule specs/times across 10 timezones. There were 6 differences of this kind, all in Lord Howe and Chatham.

### 4. The next run can resolve to the same instant as `last_run_at`

In `Australia/Lord_Howe`, if `last_run_at` is the second 01:30 during the fall-back and the schedule is `crontab(minute=30)`, the branch can return that same instant (`2026-04-04T15:00Z`) as the next run instead of advancing to the following occurrence.

That also makes `is_due` true again for a run that has already happened.

### 5. Performance for common schedules

I also did a quick `timeit` comparison of `remaining_estimate()` only (200 calls × 5 repeats, median, same machine):

| schedule                          |  `main` |   branch |
| --------------------------------- | ------: | -------: |
| `minute=0, hour=9`                |  8.3 µs | 129.9 µs |
| `minute=0, hour=9, day_of_week=1` |  8.3 µs | 126.6 µs |
| `minute=0, hour=23`               |  7.7 µs |  89.3 µs |
| `minute=0`                        |  9.2 µs |  19.7 µs |
| `minute='*'`                      | 12.1 µs |  12.0 µs |
| `day_of_month=1, month_of_year=1` | 21.8 µs |   7.9 µs |

Sparse schedules actually get faster, but a normal daily schedule is roughly 15× slower. Beat calls this for every entry on each tick, so the cost scales with the number of schedules. I haven't measured the impact on overall Beat throughput.

### 6. Daily schedules during a repeated hour

One behavior I'm not sure about is how we want to treat explicitly scheduled hours during fall-back.

Using the same Berlin transition, `crontab(minute=0, hour=2)` fires at both 02:00 CEST and 02:00 CET on the branch, whereas `main` fires once.

For `hour='*'`, firing in both instances of the repeated hour is exactly what #10107 is asking for. For an explicitly daily 02:00 schedule, though, this changes the existing semantics.

Vixie cron distinguishes these cases and avoids running fixed-time jobs twice during a repeated interval — cronie's `src/cron.c` comments the fall-back branch as "the fixed-time jobs probably have already run, and should not be repeated" — so it may be worth deciding explicitly whether Celery wants the same behavior or whether firing on both folds is intentional.

Two small things I noticed while going through the changes:

* `_next_occurence` looks like a typo; I think this should be `_next_occurrence`.
* Since some of the tests originated in #10131, it may be worth preserving attribution to the original author with a `Co-authored-by:` trailer.

Overall, the timezone-aware approach looks promising, and it does fix the repeated-hour case from #10107. The main areas that seem worth another pass are the sparse-schedule search limit, non-hour DST transitions, and the intended behavior for fixed-time schedules during a repeated hour.

Amli:
@GangEunzzang thanks a lot for the thorough review and checks !
There are definitely some things I overlooked there (I had "Australia/Lord_Howe" in mind, and forgot to check it but I was not aware of "Pacific/Chatham"...).

I'll do a first pass of fixes today.

celery/schedules.py
```diff
@@ -578,6 +500,135 @@ def __setstate__(self, state: Mapping[str, Any]) -> None:
         super().__init__(**state)
         self._orig_kwargs = dict(state)
 
+    def _next_occurence(self, current: datetime) -> datetime:
```
auvipy:
```suggestion
    def _next_occurrence(self, current: datetime) -> datetime:
```

auvipy (changes requested):

GangEunzzang:
@Amli Glad it helped!

Amli:
Runtime and ValueError exceptions should be fixed.
The cases for timezones with an half hour dst offset (Lord Howe Islands being the only one I could find) and timezones with minute level UTC offest like `Pacific/Chatham` should be good now too. I will double check this tomorrow, the day has been long already.

I will look at what can be done regarding performance after that.

On another matter, do you have a preference on rebase vs merge for keeping the branch up to date ? I also need to include `Co-authored-by:` for code attribution but an empty commit could do if all is squashed in the end.

Amli:
to verify the written tests are correct and for future reference, I ran the current tests with a simple implementation of `_next_occurrence` that cannot only be correct (adds 1 minute in UTC, check if local time matches the crontab, continue until match):

```python
    def _next_occurrence(self, current: datetime) -> datetime:
        candidate_utc = current.replace(second=0, microsecond=0).astimezone(timezone.utc)
        tz = current.tzinfo
        while range(26280000):  # 50 years * 365 days * 24 hours * 60 minutes
            candidate_utc = candidate_utc + timedelta(minutes=1)
            candidate_tz = candidate_utc.astimezone(tz)
            if (
                candidate_tz.month in self.month_of_year
                and candidate_tz.day in self.day_of_month
                and candidate_tz.isoweekday() % 7 in self.day_of_week
            ) and candidate_tz.hour in self.hour and candidate_tz.minute in self.minute:
                return candidate_tz

        raise RuntimeError(f"unable to find a next occurrence for crontab {self}, start from {current}")
```

auvipy:
can you please allow maintainer to push option?

auvipy (changes requested):
since this is a complete rewrite, I would also like to have enough integrations tests with the new unit tests as well... this will help to reduce subtle bugs

Amli:
damn, I opened the fork from an organization account which [does not allow that](https://github.com/orgs/community/discussions/5634) 😬

Switching to the same branch on my personal account should work, trying this...

edit: obviously cannot switch PR origin on an PR opened form a fork, I would need to open a new one

Amli:
@auvipy PR reopened https://github.com/celery/celery/pull/10608 and this time you should have the permission to push
(I will be AFK til monday starting more or less now)

auvipy:
no problem thanks. as mentioned, this should be covered with proper integration tests as well

MehrazRumman (changes requested):
### Bug: repeated hour missed in minute-offset zones
```python
tz = ZoneInfo("Pacific/Chatham")          # +13:45 -> +12:45, fall-back at 03:45
c = crontab(minute=45, hour=2); c.tz = tz
c._next_occurrence(datetime(2024, 4, 7, 2, 51, tzinfo=tz))
# got 2024-04-07 14:00Z (next day) / expected 2024-04-06 14:00Z (second pass)
```
Same on `main`, so not a regression — but it's the #10107 bug class this PR targets.

### Performance
`remaining_estimate()`, Berlin: `minute=0,hour=9` 11.9 µs → 85.7 µs (~7×).
Beat pays this per entry per tick. Cheap wins:
- `_has_run_hours_after()` + `_get_next_run_hour()` walk the hours twice —
  one `next(gen, None)` would halve it.
- `sorted(month_of_year/day_of_month)` on every call; precompute in `__init__`.

auvipy:
closing this in favor of the other PR

PR #10597, opened 2026-09-10
Respect map order in ConfigurationView lookups

phanky1:
## Description

`ConfigurationView.__getitem__()` currently searches all maps for each key before trying the next key form. This means a prefixed key or alias in a later map can win over a value in an earlier map.

This changes the lookup to try all key forms within each map before moving to the next one, so earlier maps take precedence.

### DIFF
--- a/celery/utils/collections.py
+++ b/celery/utils/collections.py
@@ -384,13 +384,13 @@ class ConfigurationView(ChainMap, AttributeDictMixin):
     def __getitem__(self, key):
         # type: (str) -> Any
         keys = self._to_keys(key)
-        getitem = super().__getitem__
-        for k in keys + (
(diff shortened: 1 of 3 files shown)


auvipy:
is it related to this? https://github.com/celery/celery/pull/10586

phanky1:
> is it related to this? #10586

Slightly related, but they fix different bugs.

auvipy:
ok

phanky1:
The coverage dropped because of these lines in `ChainMap`. Before this change, `ConfigurationView` went through them, but now it looks up the keys directly in each map.
  
ảnh

auvipy (approved):

auvipy (approved):

PR #10598, opened 2026-09-10
Warn when two callables collide under one task name (#10576)

### LINKED ISSUE #10576: Two callables sharing __name__ in one module collapse into one registered task, silently, and the two registration paths disagree about which survives

vasilisnasopoulos:
Closes #10576.

## What this does

A task's registered name is `f"{fun.__module__}.{fun.__name__}"`, so two distinct callables that share `__name__` inside one module collapse into a single registered task. The caller of the second one gets `SUCCESS` and the **first** function's result, and neither registration path says anything.

Task names are the wire format — they appear in messages, `task_routes`, beat schedules and monitoring — so **this changes no names**. It only makes the collapse audible, next to the `DuplicateNodenameWarning` that already covers the same class of mistake for node names:

- `DuplicateTaskNameWarning`, in the `CeleryWarning` tree, mirroring `DuplicateNodenameWarning` and `W_DUPNODE`.
- **Both** registration paths report, and each states its own outcome: the decorator path keeps the **first** callable, `register_task` replaces with the **last**. (That disagreement is the second half of the issue; this makes it visible rather than choosing a winner — happy to unify if you would rather, but that is a behaviour change and seemed yours to make.)
- The colliding callables are named with `celery.utils.imports.qualname()` — the helper that lives 94 lines above `gen_task_name` in the same module, and already prefers `__qualname__`. `Backend.prepare_exception` makes the same choice for exception types on the wire.
- One warning per name per app.

## Why identity, not the name

Detection compares the **function object**, not the name. Two closures from one factory share `__name__`, `__qualname__` **and** `__code__`, so no name-based test separates them; a module imported twice hands back the same function object. Object identity is the only thing that tells a re-registration apart from a real collision — the same distinction `DBOS.register_class` makes with `is not cls`.

## Tests

Six tests in `t/unit/app/test_app.py`, in the shape of `test_control.py::test_flatten_reply`:

- closures sharing `__name__`/`__qualname__`/`__code__`
- two classes with same-named methods
- the `register_task` path
- **negative control**: distinct explicit `name=`
- **negative control** (the near miss): identical factory shape, only `__name__` differs
- **negative control**: the *same* function registered twice, i.e. the re-import shape

`t/unit/app` + `t/unit/tasks`: 1699 passed, 0 failed. `flake8` and `isort` clean with your `setup.cfg`.

## Two things you should know before reviewing

**1. Your suite trips this 113 times, and they are real collisions.** Inside `t/unit/tasks/test_trace.py`, `TraceCase.setup_method` registers `add`, and 100+ tests then define their own local `add` in the same module. Celery keeps the first, so the "fresh" task each test believes it created is the fixture's task. I verified the consequence rather than assuming it: mutations such as `add.acks_late = True` land on `self.add` (`mutation visible on self.add -> True`). The tests pass, but the isolation is not there. I have **not** touched your suite — the fix could be explicit `name=`, or nothing at all, and that call is yours. If the output volume is the blocker, say so and I will adjust.

**2. No CI breakage, and one number in the issue thread was mine.** I first ran the full unit suite with `-W error::celery.exceptions.DuplicateTaskNameWarning` and saw 30 failures. That flag was **mine**, not yours: there is no `filterwarnings` in `setup.cfg`/`pyproject.toml` and no `-W error` in the workflows, so the warning does not fail your build.

## Scope

Python only. I did not change which callable survives, add a setting, or touch name generation. Reachability is a `__name__` collision within one module — closures from a factory, same-named methods on two classes, or two same-named functions.

If any of this was raised before, point me at the existing PR and I will close this in favour of it.

### DIFF
--- a/celery/app/base.py
+++ b/celery/app/base.py
@@ -28,14 +28,14 @@ from vine import starpromise
 from celery import platforms, signals
 from celery._state import (_announce_app_finalized, _deregister_app, _register_app, _set_current_app, _task_stack,
                            connect_on_app_finalize, get_current_app, get_current_worker_task, set_default_app)
-from celery.exceptions import AlwaysEagerIgnored, ImproperlyConfigured, OperationalError
+from celery.exceptions import AlwaysEagerIgnored, DuplicateTaskNameWarning, ImproperlyConfigured, OperationalError
 from celery.loaders import get_loader_cls
 from celery.local import PromiseProxy, maybe_evaluate
 from celery.utils import abstract
 from celery.utils.collections import AttributeDictMixin
 from celery.utils.dispatch import Signal
 from celery.utils.functional import first, head_from_fun, maybe_list
-from celery.utils.imports import gen_task_name, instantiate, symbol_by_name
+from celery.utils.imports import gen_task_name, instantiate, qualname, symbol_by_name
 from celery.utils.log import get_logger
 from celery.utils.objects import FallbackContext, mro_lookup
 from celery.utils.time import maybe_make_aware, timezone, to_utc
@@ -100,6 +100,19 @@ a valid configuration module.
 
 Example:
     {0}="proj.celeryconfig"
+"""
+
+
+W_DUPTASK = """\
+Task name {0!r} is already registered to a different callable.
+
+Existing: {1}
+New:      {2}
+
+{3}
+
+Every task must have a unique name. Pass an explicit name= to the task \
+decorator, or rename one of the callables.\
 """
 
 
@@ -392,6 +405,7 @@ class Celery:
         self.finalized = False
         self._finalize_mutex = threading.RLock()
         self._pending = deque()
+        self._duplicate_task_names_warned = set()
         self._tasks = tasks
         if not isinstance(self._tasks, TaskRegistry):
             self._tasks = self.registry_cls(self._tasks or {})
@@ -642,6 +656,7 @@ class Celery:
             add_autoretry_behaviour(task, **options)
         else:
             task = self._tasks[name]
+            self._warn_if_duplicate_task_name(task, fun, name)
         return task
 
     def register_task(self, task, **options):
@@ -657,11 +672,48 @@ class Celery:
             task_cls = type(task)
             task.name = self.gen_task_name(
                 task_cls.__name__, task_cls.__module__)
+        existing = self.tasks.get(task.name)
+        if existing is not None and existing is not task:
+            self._warn_duplicate_task_name(
+                task.name, qualname(type(existing)), qualname(type(task)),
+                'The new callable replaced the existing one; calls under this '
+                'name now reach the new callable.',
+            )
         add_autoretry_behaviour(task, **options)
         self.tasks[task.name] = task
         task._app = self
         task.bind(self)
         return task
+
+    def _warn_if_duplicate_task_name(self, task, fun, name):
+        """Warn when *name* is taken by a callable other than *fun*.
+
+        A module imported twice hands the same function object back, so
+        identity -- not the name -- is what tells a re-registration apart
+        from two distinct callables colliding under one generated name.
+        """
+        existing_fun = getattr(task, '__wrapped__', None)
+        if existing_fun is None or existing_fun is fun:
+            return
+        self._warn_duplicate_task_name(
+            name, qualname(existing_fun), qualname(fun),
+            'The new callable was not registered; calls under this name '
+            'reach the callable registered first.',
+        )
+
+    def _warn_duplicate_task_name(self, name, existing, new, consequence):
+        # one warning per name: a colliding name is a single mistake, and
+        # a test suite or module that trips it repeatedly should not bury
+        # everything else in output.
+        if name in self._duplicate_task_names_warned:
+            return
+        self._duplicate_task_names_warned.add(name)
+        warnings.warn(
+            DuplicateTaskNameWarning(
+                W_DUPTASK.format(name, existing, new, consequence),
+            ),
+            stacklevel=3,
+        )
 
     def gen_task_name(self, name, module):
         return gen_task_name(self, name, module)
--- a/celery/app/base.py
+++ b/celery/app/base.py
@@ -112,6 +112,28 @@ decorator, or rename one of the callables.\
 """
 
 
+def caller_stacklevel():
+    """Depth of the first frame outside Celery, for :func:`warnings.warn`.
+
+    A task can be registered straight from the decorator or later, when a
+    :class:`~celery.local.PromiseProxy` is first evaluated, and those two
+    paths sit at different depths -- so the level is found rather than
+    assumed. Only reached when a warning is actually emitted.
+    """
+    frame, level = sys._getframe(1), 1
+    while frame is not None:
+        module = frame.f_globals.get('__name__', '')
+        if module != 'celery' and not module.startswith('celery.'):
+            return level
+        frame, level = frame.f_back, level + 1
+    return 2
+
+
+def task_callable_qualname(task):
+    """Name the callable behind *task*, decorated or class-based."""
+    return qualname(getattr(task, '_decorated_fun', None) or type(task))
+
+
 def app_has_custom(app, attr):
     """Return true if app has customized method `attr`.
 
@@ -623,6 +645,10 @@ class Celery:
     ):
         if not self.finalized and not self.autofinalize:
             raise RuntimeError('Contract breach: app not finalized')
+        # the callable as the caller passed it, kept before the pydantic
+        # rebind below and before ``run`` turns it into a descriptor, so
+        # that a re-registration can be recognised by identity.
+        decorated_fun = fun
         name = name or self.gen_task_name(fun.__name__, fun.__module__)
         base = base or self.Task
 
@@ -640,7 +666,8 @@ class Celery:
                 '__module__': fun.__module__,
                 '__annotations__': _get_annotations(fun),
                 '__header__': self.type_checker(fun, bound=bind),
-                '__wrapped__': run}, **options))()
+                '__wrapped__': run,
+                '_decorated_fun': staticmethod(decorated_fun)}, **options))()
             # for some reason __qualname__ cannot be set in type()
             # so we have to set it here.
             try:
@@ -652,7 +679,7 @@ class Celery:
             add_autoretry_behaviour(task, **options)
         else:
             task = self._tasks[name]
-            self._warn_if_duplicate_task_name(task, fun, name)
+            self._warn_if_duplicate_task_name(task, decorated_fun, name)
         return task
 
     def register_task(self, task, **options):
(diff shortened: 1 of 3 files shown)


auvipy:
can you also check this pr please? https://github.com/celery/celery/pull/10579 , and also no need for force push

auvipy:
OK so we are going to move forward with this pr and the compatibility part from the other PR.

@MehrazRumman it would be great if you can try and review this

MehrazRumman:
Ok @auvipy !

MehrazRumman (changes requested):
## Review

I checked out the branch and ran the full unit suite: **3930 passed, 36 skipped, 3 xfailed**. Your claim about no CI breakage holds up, there is no `filterwarnings` config that would promote this to an error.

The diagnosis in #10576 is solid and the approach is the right one. Comparing function identity rather than names is the only thing that separates a real collision from a re-import, and the reasoning in the description for why `__qualname__` cannot help is correct. Warning from both paths while stating each one's actual outcome, instead of quietly picking a winner, is the right call for a first patch. The negative controls are the part I most wanted to see and they are there.

Two problems, one of which I think blocks.

### Blocking: the identity check false-positives on `bind=True` and `pydantic=True`

`__wrapped__` is set to `run`, not to `fun`, and `run` differs from `fun` on two paths:

- **`bind=True`**: `run = fun`, a plain function stored as a *class* attribute. Reading `task.__wrapped__` on the instance therefore triggers the descriptor protocol and returns a **bound method**, which is never identical to the original function.
- **`pydantic=True`**: `fun` is rebound to `pydantic_wrapper(...)` before `run` is built, so `__wrapped__` holds the wrapper while the caller passes the original.

Both make `existing_fun is fun` false for what is genuinely the same callable. I ran this on the branch:

```
same function twice, bind=False (the PR's negative control): silent (correct)
same function twice, bind=True:                              WARNED (false positive)
same function twice, pydantic=True:                          WARNED (false positive)

bind=True __wrapped__ is the original fun? False -> type: method
```

This matters because it undercuts the specific guarantee the design rests on. "A module imported twice hands the same function object back" is true, but the identity check only *sees* that on the default `bind=False`, non-pydantic path, which is exactly the one `test_duplicate_task_name_silent__same_function_twice` exercises. `@app.task(bind=True)` is extremely common, so a re-imported module or a `shared_task` registered against a second app would warn at users who did nothing wrong. A spurious warning on this is worse than the silence it replaces, because it trains people to filter the category.

The true-positive case still works, I confirmed two distinct `bind=True` closures do warn correctly, so this is a precision problem, not a detection one.

**Suggestion:** stop inferring the original callable from `__wrapped__` and record it explicitly. Capturing `fun` at the top of `_task_from_fun`, before the pydantic rebind, and stashing it in the generated class dict alongside `__wrapped__` gives you an unambiguous thing to compare. `inspect.unwrap` plus a `__func__` fallback would also work but stays sensitive to how `run` happens to be built. Please add the two missing negative controls, `bind=True` and `pydantic=True` re-registration, since those are the cases that would have caught this.

### Should fix: `stacklevel=3` is right for one path and wrong for the other

The two paths sit at different call depths, so one constant cannot serve both. Where the warning actually lands:

```
decorator path    points at: base.py line 655      <- Celery's own internals
register_task path points at: <user file> line 28  <- correct
```

The decorator path is both the more common one and the one where the user most needs a file and line, since the two colliding definitions may be far apart. As it stands they get a pointer into `celery/app/base.py`. Worth passing the right depth per call site, with a test asserting the warning resolves to the caller's frame rather than to `base.py`.

### Notes, not blocking

1. **Reconcile the warning count.** The description says the suite trips this 113 times. I counted 30 entries in the warnings summary across `t/unit`. Both numbers can be true depending on whether you count raw occurrences before the per-name dedup or distinct sites after it, but a reviewer checking the claim will hit the smaller number and wonder. Worth stating which you measured.

2. **The `test_trace.py` isolation finding is real and I would keep it out of this PR.** You verified the consequence rather than asserting it, which is the right standard, and it deserves its own issue rather than being bundled into a warning patch. Agreed that which fix to apply is the maintainers' call.

3. **`_duplicate_task_names_warned` is unbounded and unsynchronised.** It only ever grows, and concurrent registration could race on the membership check and emit twice. Both are low-impact given registration is normally single-threaded at import time, but a `set` reached from `register_task` is worth a comment saying the race is known and benign.

4. **Message asymmetry.** The decorator path reports `qualname(existing_fun)`, a function, while `register_task` reports `qualname(type(existing))`, a class. Same template, different granularity. Not wrong, but a reader comparing two warnings will notice.

Fix the identity check and the stacklevel and I think this is good to go. The framing in the description, that making the collapse audible is separable from deciding which callable should win, is the right scope for this change.

auvipy:
maybe it can also improve some test coverage

auvipy:
@GangEunzzang should we move forward with this pr as it is for now and pull the remaning part from the other pr?

GangEunzzang:
@auvipy I still favor #10598’s warning approach, with the re-registration false positives addressed. The automatic disambiguation in #10579 needs more compatibility work, particularly around registration order.

I see the warning as a useful first step. It makes collisions visible, but doesn’t prevent calls from reaching the wrong callable. If we want stronger guarantees there, I’d handle that separately with a clear compatibility plan for existing task names.

auvipy:
@vasilisnasopoulos please addressed the review suggestions to get this merged

vasilisnasopoulos:
Thank you — both findings were real and both are fixed in `4a9b14c`.

**Identity check.** You were right that `__wrapped__` is the wrong thing to compare: it holds `run`, a bound method under `bind=True` and the pydantic wrapper under `pydantic=True`. The callable is now kept as the caller passed it, before the pydantic rebind and before `run` becomes a descriptor, and stored as a `staticmethod` so reading it back does not hand out a bound method. Both false positives are gone; detection is unchanged, and two distinct closures still collide under `bind=True` and `pydantic=True` — both now covered, along with the two negative controls you asked for.

**stacklevel.** A constant cannot serve both paths, and the reason is worse than depth alone: when a task is registered through a `PromiseProxy` rather than straight from the decorator, three `celery/local.py` frames sit in between. The level is now found by walking to the first frame outside `celery`, computed only when a warning is actually emitted. A test asserts the warning resolves to the caller's file and not to `base.py`.

**On the count — my number was wrong.** The 113 included the very false positives you identified; I did not separate them. After the fix the suite reports **27** `DuplicateTaskNameWarning` entries across `t/unit`, and **13** in `test_trace.py`. Your 30 was the closer figure. The `test_trace` shadowing itself still reproduces (`mutation visible on self.add -> True`), and I agree it belongs in its own issue rather than here — happy to open it once this settles.

The warned-names set now carries a comment saying its growth and the race on its membership check are known and accepted at import time, and both paths name a colliding callable through one helper, so the two messages are no longer a function on one side and a class on the other.

Full unit suite on the updated branch: 3921 passed, 39 skipped, 3 xfailed.

auvipy (approved):

PR #10602, opened 2026-09-11
Deny access to private attributes and dunders via registered()

ndaprela:
## Description

The `Inspect.registered()` method ([docs](https://docs.celeryq.dev/en/latest/reference/celery.app.control.html#celery.app.control.Inspect.registered)) allows specifying, via `taskinfoitems`, the Task attributes to include in the response.

A possibly undesired side effect is that private attributes and dunders can also be accessed this way. For instance, requesting `__dict__` dumps all of a task's instance attributes, which — in the case of a custom `Task` subclass — could include sensitive values.

## Remediation

This PR filters out names beginning with `_` in the worker-side `registered` control handler (`celery/worker/control.py`). Public attributes remain accessible, so the documented `registered('serializer', 'max_retries')` usage is unaffected; this is the least invasive change that keeps the command out of object internals without breaking ergonomics. Unit tests included.

### DIFF
--- a/celery/worker/control.py
+++ b/celery/worker/control.py
@@ -475,6 +475,8 @@ def registered(state, taskinfoitems=None, builtins=False, **kwargs):
     """
     reg = state.app.tasks
     taskinfoitems = taskinfoitems or DEFAULT_TASK_INFO_ITEMS
+    taskinfoitems = [item for item in taskinfoitems
+                     if isinstance(item, str) and not item.startswith('_')]
 
     tasks = reg if builtins else (
         task for task in reg if not task.startswith('celery.'))
--- a/t/unit/worker/test_control.py
+++ b/t/unit/worker/test_control.py
@@ -375,6 +375,46 @@ class test_ControlPanel:
         finally:
             control.DEFAULT_TASK_INFO_ITEMS = prev
 
+    def test_dump_tasks_allows_public_attributes(self):
+        # public Task attributes (incl. non-defaults) remain available --
+        # this is the documented `registered('serializer', 'max_retries')`
+        # contract.
+        info = '\n'.join(self.panel.handle(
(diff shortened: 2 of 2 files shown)


auvipy:
@phanky1 please review if you can

phanky1 (approved):

auvipy:
@MehrazRumman please review this

MehrazRumman (commented):
Thanks for this, it is a sensible hardening and the worker-side handler is the right place for it, since it protects every client (CLI, Flower, custom scripts) rather than just `Inspect.registered()`.

I checked out the branch and ran the `dump_tasks` tests locally; all pass. The change is minimal and the three new tests cover the important cases well, especially the one proving a private property getter is never invoked.

A few non-blocking notes:

1. **Scope of the protection.** This closes the `__dict__` / `_private` route, but any *public* attribute on a custom `Task` subclass is still readable through `taskinfoitems`. That is fine given the control channel already assumes a trusted broker, but it is worth stating in the PR description or docstring that this is defence in depth, not a security boundary.

2. **Silent dropping.** Filtered names simply disappear from the output, which is consistent with how unknown attributes are already skipped, but a user asking for `_foo` gets no hint why it is missing. A `logger.debug(...)` when items are dropped would make this easier to diagnose. Optional.

3. **Docstring.** The `registered` docstring still says `taskinfoitems` is a "List of task attributes to include". A short sentence noting that names starting with `_` are ignored would keep the reference docs accurate.

4. **Behaviour change.** Anyone who did rely on underscore-prefixed attributes here will see them vanish. Unlikely, but a one-line Changelog entry would help them find the cause.

Overall looks good to me.

MehrazRumman (approved):

auvipy (approved):

PR #10603, opened 2026-09-11
Honor zero timeouts in the result drainer

phanky1:
## Description

`Drainer.drain_events_until()` treats `timeout=0` as unlimited waiting. Since #10487, ResultSet.join() passes 0 after exhausting its timeout, which can cause it to wait indefinitely.
https://github.com/celery/celery/blob/918a740497a4ce883f37eb1a3e7c2a80dfcc8b7a/celery/result.py#L813-L815

This PR changes the timeout check to treat only `None` as unlimited. It also checks readiness first so already-ready results finish without waiting or raising a timeout.

### DIFF
--- a/celery/backends/asynchronous.py
+++ b/celery/backends/asynchronous.py
@@ -254,7 +254,10 @@ class AsyncBackendMixin:
                     yield node.id, node._cache
         while bucket:
             node = bucket.popleft()
-            yield node.id, node._cache
+            if not hasattr(node, '_cache'):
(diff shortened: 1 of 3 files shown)


auvipy (changes requested):
can you please check the failing integrations tests?

phanky1:
> It also checks readiness first so already-ready

The failures were caused by the change to check readiness first. I've updated it.

auvipy (approved):

PR #10604, opened 2026-09-11
Include _keys in ConfigurationView.swap_with()

phanky1:
## Description
`ConfigurationView.swap_with()` copies the mappings, prefix, and key_t, but leaves `_keys` unchanged. When the two views use different converters, a key that works on the source view can fail after the swap.

This PR copies `_keys` too.


auvipy (approved):

auvipy:
we need to check if docker build is really failing with this or not https://github.com/celery/celery/actions/runs/34682844174

phanky1:
> we need to check if docker build is really failing with this or not https://github.com/celery/celery/actions/runs/34682844174

docker-build is okay now.

PR #10605, opened 2026-09-12
fix(backends.database): store stamping metadata with extended results (#8887)

### LINKED ISSUE #8887: Stamping doesn't work with database result backend
# Checklist

  on reporting bugs.
  for existing proposed fixes.
- [ ] I have checked the [commit log](https://github.com/celery/celery/commits/main)

     version affected).
  to reproduce this bug.

  and/or implementation.
- [x] I have tried reproducing the issue on more than one message broker and/or
  result backend.
  broker and/or result backend.

#### Related Issues

- None

#### Possible Duplicates

- None

## Environment & Settings

**Celery version**: 5.3.6, 5.4.0rc1 (opalescent)

<details>
<p>

```
software -> celery:5.4.0rc1 (opalescent) kombu:5.3.5 py:3.11.4
            billiard:4.2.0 py-amqp:5.2.0
            kernel version:23.1.0 imp:CPython
settings -> transport:amqp results:db+postgresql:///stamp_issue

deprecated_settings: None
task_default_queue: 'stamp_issue_queue'
result_extended: True
result_backend: 'db+postgresql:///stamp_issue'
```

</p>
</details>

# Steps to Reproduce

## Required Dependencies

- **Minimal Python Version**: N/A or Unknown
- **Minimal Celery Version**: N/A or Unknown
- **Minimal Broker Version**: N/A or Unknown
- **Minimal Result Backend Version**: N/A or Unknown
- **Minimal OS and/or Kernel Version**: N/A or Unknown

### Python Packages

<details>
<p>

```
amqp==5.2.0
async-timeout==4.0.3
billiard==4.2.0
celery @ git+https://github.com/celery/celery.git@9edeab6b4526a59bf699df10f1b48ac65809eaea
click==8.1.7
click-didyoumean==0.3.0
click-plugins==1.1.1
click-repl==0.3.0
greenlet==3.0.3
kombu==5.3.5
prompt-toolkit==3.0.43
psycopg2==2.9.9
python-dateutil==2.9.0.post0
redis==5.0.2
six==1.16.0
SQLAlchemy==2.0.27
typing_extensions==4.10.0
tzdata==2024.1
vine==5.1.0
wcwidth==0.2.13
```

</p>
</details>

### Other Dependencies

<details>
<p>
PostgreSQL database:

```
createdb stamp_issue
```

</p>
</details>

<details>
<p>

Stamping works with the redis result backend, but not with the database backend or the rpc backend. (The rpc backend is not suitable for me, but I tested it for completeness).

`stamp_issue.py`:
```python
from celery import Celery
from celery.canvas import group

app = Celery("stamp_issue", broker="amqp://guest@localhost:5672/")
app.conf.task_default_queue = "stamp_issue_queue"
app.conf.result_extended = True
# app.conf.result_backend = "rpc://"                       # doesn't work
# app.conf.result_backend = "redis://localhost/0"          # works
app.conf.result_backend = "db+postgresql:///stamp_issue"   # doesn't work


@app.task
def add(x, y):
    return x + y


if __name__ == "__main__":
    sig1 = add.si(2, 2)
    sig1_res = sig1.freeze()
    g = group(sig1, add.si(3, 3))
    g.stamp(stamp="your_custom_stamp")
    res = g.apply_async()
    print(res.get())
    print(sig1_res._get_task_meta()["stamp"])
```

1. Start worker: `celery -A stamp_issue worker --loglevel=INFO --hostname stamp_issue`
2. Run example: `python stamp_issue.py`

</p>
</details>

# Expected Behavior
The output

```
[4, 6]
your_custom_stamp
```

# Actual Behavior

```
[4, 6]
  File "/Users/.../stamp_issue.py", line 24, in <module>
    print(sig1_res._get_task_meta()["stamp"])
          ~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^
KeyError: 'stamp'
```

VimalN2005:
Fixes #8887

## Description

When tasks are stamped (via `signature.stamp()` or `group.stamp()`) and `result_extended = True`, the database result backend (`celery.backends.database`) dropped all stamping metadata (`stamped_headers` and individual stamp values) upon storing. This caused a `KeyError: 'stamp'` when attempting to retrieve stamps via `_get_task_meta()`.

While document/cache backends (like Redis) serialize the entire dictionary and preserve arbitrary keys, the database backend maps attributes directly to columns in `TaskExtended`. Because `TaskExtended` lacked a `stamps` column, the stamping data was filtered out and lost.

### Changes Made
1. **Model (`celery/backends/database/models.py`)**:
   - Added `stamps = sa.Column(sa.LargeBinary, nullable=True)` to `TaskExtended`.
   - Included `stamps` in `TaskExtended.to_dict()`.
2. **Backend Logic (`celery/backends/database/__init__.py`)**:
   - In `_update_result`: If `stamped_headers` are present in `meta`, packaged the stamp headers and their values into a dict, serialized with `self.encode()`, and assigned to `task.stamps`.
   - In `_get_task_meta_for`: Deserialized `data['stamps']` using `self.decode()`, restoring `data['stamped_headers']`, individual stamp keys (`data[header] = stamp_val`), and `data['stamps']`.
3. **Tests (`t/unit/backends/test_database.py`)**:
   - Added `test_store_result_with_stamps` covering both `pickle` and `json` serializers in `test_DatabaseBackend_result_extended`.

### Testing & Verification
- `pytest t/unit/backends/test_database.py -k "test_store_result_with_stamps"`: **2 passed**
- Full Database Backend test suite (`pytest t/unit/backends/test_database.py`): **83 passed**
- Full Stamping test suite (`pytest t/unit/tasks/test_stamping.py`): **439 passed**
- Code style: Flake8 passed with 0 errors/warnings.

### DIFF
--- a/docs/userguide/configuration.rst
+++ b/docs/userguide/configuration.rst
@@ -1003,7 +1003,16 @@ and client first and turn the setting on afterwards.
 Default: ``False``
 
 Enables extended task result attributes (name, args, kwargs, worker,
-retries, queue) to be written to backend.
+retries, queue, stamps) to be written to backend.
+
+.. note::
+
+    When using the database backend with :setting:`result_extended` set to ``True``,
+    task stamping metadata is stored in a ``stamps`` column. For existing database deployments
+    upgrading with an already-created ``celery_taskmeta`` table, operators should run the
+    appropriate DDL migration to add the column (e.g., ``ALTER TABLE celery_taskmeta ADD COLUMN stamps BLOB;``
+    or ``BYTEA`` on PostgreSQL). The database backend gracefully degrades to querying without the column
+    if it is not yet present.
 
 .. setting:: result_expires
 
--- a/celery/backends/database/__init__.py
+++ b/celery/backends/database/__init__.py
@@ -159,10 +159,10 @@ class DatabaseBackend(BaseBackend):
                                      traceback=traceback, request=request,
                                      format_date=False, encode=True)
 
-        # Exclude the primary key id and task_id columns
-        # as we should not set it None
+        # Exclude the primary key id, task_id, and stamps columns
+        # as we should not set it None or handle stamps separately
         columns = [column.name for column in self.task_cls.__table__.columns
-                   if column.name not in {'id', 'task_id'}]
+                   if column.name not in {'id', 'task_id', 'stamps'}]
 
         # Iterate through the columns name of the table
         # to set the value from meta.
@@ -170,6 +170,20 @@ class DatabaseBackend(BaseBackend):
         for column in columns:
             value = meta.get(column)
             setattr(task, column, value)
+
+        if hasattr(task, 'stamps'):
+            stamped_headers = meta.get('stamped_headers')
+            if stamped_headers:
+                stamps_data = {
+                    h: meta.get(h) for h in stamped_headers if h in meta
+                }
+                stamps_info = {
+                    'stamped_headers': stamped_headers,
+                    'stamps': stamps_data,
+                }
+                setattr(task, 'stamps', ensure_bytes(self.encode(stamps_info)))
+            else:
+                setattr(task, 'stamps', None)
 
     def _get_task_meta_for(self, task_id):
         """Get task meta-data for a task by id."""
@@ -187,6 +201,14 @@ class DatabaseBackend(BaseBackend):
                 data['args'] = self.decode(data['args'])
             if data.get('kwargs', None) is not None:
                 data['kwargs'] = self.decode(data['kwargs'])
+            if data.get('stamps', None) is not None:
+                stamps_info = self.decode(data['stamps'])
+                if isinstance(stamps_info, dict):
+                    if 'stamped_headers' in stamps_info:
+                        data['stamped_headers'] = stamps_info['stamped_headers']
+                    if 'stamps' in stamps_info:
+                        data.update(stamps_info['stamps'])
+                        data['stamps'] = stamps_info['stamps']
             return self.meta_from_decoded(data)
 
     def _decode_stored_result(self, payload):
--- a/celery/backends/database/__init__.py
+++ b/celery/backends/database/__init__.py
@@ -17,6 +17,7 @@ from .session import SessionManager
 
 try:
     from sqlalchemy.exc import DatabaseError, InterfaceError, InvalidRequestError
+    from sqlalchemy.orm import defer
     from sqlalchemy.orm.exc import StaleDataError
 except ImportError:
     raise ImproperlyConfigured(
@@ -136,13 +137,25 @@ class DatabaseBackend(BaseBackend):
             short_lived_sessions=self.short_lived_sessions,
             **self.engine_options)
 
+    def _query_task(self, session, task_id):
+        """Query task by id, falling back to deferring stamps if missing from database."""
+        try:
+            tasks = list(session.query(self.task_cls).filter(self.task_cls.task_id == task_id))
+            return tasks and tasks[0]
+        except DatabaseError as exc:
+            if 'stamps' in str(exc).lower() and hasattr(self.task_cls, 'stamps'):
+                tasks = list(session.query(self.task_cls).options(
+                    defer(self.task_cls.stamps)
+                ).filter(self.task_cls.task_id == task_id))
+                return tasks and tasks[0]
+            raise
+
     def _store_result(self, task_id, result, state, traceback=None,
                       request=None, **kwargs):
         """Store return value and state of an executed task."""
         session = self.ResultSession()
         with session_cleanup(session):
-            task = list(session.query(self.task_cls).filter(self.task_cls.task_id == task_id))
-            task = task and task[0]
+            task = self._query_task(session, task_id)
             if not task:
                 task = self.task_cls(task_id)
                 task.task_id = task_id
@@ -171,7 +184,7 @@ class DatabaseBackend(BaseBackend):
             value = meta.get(column)
             setattr(task, column, value)
 
-        if hasattr(task, 'stamps'):
+        if hasattr(task, 'stamps') and 'stamps' in self.task_cls.__table__.columns:
             stamped_headers = meta.get('stamped_headers')
             if stamped_headers:
                 stamps_data = {
@@ -182,15 +195,14 @@ class DatabaseBackend(BaseBackend):
                     'stamps': stamps_data,
                 }
                 setattr(task, 'stamps', ensure_bytes(self.encode(stamps_info)))
-            else:
+            elif getattr(task, 'stamps', None) is None:
                 setattr(task, 'stamps', None)
 
     def _get_task_meta_for(self, task_id):
         """Get task meta-data for a task by id."""
         session = self.ResultSession()
         with session_cleanup(session):
-            task = list(session.query(self.task_cls).filter(self.task_cls.task_id == task_id))
-            task = task and task[0]
+            task = self._query_task(session, task_id)
             if not task:
                 task = self.task_cls(task_id)
                 task.status = states.PENDING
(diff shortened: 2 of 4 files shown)


auvipy (changes requested):
the patch need more test coverage. please check that out

VimalN2005:
Hi @auvipy,

I have added comprehensive unit tests covering all edge cases for stamping in `t/unit/backends/test_database.py`:
- Stamped headers absent from metadata.
- Corrupt/non-dict stamps payloads.
- Partial stamp payloads (only stamps / only stamped headers).
- Tasks without stamps column attribute.
- Synchronized branch with latest `upstream/main`.

Patch test coverage for the modified lines is now 100% and all 85 database backend tests pass cleanly. Please take a look!

auvipy:
@MehrazRumman may be one last review in this release cycle

MehrazRumman:
OK

MehrazRumman (changes requested):
Thanks for picking up #8887 — the diagnosis in the description is correct, and the
column-based approach is the right general shape for this backend. A few things I'd
want resolved before this lands, though.

## 1. Schema change has no migration path (biggest concern)

`TaskExtended` gets a new `stamps` column, but the database backend only ever calls
`create_all()` — it never `ALTER`s an existing table. Anyone already running with
`result_extended = True` has a `celery_taskmeta` table without this column, and after
upgrading, *every* read and write will fail with something like
`OperationalError: no such column: celery_taskmeta.stamps` — not just stamped tasks.

That turns a niche bug fix into a hard break for existing deployments. At minimum this
needs:

- a release note / `Changelog.rst` entry and a line in the result-backend docs with the
  `ALTER TABLE celery_taskmeta ADD COLUMN stamps BLOB;` (per-dialect) that operators
  must run;
- ideally a test covering the "old table, new code" path, or a graceful degradation so
  a missing column doesn't take down result retrieval entirely.

Worth a maintainer call on whether the milestone this is targeting can carry a schema
change at all.

## 2. `data['stamps']` can leak the raw encoded blob into user-facing meta

In `_get_task_meta_for`, `data['stamps']` starts life as the raw bytes from `to_dict()`
and is only overwritten inside `if 'stamps' in stamps_info:`. So on the two partial
paths the new tests themselves construct — a non-dict payload, and a payload with only
`stamped_headers` — `data['stamps']` stays as the encoded bytes and gets handed to
`meta_from_decoded()`. The user ends up with `meta['stamps'] == b'\x80\x04...'`.

The existing tests don't catch this because neither case asserts on `meta['stamps']`.
Popping first fixes it and reads cleaner:

```python
raw_stamps = data.pop('stamps', None)
if raw_stamps is not None:
    stamps_info = self.decode(raw_stamps)
    if isinstance(stamps_info, dict):
        ...
```

Also, `self.decode()` here isn't guarded — a blob that fails to decode (serializer
changed between writes, truncated column) will now raise out of `get_task_meta()` and
break result retrieval for that task. Compare `_decode_stored_result()` just below,
which deliberately handles that case.

## 3. `meta['stamps']` diverges from every other backend

`BaseBackend._get_result_meta` only does:

```python
request_meta['stamped_headers'] = request.stamped_headers
request_meta.update(request.stamps)
```

i.e. stamps are flattened, and there is no `stamps` key in the meta dict. This PR adds
one, and `test_store_result_with_stamps` asserts on it — so the database backend would
return a meta shape no other backend returns, and that assertion locks it in.

Either drop `data['stamps'] = stamps_info['stamps']` so the round-trip matches the other
backends exactly, or make the case that `stamps` should be part of the meta contract
everywhere and change it in `base.py` too. I'd lean toward the former for a bug fix.

## 4. `hasattr(task, 'stamps')` as a model check

`hasattr` on the ORM instance is really asking "is `self.task_cls` `TaskExtended`?", and
it silently does nothing if the attribute name ever changes. Since the code already
computes the column list from `self.task_cls.__table__.columns` two lines above, I'd
key off that (or `self.extended_result`) instead.

Slightly bigger simplification: rather than excluding `stamps` from `columns` and then
re-implementing set-or-None below, encode the blob into `meta['stamps']` *before* the
column loop and let the generic loop handle both the set and the reset. That removes the
special case and the `else: setattr(task, 'stamps', None)` branch entirely.

On that `else` branch — worth confirming that a later `store_result` for the same task
(e.g. RETRY → SUCCESS) always carries a request with stamps, otherwise the second write
nulls out stamps recorded by the first.

## 5. Tests

Coverage is there, but a few things:

- `test_store_result_stamps_edge_cases` is five unrelated scenarios in one function with
  numbered comments. Please split them into separate named tests (or parametrize) so a
  failure points at the actual case.
- `from kombu.utils.encoding import ensure_bytes` should be a module-level import.
- Cases 2–4 write directly to the DB row and then call `tb._cache.clear()`. Reaching into
  private backend state to set up a test is fragile; if the only way to reach a branch is
  to hand-craft a row, that's a hint the branch could be simplified.
- Case 5 constructs a bare `TaskWithoutStamps` and asserts `not hasattr(dummy, 'stamps')`
  afterwards. That's covering a line, not a behaviour — the real "no stamps column" case
  is the plain `Task` model with `result_extended = False`, which would be a more
  meaningful test.
- All the tests build a `Context` by hand. Since the reported bug goes through
  `signature.stamp()` / `group.stamp()` with a `StampingVisitor`, one test that actually
  stamps a signature and reads back via `AsyncResult` would prove the original issue is
  closed, rather than proving the new column round-trips.

## 6. Minor

- `sa.LargeBinary` maps to `BLOB` (64 KB) on MySQL. Consistent with `args`/`kwargs`, so
  probably fine, but stamps are arbitrary user data and can be larger than a kwargs dict
  — worth a docs mention at least.
- A stamp header literally named `stamps` would collide with the column key on read.
  Unlikely, but the flatten-then-assign order makes the outcome arbitrary.
- No docs changes in this PR; the extended-result column list is documented and should
  mention the new column.

Happy to re-review once 1–3 are addressed — the rest is polish.

docs/userguide/configuration.rst
```diff
@@ -1003,7 +1003,16 @@ and client first and turn the setting on afterwards.
 Default: ``False``
 
 Enables extended task result attributes (name, args, kwargs, worker,
-retries, queue) to be written to backend.
+retries, queue, stamps) to be written to backend.
+
+.. note::
+
```
auvipy, line 1012: i think versionadded 5.7 annotation should be mentioned here

auvipy (changes requested):
after merging the other sqla pr of you, there are many merge conflicts. please fix them carefully

VimalN2005:
Hi @auvipy,

Thank you! I have pulled the latest `main` (including #10606) and carefully resolved all the merge conflicts in `celery/backends/database/__init__.py`.

Both `children` and `stamps` handling are now seamlessly integrated:
- `_query_task` gracefully defers missing `children` and/or `stamps` columns if absent in legacy databases.
- `_update_result` and `_get_task_meta_for` cleanly handle both features together.
- Addressed all the review feedback from @MehrazRumman (popping raw stamps, consistent flattened meta shape, model hasattr check, manual DDL documentation, and split edge case tests).

All 104 tests in `t/unit/backends/test_database.py` are passing with 99% coverage. The branch is now clean and ready to merge!

docs/userguide/configuration.rst
auvipy: @VimalN2005 did you address this?

auvipy:
@MehrazRumman another round please

MehrazRumman:
ok

MehrazRumman (commented):
Re-reviewed at `b6f9065`. Items 2–5 from my last pass are properly addressed: `raw_stamps` is popped before decode so the blob can't leak into meta, the decode is guarded, `data['stamps']` is gone so the meta shape now matches `BaseBackend` exactly, the edge cases are split into named tests, and `test_stamped_signature_roundtrip_with_async_result` exercises the real `signature.stamp()` path from #8887. The `stamps` handling also mirrors the existing `children` precedent from #10606 line for line, which is the right call.

One thing left, and it's just the docs wording:

> The database backend gracefully degrades to querying without the column if it is not yet present.

That holds for reads only. `_query_task()` has the `defer` fallback, but the write path has none — `_update_result()` guards on `'stamps' in self.task_cls.__table__.columns`, which is the model's metadata and is always true, so it sets `task.stamps` and the `UPDATE` at `session.commit()` still fails with `no such column` on an un-migrated table.

So the `ALTER TABLE` is mandatory for existing deployments, not optional. Suggest dropping that sentence and saying the migration is required before upgrading; otherwise an operator reads this as safe to skip and loses result writes.

Worth noting this gap is inherited from the `children` fallback, not introduced here — so it's a doc fix, not a code change. With that reworded I'm happy for this to land.

auvipy:
@VimalN2005 please address the remaining concerns

auvipy:
@VimalN2005 can you please address the remaining suggestions?

VimalN2005:
Hi @auvipy @MehrazRumman,

Thank you! I have updated the documentation in `docs/userguide/configuration.rst` as suggested:
- Dropped the sentence implying graceful degradation on writes.
- Explicitly noted that executing the `ALTER TABLE` migration is required before upgrading to avoid write failures.

All 104 tests are passing with 99% coverage, and the commit has been pushed. The PR is now ready to land!

auvipy (approved):

PR #10606, opened 2026-09-12
fix(backends.database): store task children with database backend (#8336)

### LINKED ISSUE #8336: database backend does not store children
The [`Task`](https://github.com/celery/celery/blob/main/celery/backends/database/models.py#L20-L27) and [`TaskExtended`](https://github.com/celery/celery/blob/main/celery/backends/database/models.py#L57-L62) models for the `database` backend do not include `children`. This means that when using any `database` backend, [`AsyncResult.children`](https://github.com/celery/celery/blob/main/celery/result.py#L424) is always empty, even if a task does have children.

VimalN2005:
## Summary

Fixes #8336

Resolves the issue where `AsyncResult.children` was not populated or retained when using the database result backend (`celery.backends.database:DatabaseBackend`).

### Background & Context
A previous attempt to store children (#8338) was reverted in #8475 because:
1. It used `sa.Column(PickleType)` directly in the model, forcing Python pickle regardless of `result_serializer`.
2. Existing database installations without the `children` column crashed on queries (`OperationalError: no such column: celery_taskmeta.children`).

### Solution
1. **Model Storage:** Added `children = sa.Column(sa.LargeBinary, nullable=True)` to `Task` in `celery/backends/database/models.py`.
2. **Serializer Compatibility:** Children are serialized using `self.encode(children)` on store and `self.decode(meta['children'])` on retrieval, respecting `result_serializer` (JSON, pickle, etc.).
3. **Graceful Auto-Migration:** Added `_migrate_missing_columns` helper in `SessionManager.prepare_models()` that checks existing schema and automatically executes `ALTER TABLE celery_taskmeta ADD COLUMN children BLOB/BYTEA` if the column is missing, gracefully catching existing setups and permission constraints.
4. **Child Result Reconstruction:** Reconstructs `[self.app.AsyncResult(cid) for cid in children]` when `children` is fetched.

### Test Coverage
Added comprehensive unit tests in `t/unit/backends/test_database.py`:
- `test_store_result_with_children`: Validates storing and retrieving children with both JSON and Pickle serializers.
- `test_async_result_children`: End-to-end test verifying `AsyncResult.children` returns list of `AsyncResult` objects.
- `test_migrate_missing_columns`: Verifies database auto-migration adds `children` column to legacy tables without data loss.
- All 86 database backend tests pass and Flake8 linter reports 0 errors.

cc @auvipy


auvipy (changes requested):
the patch need more coverage

VimalN2005:
Hi @auvipy,

I have added comprehensive unit test coverage in `t/unit/backends/test_database.py`:
- Task instances without children attribute fallback.
- Auto-migration exception/permission handling (`inspect` error tolerance).
- Auto-migration handling when tables do not exist.
- Synchronized branch with latest `upstream/main`.

Patch test coverage for the modified lines is now 100% and all 89 database backend tests pass cleanly. Please take a look!

auvipy:
@MehrazRumman feel free to try and review it. our main concern is backward compatibility here

VimalN2005:
Hi @auvipy @MehrazRumman,

Regarding **backward compatibility**, this PR specifically avoids the regression that caused #8338 to be reverted in #8475:

1. **Serializer Agnostic (No Pickle Enforcement):**
   - The column is defined as `sa.LargeBinary, nullable=True` rather than `sa.Column(PickleType)`.
   - Serialization goes through `self.encode(children)` / `self.decode(children)`, strictly honoring the user's configured `result_serializer` (JSON, pickle, msgpack, etc.).

2. **Existing Tables Without the `children` Column:**
   - Previous PR failed on existing deployments with `OperationalError: no such column: celery_taskmeta.children`.
   - Here, `_migrate_missing_columns` automatically runs `ALTER TABLE celery_taskmeta ADD COLUMN children ...` at model preparation time, gracefully catching database permission limits without crashing.

3. **Legacy Rows:**
   - When reading older rows where `children` is `None` or absent, `meta['children']` is safely `None` and `AsyncResult.children` returns `None` without error.
   - All legacy database unit tests continue to pass (89/89 tests passing, 100% diff coverage).

auvipy (dismissed):

MehrazRumman (changes requested):
## Review

Thanks for picking this up and for the careful write-up of why #8338 was reverted. The serializer handling is correct now: children go through `self.encode` / `self.decode`, so `result_serializer = json` works and there is no forced pickle. I checked out the branch and ran `t/unit/backends/test_database.py` locally, all 89 tests pass.

Since backward compatibility is the main concern, I focused on that, and I think there is one gap that needs fixing before merge.

### Blocking

**1. If the automatic migration fails, the backend is broken exactly as in #8475, but now silently.**
`_migrate_missing_columns` wraps everything in `except Exception: pass`. If `ALTER TABLE` is refused (a read-only DB role, a DBA-managed schema, a dialect quirk), nothing is logged and the worker starts up looking healthy. The very next `store_result` or `get_task_meta` then fails, because the ORM always selects the `children` column:

```
OperationalError: (sqlite3.OperationalError) no such column: celery_taskmeta.children
```

I reproduced this by stubbing the migration to a no-op against a pre-existing table without the column. Suggested changes:

- Log at `warning` (or `error`) level when the ALTER fails, with the exact statement so the operator can run it by hand.
- Document the manual migration in the changelog / database backend docs for users who do not allow DDL from the app.
- Ideally, make the read path tolerant: if the column is known to be missing, defer `children` on the mapped class or fall back to a select that excludes it, so a failed migration degrades to "children not stored" rather than "backend unusable".

### Worth discussing

**2. Running DDL on existing production tables at worker start-up.** Celery already does `create_all`, so there is precedent, but altering an existing table is more intrusive. On MySQL before 8.0 `ADD COLUMN` rebuilds the table and takes a lock, which can be a very long stall on a large `celery_taskmeta`. Multiple workers starting at once will also race on the ALTER; on most databases the losers just error and get swallowed, but it is worth a comment in the code and a note in the docs. Maintainers may prefer an opt-in setting for the auto-migration.

**3. Unquoted identifiers in the ALTER statement.** `table.fullname` and `col.name` are built with an f-string. They are not user-controlled in the injection sense, but `database_table_names` / `database_table_schemas` let users pick names that may need quoting. Using `engine.dialect.identifier_preparer.quote(...)` would be safer and costs nothing.

### Minor

- The `hasattr(task, 'children')` guard and the `TaskWithoutChildren` test cover a path that cannot happen in practice, since `_update_result` is always called with the ORM model. Not harmful, just noise.
- The PR description says children are reconstructed as `AsyncResult` objects in this backend; in the code that actually happens in the base `AsyncResult.children` path via `result_from_tuple`. Worth correcting the description so it does not mislead future readers.
- A changelog entry describing the new column and the manual `ALTER TABLE` statement would help operators who upgrade.

Once point 1 is addressed (at minimum, log the failure and document the manual migration), I think this is in good shape.

auvipy (approved):

auvipy (approved):

PR #10608, opened 2026-09-12
Allow crontab to correctly handle daylight switching time (correct fork...)

### LINKED ISSUE #10107: Hourly crontab skips execution during fall-back DST transition

Amli:
(same as https://github.com/celery/celery/pull/10585 from a more open fork, see original PR for a first batch of review and comments)

## Description
This pull request is an attempt to get the crontab handle daylight saving correctly:
- do not produce a datetime that does not exists (at DST start, when an hour disappears)
- do not produce a date an hour late (this currently happens at DST end, when an hour appears twice the same day)
- do not produce a next date in the past (remaining_estimate does that on dst end)

After [trying to hack](https://github.com/opendatasoft/celery/commit/c8c73a3a2e643ac80360dba16b85d094e6ec57ac) the current code, I ended up with a complete rewrite of the main function that computes the next run date for a crontab because the existing code was entirely ignoring timezone and makes it very difficult to just hack around it to get a correct result in all cases.

## Disclaimer
That's a relatively big change on the crontab code (even more for a first time contribution), I hesitated quite some time before opening this and I would understand if was too risky but, given the number of issues involving timezones and DST, I think it could be a good thing to have a code that handles timezone at its heart.
It is also highly possible I overlooked some cases (not aware datetime do not seem to be possible in this part of the code but I wonder), so I would be happy to get more insight.
I'm obviously available to discuss all this and ready to change the approach if needed.

## Changes
The new `crontab._next_occurence()` function produces a correct timezone aware date for the next expected crontab run.
It replaces the uses of `remaining()` and `remaining_delta()` (and the internal `_delta_to_next()`).

It does not work with delta at all, the delta is simply computed in `remaining_estimate()` as a simple difference between now and next_run_at (in utc).

The logic inside `_next_occurence()` is a mix of the existing `remaining_delta()`, for the part where we check the current day first to be as fast as possible and `_delta_to_next()` where we look for candidate days.

The code tries to be as fast as possible while only producing correct datetimes, so it is possibly a bit slower, but as far as I can tell not significantly slower.

The existing `remaining_delta()` was kept and modified to use the new code but still produce an `ffwd` instance.
AFAIK it is not used in the codebase except in an assert in the tests but I don't know if it considered to be a visible method or not.

The changes here do not try to modify the cron behaviour, if a crontab is scheduled at `2` on a day that skips that hour in this timezone, I will just not run, and the opposite if the `2` hour happens 2 times, it will run twice that day.

## Tests
I added a whole bunch of tests for cases around DST in multiple timezone to ensure my implementation was correct.
Some tests come from https://github.com/celery/celery/pull/10131 and in the end, there are probably too many tests and duplicated cases...

Fixes #10107

First time contribution, do not hesitate to tell me if things should be done differently.

### DIFF
--- a/celery/schedules.py
+++ b/celery/schedules.py
@@ -500,7 +500,7 @@ class crontab(BaseSchedule):
         super().__init__(**state)
         self._orig_kwargs = dict(state)
 
-    def _next_occurence(self, current: datetime) -> datetime:
+    def _next_occurrence(self, current: datetime) -> datetime:
         """returns the date of next crontab execution after given current date.
 
         The returned date is timezone aware and uses the same timezone as the given
@@ -644,7 +644,7 @@ class crontab(BaseSchedule):
         last_run_at = self.maybe_make_aware(last_run_at).astimezone(schedule_tz)
         now = self.maybe_make_aware(self.now()).astimezone(schedule_tz)
 
-        _next = self._next_occurence(last_run_at)
+        _next = self._next_occurrence(last_run_at)
         delta = ffwd(
             year=_next.year if last_run_at.year != _next.year else None,
             month=_next.month if last_run_at.month != _next.month else None,
@@ -672,7 +672,7 @@ class crontab(BaseSchedule):
         last_run_at = self.maybe_make_aware(last_run_at).astimezone(schedule_tz)
         now = self.maybe_make_aware(self.now()).astimezone(schedule_tz)
 
-        next_run_at = self._next_occurence(last_run_at)
+        next_run_at = self._next_occurrence(last_run_at)
 
         if C_REMDEBUG:  # pragma: no cover
             print(
--- a/celery/schedules.py
+++ b/celery/schedules.py
@@ -521,11 +521,10 @@ class crontab(BaseSchedule):
             """
             sorted_month_of_year = sorted(self.month_of_year)
             sorted_day_of_month = sorted(self.day_of_month)
-            # we cannot have more than 8 years between 2 leap years
-            # which should be the biggest possible gap between 2 runs
-            # (this is technically false if a day_of_week constraint is added to a
-            # crontab specifying month=2 day=29...)
-            for y in range(start.year, start.year + 8):
+
+            # 50 years search should above possible time between 2 leap days
+            # with day of week specified in the crontab
+            for y in range(start.year, start.year + 50):
                 for m in [
                     _m for _m in sorted_month_of_year
                     if y != start.year or _m >= start.month
--- a/celery/schedules.py
+++ b/celery/schedules.py
@@ -550,9 +550,9 @@ class crontab(BaseSchedule):
             """
             original_day = d.day
             d = d.replace(minute=0)
-            # we should never try more than 25 hours for a complete day
-            # (24h + 1 potential duplicated hour during dst time shift)
-            for _ in range(25 - d.hour):
+            # we should never try more than 24 hours for a complete day
+            # (23h til last hour + 1 potential duplicated hour during dst time end)
+            for _ in range(24 - d.hour):
                 d = _move_forward(d, timedelta(hours=1))
                 # resolve time that do not exist in current timezone
                 # and move to next possible time
--- a/celery/schedules.py
+++ b/celery/schedules.py
@@ -551,18 +551,16 @@ class crontab(BaseSchedule):
             We use a simple loop over the hours here to correctly handle DST changes.
             """
             original_day = d.day
-            d = d.replace(minute=0)
             # we should never try more than 24 hours for a complete day
             # (23h til last hour + 1 potential duplicated hour during dst time end)
             for _ in range(24 - d.hour):
-                d = _move_forward(d, timedelta(hours=1))
+                d = _move_forward(d, timedelta(minutes=60 - d.minute))
                 # resolve time that do not exist in current timezone
-                # and move to next possible time
-                d = resolve_imaginary(d)
+                # d = resolve_imaginary(d)
                 # check if this can make us change the day and get out if it does
                 if original_day != d.day:
                     return
-                if d.hour in self.hour:
+                if d.hour in self.hour and any(True for m in self.minute if m >= d.minute):
                     yield d
 
         def _has_run_hours_after(d: datetime) -> bool:
@@ -603,7 +601,10 @@ class crontab(BaseSchedule):
 
         # first we do a quick check on current day
         if _may_run_today:
-            _may_run_this_hour = candidate.hour in self.hour and any(m for m in self.minute if m > candidate.minute)
+            _may_run_this_hour = (
+                candidate.hour in self.hour
+                and any(True for m in self.minute if m > candidate.minute)
+            )
             if _may_run_this_hour:
                 # if there are slots later this hour, we can safely add one minute with worrying about hour change.
                 # the search for a correct minute will happen right after
--- a/celery/schedules.py
+++ b/celery/schedules.py
@@ -541,7 +541,9 @@ class crontab(BaseSchedule):
                             continue
                         # check if we have constraint on day of week
                         if not _crontab_has_day_of_week or new_date.isoweekday() % 7 in self.day_of_week:
-                            yield new_date
+                            # at dst start for timezone that trigger the change at midnight,
+                            # we could be generating a non existent date
+                            yield resolve_imaginary(new_date)
 
         def _generate_run_hours_after(d: datetime) -> Generator[datetime]:
             """yields candidate hours for task execution.
--- a/celery/schedules.py
+++ b/celery/schedules.py
@@ -2,20 +2,19 @@
 from __future__ import annotations
 
 import re
-from bisect import bisect, bisect_left
 from collections import namedtuple
 from datetime import datetime, timedelta, tzinfo
-from typing import Any, Callable, Iterable, Mapping, Sequence, Union
+from typing import Any, Callable, Generator, Iterable, Mapping, Sequence, Union
 
+from dateutil.tz import resolve_imaginary
 from kombu.utils.objects import cached_property
 
 from celery import Celery
 
 from . import current_app
 from .exceptions import ImproperlyConfigured
-from .utils.collections import AttributeDict
-from .utils.time import (ffwd, humanize_seconds, localize, maybe_make_aware, maybe_timedelta, remaining, timezone,
-                         weekday, yearmonth)
+from .utils.time import (C_REMDEBUG, ffwd, humanize_seconds, localize, maybe_make_aware, maybe_timedelta, remaining,
+                         timezone, weekday, yearmonth)
 
 __all__ = (
     'ParseException', 'schedule', 'crontab', 'crontab_parser',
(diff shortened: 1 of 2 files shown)


GangEunzzang (commented):
@Amli @auvipy I re-ran the cases from my earlier review against `1880790f5`, with `main` at `5ea4a4919`. The seven earlier correctness repros now pass: the three sparse schedules, the Lord Howe exception and repeated-instant cases, and the Santiago and Chatham skipped-hour cases. I also ran `test_schedules.py` and `test_beat.py` together on macOS 15.6 / Python 3.12.13: 442 passed, 11 skipped on this branch.

While checking other DST transitions, I found three regressions. In each case, `main` and a simple UTC-minute search return the expected time:

| Timezone / schedule | `last_run_at` | Expected next run; also returned by `main` | This branch |
| --- | --- | --- | --- |
| `Pacific/Chatham`, `hour=3, minute=50` | `2026-09-26 03:50+12:45` | `2026-09-27 03:50+13:45` | `2026-09-28 03:50+13:45` |
| `Pacific/Chatham`, `hour=3, minute=50` | `2026-04-04 03:50+13:45` | `2026-04-05 03:50+12:45` | `2026-04-05 02:50+12:45` |
| `Antarctica/Troll`, `hour=23, minute=0` | `2026-10-24 23:00+02:00` | `2026-10-25 23:00+00:00` | `2026-10-26 23:00+00:00` |

These reproduce through `remaining_estimate()` with `now == last_run_at`:

```python
from datetime import datetime, timezone
from zoneinfo import ZoneInfo


def next_run(zone, parts, hour, minute):
    tz = ZoneInfo(zone)
    last = datetime(*parts, tzinfo=tz)
    c = crontab(hour=hour, minute=minute, nowfun=lambda: last)
    c.tz = tz
    return (last.astimezone(timezone.utc)
            + c.remaining_estimate(last)).astimezone(tz)


print(next_run("Pacific/Chatham", (2026, 9, 26, 3, 50), 3, 50))
print(next_run("Pacific/Chatham", (2026, 4, 4, 3, 50), 3, 50))
print(next_run("Antarctica/Troll", (2026, 10, 24, 23, 0), 23, 0))
```

For Chatham's spring transition, advancing from 02:00 by 60 elapsed minutes reaches 04:00, passing over the valid 03:45–03:59 portion of the hour. At fall-back, `_get_run_minute()` advances from the first 03:00 by 50 elapsed minutes and returns the second 02:50 without rechecking `hour=3`. For Troll, the two-hour fall-back makes the local day 26 hours long, and the `range(24 - d.hour)` limit ends the search before 23:00.

I think the hour traversal still needs to account for these partial or longer repeated intervals, and the final candidate should satisfy the cron fields after advancing in UTC. Would these three cases be useful additions to the regression tests?

Separately, I confirmed that Berlin's explicit daily 02:00 schedule still runs on both folds, as the PR description intends; `main` runs it once. I would describe that as a deliberate behavior change. The wrong-hour and skipped-date cases above occur independently of that policy choice.

These checks cover schedule calculations and the existing Beat unit tests; I haven't run a live Beat/broker integration test for this revision.

auvipy:
@Amli when you are available please also add integrations tests for this

MehrazRumman (changes requested):
Thanks for taking this on, the timezone-first approach is the right direction and the test coverage is much better than what `main` has. I ran the new `_next_occurrence()` at the current head (`5226767`) against a brute-force oracle (step one minute at a time in UTC, convert to the schedule tz, accept the first instant whose local fields match the cron spec) across every DST transition in 2024–2026 for 15 zones, 30 schedules, and `current` values from -150 to +150 minutes around each transition: 63,210 cases.

**The good news:** zero mismatches, zero exceptions, and zero non-existent results for the 1-hour zones (Paris, London, LA, New York, Santiago, Havana, Asuncion, Azores, Sydney, Auckland, Nuuk), the 30-minute zones (St. John's, Lord Howe), including the midnight-transition cases. That is a real improvement over `main`.

**Requesting changes because the three regressions @GangEunzzang reported on `1880790f5` are still present** (no author commits since; only merges from `main`), and the sweep shows they are symptoms of three general gaps rather than one-offs:

| Zone | Mismatches | Root cause |
| --- | --- | --- |
| `Antarctica/Troll` (2 h fall-back) | 150 | `_generate_run_hours_after` bounds the loop with `range(24 - d.hour)`, i.e. by wall-clock hours. A 26-hour local day ends the search before `23:00`, so `hour=23` schedules skip a whole day. |
| `Pacific/Chatham` (45-min offset) | 201 | Two things. (a) `_move_forward(d, timedelta(minutes=60 - d.minute))` advances a full UTC hour, so on spring-forward it jumps from `02:00` straight to `04:00` and never visits the valid `03:45–03:59` window. (b) `_get_run_minute()` and the top-of-function quick path advance minutes in UTC but never re-check `d.hour in self.hour`, so when the shift lands mid-hour the result is in the wrong hour (`hour=3` returns `02:50`; `hour=2` returns `03:45`). |

120 of the 351 mismatches return a datetime whose local fields do not satisfy the cron spec at all, which is worse than being a day late: Beat will fire a job at an hour the user never asked for.

Reproduction for the three original cases (still fail at `5226767`):

```python
from datetime import datetime
from zoneinfo import ZoneInfo

def nxt(zone, parts, hour, minute):
    tz = ZoneInfo(zone); last = datetime(*parts, tzinfo=tz)
    c = crontab(hour=hour, minute=minute); c.tz = tz
    return c._next_occurrence(last)

nxt("Pacific/Chatham",  (2026, 9, 26, 3, 50), 3, 50)   # 09-28 03:50, expected 09-27 03:50
nxt("Pacific/Chatham",  (2026, 4, 4, 3, 50),  3, 50)   # 04-05 02:50, expected 04-05 03:50
nxt("Antarctica/Troll", (2026, 10, 24, 23, 0), 23, 0)  # 10-26 23:00, expected 10-25 23:00
```

**Suggested fix shape.** Rather than patching each path, make the final step self-validating: whatever candidate the fast paths produce, verify every field in local time, and if it fails, fall back to a bounded search (hour by hour in UTC until the local day changes, then minute by minute inside the matching hour). Bounding the hour loop by "local day changed" instead of `24 - d.hour` fixes Troll by itself. A parametrized test over `Pacific/Chatham` and `Antarctica/Troll` with the oracle above would lock all of this in and is cheap to run.

**Behaviour change that needs an explicit decision.** With this PR a daily `hour=2, minute=0` in `Europe/Paris` runs twice on 2024-10-27 (`02:00+02:00` and again `02:00+01:00`); `main` runs it once, and Vixie cron also skips wall-clock jobs that fall into a repeated interval when clocks go back by less than 3 hours. Running twice may be acceptable, but it should be a deliberate, documented choice (release note, and ideally a setting), not a side effect of the rewrite, since it changes production Beat behaviour for every DST zone.

**Smaller items:**
- `Generator[datetime]` is invalid on Python < 3.13 (`typing.Generator` needs three args there). It only works because of `from __future__ import annotations`; `Iterator[datetime]` is the safe spelling.
- Docstring of `_generate_run_days_after` says "8 years" but the code searches 50; the comment above the loop should carry the reasoning about the 40-year Feb-29 weekday gap instead.
- Two commented-out lines (`# d = resolve_imaginary(d)`, `# candidate = _move_forward(...)`) should go.
- Typo: "with worrying about hour change" → "without".
- `_has_run_hours_after()` followed by `_get_next_run_hour()` runs the same generator twice; a single `next(gen, None)` covers both.
- CI has only run 2 checks on this PR; the full unit and integration matrix does not appear to have executed yet, so the new tests are not verified on the project's supported Python versions.

Happy to re-run the sweep on a revised head.

Amli:
Thank you for your support, I'm discovering so many edge cases here.

I managed to fix a few things and add some integration tests:
#### smaller items from @MehrazRumman review
they should be fixed or cleaned up:
- typo and invalid comments
- incorrect typing for Generator
- double call to the same generator

#### next occurrence computation correctness
The missed day for `Antartica/Troll` cases should be fixed and a test was added for this specific case.
The incorrect estimates for `Pacific/Chatham` should be fixed too (and tested).

Overall, correctness should be good for all timezones now 🤞

#### Integration tests
I added some integration tests, only one fail in `main`, because remaining estimate returns a negative value that is considered as due
@auvipy I'll let you tell me if that's enough or if you have other kind of tests in mind.

### What remains 

#### Performance
As mentioned in the initial review, the time required to find the next occurrence is generally slow, most cases are still between 2 and 10 times slower.
The hour search is the main culprit but I think the loop on every hour could be skipped, I'll work on that today.

#### Behaviour change
I have no strong position on one or the other way of behaviour when the hour is explicitly defined in the crontab:
- running twice when the time goes back if the hour (the current behaviour in this PR)
- running only once on the first occurrence (the behaviour in main)

I'm all for not introducing breaking changes here and can modify the code to skip the occurrence if `hour != "*"` and the occurrence would be in the folded time (approximative condition).

This behaviour could also be defined through an option but I'm not sure it would be worth the added complexity.

auvipy:
backward compatibility is also a concern. and if the change introduce severe performance degrade, that that is a big concern. I would suggest try to keep the API as close as possible

Amli:
I'll let reviewers verify my claims but last commits should address the remaining concerns:

**Behaviour change**, crontab defined with anything other than `hour=*` will now not run during folded time, which should match the current `main` behaviour.
Crontab with `hour=*/2` and such also won't run in duplicated hours.

I coded the exception in `remaining_estimate()` that calls `_next_occurrence()` again when an unwanted case happens because I wanted to keep the main next occurrence logic clean from exception, as it is complicated enough as it is.


**Performance improvements**, new code is now somewhere between 1x -> 2.5x the time in `main`, the slowest being the search for day_of_week alone (possibly could be optimized specifically if needed).

Benchmark on my machine for `ct.remaining_estimate(datetime(year=2026, month=1, day=1, tzinfo=timezone.utc))` , 10000 loops, 5 repeat, timing per loop, best of 5
|crontab|main|branch|
|-------|-----|-------|
|hour=23, minute=59|3.96 usec|5.37 usec|
|minute=0, hour=9, day_of_week=1|4.12 usec|10.2 usec|
|minute=0|4.62 usec|6.05 usec|
|minute='*'|5.54 usec|5.93 usec|
|day_of_month=1, month_of_year=1|5.64 usec|5.96 usec|
|day_of_month=25, month_of_year=10|7.19 usec|6.94 usec|
|day_of_month=25, month_of_year=10, day_of_week=1|9.62 usec|7.44 usec|

@MehrazRumman @GangEunzzang I'm curious to see if I missed some more points, if would be kind enough to run your tests cases once again

GangEunzzang (commented):
@Amli @auvipy Re-ran everything on `0fe728a3`. The three Chatham and Troll regressions I reported are gone, the older repros still pass, and I see regression tests for all three landed — thanks for adding them.

I also ran a broader check this time: 24,000 schedule/instant combinations across ten zones, each compared against a brute-force UTC-minute search. The branch agrees with that search on all 24,000. `main` disagrees on 14, all spring-forward cases in Lord Howe, Troll, New York, and St John's.

The fold behavior also matches the PR description. With an explicit hour, the second occurrence of a repeated hour is skipped in Troll, Berlin, and New York alike; `hour="*"` is the case that fires on both folds. That distinction makes sense to me.

One small thing: `test_antartica_troll_dst_end_45_minute` asserts `_next_occurrence()`, which still returns the folded `02:45`. `remaining_estimate()` then advances that to the 26th, so an `hour=2` schedule won't actually fire on the fold. I'd assert this through `remaining_estimate()` instead, since that's the path Beat ultimately uses. Not blocking, but it would make the regression test match the behavior we're actually relying on.

Performance matches your table too. `day_of_week=1` by itself is the worst case I measured at roughly 2.5x `main`, down from about 15x on #10585.

As before, this covers the schedule calculations and unit tests; I haven't run a live Beat/broker setup.

LGTM!

Amli:
indeed, I'll add an assert for `remaining_estimate()` in `test_antartica_troll_dst_end_45_minute` too 👍

GangEunzzang (commented):
@Amli The new assertion checks out, but the two `print()` calls in `remaining_estimate()` also run with debugging off. Could we remove them or put them behind `C_REMDEBUG`? Otherwise looks good.

Amli:
Oops, debug `print()` removed, thanks for spotting them

PR #10614, opened 2026-09-12
Accept timedelta for countdown, expires and time limits

### LINKED ISSUE #10367: Allow passing durations as timedeltas
# Checklist

  to find out if the same enhancement was already implemented in the
  main branch.

#### Related Issues

- None

#### Possible Duplicates

- None

# Brief Summary

It would be a great improvement to DX if the duration arguments (`expires`, `time_limit`, `soft_time_limit`, etc) accepted a timedelta. In that way we avoid the guessing about what unit the argument should be in. At least with my IDE, on hover I don't get a description of the argument, so it's not immediately clear the argument should be in seconds.

In the codebases I've worked with is quite common to see code like the following:

```python
@shared_task(
    soft_time_limit=60 * 60 * 2,  # 2 hours
    time_limit=60 * 60 * 2 + 60,  # 2 hours + 1 minute
    expires=60 * 60 * 4,  # 4 hours
)
def task():
  ...
```

And this could be much better be expressed as 

```python
@shared_task(
    soft_time_limit=timedelta(hours=2),
    time_limit=timedelta(hours=2, minutes=1),
    expires=timedelta(hours=4),
)
def task():
  ...
```

Passing the timedeltas is possible today, but it quite verbose because many of those arguments are documented as ints.

```python
@shared_task(
    soft_time_limit=int(timedelta(hours=2).total_seconds()),
    time_limit=int(timedelta(hours=2, minutes=1).total_seconds()),
    expires=int(timedelta(hours=4).total_seconds()),
)
def task():
  ...
```
# Design

None

## Proposed Behavior

The changes to support the changes depend on whether timedeltas are already used. For example, for `expires`, the [current logic](https://github.com/celery/celery/blob/aef7f130e3ca547c49d3d581a831927458fc1a71/celery/app/amqp.py#L348-L354) already converts seconds to a timedelta after validation.

This could be easily changed as following
```python
if isinstance(expires, numbers.Real):
    expires = timedelta(seconds=expires)

if isinstance(expires, timedelta):
    self._verify_seconds(int(expires.total_seconds()), 'expires') # the int casting is optional
    now = now or self.app.now()
    timezone = timezone or self.app.timezone
    expires = maybe_make_aware(
        now + timedelta(seconds=expires), tz=timezone,
    )
```

For fields, like `time_limit`, the protocol already supports floats. So they can be converted to floats using `total_seconds`.

If any other fields only allow for integer seconds, then if we accepted timedeltas we should raise a warning when passing timedeltas with a fractional seconds.
## Diagrams

N/A

## Alternatives

None

rmnvg:
## Description

Fixes #10367.

`countdown`, `expires`, `time_limit` and `soft_time_limit` are documented as a number of seconds, which makes longer durations awkward to write and easy to misread:

```python
@shared_task(
    soft_time_limit=60 * 60 * 2,       # 2 hours
    time_limit=60 * 60 * 2 + 60,       # 2 hours + 1 minute
    expires=60 * 60 * 4,               # 4 hours
)
def task(): ...
```

Passing a `timedelta` instead currently fails, in two different ways:

* **`countdown` / `expires`** raise `TypeError: '<' not supported between instances of 'datetime.timedelta' and 'int'`, from the seconds range check — far from the call site that caused it.
* **`time_limit` / `soft_time_limit`** are accepted silently and placed in the message headers as `timedelta` objects, then fail later when the message is serialized:

  ```
  timelimit header -> [datetime.timedelta(seconds=30), datetime.timedelta(seconds=10)]
  TypeError: Object of type timedelta is not JSON serializable
  ```

After this change:

```python
@shared_task(
    soft_time_limit=timedelta(hours=2),
    time_limit=timedelta(hours=2, minutes=1),
    expires=timedelta(hours=4),
)
def task(): ...
```

## Implementation

Adds `celery.utils.time.maybe_seconds()`, the inverse of the existing `maybe_timedelta()`, and uses it to normalise these four arguments in:

* `Celery.send_task()`
* `AMQP.as_task_v1()` and `AMQP.as_task_v2()`

Normalising in `send_task()` matters as well as in `amqp.py`: the native delayed-delivery routing and the expiration check both do arithmetic and comparisons on these values before the message is built, so converting only at message-creation time would leave those paths broken.

A `timedelta` and the equivalent number of seconds now produce identical task messages. Numbers, `None` and `datetime` values are unaffected, so this is backwards compatible.

Docstrings for `apply_async()` and `retry()` and the *ETA and Countdown* / *Expiration* sections of the calling guide are updated to mention the accepted types.

## Verification

New tests, all passing:

* `t/unit/utils/test_time.py` — `maybe_seconds()` over timedeltas, numbers and `None`
* `t/unit/app/test_amqp.py` — timedelta `countdown`/`expires` for `as_task_v2`; an equivalence test asserting a timedelta and the equivalent seconds produce identical messages for both protocols; a JSON-serialisability test for `timelimit`
* `t/unit/app/test_app.py` — `send_task()` normalises all four arguments

Checklist from `CONTRIBUTING.rst`:

* `pytest t/unit` — same 3 pre-existing failures as `main`, +14 passing tests, no new errors (the remaining errors are a pre-existing teardown cascade from a leaked thread in `t/unit/contrib/test_rdb.py`; running the affected modules without it gives 789 passed, 0 errors)
* `pre-commit run --files <changed>` — passed, including mypy and isort
* `bandit -b bandit.json celery/` — 0 issues
* `sphinx-build -b apicheck` / `-b configcheck` — no new warnings

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/celery/app/amqp.py
+++ b/celery/app/amqp.py
@@ -14,7 +14,7 @@ from celery import signals
 from celery.utils.nodenames import anon_nodename
 from celery.utils.saferepr import saferepr
 from celery.utils.text import indent as textindent
-from celery.utils.time import maybe_make_aware
+from celery.utils.time import maybe_make_aware, maybe_seconds
 
 from . import routes as _routes
 
@@ -339,6 +339,10 @@ class AMQP:
             raise TypeError('task args must be a list or tuple')
         if not isinstance(kwargs, Mapping):
             raise TypeError('task keyword arguments must be a mapping')
+        countdown = maybe_seconds(countdown)
+        expires = maybe_seconds(expires)
+        time_limit = maybe_seconds(time_limit)
+        soft_time_limit = maybe_seconds(soft_time_limit)
         if countdown:  # convert countdown to ETA
             self._verify_seconds(countdown, 'countdown')
             now = now or self.app.now()
@@ -432,6 +436,10 @@ class AMQP:
             raise TypeError('task args must be a list or tuple')
         if not isinstance(kwargs, Mapping):
             raise TypeError('task keyword arguments must be a mapping')
+        countdown = maybe_seconds(countdown)
+        expires = maybe_seconds(expires)
+        time_limit = maybe_seconds(time_limit)
+        soft_time_limit = maybe_seconds(soft_time_limit)
         if countdown:  # convert countdown to ETA
             self._verify_seconds(countdown, 'countdown')
             now = now or self.app.now()
--- a/celery/app/base.py
+++ b/celery/app/base.py
@@ -38,7 +38,7 @@ from celery.utils.functional import first, head_from_fun, maybe_list
 from celery.utils.imports import gen_task_name, instantiate, symbol_by_name
 from celery.utils.log import get_logger
 from celery.utils.objects import FallbackContext, mro_lookup
-from celery.utils.time import maybe_make_aware, timezone, to_utc
+from celery.utils.time import maybe_make_aware, maybe_seconds, timezone, to_utc
 
 from ..utils.annotations import annotation_is_class, annotation_issubclass, get_optional_arg
 from ..utils.quorum_queues import detect_quorum_queues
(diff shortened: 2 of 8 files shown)


auvipy:
we already have an open pr for this https://github.com/celery/celery/pull/10377 . please dont just open anything and everything with ai before verifying....

PR #10618, opened 2026-09-13
docs: fix awkward 'exception occurred' phrasing in task states

mrchatam:
*Note*: Before submitting this pull request, please review our [contributing guidelines](https://docs.celeryq.dev/en/main/contributing.html).

## Description

Clarify awkward phrasing in the FAILURE task state docs:

`contains the exception occurred` → `contains the exception that occurred`.

## Test plan

- [x] Confirmed surrounding sentence still reads correctly after the change.

### DIFF
--- a/docs/userguide/tasks.rst
+++ b/docs/userguide/tasks.rst
@@ -1306,7 +1306,7 @@ FAILURE
 
 Task execution resulted in failure.
 
-:meta-data: `result` contains the exception occurred, and `traceback`
+:meta-data: `result` contains the exception that occurred, and `traceback`
             contains the backtrace of the stack at the point when the
             exception was raised.
 :propagates: Yes


auvipy (changes requested):
instead of opening small 1 word pr, please fix more doc erros in a single pr

MehrazRumman (commented):
The one-word change is correct and reads fine in context.

To address @auvipy's request to batch more fixes into this PR, here are three concrete ones in the same "Task states" section of `docs/userguide/tasks.rst`, a few lines below your change:

1. **RETRY state, line 1323:** same kind of slip as the one you fixed:
   `when the exceptions was raised.` → `when the exception was raised.`

2. **FAILURE state:** the entry lists `:propagates: Yes` but is missing `:ready: Yes`. `FAILURE` is in `READY_STATES` in `celery/states.py`, so the doc entry is currently inconsistent with `SUCCESS`, which does list it.

3. **REVOKED state:** same gap. `REVOKED` is in both `READY_STATES` and `PROPAGATE_STATES`, so the entry should carry `:ready: Yes` alongside its existing `:propagates: Yes`.

For reference, the code is:

```python
READY_STATES = frozenset({SUCCESS, FAILURE, REVOKED})
PROPAGATE_STATES = frozenset({FAILURE, REVOKED})
```

With those three folded in, the state reference would be internally consistent and match the implementation, which seems like a reasonable scope for a single docs PR.

PR #10619, opened 2026-09-13
Prevent spawned worker crashes with explicit lazy=True

phanky1:
## Description

A spawned worker crashes if a task explicitly sets `lazy=True` because 
https://github.com/celery/celery/blob/0c950f6af4f4fd55d409e69077c9b057a51f0c22/celery/app/base.py#L560
passes `lazy=False` alongside `**opts` supplies the keyword twice and raises a TypeError.

This sets `lazy=False` in the options dictionary before forwarding it to shared_task(), ensuring the keyword is passed only once.

### DIFF
--- a/celery/app/base.py
+++ b/celery/app/base.py
@@ -557,7 +557,8 @@ class Celery:
             # the task instance from the current app.
             # Really need a better solution for this :(
             from . import shared_task
-            return shared_task(*args, lazy=False, **opts)
+            opts['lazy'] = False
+            return shared_task(*args, **opts)
 
         def inner_create_task_cls(shared=True, filter=None, lazy=True, **opts):
             _filt = filter
--- a/t/smoke/tasks.py
+++ b/t/smoke/tasks.py
@@ -8,7 +8,7 @@ from signal import SIGKILL
 from time import sleep
 
 import celery.utils
-from celery import Task, shared_task, signature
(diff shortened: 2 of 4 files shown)


auvipy:
@GangEunzzang please give this a try

GangEunzzang (commented):
@auvipy @phanky1 Thanks for catching this. I missed the explicit `lazy=True` case when I tested the spawn path in #10583.

I tested `dc57a1d5` against main at `86ee0930`. On macOS 15.6 / Python 3.12.13 with Redis, I could reproduce the TypeError on main with a real spawned worker. This branch handles both `@app.task(lazy=True)` and `@current_app.task(lazy=True)` correctly; the request IDs match and both tasks are bound to the current app as expected.

I checked both configured spawn and automatic detection of billiard’s spawn start method. Each processed 20 tasks, then another 12 after a hard time limit replaced a child. The implicit-lazy spawn and explicit-lazy fork controls passed as well.

The new unit regression fails on main as expected. The combined worker/app/prefork tests pass here: 974 passed, 2 skipped. All 6 Linux Docker spawn smoke tests also passed on Python 3.13.15, covering RabbitMQ and Redis brokers with a Redis backend, explicit `lazy=True`, and child replacement.

Looks good to me.

auvipy (approved):

PR #10621, opened 2026-09-13
Prevent add_defaults() from crashing when config_from_object() fails on an already configured app

phanky1:
## Description

If `config_from_object(..., silent=True)` fails on an already configured app, `self._conf` has already been cleared, but `configured` is still True. Calling `add_defaults()` then raises an AttributeError.

This PR only clears self._conf after the configuration loads successfully.

```python
def add_defaults(self, fun):
    """Add default configuration from dict ``d``.

    If the argument is a callable function then it will be regarded
    as a promise, and it won't be loaded until the configuration is
    actually needed.

    This method can be compared to:

    .. code-block:: pycon

        >>> celery.conf.update(d)

    with a difference that 1) no copy will be made and 2) the dict will
    not be transferred when the worker spawns child processes, so
    it's important that the same configuration happens at import time
    when pickle restores the object on the other side.
    """
    if not callable(fun):
        d, fun = fun, lambda: d
    if self.configured:
>           return self._conf.add_defaults(fun())
               ^^^^^^^^^^^^^^^^^^^^^^^
E           AttributeError: 'NoneType' object has no attribute 'add_defaults'
```

### DIFF
--- a/celery/app/base.py
+++ b/celery/app/base.py
@@ -729,8 +729,8 @@ class Celery:
         self._config_source_silent = silent
         self.namespace = namespace or self.namespace
         if force or self.configured:
-            self._conf = None
             if self.loader.config_from_object(obj, silent=silent):
+                self._conf = None
                 return self.conf
 
     def config_from_envvar(self, variable_name, silent=False, force=False):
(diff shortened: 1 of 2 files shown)


auvipy:
any relevant open issue for this?

phanky1:
> any relevant open issue for this?

No, I just came across it while reading the code.

auvipy:
ok cool

MehrazRumman (approved):
Everything else looks goot to me, Just followup this issues !

### Test
Fails on `main` with the reported `AttributeError`, passes here. Suggestions:
1. Assert `app._conf is not None` (the actual invariant).
2. Add a non-silent companion test.

auvipy (approved):

PR #10623, opened 2026-09-13
Preserve configuration when copying settings

GangEunzzang:
## Description

Copying `app.conf` produces a settings object that cannot be read:

```python
from celery import Celery

app = Celery('example')
assert app.conf.task_always_eager is True

copied = app.conf.copy()
assert copied.task_always_eager is True  # TypeError: 'str' object is not callable
```

`copy.copy(app.conf)` fails in the same way. The inherited `ChainMap.copy()` expands the default mappings into positional arguments, but `ConfigurationView` expects a list of defaults followed by key converters and a prefix.

This adds a copy implementation that matches `ConfigurationView`'s constructor and preserves its key conversions. Changes get their own mapping, while default mappings and nested values remain shared. `Settings` also loads pending configuration before copying and preserves the deprecated-settings value without adding it back after `clear()`.

## Validation

- The 30 new cases give **26 failures and 4 passes** on main at `cdd516f71d98121d8a4a659d90d400a78b13dcd3`, and **30 passes** with the fix.
- `pytest -p celery.contrib.pytest t/unit/utils t/unit/app t/unit/test_generics.py`: **1,301 passed, 14 skipped, 1 xfailed**, with 6 subtests passed.
- `pre-commit run --all-files`: passed, including mypy.
- Four memory-broker worker checks using copied settings passed, covering both copy APIs before and after configuration was loaded. Changes to the copy left the original settings unchanged.
- The new copy methods have full line and branch coverage in the focused tests.

These checks used CPython 3.12.13 on macOS. I also ran `apicheck`, `configcheck`, and Bandit; the base and this branch reported the same documentation failures and Bandit findings. The existing `bandit.json` could not be parsed in either tree. I have not run the full upstream platform matrix locally.

### DIFF
--- a/celery/app/utils.py
+++ b/celery/app/utils.py
@@ -81,6 +81,18 @@ class Settings(ConfigurationView):
 
         self.deprecated_settings = deprecated_settings
 
+    def copy(self):
+        self.finalize()
+        copied = super().copy()
+        if 'deprecated_settings' in self.changes:
+            copied.changes['deprecated_settings'] = (
+                self.changes['deprecated_settings']
+            )
+        else:
+            copied.changes.pop('deprecated_settings', None)
+        return copied
+    __copy__ = copy
+
     @property
     def broker_read_url(self):
         return (
--- a/celery/utils/collections.py
+++ b/celery/utils/collections.py
@@ -373,6 +373,15 @@ class ConfigurationView(ChainMap, AttributeDictMixin):
             _keys=keys,
         )
 
+    def copy(self):
+        # type: () -> 'ConfigurationView'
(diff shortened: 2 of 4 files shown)


auvipy:
@phanky1  please check this

phanky1 (approved):
LGTM.

auvipy:
can you please check why the codeQL is failing?

GangEunzzang:
@auvipy Could you try re-running the CodeQL job with debug logging enabled? The logs stop at `Uploading results`, but don’t show why it failed.

auvipy (approved):

PR #10624, opened 2026-09-13
Prevent apply_async() from mutating cached execution options

phanky1:
## Description

When `apply_async()` receives an empty `options`, it reuses and mutates the task's cached options 
https://github.com/celery/celery/blob/ea1db4a551a40c3ec1192abeed4785853833cda5/celery/app/task.py#L686-L687
while adding value such as ignore_result and priority
https://github.com/celery/celery/blob/ea1db4a551a40c3ec1192abeed4785853833cda5/celery/app/task.py#L689-L691

Subsequent `send_task()` calls can then behave differently depending on earlier dispatches because they use the
mutated cached options as defaults.

This change always copies the cached options before adding defaults.


auvipy:
any open issues related to this?

phanky1:
> any open issues related to this?

I couldn't find any open issues for this. It's just something I found while reading through the codebase.

auvipy:
ok no prob

auvipy:
@GangEunzzang please review

GangEunzzang (commented):
@phanky1 @auvipy Tested this locally on `d0692a20` — unit suite passes and I couldn't find anything it misses. Looks good to me!

MehrazRumman:
LGTM, verified against `main` (`ea1db4a`).

**Root cause:** with empty `options`, `options = preopts` aliased the cached `_exec_options` dict, so `options.setdefault('ignore_result', ...)` wrote straight into the cache. `ignore_result` is the only key that actually leaked: `priority` is already part of `extract_exec_options`, so that `setdefault` is a no-op either way.

**Why it is user-visible:** `send_task()` merges `task_type._get_exec_options()` for plain `app.send_task("name")` calls (the `resolved_from_registry` block in `celery/app/base.py`). So one bare `task.apply_async()` on an `ignore_result=True` task turned every later `app.send_task(task.name)` into an ignored result, where `.get()` returns `None` immediately. The new smoke test reproduces exactly that sequence and would fail fast on `main` (`None == 11`), and the unit test would fail there too since the cached dict gains `ignore_result: True`. CI is green across the matrix, including the `t/smoke/tests/test_tasks.py` jobs.

Dropping the `if options` fast path costs a ~12-key shallow copy per call, which is negligible.

Optional, non-blocking: have `_get_exec_options()` return `dict(self._exec_options)` so no future caller can mutate the cache either.

MehrazRumman (approved):
LGTM, verified against main (ea1db4a).

Root cause: with empty options, options = preopts aliased the cached _exec_options dict, so options.setdefault('ignore_result', ...) wrote straight into the cache. ignore_result is the only key that actually leaked: priority is already part of extract_exec_options, so that setdefault is a no-op either way.

Why it is user-visible: send_task() merges task_type._get_exec_options() for plain app.send_task("name") calls (the resolved_from_registry block in celery/app/base.py). So one bare task.apply_async() on an ignore_result=True task turned every later app.send_task(task.name) into an ignored result, where .get() returns None immediately. The new smoke test reproduces exactly that sequence and would fail fast on main (None == 11), and the unit test would fail there too since the cached dict gains ignore_result: True. CI is green across the matrix, including the t/smoke/tests/test_tasks.py jobs.

Dropping the if options fast path costs a ~12-key shallow copy per call, which is negligible.

Optional, non-blocking: have _get_exec_options() return dict(self._exec_options) so no future caller can mutate the cache either.

auvipy (approved):

PR #10626, opened 2026-09-14
Task.replace() crashes with tuple-backed chains

phanky1:
## Description
Fixes a regression in #9604 since `sig.tasks` may be a tuple.
https://github.com/celery/celery/blob/ea1db4a551a40c3ec1192abeed4785853833cda5/celery/app/task.py#L1068-L1069


mohithvardhan002 (approved):
Reviewed the fix and regression coverage. Converting sig.tasks to a list before mutating the final group correctly handles tuple-backed chains, and the added tests cover both tuple-backed and normal positional chain construction.

auvipy (approved):

Issue #6603, opened 2021-01-18
Awaitable task result


bblanchon:
# Checklist

- [X] I have checked the [issues list](https://github.com/celery/celery/issues?utf8=%E2%9C%93&q=is%3Aissue+label%3A%22Issue+Type%3A+Feature+Request%22+)
  for similar or identical feature requests.
- [X] I have checked the [pull requests list](https://github.com/celery/celery/pulls?utf8=%E2%9C%93&q=is%3Apr+label%3A%22PR+Type%3A+Feature%22+)
  for existing proposed implementations of this feature.
- [X] I have checked the [commit log](https://github.com/celery/celery/commits/master)
  to find out if the if the same feature was already implemented in the
  master branch.
- [X] I have included all related issues and possible duplicate issues
  in this issue (If there are none, check this box anyway).

#### Related Issues

- #4353

#### Possible Duplicates

- #4353

#### Also related:

- [How to combine Celery with asyncio?](https://stackoverflow.com/questions/39815771/how-to-combine-celery-with-asyncio/43325237) (StackOverflow question)

# Brief Summary

Being able to `await` an `AsyncResult`

The goal is to offload CPU intensive tasks to dedicated workers *and* return a result to the client once the task is completed.

# Design

## Architectural Considerations

I don't know how to implement this feature, but it seems that this project does:
https://github.com/kai3341/celery-pool-asyncio
However, it's not compatible with Celery 5 (see https://github.com/kai3341/celery-pool-asyncio/issues/29)

## Proposed UI/UX

In a [Django 3.1 async view](https://docs.djangoproject.com/en/3.1/releases/3.1/#asynchronous-views-and-middleware-support):

```python
async def my_view(request):
    result = await my_cpu_intensive_task.delay()
    return HttpResponse(f"Result = {result}")
```

## Diagrams

N/A

## Alternatives

I've considered polling `AsyncResult.state` every second, but it's a very inelegant solution.

auvipy:
we can try this in django-celery-results along with celery core as a POC first. but it is a part of a bigger design decision which we are planning for celery 6.0. additionally we can try to use an aio-libs in one of our result back end and see how things need to be done as experiments.

philmday:
this would be a very useful feature.

bblanchon:
As [I wrote on stackoverflow](https://stackoverflow.com/a/66318397/1164966), I currently use the following helper as a workaround:

```python
import asyncio
from asgiref.sync import sync_to_async

[...]

Issue #6552, opened 2020-12-19
Support async function


SeyyedKomeil:
# Checklist

- [x] I have checked the [issues list](https://github.com/celery/celery/issues?utf8=%E2%9C%93&q=is%3Aissue+label%3A%22Issue+Type%3A+Feature+Request%22+)
  for similar or identical feature requests.
- [x] I have checked the [pull requests list](https://github.com/celery/celery/pulls?utf8=%E2%9C%93&q=is%3Apr+label%3A%22PR+Type%3A+Feature%22+)
  for existing proposed implementations of this feature.
- [x] I have checked the [commit log](https://github.com/celery/celery/commits/master)
  to find out if the if the same feature was already implemented in the
  master branch.
- [x] I have included all related issues and possible duplicate issues
  in this issue (If there are none, check this box anyway).

# Brief Summary

Support async celery tasks & call async function in celery tasks

# Design
```
async def foo(x, y):
    return x + y

@app.task(
    name='tasks.add',
    queue='custom',
    rate_limit='10/s',
)
async def add(x, y):
    await foo(x, y)
```

auvipy:
its a big undertake

SeyyedKomeil:
> its a big undertake

So any idea to call async function in celery task?
I use asyncio.run but sometimes got thread error.
@auvipy

auvipy:
it's not currently supported but we are working on implementing it as soon as possible

tasn:
Would be a dream. A lot of my tasks are IO bound rather than CPU bound. I was looking at maybe using eventlet, but it's not entirely obvious to me how it'll interact with asyncio (aiohttp and aiopg) which I'm using.

auvipy:
you can read celery CEP on this. though I need to write a CEP with some POC

Kludex:
> you can read celery CEP on this. though I need to write a CEP with some POC

Where is the CEP?

auvipy:
we have to write one but the high level ones are here https://github.com/celery/ceps

Kludex:
Is this open for work, or you'll personally create the CEP and the mentioned POC at some point? (not saying that I want to do it or will, just asking)

u23a:
Will it appear in 6.0.0 Milestone?

GuptaAman08:
Any rough estimation of when celery will support this going forward ??

aleGpereira:
This will be nice to have, for sure.

auvipy:
[...]

Issue #6483, opened 2020-11-16
Submitting celery tasks from asyncio


mbirtwell:
I posted on the Celery mailing list:

> I've seen some suggestions that asyncio support was hoped for Celery 5.0, but it doesn't seem to have happened. Does anyone know if this is being actively worked on?
> 
> I'm particularly interested in submitting tasks and waiting for results from an asyncio context as opposed to writing asyncio code in a celery worker. Having had a bit of a poke around it seems like writing a separate celery client just to cover the most obvious use-cases in this area would be quite doable. But I wanted to check this wasn't coming to main celery soon or if anyone knew of anything similar.

And it was suggested that I create an issue here to get some feedback on my approach. In brief I'm writing a separate client that speaks the celery protocol for submitting tasks. Where possible I'm making use of existing celery functions, which don't do any networking to help with that and re-implementing anything that talks to the network using aioamqp and aioredis. 

I've got some code, which is only really demoware at the moment: https://github.com/mbirtwell/asyncio_celery_client

xirdneh:
Looks promising.
I think the example is simple but we would need a bit more work to implement this in Celery's code base.

@thedrow What do you think?

thedrow:
I'm not against any other clients for Celery but this is not part of the design I had in mind. You should read the HLA CEP for Celery NextGen.

caniko:
Why asyncio exclusively? There are other async packages like `trio`

auvipy:
we might go with anyio

lucasgadams:
Any update on this?

PR #10634, opened 2026-09-15
asyncio pool: run coroutine task bodies on a shared event loop

ghirailghiro:
## Problem

Celery cannot run a task defined with `async def`. Calling a coroutine
function only builds a coroutine object — it never runs — and Celery
treats that object as the task's return value. Depending on where the
result goes, the task either reports `SUCCESS` with a coroutine object as
its result (with no result backend, or `ignore_result`), or fails with
`EncodeError: Object of type coroutine is not JSON serializable` (with a
JSON backend) — an error that names the serializer rather than the cause.

The reason is structural: the worker never hands a task's function to the
pool, it hands the *tracer* built by `build_tracer()` in
`celery/app/trace.py` — a synchronous closure that decodes the message,
fires signals, calls the body, stores the result and acks, all on one
straight line. As long as that line is synchronous, no pool can `await` a
task body without patching it. That's why the two existing out-of-tree
packages
([celery-aio-pool](https://github.com/the-wondersmith/celery-aio-pool),
[celery-pool-asyncio](https://github.com/kai3341/celery-pool-asyncio))
monkeypatch `build_tracer` rather than shipping as ordinary pools.

## Example

```console
$ celery -A proj worker --pool=asyncio --concurrency=100
```

`--concurrency` here means something different from `prefork`: it's how
many tasks may be in flight at once on the shared loop, not a process
count. Size it for how many concurrent requests you want, not for CPU
cores — the worker's own default (CPU count) is too low for this pool.

```python
@app.task
async def fetch(url):
    async with session.get(url) as response:
        return await response.text()
```

No special decorator, base class, or setting needed — that's the whole thing.

## What this does

Five independent, individually reviewable commits:

1. **The tracer resolves a coroutine returned by a task body**, inside
   the existing exception-handling block, so `Retry`/`Reject`/`Ignore`
   raised from a coroutine are handled exactly like their synchronous
   counterparts. How the coroutine is run is left to a small new module,
   `celery.utils.coroutines` — an execution pool that owns a loop
   installs its own runner there.
2. **`Local` storage (`celery.utils.threads`) can now hold more than one
   identity per thread**, via a `ContextVar`. Needed because `self.request`
   is normally keyed by thread, and many coroutines share one thread on an
   event loop — without this, `self.retry()` and `self.request.id` would
   read a neighbouring task's request. Mirrors what Celery already does
   for eventlet/gevent (identity = the current greenlet).
3. **A new `asyncio` execution pool** (`--pool=asyncio`). One event loop
   shared by the whole worker process, so a client session, connection
   pool or engine created once is valid for every task — unlike wrapping
   each task body in `asyncio.run()` or `asgiref.sync.async_to_sync`,
   which build a loop per call and break the second task that touches
   the same resource. The synchronous parts of tracing (decode, signals,
   result store, ack) still run in a bounded thread pool of
   `--concurrency` threads, so this pool needs neither an async broker
   nor an async result backend and doesn't prejudge how those get built
   later. Time limits and `revoke(terminate=True)` map onto asyncio
   cancellation, with both soft and hard limits implemented — neither
   `gevent` nor `threads` currently has both.
4. **`autoretry_for` now works on coroutine task bodies.** The wrapper it
   installs is synchronous, so given a coroutine body it only ever saw
   the coroutine being *built* — the exception is raised later, when the
   tracer awaits it. Fixed by deciding on the returned value at runtime,
   the same way the tracer does, rather than on the function's static
   shape (which would wrongly treat a `functools.wraps`-based
   `async_to_sync` adapter as a coroutine function and try to await
   something built to already be synchronous).
5. **The behaviour change is gated by a new setting with a deprecation
   period**, following the existing precedent of
   `broker_connection_retry_on_startup`: `worker_resolve_coroutines`
   unset (the default) keeps a coroutine task running on a private
   per-task loop on pools without one, with a
   `CPendingDeprecationWarning`; explicit `True` keeps that silently;
   explicit `False` raises `ImproperlyConfigured` now instead of in 6.0.

Standard library only — no new dependencies.

### How the pool fits together

```mermaid
flowchart TB
    A["Request.execute_using_pool()"] --> TP

    subgraph TP["ThreadPoolExecutor — one thread per task in flight, up to --concurrency"]
        direction LR
        T1["tracer thread 1<br/>decode · signals<br/>store_result · ack"]
        T2["tracer thread 2<br/>decode · signals<br/>store_result · ack"]
        T3["tracer thread N<br/>decode · signals<br/>store_result · ack"]
    end

    T1 -- "resolve_coroutine()<br/>run_coroutine_threadsafe()" --> EL
    T2 -- "resolve_coroutine()<br/>run_coroutine_threadsafe()" --> EL
    T3 -- "resolve_coroutine()<br/>run_coroutine_threadsafe()" --> EL

    subgraph EL["one asyncio event loop, one dedicated thread"]
        direction LR
        C1((coro 1))
        C2((coro 2))
        C3((coro N))
        SHARED[("shared client session /<br/>connection pool")]
        C1 -.-> SHARED
        C2 -.-> SHARED
        C3 -.-> SHARED
    end
```

## What this does not do

To place this in the bigger picture:

```mermaid
flowchart LR
    Client(["Client"])
    Broker[("Broker")]
    Body["async def task body<br/>THIS PR"]
    Backend[("Result backend")]

    Client -- "task.delay()<br/>sync, blocking" --> Broker
    Broker --> Body
    Body -- "store_result()<br/>sync, blocking" --> Backend
    Body -- "ack" --> Broker
    Backend -- "result.get()<br/>sync, blocking (polling)" --> Client

    classDef thispr fill:#2ea44f,color:#fff,stroke:#1b7f37;
    class Body thispr;
```

- **Publishing stays synchronous.** `task.delay()` from inside a
  coroutine still blocks the loop for the duration of the broker
  publish. Fixing that needs an async kombu client, out of scope here.
- **The result backend stays synchronous.** Storing a result still costs
  a blocking call on a tracer thread — it doesn't block the shared loop,
  but it is a cost. A fully async tracer would remove that; this doesn't
  block that follow-up.
- **No `await result`.** `AsyncResult.get()` is unchanged — that's
  #6603's issue, not this one.
- On very short tasks this pool is slower than prefork — not from
  anything in the pool itself, but because an event loop is a single
  thread: any CPU an async client spends per call serialises there. The
  usual asyncio fix applies here too — more worker processes, one loop
  each. On IO-bound tasks it used ~5% of prefork's memory at the same
  concurrency in a quick benchmark.

## Scope

This is deliberately narrow: it addresses #6552, #7874 and #3884 (running
coroutine task bodies), not #6603 or #6483 (an async client — publishing
a task or awaiting a result without blocking), which need work in kombu
this PR doesn't attempt.

Since it doesn't touch the broker or the result backend, I don't think it
conflicts with whatever native design 6.0 eventually settles on for that
side — it's an addition alongside it, not a competing implementation. On
that basis I'd suggest this is worth landing as an opt-in pool in a 5.x
release rather than waiting on that design, but happy to hear if you'd
rather treat it as a first step toward it instead. I've kept it as five
separate commits (one per numbered point above) in case that's easier to
review than the combined diff, or in case only part of it makes sense to
take now.

I'm not in a hurry to see this merged. It touches a sensitive, central
part of the worker (the tracer, `Local` identity), and I'd rather it get
a properly unhurried review — including anyone who wants to poke at the
concurrency edge cases — than get rushed through. Happy to keep iterating
on feedback for as long as that takes.

## Testing

Full unit suite passes, plus dedicated coverage for the new pool's
lifecycle, time limits, revocation, and request-context propagation
across concurrent coroutines sharing one thread.


ghirailghiro:
@auvipy  if there's appetite to go further than this PR, I'd also be glad to put
more time in: a fully async tracer (which would remove the per-task
thread this pool still needs) or the kombu-side work would
need are both natural next steps, and I'd rather help build toward
whatever direction the project wants than just leave this here and move
on.

auvipy (changes requested):
lets hold off this for a while

PR #10641, opened 2026-09-15
Rebuild the GCS storage client after a fork when Firestore is touched first

Dev-next-gen:
## Description

`GCSBackend` holds two clients: the Cloud Storage one on `_client` and the
Firestore one used for the chord counter on `_firestore_client`. Both
properties guard against a fork with the same check — compare `self._pid` to
`getpid()`, and rebuild if they differ — and both of them wrote back to that
one `_pid` attribute. So whichever client is rebuilt first clears the fork
flag for the other one:

```python
backend.firestore_client   # rebuilds, and sets _pid to the child's pid
backend.client             # _pid now matches, so the parent's client is returned
```

I ran into this reading `on_chord_part_return`, because that is the order it
uses: `incr()` reaches Firestore first through `_firestore_document`, then
`_restore_deps` reaches GCS through `client`. In a prefork worker the child
therefore keeps using the storage client the parent built, along with the
`requests` session and the sockets mounted on it in the `client` property —
which is the situation the pid check exists to avoid.

The fix is to give the Firestore client its own `_firestore_pid`, so the two
properties stop overwriting each other's marker.

There are already two tests for this, `test_new_client_after_fork` and
`test_new_firestore_client_after_fork`, but each one only ever asks for a
single client, so neither can observe the interaction. I added one that asks
for both after the pid changes; it fails on main with

```
assert backend.client is not storage_client
E  AssertionError: assert <MagicMock name='Client()' id='...'> is not <MagicMock name='Client()' id='...'>
```

and passes with the change. `t/unit/backends/` is green (633 passed, 19
skipped). I also moved the two `_pid` assertions in
`test_new_firestore_client_after_fork` onto `_firestore_pid`, since that is
the attribute that test is actually about.


auvipy:
@MehrazRumman can you try this plz?

MehrazRumman (approved):
Verified locally, LGTM. One suggestion on the test.

The diagnosis is right: `GCSBackendBase.client` and `GCSBackend.firestore_client` both used `self._pid` as their fork marker, so whichever property was rebuilt first stamped the child's pid and the other property then happily returned the parent's client. Giving Firestore its own `_firestore_pid` is the correct minimal fix.

I mocked `getpid` and both `Client` constructors and drove the two properties in both orders after a simulated fork:

| order after "fork" | `main` | this PR |
|---|---|---|
| `firestore_client`, then `client` | storage **not** rebuilt | both rebuilt |
| `client`, then `firestore_client` | firestore **not** rebuilt | both rebuilt |

So the bug is symmetric on `main`, not just the Firestore-first order described in the PR. The reverse order is arguably the more common one in a prefork worker: a child will usually hit `client` via `set()`/`get()` on an ordinary result before it ever reaches `on_chord_part_return`, at which point `firestore_client` hands back the parent's gRPC channel. The fix covers both, which is good.

The new test only exercises the Firestore-first order, though. Since the reverse order was equally broken and is the one a real worker is more likely to hit, I'd add the mirror case (or parametrize the order) so a future refactor can't reintroduce it on one side:

```python
mock_pid.return_value = 456
mock_client.return_value = Mock()
mock_firestore_client.return_value = Mock()

assert backend.client is not storage_client
assert backend.firestore_client is not firestore_client
```

Everything else checks out: the new test fails on `main` with the `is not` assertion from the description and passes here, the two moved `_firestore_pid` assertions are correct (the Firestore property no longer touches `_pid`), `t/unit/backends/test_gcs.py` is 38 passed, and flake8 is clean. No other backend in `celery/backends/` shares a pid marker between two clients, so this is the only instance of the pattern.

Dev-next-gen:
@MehrazRumman you're right that the reverse order is the one a worker hits first, and your table made that plain. I pushed c687af0 that parametrizes the test on the order, so it now asks for `firestore_client` then `client`, and `client` then `firestore_client`.

Against `celery/backends/gcs.py` from `main` (c1fd34e) both cases fail, `order0` on `client` and `order1` on `firestore_client`, each time with the same `assert <MagicMock name='Client()' ...> is not <MagicMock name='Client()' ...>`. With the fix, `t/unit/backends/test_gcs.py` gives 39 passed and flake8 is clean.

auvipy (approved):

PR #10643, opened 2026-09-16
Fix celery multi crashing on a single node name containing a dash

Dev-next-gen:
## Description

I was reading `celery/apps/multi.py` and noticed that `MultiParser.parse()` decides whether the
arguments describe a range *before* it knows what the single argument actually is:

```python
ranges = len(names) == 1
...
if ranges:
    try:
        names, prefix = self._get_ranges(names), range_prefix
    except ValueError:
        pass
self._update_ns_opts(p, names)
self._update_ns_ranges(p, ranges)
```

`_get_ranges()` calls `int(names[0])`, so for `celery multi start worker-1` it raises `ValueError`
and the names are left alone — but `ranges` is still `True`. `_update_ns_ranges()` then hands that
flag to `_parse_ns_range()`, which treats any dash in a namespace as a range separator and calls
`int()` on both halves.

So starting a single worker whose name contains a dash, with a namespaced option for it, crashes:

```
$ celery multi show worker-1 -c:worker-1 5
...
  File "celery/apps/multi.py", line 363, in _parse_ns_range
    str(n) for n in range(int(start), int(stop) + 1)
ValueError: invalid literal for int() with base 10: 'worker'
```

The same thing happens with the index form (`-c:1`), because `_update_ns_opts()` copies the options
onto the `worker-1` namespace first. Two or more node names on the command line are fine, since
`ranges` is then `False`.

The fix is to clear the flag when the range expansion didn't apply — if the lone argument isn't a
node count, it's a plain node name and its dashes aren't range separators.

After the change:

```
$ celery multi show worker-1 -c:worker-1 5 --pidfile=/tmp/%n.pid --logfile=/tmp/%n.log
python -m celery worker --detach --pidfile=/tmp/worker-1.pid --logfile=/tmp/worker-1.log \
    -c 5 -n worker-1@shwp-ai --executable=python
```

and the numeric form still expands as before:

```
$ celery multi show 3 -c:1-2 10 ...
... -c 10 -n celery1@shwp-ai ...
... -c 10 -n celery2@shwp-ai ...
... -n celery3@shwp-ai ...
```

I added a unit test in `t/unit/apps/test_multi.py` that reproduces the crash: it fails with the
`ValueError` above on main and passes with the fix. `t/unit/apps/` and `t/unit/bin/` are green
(62 passed).


auvipy:
@GangEunzzang @phanky1 @MehrazRumman would you mind verifying this, please?

GangEunzzang (commented):
@Dev-next-gen @auvipy Tried this out — the crash reproduces on `main` and this fixes it, and the other `celery multi` forms I tried behave the same as before.

Worth mentioning: the same `int()` still trips when the first argument is a node count, on `main` and here both:

```
$ celery multi show 3 -c:my-ns 10
ValueError: invalid literal for int() with base 10: 'my'
```

Different entry point, so probably out of scope here.

LGTM!

auvipy:
> @Dev-next-gen @auvipy Tried this out — the crash reproduces on `main` and this fixes it, and the other `celery multi` forms I tried behave the same as before.
> 
> Worth mentioning: the same `int()` still trips when the first argument is a node count, on `main` and here both:
> 
> ```
> $ celery multi show 3 -c:my-ns 10
> ValueError: invalid literal for int() with base 10: 'my'
> ```
> 
> Different entry point, so probably out of scope here.
> 
> LGTM!

@Dev-next-gen may be fixing the related issue would make sense in this pr

MehrazRumman (changes requested):
Verified locally, LGTM.

The analysis in the description is right: `ranges = len(names) == 1` is decided before `_get_ranges()` gets to reject the argument, so a lone non-numeric node name leaves `ranges=True` and every dash in a namespaced option is then fed to `int()`. Clearing the flag in the `except ValueError` branch is the minimal correct fix.

What I checked on `main` vs this branch (all with `--pidfile`/`--logfile` pointed at a temp dir):

| argv | main | this PR |
|---|---|---|
| `worker-1 -c:worker-1 5` | `ValueError: 'worker'` | `worker-1@ -c 5` |
| `worker-1 -c:1 5` | `ValueError: 'worker'` | `worker-1@ -c 5` |
| `w-a-b -c:w-a-b 5` | `ValueError: too many values to unpack` | `w-a-b@ -c 5` |
| `worker-1 -Q:worker-1,other q1` | `ValueError: 'worker'` | `worker-1@ -Q q1` |
| `-c:worker-1 5 worker-1` (option first) | `ValueError: 'worker'` | `worker-1@ -c 5` |
| `1-3 -c:1-3 10` | `1-3@` with the `-c` silently dropped | `1-3@ -c 10` |
| `3 -c:1-2 10` | celery1/2 get `-c 10`, celery3 doesn't | unchanged |
| `3 -c:1 10` | celery1 gets `-c 10` | unchanged |
| `worker-1 worker-2 -c:worker-1 5` | fine | unchanged |
| `worker-1` / `worker-1 -c 5` | fine | unchanged |

So the numeric-range forms are untouched, the multi-name form is untouched, and the single-dashed-name form goes from crashing to doing the obvious thing. The `1-3 -c:1-3 10` row is a small bonus: on `main` the option was expanded into namespaces `1,2,3` that matched nothing, so it vanished silently; now it lands on the node.

The new test fails on `main` with the `ValueError` from the description and passes here. `t/unit/apps` is 31 passed and flake8 is clean on both changed files.

Agree with @GangEunzzang that `3 -c:my-ns 10` still raises the same `ValueError` on both `main` and this branch. That's the genuinely-a-range path (`ranges=True` because the argument really is a count), so a dashed namespace there is ambiguous by design and belongs in a separate change if anyone wants to make `_parse_ns_range` fall back to a literal namespace when `int()` fails.

MehrazRumman (approved):

auvipy (approved):

PR #10645, opened 2026-09-16
Set default max priority in Queues.__setitem__() when the queue has none 

phanky1:
## Description

There is a discrepancy between `Queues.add()` and `Queues.__setitem__()` where `add()` applies the default `x-max-priority`, but `__setitem__()` doesn't. I think users would expect these two ways of adding a queue to handle priority the same way.


auvipy (changes requested):
should we also consider adding some integration test for this?

auvipy (approved):

PR #6120, opened 2020-05-27
issue 6108 fix filesystem backend cannot not be serialized by picked

singlaive:
https://github.com/celery/celery/issues/6108

*Note*: Before submitting this pull request, please review our [contributing
guidelines](https://docs.celeryproject.org/en/master/contributing.html).

## Description

This PR is to fix issue-6108, that when a instance of FilesystemBackend class is pickled, it cannot be deserialised because a mandatory parameter has not been provided.

The unit test still needs further working, since the backend complains it needs a live app to be pickled.

Issue #6108, opened 2020-05-18
[FilesystemBackend] Failure in unpickling a FilesystemBackend instance for a group


singlaive:
# Checklist

- [ x] I have verified that the issue exists against the `master` branch of Celery.
- [ ] This has already been asked to the [discussion group](https://groups.google.com/forum/#!forum/celery-users) first.
- [x ] I have read the relevant section in the
  [contribution guide](http://docs.celeryproject.org/en/latest/contributing.html#other-bugs)
  on reporting bugs.
- [x ] I have checked the [issues list](https://github.com/celery/celery/issues?q=is%3Aissue+label%3A%22Issue+Type%3A+Bug+Report%22+-label%3A%22Category%3A+Documentation%22)
  for similar or identical bug reports.
- [ ] I have checked the [pull requests list](https://github.com/celery/celery/pulls?q=is%3Apr+label%3A%22PR+Type%3A+Bugfix%22+-label%3A%22Category%3A+Documentation%22)
  for existing proposed fixes.
- [ ] I have checked the [commit log](https://github.com/celery/celery/commits/master)
  to find out if the bug was already fixed in the master branch.
- [ ] I have included all related issues and possible duplicate issues
  in this issue (If there are none, check this box anyway).

## Mandatory Debugging Information

- [ ] I have included the output of ``celery -A proj report`` in the issue.
    (if you are not able to do this, then at least specify the Celery
     version affected).
- [ ] I have verified that the issue exists against the `master` branch of Celery.
- [ ] I have included the contents of ``pip freeze`` in the issue.
- [ ] I have included all the versions of all the external dependencies required
  to reproduce this bug.

## Optional Debugging Information

- [ ] I have tried reproducing the issue on more than one Python version
  and/or implementation.
- [ ] I have tried reproducing the issue on more than one message broker and/or
  result backend.
- [ ] I have tried reproducing the issue on more than one version of the message
  broker and/or result backend.
- [ ] I have tried reproducing the issue on more than one operating system.
- [ ] I have tried reproducing the issue on more than one workers pool.
- [ ] I have tried reproducing the issue with autoscaling, retries,
  ETA/Countdown & rate limits disabled.
- [ ] I have tried reproducing the issue after downgrading
  and/or upgrading Celery and its dependencies.

#### Related Issues

- None

#### Possible Duplicates

- None

## Environment & Settings

**Celery version**:

<details>
<summary><b><code>celery report</code> Output:</b></summary>
<p>

```
[...]

Issue #2655, opened 2015-06-15
Celery Kickstarter Ideas


thedrow:
In order to get this project going we're going to raise funds using Kickstarter.

**If you'd like to propose yourself as a candidate for funded work please comment here.**

List of tasks we need to do before launching the Kickstarter campain:
- [ ] Define goals.
- [ ] Define stretch goals.
- [ ] Create a video with our proposal.
- [ ] Write the proposal text.
- [ ] Estimate our required budget.
- [ ] Create issues for each goal we agree on.

List of proposed goals as discussed with @malinoff on IRC:
- [ ] Kafka broker support.
- [ ] Using Etcd for discovery of Celery workers instead of using the broker.
- [ ] Using Consul for discovery of Celery workers instead of using the broker.
- [ ] librabbitmq rewrite that will allow it to support Python 3 and PyPy. https://github.com/celery/librabbitmq/issues/13
- [ ] Bug fixes for critical bugs in Celery that cause Celery to stop consuming tasks. See the **[Worker Hangs label](https://github.com/celery/celery/labels/Worker%20Hangs)** and **[Worker Crashes label](https://github.com/celery/celery/labels/Worker%20Crashes)**
- [ ] AWS Kinesis broker support.
- [ ] Better support for our experimental brokers.
- [ ] Increasing test coverage, general refactoring, documentation improvements.
- [ ] Automatic autoscaling across the cluster when the task backlog is high or low.
- [ ] Manually increasing/decreasing the pool size of a worker for a specified period of time.
- [ ] An official Docker container for Celery.
- [ ] A deployment tool for Celery and the different brokers.
- [ ] Deprecation of the old threadpool worker pool and introduction of a billiard based thread pool (Which eventually be valuable for PyPy-STM).
- [ ] Broker health checks.
- [ ] Automatic detection if a worker might be hanging. #2668

**If you'd like to propose a goal please comment here**

thedrow:
RabbitMQ has a healthcheck option if the management API is alive.
Performing a GET request on /api/aliveness-test/vhost will enqueue a message on a test queue and consume it in order to verify that everything works correctly.

minuscorp:
I think another interesting goal might be the programatically spawn of new workers on-demand. Many web applications are Platform as Service oriented where the user can choose the number of workers to make some tasks. Maybe you can tag some machines as deployers of workers (with maximum workers per machine) and let the user manage how many workers and for how much time  are running to process a bunch of tasks.

[...]

PR #10647, opened 2026-09-16
Keep the Azure Block Blob backend picklable

Dev-next-gen:

## Description

I noticed that the Azure Block Blob result backend can't survive a pickle round trip. `AsyncResult.__reduce_args__` includes `self.backend`, and `unpickle_backend` rebuilds the backend from whatever kwargs its `__reduce__` returns. `AzureBlockBlobBackend` never overrides `__reduce__`, so it gets rebuilt with `url=None`, and `_parse_url` fails on `None[len(prefix):]`. This is the same failure #6108 reported for the filesystem backend (fixed in #6120).

This is enough to trigger it, with azure-storage-blob 12.30.1 installed and no network needed:

```python
import pickle
from celery import Celery
from celery.result import AsyncResult

app = Celery(set_as_current=True)
app.conf.result_backend = (
    'azureblockblob://DefaultEndpointsProtocol=https;AccountName=name;'
    'AccountKey=key;EndpointSuffix=core.windows.net')
pickle.loads(pickle.dumps(AsyncResult('some-task-id', app=app)))
```

On main it raises `TypeError: 'NoneType' object is not subscriptable`. With this change it gives back an `AsyncResult` backed by an `AzureBlockBlobBackend`. In practice, this breaks a task that returns an `AsyncResult` when `result_serializer='pickle'`: the result is stored fine, but `.get()` fails with a `DecodeError` when it's read back.

The fix passes the connection string (with its `azureblockblob://` prefix) and the container name through `__reduce__`, the same way the filesystem, redis and mongodb backends pass their URL. The signature matches `Backend.__reduce__`. The pickle format of this backend does change, but before this change unpickling it always failed, so nothing that used to load can stop loading. As with redis and mongodb, the pickled backend carries its credentials, since it can't reconnect without them.

I added two tests to `t/unit/backends/test_azureblockblob.py`: a plain pickle round trip, and an `AsyncResult` encoded and decoded through the backend with the pickle serializer. Without the fix, both fail with the `TypeError` above (the second one wrapped in `DecodeError`). With the fix, the file passes (24 tests), `t/unit/backends/` passes (636 passed, 19 skipped), and flake8 reports nothing on the two files.


auvipy:
@MehrazRumman please review this humanisticly...

MehrazRumman (approved):
Thanks for the detailed writeup, this one is easy to follow and the diagnosis holds up.

I traced the same path independently. `AsyncResult.__reduce_args__` returns
`self.id, self.backend, None, None, self.parent` (celery/result.py:442), so pickling any
AsyncResult pickles its backend. `Backend.__reduce__` returns
`(unpickle_backend, (cls, args, kwargs))` with empty kwargs when nothing overrides it, and
`unpickle_backend` calls `cls(*args, app=current_app, **kwargs)`. With no `url` in kwargs,
`AzureBlockBlobBackend.__init__` runs `_parse_url(None)`, which does `url[len(prefix):]` and
raises `TypeError: 'NoneType' object is not subscriptable`. Same shape as the filesystem case
you link.

The fix matches what redis (celery/backends/redis.py:722) and mongodb
(celery/backends/mongodb.py:293) already do, so it reads as the obvious one.

One detail worth calling out in the fix's favor, since it isn't obvious: rebuilding the url as
`PREFIX + self._connection_string` is lossless even when the original url did not start with
`azureblockblob://`. `_parse_url` strips a fixed 17 characters rather than matching the prefix,
so re-prefixing and re-parsing always lands back on the same `_connection_string`.

Two things I'd like your take on, neither blocking:

1. `expires`. redis and mongodb both pass `expires=self.expires` through `__reduce__` alongside
   the url. This PR passes only the url and the container name, so an explicitly constructed
   `AzureBlockBlobBackend(expires=...)` loses that value on a round trip and falls back to
   `app.conf.result_expires` via `prepare_expires`. Same-process round trips are unaffected,
   since the conf is identical. It's a one line addition if you want full parity.

2. `base_path`. It's read from `azureblockblob_base_path` in `__init__` and prefixes every blob
   key in `get`, `set`, `mget` and `delete`, but it isn't an `__init__` parameter, so it can't
   ride along in `__reduce__` and is re-derived from `current_app` on unpickle. Fine in
   practice, but it means the reconstructed backend is faithful to the conf rather than to the
   original instance. Probably right to leave as is, flagging it for the record.

Optional nit: `as_uri(include_password=True)` already returns exactly
`AZURE_BLOCK_BLOB_CONNECTION_PREFIX + self._connection_string`, so `__reduce__` could call it
instead of rebuilding the same string. I'm genuinely torn. It removes the duplication, but it
couples reconstruction to a display method, and a future change to redaction in `as_uri` would
then break unpickling silently. I'd keep the duplication, but the two expressions need to stay
in sync.

On the credentials point in your description: agreed, and I don't see an alternative. The
backend can't reconnect without the connection string, and redis and mongodb embed theirs the
same way. The redaction in `as_uri` is about display, so this doesn't undercut it.

Tests look right. Setting `accept_content` before constructing the backend matters, since
`self.accept` is resolved in `Backend.__init__`, and you have that ordering.
`depends_on_current_app` is needed because `unpickle_backend` reaches for `current_app`, and it
matches `test_pickleable` in the filesystem tests.

CI is green across all 375 checks, and codecov reports both new lines covered.

LGTM.

Dev-next-gen:
@MehrazRumman your point that re-prefixing is lossless because `_parse_url` strips a fixed 17 characters is the part I was least sure a reader would accept without checking, so I'm glad you traced it through.

I took the `expires` suggestion in 3182346. `test_pickleable` now builds the backend with `expires=1234` and checks it on the unpickled copy. Without the one-line change it fails with `86400.0`, the `result_expires` default, and with it the file passes (24 tests) and `t/unit/backends/` gives 636 passed, 19 skipped. I left `base_path` and the `as_uri` duplication as they are, for the reasons you gave.

The red Smoke-tests on the previous head don't come from this change. On main at 141c636, `test_worker.py` fails on the same redis `test_soft_shutdown_reset_visibility_timeout*` cases (https://github.com/celery/celery/actions/runs/35212199251/job/105173741248), and the `test_tasks.py` jobs were cancelled there too.

MehrazRumman:
@Dev-next-gen  You are using AI for almost everything even for a comment ! I am not fully agree with this approach !

MehrazRumman (commented):
Almost all the things are AI generated !

auvipy:
closing as AI generated shits

Dev-next-gen:
@auvipy @MehrazRumman Hello. I am Leo, the person behind this. You are right that the tone reads like a machine, and I would rather explain than argue.

I run a study on an autonomous defect-hunting pipeline. It looks for bugs that technical debt leaves behind in open source, proves each one with a reproduction that fails before the patch and passes after, and opens the fix upstream under each project's own rules. The discipline I impose on it, always the same structure and the evidence before the claim, is exactly what gives it that robotic voice. That criticism is fair. What it does, what it measures and where it falls short is written down and kept current here: https://github.com/Dev-next-gen/defect-hunting-pipeline

My reason for building it is simple. Reviewing, correcting and integrating are already moving towards automation on the maintainer side too. If a review can be automated, then a program that finds real defects and submits them to that same review seems to me a logical thing to appear. I would rather build it in the open, with its limits documented, than pretend a human typed every line.

That said, whether it is welcome here is your call, not mine. I have removed celery from my program's scan list, so you will not receive further pull requests from me. I am also adding a way for any maintainer to opt out directly, so that nobody has to ask twice.

My apologies for the friction.

Leo CAMUS

(@Dev-next-gen) This comment was written by me, not by the pipeline. I used an LLM to correct it and translate it into English, because my own English is not good enough to write it properly myself.

auvipy:
@MehrazRumman @Dev-next-gen may be this pr is related ? https://github.com/celery/kombu/issues/2654

Dev-next-gen:
@auvipy You are right to connect them. It is the same failure, one layer down.

This PR is the result backend: `AzureBlockBlobBackend` comes back with `url=None` after a pickle round trip, because it never says how to rebuild itself. kombu#2654 is the Service Bus transport, where lock renewal leaves the received message unpicklable and the prefork pool fills up and stalls. Different object, same cause — an Azure piece that does not survive being pickled to a child process. jerbob's #2655 already covers the transport side and adds a prefork-pickling test, which is the right guard; this PR is the result-backend half of the same problem.

I took celery off my scan list and I meant it, so this is just me answering your question, not pushing anything. If it ever helps, a small pickle round-trip test on each Azure integration is a cheap way to catch the next one, but that is your call.

(@Dev-next-gen) This comment was written by me, not by the pipeline. I used an LLM to correct it and translate it into English, because my own English is not good enough to write it properly myself. 😉

auvipy:
Ok forgiving you for now 😁

MehrazRumman:
> _Note_: Before submitting this pull request, please review our [contributing guidelines](https://docs.celeryq.dev/en/main/contributing.html).

Please look at the PR Description !

auvipy:
@GangEunzzang would you mind cross checking this? as you reviewed relevant one in kombu lately

GangEunzzang (approved):
Cross-checked locally: the repro from the description fails on main with the `TypeError` and passes here, the two new tests fail without the fix and the file passes with it (24), and the `__reduce__` shape matches the redis/mongodb backends with `container_name` added since it's a constructor argument. LGTM.

auvipy (approved):

PR #10657, opened 2026-09-16
Publish the Cassandra session only once the connection is ready

Dev-next-gen:

## Description

I was reading `CassandraBackend._get_connection()` and noticed that callers leave early as soon as `self._session is not None`, before taking the lock. Inside the lock, though, `self._session` is set right after `Cluster.connect()`, and only then are the write/read statements prepared and, for writers, `CREATE TABLE` issued. So another thread (a `threads` pool worker where several tasks finish right after start, for instance) can see the session and run `_store_result` while the first thread is still waiting on `CREATE TABLE`, which on a fresh keyspace means an insert into a table that doesn't exist yet. It can also land in the short window where `_write_stmt` is still `None`.

The fix builds the session and the statements in local variables and assigns `_write_stmt`, `_read_stmt` and `_session` at the end, once the table exists. Nothing else changes: the statements, the order in which they run and the lock are the same, and only private attributes are involved. As a side effect, an `OperationTimedOut` raised by `CREATE TABLE` now leaves `_session` unset, which is what the `except` branch was already trying to do.

The new test starts a second thread from inside the fake `CREATE TABLE` call. Before the change it fails with `assert ['INSERT', 'CREATE'] == ['CREATE', 'INSERT']`: the second thread wrote first. After the change the second thread waits on the lock, and `t/unit/backends/test_cassandra.py` passes (17 tests), as does the rest of `t/unit/backends` (635 passed, 19 skipped).

A review pointed out that the barrier was still bypassed when the first caller was a read: `_get_connection(write=False)` published the session without issuing `CREATE TABLE`, and the early return only looked at `self._session`, so the first `_store_result` behind a read-opened session returned before the table existed. The second commit tracks that with a `_table_created` flag, checked in the early return and assigned after `_session`, so a writer takes the lock while the flag is false. `test_table_is_created_for_a_write_through_a_read_session` reads then writes and fails without it with `assert ['SELECT', 'INSERT'] == ['SELECT', 'CREATE', 'INSERT']`.


sylvesterkaczmarek (commented):
The publication barrier is still bypassed if the first caller is a read. `_get_connection(write=False)` publishes `self._session` without creating the table; a later `_store_result()` calls `_get_connection(write=True)`, sees the existing session, and returns before `CREATE TABLE`. On a fresh keyspace that can still reach the INSERT with no table. Could the write path track/ensure schema initialization separately even when a session already exists, with a read-first-then-write regression test?

Dev-next-gen:
You are right, and it is not something my first commit introduced: the early return in `_get_connection` only looked at `self._session`, so a backend whose session had been opened by `_get_task_meta_for` never issued `CREATE TABLE`, and the first `_store_result` returned before reaching it. Pushed 799209c52, which tracks that separately with a `_table_created` flag: a writer goes through the lock while the flag is false even when the session is already published, and the flag is assigned after `_session`, so a writer arriving concurrently blocks on the lock instead of seeing a half-initialised connection. `CREATE TABLE` still runs at most once per backend instance.

The regression test you asked for is `test_table_is_created_for_a_write_through_a_read_session`: it opens the session with a read, then writes, and records the first word of every statement the fake session executes. On the previous head it fails with `assert ['SELECT', 'INSERT'] == ['SELECT', 'CREATE', 'INSERT']`, which is the insert-with-no-table you described; with the commit it gives `['SELECT', 'CREATE', 'INSERT']`, and a second write appends only another `'INSERT'`. `t/unit/backends/test_cassandra.py` is 18 passed, and the whole of `t/unit/backends` 735 passed and 13 skipped, on 3.12 with the backend extras installed.

One knock-on worth flagging since it touches the test file rather than the fix: the tests that hand a `FakeCassandraTable` to a `_session` they stubbed by hand now reach the `CREATE TABLE` branch on their first write, and without a change to that helper three ids fail (`test_store_and_get_result_compressed[json]`, `[pickle]` and `test_get_result_written_before_compression`). The helper now returns early when `parameters is None`, which is what the real session is given for a DDL statement.

Dev-next-gen:
@auvipy @MehrazRumman @sylvesterkaczmarek Following up on what I wrote in #10647 and #10661.

One point of honesty about the timing, because it looks bad and I would rather explain it than let it stand. The commit and the comment above landed seven minutes after I wrote that I would not push anything further here. That was not a change of mind. I had put celery on my exclusion list, but I had not yet loaded that list into the running program, so the pipeline answered the review exactly as it would have for any other project. The exclusion is in force now, and I checked it on every entry point: no scan, no pull request, no reply.

@sylvesterkaczmarek, your review deserved an answer and it has one above, so I will leave it there. You found a real gap that my first commit did not close: a read publishes the session, and a later write returned before CREATE TABLE, so an insert could reach a missing table on a fresh keyspace. The fix and the read-then-write regression test are in 799209c52 if they are of any use to you. Whatever happens to this pull request, the finding was yours.

This is my last message here. Close both pull requests if you prefer, I will not push to them again.

Leo CAMUS

(@Dev-next-gen) Written by me, not by the pipeline. Corrected and translated into English with an LLM, because my own English is not good enough to write it properly myself.

sylvesterkaczmarek (approved):
Rechecked `799209c52`. This closes the read-first/write gap I flagged: session readiness and table readiness are now tracked separately, so a writer cannot return early just because a reader published the session. The write path re-enters the same lock until `_table_created` is true, and the new regression exercises the exact `SELECT -> CREATE -> INSERT` sequence plus a second write without another CREATE. My blocker is resolved.

Issue #457, opened 2011-08-27
[gevent] Celery improper exit behavior on KeyboardInterrupt exception in task body.


linnik:
Here is a sample:
http://pastebin.com/QnwT00G3
If i run task with sample.apply_async() and then hit Ctrl-C at console with worker - it's exit without exec any postrun signal and without sending result's state, return value, etc... 
Gevent.sleep(value) doesn't make any sense, i think. That's all around KeyboardInterrupt. Usually, if you raise some exception during task execution, Celery will return 'Failed' state and will run signals, etc. But if we raise KeyboardInterrupt - we see this strange reaction. So it's all about wrong KeyboardInterrupt handling, i guess.
Here is a stacktrace: http://pastebin.com/YyHz2Mbq

Issue #452, opened 2011-08-16
receiving KeyError when requesting inspect().active()


heckj:
Invoking celery.task.control.inspect().active() and getting a KeyError response:

Celery is being loaded & registered through Flask-Celery (master/trunk, not released) with Celery 2.3.1

(also: http://paste.openstack.org/show/2183/)

> > > from celery.task.control import inspect
> > > inspect().active()
> > > Traceback (most recent call last):
> > >   File "<console>", line 1, in <module>
> > >   File "/home/heckj/src/maestro/.maestro-venv/lib/python2.7/site-packages/celery/task/control.py", line 41, in active
> > >     return self._request("dump_active", safe=safe)
> > >   File "/home/heckj/src/maestro/.maestro-venv/lib/python2.7/site-packages/celery/task/control.py", line 38, in _request
> > >     timeout=self.timeout, reply=True))
> > >   File "/home/heckj/src/maestro/.maestro-venv/lib/python2.7/site-packages/celery/task/control.py", line 218, in broadcast
> > >     channel=channel)
> > >   File "/usr/lib/python2.7/contextlib.py", line 24, in __exit__
> > >     self.gen.next()
> > >   File "/home/heckj/src/maestro/.maestro-venv/lib/python2.7/site-packages/celery/app/base.py", line 287, in default_connection
> > >     yield connection
> > >   File "/home/heckj/src/maestro/.maestro-venv/lib/python2.7/site-packages/kombu/connection.py", line 472, in **exit**
> > >     self.release()
> > >   File "/home/heckj/src/maestro/.maestro-venv/lib/python2.7/site-packages/kombu/connection.py", line 562, in _release
> > >     self.release(resource)
> > >   File "/home/heckj/src/maestro/.maestro-venv/lib/python2.7/site-packages/kombu/connection.py", line 589, in release
> > >     self.close_resource(resource)
> > >   File "/home/heckj/src/maestro/.maestro-venv/lib/python2.7/site-packages/kombu/connection.py", line 571, in close_resource
> > >     resource.close()
> > >   File "/home/heckj/src/maestro/.maestro-venv/lib/python2.7/site-packages/kombu/connection.py", line 158, in release
> > >     self._close()
> > >   File "/home/heckj/src/maestro/.maestro-venv/lib/python2.7/site-packages/kombu/connection.py", line 145, in _close
> > >     self.transport.close_connection(self._connection)
[...]

PR #10659, opened 2026-09-17
Update billiard version requirement to 4.3.0

auvipy:
test the new release


auvipy:
@GangEunzzang tagging you as you contributed some recent billiard and relevant contributions in celery

auvipy:
primary findings

The failure is a real Redis connection leak, not an infrastructure failure.

### Cause

The failing test retains every `TestApp` instance in `apps`:

- [`test_backend_leak_in_fixture_teardown.py:13-25`](https://github.com/celery/celery/blob/aed060208bec56e94fad7dc998e0d539175fbef9/t/integration/test_backend_leak_in_fixture_teardown.py#L13-L25)
- [`test_backend_leak_in_fixture_teardown.py:28-40`](https://github.com/celery/celery/blob/aed060208bec56e94fad7dc998e0d539175fbef9/t/integration/test_backend_leak_in_fixture_teardown.py#L28-L40)

Each task result causes the Redis backend to create:

1. A result-consumer Pub/Sub connection.
2. A connection-pool connection.

`ResultConsumer.stop()` closes only the Pub/Sub object ([`celery/backends/redis.py:177-180`](https://github.com/celery/celery/blob/aed060208bec56e94fad7dc998e0d539175fbef9/celery/backends/redis.py#L177-L180)). The Redis connection pool itself is not disconnected when the worker fixture exits. Because the test keeps the apps alive, garbage collection cannot clean up those pools. The result is a repeatable increase of 27 file descriptors after 20 cycles:

```text
assert 27 < 20
```

The same failure occurred on all three retry attempts, so increasing retries or changing the assertion would only hide the leak.

### Recommended fix

Explicitly close the backend consumer and disconnect its Redis pool when each temporary app cycle ends:

```python
def close_backend(app):
    backend = app.backend

    # Close the result-consumer Pub/Sub connection.
    backend.result_consumer.stop()

    # Do not access backend.client here: cached_property access would create
    # a client even when the test did not use one.
    client = backend.__dict__.get("client")
    if client is not None:
        client.connection_pool.disconnect()
```

Use it in `run_one_fixture_cycle()`:

```python
def run_one_fixture_cycle():
    app = TestApp(config={'result_backend': TEST_BACKEND})

    @app.task
    def store_result():
        return 'stored'

    try:
        with setup_default_app(app):
            with start_worker(app):
                result = store_result.delay()
                assert result.get(timeout=10) == 'stored'
        return app
    finally:
        close_backend(app)
```

For the Redis-specific integration test, an inline version is also sufficient:

```python
    finally:
        app.backend.result_consumer.stop()
        client = app.backend.__dict__.get("client")
        if client is not None:
            client.connection_pool.disconnect()
```

This preserves the purpose of the test: applications remain referenced in `apps`, while all backend-owned sockets are explicitly released during fixture teardown. The unrelated `systemd-manager.service` and Redis memory-overcommit messages in the job log are warnings from the runner and are not the cause of the failure.

auvipy:
@phanky1 may be you can also dig in?

auvipy:
interestingly, tintegration tests seems to be passing after restarting the job but smoke tests are failing

auvipy:
The failure is in the Redis visibility-timeout shutdown tests, not in Redis startup or task execution. All failing cases eventually log:

```text
worker: Cold shutdown (MainProcess)
```

but the Docker worker container does not reach `exited` within the 60-second wait. The tests run a prefork worker with:

- `worker_concurrency = 10`
- `prefetch_multiplier = 2`
- several 15-second tasks
- multiple Docker-backed tests in parallel via `pytest -n auto`

That creates excessive worker-process and container contention on the 4-vCPU runner. The repeated retries and failures across the same test variants indicate shutdown/resource contention rather than a functional visibility-timeout failure.

## Recommended fix

Reduce the worker pool size for this test class. Two processes are sufficient: the group tests contain at most two tasks, and the tests do not require ten concurrent workers.

Update `t/smoke/tests/test_worker.py`:

```diff
 class test_reset_visibility_timeout(SuiteOperations):
     @pytest.fixture
     def default_worker_app(self, default_worker_app: Celery) -> Celery:
         app = default_worker_app
         app.conf.prefetch_multiplier = 2
-        app.conf.worker_concurrency = 10
+        # Two workers are sufficient for the two-task group tests and avoid
+        # spawning ten prefork children for every Docker-backed smoke test.
+        app.conf.worker_concurrency = 2
         app.conf.visibility_timeout = 3600  # 1 hour
         app.conf.broker_transport_options = {
             "visibility_timeout": app.conf.visibility_timeout,
             "polling_interval": 1,
         }
```

The relevant fixture is at [`t/smoke/tests/test_worker.py`, lines 389–404](https://github.com/celery/celery/blob/aed060208bec56e94fad7dc998e0d539175fbef9/t/smoke/tests/test_worker.py#L389-L404).

## Additional hardening

The test helper currently waits only 30 seconds for a container to exit:

```python
RESULT_TIMEOUT = 30
```

and the worker-container fixture itself reports a 60-second Docker timeout. Under a busy CI runner, allow a little more time for process reaping:

```diff
-RESULT_TIMEOUT = 30
+RESULT_TIMEOUT = 60
```

This should be treated as secondary hardening, not the primary fix. Increasing the timeout alone would only hide the process-contention problem.

The workflow runs the smoke tests with automatic xdist parallelism:

```text
pytest -xsvv t/smoke/tests/test_worker.py --reruns 5 --reruns-delay 10 -n auto
```

Reducing `worker_concurrency` from 10 to 2 lowers the number of prefork children created across parallel Docker workers while preserving the test semantics:

- single-task tests still execute normally;
- the two-task group tests can still run both tasks concurrently;
- the long-running task remains active when shutdown occurs;
- Redis can still restore unacknowledged messages after the worker exits.

After this change, the existing visibility-timeout assertions should remain unchanged.

auvipy:
initial co pilot based analysis, although not sure how real they are

GangEunzzang (commented):
@auvipy Ran Celery `main` against the new release on macOS arm64 / Python 3.12.

4.3.0 makes `spawn` the default on Darwin, where 4.2.1 used `fork`, and the prefork pool handles it: tasks run, a hard time limit still fires on time, and the pool keeps working afterwards. `t/unit` gives the same numbers on both versions, and the Python floors line up — Celery is already `>=3.10`, and so is 4.3.0.

I did find two behavior changes that come along with the upgrade, both from the exception-propagation work in this release. I bisected each one to the commit that introduced it, with 3 runs per point.

`celery/billiard@001b9f08` (`except MemoryError: raise` in the pool callback): a task raising `MemoryError` used to be reported as an ordinary failure and the worker kept running. Now it reaches `raise MemoryError(f'Process got: {exc}')` in `worker/request.py`, which escapes the consumer loop — the worker logs `Requeuing undeliverable message for queue 'celery': No consumers` and stops consuming.

`celery/billiard@bd2f803b` (`except Exception` → `except BaseException` in the worker loop): a task raising a `BaseException` — either a bare one or `asyncio.CancelledError` — used to come back as `WorkerLostError` within 0.1s. Now the exception itself is propagated, Celery can't store it (`EncodeError: Object of type BaseException is not JSON serializable`), and the caller waits out its own timeout with no result.

Neither reproduces on the parent commit; both start reproducing on the commits above, and both still reproduce on 4.3.0 regardless of start method. Neither is caused by this one-line change, but raising the floor makes both behaviors unavoidable.

On the smoke failures — I couldn't reproduce the shutdown hang locally on either version, at concurrency 10 or 2, warm or cold; the worker exits in under 4s each time. That's without Docker or `-n auto` though, so it doesn't say much about the runner.

The bump itself looks fine to me. It's those two behavior changes I'd want a second opinion on before moving the floor. I'm happy to dig into either on the Celery or Billiard side, depending on where you think the fixes belong.

auvipy:
on main branch, without this billiard update, the smoke tests fails https://github.com/celery/celery/actions/runs/35212199251/job/105173741390

phanky1:
> interestingly, tintegration tests seems to be passing after restarting the job but smoke tests are failing

For the intermittent integration test failure, I think the assertion I chose in `test_fixture_teardown_releases_backend_connections` is the problem. Removing the explicit `gc.collect()` made it depend on automatic GC timing.

I'll open a PR to address this.

auvipy:
> > interestingly, tintegration tests seems to be passing after restarting the job but smoke tests are failing
> 
> For the intermittent integration test failure, I think the assertion I chose in `test_fixture_teardown_releases_backend_connections` is the problem. Removing the explicit `gc.collect()` made it depend on automatic GC timing.
> 
> I'll open a PR to address this.

yeah I also think so. please come with a potential fix.

phanky1:
> yeah I also think so. please come with a potential fix.

https://github.com/celery/celery/pull/10662
Here it is. The test should now be stable and more realistic.

requirements/default.txt
```diff
@@ -1,4 +1,4 @@
-billiard>=4.2.1,<5.0
+billiard>=4.3.0,<5.0
```
auvipy:
```suggestion
billiard==4.3.0rc1
```

auvipy (commented):
lets see if the previous rc release improve the situation or not

auvipy:
> The bump itself looks fine to me. It's those two behavior changes I'd want a second opinion on before moving the floor. I'm happy to dig into either on the Celery or Billiard side, depending on where you think the fixes belong.

@GangEunzzang please investigate billiard then celery

auvipy:
rolling back to 4.3.0rc1 make the integrations tests fail again

requirements/default.txt
```diff
@@ -1,4 +1,4 @@
-billiard>=4.2.1,<5.0
+billiard==4.3.0rc1
```
auvipy:
```suggestion
billiard==4.3.0
```

GangEunzzang:
@auvipy Looked into both, and I have to correct myself on the first one.

**MemoryError — my finding was wrong.** celery/celery#10482 says it plainly: Celery intentionally raises `MemoryError` from that path and Billiard used to swallow it, so celery/billiard#451 is the complementary fix rather than a regression. `test_on_failure_propagates_MemoryError` and `test_on_success__propagates_MemoryError` pin that behavior, and routing it through `WorkerTerminate` instead breaks exactly those two. I also described the symptom wrong: the worker doesn't linger, it exits with code 1 about two seconds later. Sorry for the noise.

**BaseException — this one holds, and it looks like a one-liner on the Celery side.** `Backend.encode_result` in `celery/backends/base.py`:

```python
if state in self.EXCEPTION_STATES and isinstance(result, Exception):
    return self.prepare_exception(result)
```

A `BaseException` that isn't an `Exception` fails that check, so it skips `prepare_exception()` and the raw object reaches the encoder — which is where `EncodeError: Object of type BaseException is not JSON serializable` comes from, and why the result is never stored. `prepare_exception()` itself only reads `__qualname__`, `args` and `__module__`, so it handles any exception type fine.

Changing the check to `isinstance(result, BaseException)` is enough:

| task raises | before | after |
| --- | --- | --- |
| bare `BaseException` | caller times out after 20s | `BaseException` raised in 0.2s |
| `asyncio.CancelledError` | caller times out after 20s | `CancelledError` raised in 0.2s |

3 runs each, and the worker keeps consuming in both. Full `t/unit` is 4081 passed with and without the change.

**Smoke failures — Billiard 4.3.0 is the cause.** I kept digging into them locally. The main run that seemed to clear it had installed 4.3.0 too: it ran two hours after the release, and main's >=4.2.1 floor resolved to the newest version. Your ==4.3.0rc1 commit turned out to be the clean control. Same tree, one-line difference: the rc1 pin passes test_worker.py, while the 4.3.0 pin fails all five test_reset_visibility_timeout cases, both locally and on CI.

What actually happens isn't a hang. The worker exits normally, but the task that was running gets acknowledged and dropped instead of being requeued. That's also why it looks fine if you only check whether the worker exits: it does, and the message loss is silent.

The path is:

1. Cold shutdown cancels the active request, which sends SIGTERM to the child running it. Billiard's handler in the child turns that into SystemExit.
2. Before 4.3.0, SystemExit escaped the worker loop, the child died, and Celery saw a Terminated job. Request.on_failure returns early for those, so the message stays unacked and Kombu restores it on close: Restoring 1 unacknowledged message(s).
3. Since celery/billiard#452, the worker loop catches BaseException, so the same SystemExit comes back as an ordinary task failure. on_failure takes the acks_late path, acks_on_failure is on by default, and the message gets acked.

Bisecting Billiard between rc1 and 4.3.0 puts the change at bd2f803b (#452): 001b9f08 is still fine, bd2f803b is where the message starts disappearing, and it stays that way through the release. I reproduced the same thing on macOS with a plain prefork worker against Redis.

I tried both sides of the fix:

* Billiard: re-raise SystemExit and KeyboardInterrupt in the worker loop before the BaseException catch, so the child dies as it did before. This restores the old behavior, but conflicts with test_system_exit_from_task_replaces_worker from #457, which expects SystemExit to be marshaled back to the caller. So changing it there would mean deciding which Billiard semantics we want.
* Celery: in Request.on_failure, treat a SystemExit/KeyboardInterrupt result like Terminated when the request was cancelled or revoked by the worker itself, and return without acking. This is a small change, restores the message, leaves t/unit/worker unaffected, and makes the smoke tests pass with 4.3.0 pinned.

I lean toward the Celery-side fix because it preserves Billiard's new behavior and only changes how Celery handles the result of a task it killed itself. I can open that unless you'd rather handle the semantics in Billiard.

auvipy:
Dont we need fix in billiard side? Or it is on celery? Or we need change in both?

auvipy:
Lets fix on the celery side first then

GangEunzzang:
@auvipy Opened the Celery-side fix as #10667. With that in, `test_worker.py` passes with 4.3.0 pinned.

auvipy:
merging the billiard release to fix compatibility bugs

PR #10660, opened 2026-09-17
Fix package version specifications in test.txt and upgrade moto

auvipy:
Updated package versions in requirements/test.txt for consistency. mainly moto upgrade


auvipy (commented):
merging as CI was green

PR #10661, opened 2026-09-17
Apply index lists like -c:1,2 to named nodes in celery multi

Dev-next-gen:

## Description

I was reading `MultiParser.parse` after #10643 and noticed that an index list is silently dropped when the nodes have names. `celery multi start foo bar baz -c 3 -c:1 10` gives `foo` 10 processes, as the comment in `_update_ns_opts` describes, but `celery multi start foo bar baz -c 3 -c:1,2 10` leaves all three nodes at 3, with no error. The docstring of `celery multi` says `-c:1-3` can also be written as `-c:1,2,3`, and it shows lists working with node names (`-c:foo,bar,baz`), so a list of indexes should reach the named nodes the same way a single index does.

The cause is the order of the two passes: `_update_ns_opts` maps numeric namespaces to node names first, and only then does `_update_ns_ranges` split `1,2` into `1` and `2`. By that point the mapping has already run, so the new namespaces never reach `foo` or `bar`. When the arguments are node names, I now split the lists first and map the indexes afterwards. The order stays the same when the nodes are given as a count (`celery multi start 3 ...`), because there the node names are the indexes themselves, so the out-of-range handling for that form is unchanged. With named nodes, an out-of-range index inside a list (`-c:1,5` with two nodes) now raises the same `KeyError` that `-c:5` already raised, instead of being ignored.

The new test in `t/unit/apps/test_multi.py` fails on main with `assert ['3', '3', '3'] == ['10', '10', '3']` and passes with the change. The rest of `t/unit/apps/test_multi.py` and `t/unit/bin` still pass, and flake8 is clean on both files.


Dev-next-gen:
The red Smoke-tests jobs here aren't caused by this change. The failures are in `test_worker.py::test_worker_shutdown::...::test_soft_shutdown_reset_visibility_timeout*`: "Worker container ... did not log -> Restoring 1 unacknowledged message(s) within 60 seconds" ([job](https://github.com/celery/celery/actions/runs/35242537021/job/105280506021)). The same jobs fail on main at 141c636, the base of this PR, with that same assertion in the same tests ([job](https://github.com/celery/celery/actions/runs/35212199251/job/105173741248)), and `test_tasks.py` ends as cancelled on both runs. None of the smoke tests import `celery.apps.multi`, which is the only non-test file this PR touches.

Dev-next-gen:
@auvipy @MehrazRumman Following up on what I wrote in #10647, so that nothing is left hanging on your side.

This pull request and #10657 were both opened before I removed celery from my scan list. They are the last two you will see from me. Either outcome suits me: close them without further discussion if you prefer, or leave them open if the content is useful to someone. I will not push anything new to them, and I will not bring the subject back.

One thing I would rather not see lost in the disagreement. On #10657, @sylvesterkaczmarek pointed out that the fix is incomplete: if the first caller is a read, the session is published without the table being created, and a later write returns early because a session already exists, so an insert can still reach a missing table on a fresh keyspace. That observation stands on its own, whatever you decide about the pull request itself, and it is worth keeping.

Thank you for the reviews you gave earlier.

Leo CAMUS

(@Dev-next-gen) Written by me, not by the pipeline. Corrected and translated into English with an LLM, because my own English is not good enough to write it properly myself.

sylvesterkaczmarek (approved):
Reviewed current 23310126. For named nodes, comma-separated numeric namespaces are now expanded before the existing index-to-name mapping, so -c:1,2 behaves consistently with -c:1 without changing range-mode handling. The focused regression covers foo/bar/baz assignment and the full CI matrix is green. No blocker from me.

MehrazRumman (commented):
## Review

The diagnosis is correct and the fix is the minimal one. I traced it:

`celery multi start foo bar baz -c 3 -c:1,2 10` gives `namespaces = {'1,2': {'-c': '10'}}` and `ranges = False`. On `main`, `_update_ns_opts` runs first, skips `'1,2'` because `'1,2'.isdigit()` is `False`, and then `_update_ns_ranges` creates `'1'` and `'2'` — after the only pass that could have mapped them to `foo` and `bar`. Nothing consumes them, so the option evaporates. Swapping the two passes for the named-node case makes `'1'` and `'2'` exist before the mapping runs, and `foo`/`bar` pick them up. The new test's `['10', '10', '3']` is right, and it does fail on `main`.

Keeping the original order for the count form is also the right call, and more important than the PR description makes it sound — see point 1.

So: no objection to the approach. What I'd want before merge is mostly tests, plus one comment.

---

### 1. The asymmetry needs a comment explaining itself

Right now only the `else` branch is commented. The `if ranges:` branch looks like dead weight — the mapping pass is an *identity* operation in range mode (`names` is `['1', '2', '3']`, so `names[int(ns)-1] == ns`), which makes "why not just use one order?" the obvious next thought. It isn't safe:

```console
$ celery multi start 3 -c:1-5 10
```

With the current order, `_update_ns_opts` sees `'1-5'`, skips it (not a digit), and the split then creates namespaces `4` and `5` that no node consumes — silently ignored. Unify the order and `_update_ns_opts` now sees `'4'`, does `names[3]`, and raises `KeyError: 'No node at index 4'`. That's a behaviour change for a documented syntax (`-Q:1-3` in the `celery multi` docstring), triggered by a refactor that looks like pure cleanup.

Please put that in the code, not just in the PR description — something like:

```python
if ranges:
    # In range mode the node names *are* the indexes, so the mapping
    # pass is an identity. Keep it first regardless: reordering would
    # make out-of-range members of ``-c:1-5`` raise instead of being
    # ignored, which is a behaviour change for the count form.
    self._update_ns_opts(p, names)
    self._update_ns_ranges(p, ranges)
else:
    ...
```

### 2. The new `KeyError`s aren't pinned by any test

The change introduces at least two new user-visible errors for input that used to pass silently, and nothing tests either:

* `celery multi start foo bar -c:1,5 10` → now `KeyError: "No node at index '5'"`. The description mentions this; no test covers it.
* `celery multi start foo bar -c:0,1 10` → now `KeyError: "Indexes start at 1 got: '0'"`. This one isn't in the description at all. On `main`, `'0,1'.isdigit()` is `False` so the whole thing is skipped; after the reorder, `'0'` reaches the `ns_index < 0` guard. It's the consistent outcome (`-c:0` already raises), but it's a new error and it should be stated and tested.

### 3. Nothing guards the branch the fix depends on

The entire justification for the `if/else` is "the count form is unchanged". There's no test asserting that. Two cheap ones:

* `start 3 -c:1-5 10` still assigns `['10', '10', '10']` without raising;
* `start 3 -c:5 10` still raises `KeyError`.

Without those, the next person who simplifies `parse()` gets a green suite.

### 4. Remaining inconsistencies, worth at least documenting

The fix closes one hole and leaves adjacent ones open, which is fine for a focused bugfix but shouldn't go unstated:

* **`-c:1-2` with named nodes is still silently dropped.** `_update_ns_ranges` only expands `-` when `ranges` is true, so with `foo bar baz` the namespace `'1-2'` survives, isn't a digit, and never reaches a node. Same failure mode as the one being fixed. I don't think you can fix it by simply allowing `-` here — #10643 exists because `worker-1` is a legal node name — but an all-digits check on the components would distinguish `1-2` from `worker-1` if someone wants to follow up.
* **Out-of-range now diverges between the two forms.** With names, `-c:1,5` raises; with a count, `-c:1,5` still silently ignores the `5` (only the bare `-c:5` raises). So the same argument behaves differently depending on whether you wrote `foo bar` or `2`.

### 5. The root cause is that unmatched namespaces are silent

Both this bug and the two leftovers above are instances of one thing: a namespace that matches no node is discarded without a word. A warning at the end of `parse()` for any namespace that was never consumed by a node would have surfaced this to users years ago, and would cover `-c:1-2` with names, `-c:1,5` in count mode, and typos like `-c:qux`. That's a bigger change than this PR and shouldn't block it, but it's the thing actually worth doing here and it'd be good to open an issue for it.

### 6. Precedence wart this makes reachable

```console
$ celery multi start foo bar -c:1,2 10 -c:1 20
```

`foo` gets `10`, not `20` — the list expansion writes into `namespaces['1']` and clobbers the more specific option, in either argument order. That's pre-existing behaviour in `_update_ns_ranges`, but on `main` it was unobservable for named nodes because the list never reached a node at all. Now it does. Most people would expect the specific index to win. Worth a test pinning whatever you decide, even if the decision is "list wins, documented".

### Nits

* In the `else` branch, `ranges` is provably `False`, so `self._update_ns_ranges(p, ranges)` reads as though it might not be. Not wrong, but a reader has to walk back up through the `try/except ValueError` to confirm. `False` with the existing comment would be clearer.
* `_update_ns_opts` iterates a snapshot of `items()`, but the `ns_opts` values are live dict references, so if `names[ns_index]` collides with a namespace key processed later in the same loop, updates cascade. Reachable with nodes literally named `'1'`, `'2'`… Esoteric and pre-existing, but the reorder populates digit namespaces in bulk now, so it's slightly more reachable than before.
* The assertion checks `n.options['-c']`; neighbouring tests in this file check the rendered `argv` (`'-c 5' in nodes[0].argv`). Checking both would catch a regression between option merge and command construction.

### Housekeeping

* No changelog entry. This is a user-visible CLI behaviour change in both directions (an option that used to be ignored now applies; input that used to be ignored now raises).
* The usage docstring in `celery/bin/multi.py` is the de-facto documentation for this syntax. If `-c:1,2` now works with named nodes but `-c:1-2` still doesn't, that distinction belongs there.
* Four commits, three of them merges from `main`. Fine assuming a squash.

### One practical note for maintainers

The author has stated in the thread that they've stopped scanning this project and won't push further to this PR. So if the tests in points 2 and 3 are wanted — and I think at least point 3 is, since it protects the branch the fix relies on — someone else will need to add them. The change itself is eight lines and independently verifiable, so that's a small amount of work to finish rather than a reason to drop it.

---

**Verdict:** correct fix, right scope. I'd merge after the comment in point 1 and the regression tests in points 2 and 3; points 4–6 are follow-ups worth recording in an issue rather than blocking on.

auvipy:
@Dev-next-gen please check the remaining review suggestions

Dev-next-gen:
Thanks for the ping, @auvipy of course, happy to look.

I'd gone quiet on purpose: I'd turned celery off in my automated scan, which is why I said #10661 and #10657 would be the last PRs you'd see from me, and I didn't want to keep pushing to this one without a sign it was still wanted.
Since you're asking, I'll go back through the review points here and address whatever's left.

Give me a short while to do it properly and I'll follow up in this thread.

(@Dev-next-gen)

Dev-next-gen:
@MehrazRumman, your review was thorough and the three blocking items were exactly right. I pushed 7d6fa8f addressing them:

**Point 1** — added a comment in the `if ranges:` branch explaining why the original order must stay: reordering would make out-of-range members of `-c:1-5` raise instead of being silently ignored, which is a behaviour change for the count form.

**Point 2** — two new tests pin the `KeyError` paths: `-c:1,5` with two named nodes raises on index 5, and `-c:0,1` raises on index 0.

**Point 3** — two new tests guard the count-form branch: `start 3 -c:1-5 10` assigns `['10', '10', '10']` without raising, and `start 3 -c:5 10` still raises `KeyError`.

All 33 tests in `t/unit/apps/test_multi.py` pass. A changelog entry is included.

Points 4–6 (the `-c:1-2` named-node gap, count-vs-names divergence on out-of-range, and the silent-namespace-discard root cause) are genuine and worth an issue; they shouldn't block this PR, as you said.

sylvesterkaczmarek (approved):
Rechecked current 7d6fa8f. The follow-up now documents why the count-form ordering must remain unchanged and adds regressions for the named-node out-of-range/zero cases plus the count-form range behavior. The original comma-separated index-list fix remains intact, and the current CI matrix is green. No blocker from me.

PR #10662, opened 2026-09-17
Fix intermittent failure in backend teardown integration test

phanky1:
## Description
This PR updates the backend teardown regression test to track Redis backend sockets using weak references and check that some are reclaimed during repeated fixture cycles.


t/integration/test_backend_leak_in_fixture_teardown.py
```diff
@@ -35,6 +36,7 @@ def test_fixture_teardown_releases_backend_connections():
     for _ in range(iterations):
         apps.append(run_one_fixture_cycle())
 
-    # Repeated cycles give GC a chance to collect orphaned backends along the way.
+    # Simulate automatic garbage collection.
+    gc.collect()
```
auvipy: can we try to find other ways to improve this test? is gc a must need here?

phanky1: IMO this gc.collect() call makes the test more reliable and better reflects normal usage.

As for @Nusnus's concern, I think he meant we shouldn't use `gc.collect(`) in application code, not that we shouldn't use it in tests.

auvipy: i understand that, but this is not the best fix, it is just a work around IMHO

auvipy:
lets wait for the CI

t/integration/test_backend_leak_in_fixture_teardown.py
phanky1: ok, let me see if there is a better appoach :D

phanky1: I updated this PR with a new appoach.

auvipy:
i was about to suggest kind of related change suggested by my copilot analysis

phanky1:
> lets wait for the CI

These smoke tests still fail because the current requirement, billiard>=4.2.1,<5.0, still allows billiard 4.3.0.

ảnh

auvipy:
so the billiard new release has the source of new regressions or incompatibility? @GangEunzzang

auvipy:
also the smoke tests taking so long to execute is not a great sign

auvipy:
lets try billiard 4.3.0rc1 to see if that can stablize or not, otherwise will roll back to 4.2.x https://github.com/celery/celery/pull/10659/changes

phanky1:
> lets try billiard 4.3.0rc1 to see if that can stablize or not, otherwise will roll back to 4.2.x https://github.com/celery/celery/pull/10659/changes

I've tried Billiard 4.3.0rc1, and the entire test suite passed twice.

auvipy:
Ok so my one with billiard 4.3.0rc1 failed as it didnt have the change of this pr @GangEunzzang

auvipy (approved):

auvipy (approved):

GangEunzzang:
@auvipy Yes, the teardown failure was separate from the Billiard 4.3.0 smoke failures; this PR addressed the former, and #10667 fixed the shutdown and SystemExit cases.

PR #10663, opened 2026-09-17
Fix escaped percent signs in node formatting

jakezwang:
## Description

Recognize `%%` as an escaped percent sign in `simple_format()`. The formatter already registers `%` as a literal value, but its pattern only matches word characters, so `%%n` expands the node name and `%%x` raises an unknown-format error.

For example, `host_format("logs/%%n-%n.log", host="worker.example.com")` now returns `logs/%n-worker.log`.

Tests: 598 utility tests passed, with 2 skips and 1 expected failure; all pre-commit checks passed, including mypy. Seven regression cases fail before the fix and pass afterward.


auvipy:
@MehrazRumman can you check this?

MehrazRumman (approved):
## Review: Fix escaped percent signs in node formatting

Thanks, this is a clean, minimal fix. The intent was already in the code: `simple_format()` does `keys.setdefault('%', '%')`. But `RE_FORMAT = r'%(\w)'` can never match a second `%`, so that line was dead code. Widening the pattern to `%([%\w])` makes the existing mapping take effect. Because `re.sub` scans left to right and doesn't overlap matches, `%%%n` → `%` + expanded `n` and `%%%%n` → `%%n` behave as you'd expect. The parametrized cases cover this well.

I traced all callers (`host_format`, `node_format` in `bin/worker.py`, `app/log.py`, `apps/multi.py`). A few things to consider before merging.

### 🟡 1. Escapes don't survive `celery multi` (double expansion)

`celery multi` expands option values through `build_expander()` → `node_format()`. The worker then formats the same value **again** (`setup_logging_subsystem()` → `node_format(logfile, hostname)`). With this change, an escape is consumed by the first pass and the second pass expands it:

| input | after `multi` pass | after worker pass |
|---|---|---|
| `logs/%%n-%n.log` | `logs/%n-w1.log` | `logs/w1-w1.log` ❌ |
| `logs/50%%d.log` | `logs/50%d.log` | `logs/50example.com.log` ❌ |
| `logs/%%x.log` | `logs/%x.log` | `ValueError: Unknown format %x` ❌ |

This isn't a regression, since `%%n` was already broken before. But the PR description presents `%%` as *the* way to get a literal `%`, and that only holds for single-pass paths (`celery worker` directly). Under `multi` you'd need `%%%%`. The existing `UNKNOWN_SIMPLE_FORMAT_KEY` message already hints at this ("did you escape and the value was expanded twice? (%%N -> %N -> %hostname)"). `build_expander()` also deliberately re-emits `i='%i'` / `I='%I'` so those survive to the second pass.

Suggestions, any of these would do:
- Document it: literal `%` is `%%` for `celery worker`, and `%%%%` when going through `celery multi`.
- Or, preferably, have `build_expander()` re-emit the escape as well (`**{'%': '%%'}`), the same way it already preserves `%i` / `%I`. Then `%%` means the same thing everywhere. Note that `node_format(..., **extra)` can't take `%` as a keyword, so this needs a small change to how the extra keys are passed.
- Add a test in `t/unit/apps/test_multi.py` asserting the final worker argv for an escaped value, whichever behavior you pick.

### 🟡 2. Small behavior change worth a changelog line

Anyone who currently has `%%h` (or `%%n` / `%%d`) reaching Celery unprocessed gets `%<hostname>` today and will get a literal `%h` after this change. This is the correct behavior and unlikely to be relied on, but it's user-visible for log and pid file paths. Note that `docs/userguide/workers.rst` tells supervisor users to write `%%h`: supervisor unescapes that to `%h` before Celery sees it, so those users are unaffected. I'd still mention it in the 5.7 changelog, e.g. "`%%` in node/host format strings now produces a literal `%`".

### 🟢 Nits

- **Docs:** the variable table in `docs/userguide/workers.rst` (the `%h` / `%n` / `%d` section) would benefit from a `%%` row ("literal `%`"). Currently the only mention of `%%` is the supervisor note, which is about supervisor's escaping, not Celery's, and reads confusingly next to this change.
- **Test placement:** `test_host_format_percent_escape` is a module-level function in `test_nodenames.py`, while the existing tests there are grouped in a class. Minor, just for consistency.
- **Extra cases:** consider covering an escape adjacent to the process-index keys (`%%i` → `%i`, and `%%I`). Those are the ones most likely to be passed through multiple layers (see `prefork.process_initializer`, which re-runs logging setup when `%i` is present in `CELERY_LOG_FILE`).
- `test_simple_format_unknown_key` passes on `main` too, so it's a guard rather than a regression case. That's fine to keep, just noting it for accuracy against the "seven regression cases" count.

### Summary

LGTM on the core change: it's correct, minimal, and restores clearly intended behavior. I'd like to see **(1)** either handled in `build_expander()` or documented, plus a changelog entry for **(2)**. The nits are optional.

auvipy:
@jakezwang please cross check the remaining review suggestions. if possible add some smoke or integrations tests which are relevant in this case

auvipy:
@GangEunzzang you review will help to proceed with this

GangEunzzang (commented):
Nice fix. One thing I ran into: with `celery multi`, a pidfile like `--pidfile=%%%%n-%n.pid` gets written as `%n-w1.pid`, but multi looks for `%%n-w1.pid`, so `stopwait` can't find the worker and leaves it running. It's not a regression since main can't even start with that, so a follow-up is fine too. Wrapping `Node.pidfile` in `node_format()` fixed it for me.

PR #10667, opened 2026-09-18
Keep a cancelled task unacked when the child reports SystemExit

GangEunzzang:
Follow-up to the investigation in #10659.

With Billiard 4.3.0 the `test_reset_visibility_timeout` smoke tests fail: after `worker: Cold shutdown (MainProcess)` the worker never logs `Restoring 1 unacknowledged message(s)`. The worker isn't stuck. It exits normally, and the task that was still running has been acknowledged and dropped instead of being requeued.

Cold shutdown cancels the active request by sending `SIGTERM` to the child running it, and Billiard's signal handler in the child raises `SystemExit`. Up to Billiard 4.3.0rc1 that killed the child, the job came back as `Terminated`, and `Request.on_failure` returned early for it, so the message stayed unacked and kombu restored it on close. Since celery/billiard#452 the worker loop catches `BaseException`, so the same `SystemExit` is marshaled back as an ordinary task failure. `on_failure` then takes the `acks_late` branch, `acks_on_failure` is on by default, and the message gets acked. Bisecting Billiard between rc1 and 4.3.0 lands on that commit.

This change treats a `SystemExit`/`KeyboardInterrupt` result like `Terminated` when the request was cancelled or revoked by the worker itself. A task that exits on its own is still a failure and goes through the usual path.

Checked with Billiard 4.3.0 pinned: the five failing `test_reset_visibility_timeout` cases pass, the full `t/smoke/tests/test_worker.py` passes, and `t/unit/worker` is unchanged apart from the three new tests. Also reproduced and re-checked on macOS with a plain prefork worker against Redis, where the in-flight message now comes back to the queue after a cold shutdown.

Fixes the smoke failures blocking #10659.


auvipy (changes requested):
test task smoke tests are failing here

GangEunzzang:
> test task smoke tests are failing here

Looking into it now!

auvipy:
merged the other integration test upgrade and also merged billiard 4.3 to main branch to find out all compatibility issues. only in dev billiard 4.3rc1 remains, which we can change after this task tests are working

GangEunzzang:
@auvipy The `test_tasks.py` job wasn't failing — it was hanging in `test_task_termination[SYSTEM_EXIT]` until the 60-minute job timeout. This comes from the same Billiard change.

A task calling `sys.exit(1)` used to kill the child and come back as `WorkerLostError`. With 4.3.0, the `SystemExit` is reported as the result first, so `on_failure` tries to store it. The JSON serializer rejects it (`Object of type SystemExit is not JSON serializable`), no result is written, and the test's `.get()` waits indefinitely.

It didn't show up on `main` because `test_worker.py` fails first there, which cancels the remaining smoke jobs.

I pushed a second commit that reports this as `WorkerLostError` like before. The child exits either way — Billiard exits it immediately after reporting the result — and a task can't meaningfully fail with `SystemExit`. With both commits and 4.3.0 pinned, `test_task_termination` and the `test_worker.py` shutdown tests pass locally, and both smoke jobs are green now.

The remaining red check is `docker-build`: `requirements/dev.txt` still has `billiard.git@v4.3.0rc1` from #10662, while `default.txt` now pins `==4.3.0`, so pip can't resolve both requirements together. Happy to drop the rc1 pin here as well if you want that included in this PR.

auvipy:
yes I already opened a pr to change that wait a min

auvipy (approved):

Issue #6075, opened 2020-05-06
Revoked tasks storage & communication improvements


jheld:
# Checklist

- [x] I have checked the [issues list](https://github.com/celery/celery/issues?q=is%3Aissue+label%3A%22Issue+Type%3A+Enhancement%22+-label%3A%22Category%3A+Documentation%22)
  for similar or identical enhancement to an existing feature.
- [x] I have checked the [pull requests list](https://github.com/celery/celery/pulls?q=is%3Apr+label%3A%22Issue+Type%3A+Enhancement%22+-label%3A%22Category%3A+Documentation%22)
  for existing proposed enhancements.
- [x] I have checked the [commit log](https://github.com/celery/celery/commits/master)
  to find out if the if the same enhancement was already implemented in the
  master branch.
- [x] I have included all related issues and possible duplicate issues in this issue
      (If there are none, check this box anyway).

#### Related Issues

- None

#### Possible Duplicates

- None

# Brief Summary

Part of this is a question about the current implementation around revoked task storage and communication, and then depending on the discussion, possible improvements (including documentation).

This conversation is more related to the redis broker (implementation discussion may be more related to it, for brevity & consistency), but I imagine the same issue can be seen on other brokers.

When a worker starts up, it asks (via `mingle`) all of the workers in the system for their revoked task lists. It also sends it own in that message. It then receives the revoked task lists from all workers that were able to respond, and then it merges those lists one by one into its own.

For a fairly healthy celery cluster, this means that it sends it revoked tasks to all other workers (if persistent mode is disabled there will be _no_ revoked tasks yet), and it also then receives what will likely be 99% identical revoked tasks lists from all other workers in the cluster (other than the ones that are _just_ coming up like this worker).

If for instance you have the default 500k revoked tasks list fully saturated (or really, just any large number), that is a lot of messages and data flying through the system at about the same time. In redis' case, depending on the number of existing workers in the cluster (whether they're being killed or not), the number of new workers coming up, and how many revoked tasks are generally in each [existing and fully started up] worker, if the redis' size isn't "large" (under several GB), we risk a redis OOM error (my company has seen this so many times, and now I'm working through extensive debugging, tests, and improvements). This can cripple a celery cluster, and any other process that relies on celery or uses the same broker.

So, part of this is a question in implementation, and the other part is, assuming we wish to make alternative implementations, the discussion of those.

# Design

## Architectural Considerations

None

[...]

PR #10668, opened 2026-09-18
Fix revoked tasks running anyway after mingle merged revoke stamps from another host

### LINKED ISSUE #4300: Revoked task still executes?
This issue is a continuation of issues https://github.com/celery/celery/issues/4295 (see for Celery details) and https://github.com/celery/celery/issues/4299.

I am trying to [cancel a scheduled task](http://docs.celeryproject.org/en/latest/faq.html#can-i-cancel-the-execution-of-a-task) using [revoke()](http://docs.celeryproject.org/en/latest/reference/celery.result.html#celery.result.AsyncResult.revoke), but the task still executes. Here’s the Celery log:

```
[2017-09-28 17:52:18,811: INFO/MainProcess] Received task: app.tasks._do_singlething[0e4f0131-b4e6-4209-865e-e91b73e92f28]  ETA:[2017-09-28 07:53:18.766382+00:00] 
[2017-09-28 17:52:32,091: INFO/MainProcess] Tasks flagged as revoked: 0e4f0131b4e64209865ee91b73e92f28
[2017-09-28 17:53:18,831: WARNING/ForkPoolWorker-1] Hello from executing task 0e4f0131-b4e6-4209-865e-e91b73e92f28
[2017-09-28 17:53:18,841: INFO/ForkPoolWorker-1] Task app.tasks._do_singlething[0e4f0131-b4e6-4209-865e-e91b73e92f28] succeeded in 0.06189309403998777s: None
```

Code snippet that revokes is the same as in issue https://github.com/celery/celery/issues/4299.

yauhpl-alt:
## Description

Fixes #4300 — a revoked task with an ETA/countdown runs anyway, although the worker logged `Tasks flagged as revoked` for it. This is the mechanism @tahamr83 guessed at in https://github.com/celery/celery/issues/4300#issuecomment-951743011, confirmed on a production cluster and reproduced in unit tests at the `hello`, mingle and `--statedb` levels.

### The bug

`celery.worker.state.revoked` is a `LimitedSet`, and `LimitedSet.add()` stamps every item with `time.monotonic()` — seconds since the boot of *this* host. `mingle` copied the whole set between workers on start-up (`hello` sent and received `revoked._data`, stamps included), and the stamps were merged as they were.

A stamp taken on a host with a longer uptime is ahead of the local clock of the worker that receives it, so for that worker:

- it never expires — `purge()` stops at the first item whose `stamp + expires > now`, and these are forever "not expired yet";
- once such stamps fill the set up to `maxlen`, every `add()` triggers `purge()`, which drops the item with the *smallest* stamp — the id just revoked locally.

So the revoke is logged and lost in the same call. Any worker started while the cluster is in this state inherits the stamps on mingle, so a rolling restart never clears it; it recovers only when the local uptime catches up with the foreign stamps (which is why the reporter saw it "disappear the next morning"). A state db (`--statedb`) read after a reboot of the host does the same to a single worker.

Observed on Celery 5.6.3, Kubernetes, ~3800 revokes/hour cluster-wide: every worker had exactly `maxlen` (50 000) revoked ids, the oldest 15 days old with `worker_revoke_expires` at 3 hours, and 8 out of 8 revoked ETA tasks whose ETA fell inside the hour I looked at ran anyway, with no `Discarding revoked task` in the log.

### The fix

Foreign stamps are never merged: a set received on mingle or read from the state db contributes its **ids only**, stamped with the local clock (`merge_revoked()` in `celery.worker.state`, used by the `hello` command, `Mingle.on_revoked_received` and `Persistent`). `hello` and `Mingle.send_hello` send only the ids as well, which halves the payload.

- `LimitedSet` keeps `time.monotonic()`, so a step of the wall clock cannot reproduce the bug on a single host.
- The ids received live `REVOKE_EXPIRES` seconds from their receipt rather than from the revoke — the concern of the `# XXX AVOID THIS` comment on that path of `LimitedSet.update()`, and a far better failure than losing revokes. Documented in the workers guide.
- A mixed cluster is safe in both directions: an older worker merges a list of ids through the same iterable path (stamping them with its clock), and a newer worker takes the ids out of the dict an older one sends. Same for a state db written by either version.

Independently of the clock, the heap entries of `LimitedSet` are now ordered by `(stamp, sequence number, item)`: items added within the clock resolution keep their insertion order, and the heap never compares the items themselves (a tie between an `int` and a `str` raised `TypeError` — the flakiness #3892 papered over). `add()`/`purge()` take `now=0` as a time rather than as "unset", and `__eq__` returns `NotImplemented` for a non-`LimitedSet` operand.

### Tests

- `test_hello_stamps_received_revoked_items_locally` (`test_control.py`), `test_start_stamps_received_revoked_items_locally` (`test_consumer.py`) and `test_merge_stamps_saved_items_locally` (`test_state.py`) reproduce the bug on the real paths: ids arriving with stamps a million seconds ahead of the local clock are stamped locally, expire with the local ones, and — at the mingle level, with `maxlen=1` — no longer evict an id revoked afterwards.
- `test_add_orders_items_within_the_clock_resolution`, `test_add_at_time_zero`, `test_update_from_another_clock` and `test_eq` cover the `LimitedSet` changes.

All of them fail on `main`. `t/unit` passes with the same set of failures as `main` on my machine (15 pre-existing `test_prefork` failures on macOS, unrelated); `pre-commit` is clean on the changed files, `apicheck`/`configcheck` report nothing new.

### Notes for the 5.7 changelog

- Fixes #4300: revoked tasks ran anyway once `mingle` had filled a worker's revoked set with stamps from a host with a longer uptime. Ids received on mingle or read from `--statedb` are now stamped with the local clock and expire `CELERY_WORKER_REVOKE_EXPIRES` seconds after receipt.
- The `revoked` value in the `hello` remote-control request and reply is a list of ids, not a `{id: [stamp, id]}` dict; both versions merge the other's format, so a rolling upgrade is safe.
- `LimitedSet._data`/`_heap` entries are `(stamp, seq, item)` 3-tuples; `now=0` is a time, not "unset"; `__eq__` returns `NotImplemented` for other types. `as_dict()`, pickling and `update()` are unchanged.


auvipy:
@GangEunzzang @MehrazRumman please check this

MehrazRumman (changes requested):
## Review

Thanks for this — the diagnosis is the best part of the PR. The causal chain (monotonic stamp → merged verbatim by `hello` → `stamp + expires > now` forever → set saturates at `maxlen` → every `add()` purges the *newest local* revoke as the "oldest") is correct, matches #4300, and the production numbers (50 000 ids, oldest "15 days" old with `worker_revoke_expires=3h`, 8/8 revoked ETA tasks executed) are convincing. The `(stamp, seq, item)` tie-break is also a genuinely good fix for the `TypeError`-on-tie problem that #3892 papered over by switching to `monotonic()`.

My concerns are with the *choice of clock*, not with the analysis. I'd like a couple of them addressed (or explicitly accepted and documented) before this lands in 5.7.0.

---

### 1. Moving to `time.time()` trades a cross-host bug for an on-host one

`monotonic()` was not only about test resolution — it is also immune to wall-clock steps. After this change, a backwards step of the local clock (chrony `makestep` after a bad container start, VM snapshot restore, a bad NTP source, manual correction) reproduces **exactly bug #4300 on a single host, with no mingle involved**:

* items added before the step now have `inserted > now`, so they never expire;
* items added after the step get *smaller* stamps, so once `maxlen` is reached, `purge()` pops them first.

Worse, the clamp in `add()` doesn't help here. It only protects against stamps that are in the future *at insertion time*; it cannot protect against entries that become "in the future" later. And `purge()` only inspects `self._heap[0]`, so the future-stamped entries sit at the tail of the heap where the head-only loop never reaches them.

Forward steps are less dramatic but lose revokes early (everything older than `now - expires` after the jump is dropped in one purge).

This isn't hypothetical in the Kubernetes environment the PR was debugged in — clock steps at container start are common.

**Suggested alternative: keep `monotonic()` internally, use the wall clock only on the wire.** The internal stamp stays step-proof; only the exchange is made comparable:

```python
def add(self, item, now=None):
    now = time.monotonic() if now is None else now
    ...

def update(self, other):
    ...
    # `inserted` arrives as a wall-clock stamp from another host
    age = max(0.0, time.time() - inserted)
    self.add(key, now=time.monotonic() - age)

def as_dict(self):
    wall, mono = time.time(), time.monotonic()
    return {key: wall - (mono - inserted)
            for inserted, _, key in self._data.values()}
```

Cost to be honest about: `as_dict()` round-trips are no longer bit-exact, so the `r.update(s.as_dict()); r == s` doctest and `__eq__` would need a tolerance or a key-only comparison. That's a real wrinkle, but it's contained inside this class, whereas clock sensitivity is not.

**Even simpler alternative worth considering:** don't merge foreign stamps at all — on mingle, add received ids with the local `now` and discard the sender's stamp. Two lines, no clock question, no wire-format question, and it fixes #4300 outright. The cost is that a received id lives up to `expires` from the moment of receipt rather than from the original revoke, which is the risk the existing `# XXX AVOID THIS, it could keep old data if more parties exchange them all over and over again` comment already flags. Strictly safer than the current failure mode, which is *losing* revokes.

If you'd rather keep `time.time()` as-is, that's a defensible call — but please say so explicitly in the PR description and add the rationale as a comment, so the next person doesn't flip it back for the same reason #3892 did.

### 2. Mixed-version clusters are affected in both directions, and `--statedb` makes it persistent

The PR says the wire format is unchanged. That's true *structurally* (`update()` still reads `inserted[0]`), but not *semantically* — the units of the first element change, and both directions misbehave during a rolling upgrade:

* **new worker ← old worker:** it receives monotonic stamps (≈ host uptime, say `10^5`–`10^6`). The clamp is a no-op because they're far *below* `time.time()`, so every inherited id looks ~55 years old and is dropped at the first `purge()`. The new worker silently starts with an empty revoked set, and revoked ETA tasks can run on it.
* **old worker ← new worker:** it receives wall-clock stamps (`≈1.76e9`) against its monotonic clock. Those entries never expire — effectively permanently, not the "recovers next morning" version in the issue. So a partial rollout makes #4300 *worse* on the not-yet-upgraded workers.

The same applies to `--statedb`: `__reduce__` goes through `as_dict()`, so a state file written by an old worker loses its whole revoked set on first load by a new worker, and a state file written by a new worker poisons an old worker on downgrade — and unlike mingle, that survives restarts.

Two things would help:

1. A legacy-stamp heuristic in `add()`/`update()`: a stamp below a plausible epoch floor (or `< time.time() - some_max_age`) almost certainly came from a pre-5.7 monotonic worker; re-stamp it to `now` instead of letting it expire instantly. Combined with the existing future-clamp, both directions then degrade to "kept for up to `expires`" rather than "lost" or "kept forever".
2. Either way, an entry in the changelog and in the 5.7 upgrade notes — at minimum recommending `--without-mingle` during the rollout, and noting the `--statedb` implication.

### 3. `update()` no longer does a bulk merge

```python
if isinstance(other, LimitedSet):
    for inserted, _, key in other._data.values():
        self.add(key, inserted)
```

This replaces `_data.update()` + one `_refresh_heap()` + one `purge()` with N × (`time.time()` + `heappush` + a `purge()` call once `len(_data) >= maxlen`). It's correct and the clamp requires going through `add()`, but on a `REVOKE_LIMIT=50000` mingle it's `O(n log n)` with a `purge()` per item where it used to be a `heapify` and a single purge. Worth either measuring or adding a bulk path that stamps once:

```python
now = time.time()   # hoist: one clock read for the whole merge
```

Also: `s.update(s)` now mutates `self._data` while iterating it (`RuntimeError`). A `if other is self: return` guard is cheap.

### 4. Smaller things

* **`add()`: `if now` should be `if now is None`.**
```python
  now = min(now, time.time()) if now else time.time()
```
  `now=0` / `0.0` is silently replaced by the current time. Under `monotonic()` that was near-harmless; under wall clock, `0` is a legitimate (epoch) timestamp and callers passing `now=0` in tests or code get the opposite of what they asked for. Prefer:
```python
  now = time.time() if now is None else min(now, time.time())
```
  `purge()` has the same `now = now or time.time()` pattern.
* **`add()` reads the clock even when `now` is supplied.** Minor, but it's a syscall per item on the merge path above.
* **`__eq__` via `as_dict()`** is the right call given the sequence numbers, but it now builds two dicts per comparison, and it still raises `AttributeError` on a non-`LimitedSet` operand. Returning `NotImplemented` for non-`LimitedSet` would be a small improvement while you're here.
* **Wire payload grows.** `hello` sends `revoked._data`, whose values are now 3-element entries, so every id costs one extra integer on the wire. It's small next to the id duplication that's already there, but given #6075 (mingle payloads and broker OOM), it's worth noting — and it raises the question of why `hello` sends `_data` at all rather than `as_dict()`, which `update()` already accepts and which would roughly halve the payload by not repeating each id. Reasonable follow-up, possibly out of scope here.
* **`_seq` isn't restored by `__reduce__`.** Fine in practice (the counter restarts and items are re-added in order), but worth a one-line comment so it doesn't look like an oversight.
* **Entry arity is a breaking change for anything reading `_heap`/`_data`.** Both are underscore-private, but `_data` is on the wire, which makes it semi-public. A changelog line covers this.

### 5. Tests

The three new tests are on point, and `test_update_from_a_clock_ahead_does_not_evict_new_items` is a genuine regression test for #4300. A few suggestions:

* `test_add_orders_items_within_the_clock_resolution` patches `celery.utils.collections.time.time`, which is the stdlib `time` module object — so this patches `time.time` **process-wide** for the duration, including for pytest internals and logging. Since `add()` now clamps rather than ignores `now`, you can get the same coverage without any patching:
```python
  s.add('b', now=1.0)
  s.add(1, now=1.0)   # no comparison between the items themselves
  s.add('a', now=1.0)
```
* Missing: a test for a **backwards** clock step (§1) — add items, purge with `now` *before* the insertion stamps, assert the set doesn't start evicting new items.
* Missing: a test for a **legacy monotonic stamp** arriving via `update()` (§2), whatever the chosen behaviour ends up being.
* Worth adding a test at the `hello` level in `t/unit/worker/test_control.py` that goes through the actual mingle path, not just `LimitedSet` directly — that's where the bug actually bites, and a unit-level test on the collection won't catch a future regression in `control.py`.
* `assert len(s) == 3` in the new test is a bit brittle; asserting on membership of the specific ids would survive a `maxlen` tweak better.

### 6. Housekeeping

* Codecov reports 100% of modified lines covered and project coverage flat at 89.14% — good.
* The branch has three merge commits from `main` in it; assuming maintainers squash on merge that's fine, otherwise a rebase would tidy the history.
* Please add a changelog entry, and update the `worker_revoke_expires` / mingle documentation if it mentions the stamping behaviour.

---

**Verdict:** the bug is real, well evidenced and worth fixing for 5.7.0, and the sequence-number tie-break should go in regardless. I'd like to see §1 (clock-step behaviour) either fixed or explicitly accepted with a rationale, and §2 (mixed-version rollout + `statedb`) at least documented, before merge.

yauhpl-alt:
Thanks for the thorough review — you're right on §1, and I went with your simpler alternative. Pushed as a single rebased commit (the merge commits were GitHub's "Update branch"; gone now), PR title and description updated to match.

**§1 (clock).** Agreed: a backwards step of the wall clock would have reproduced #4300 on a single host, and the clamp in `add()` couldn't have caught it. `LimitedSet` keeps `time.monotonic()`. Foreign stamps are now never merged at all: `merge_revoked()` in `celery.worker.state` takes only the ids out of whatever arrives and stamps them with the local clock. It's used by the `hello` command, by `Mingle.on_revoked_received` and by `Persistent` — the state db is the same bug after a reboot of the host, since the saved stamps are from the previous uptime. The cost you describe is real and documented (the received ids live `REVOKE_EXPIRES` from receipt; the `# XXX AVOID THIS` comment on that path of `update()` now says why the path exists and what the trade-off is), and there's a sentence in the workers guide.

**§2 (mixed clusters, statedb).** With stamps ignored on receipt, no heuristic is needed: a new worker takes the ids out of the dict an old one sends, and `hello`/`send_hello` now send only the ids, which an old worker merges through the same iterable path, stamping them with its own clock (checked against 5.6.3: `hello` and `Mingle.sync_with_node` fed a plain list add the ids with local stamps) — so the not-yet-upgraded side of a rollout actually ends up in a better state than before, not a worse one. Same for a state db written by either version (`__reduce__` goes through `as_dict()`, and `Persistent` drops the stamps on load whatever the format). Sending only the ids also halves the `hello` payload you mentioned under #6075.

**§3 (bulk merge).** `update(LimitedSet)` is back to the bulk path (`_data.update()` + one `_refresh_heap()` + one `purge()`), unchanged from `main`; `s.update(s)` is fine again.

**§4.** `now is None` in both `add()` and `purge()` (`test_add_at_time_zero`); the clock is read only when `now` isn't supplied; `__eq__` returns `NotImplemented` for a non-`LimitedSet` operand (`test_eq`); a comment on `__reduce__` about the sequence numbers. The wire payload is now smaller than before rather than larger, and the entry arity is called out in the commit message.

**§5 (tests).** The tie-break test passes `now=1.0` instead of patching `time.time`. The regression tests now sit on the real paths: `test_hello_stamps_received_revoked_items_locally` (`test_control.py`), `test_start_stamps_received_revoked_items_locally` (`test_consumer.py`, with `maxlen=1` to show the eviction no longer happens) and `test_merge_stamps_saved_items_locally` (`test_state.py`). All six new tests fail on `main`. The backwards-clock-step and legacy-stamp tests you asked for don't apply any more: the clock is monotonic again, and there's no stamp interpretation left to get wrong — `test_update_from_another_clock` pins down that `update(iterable)` stamps locally while `update(LimitedSet)`/`update(dict)` keep the stamps, which is the contract the worker relies on.

**§6.** Rebased onto `main` in one commit. On the changelog: `Changelog.rst` here seems to be assembled from the PR titles at release time ("Prepare for release" commits) — happy to add an entry if you'd rather have it in the PR, tell me where.

auvipy:
we need more reviewer here @GangEunzzang @phanky1

auvipy:
@yauhpl-alt also, no more forced push please.it is better to have commit diffs for easier review

MehrazRumman (approved):
## Re-review

Thanks for the quick and complete turnaround — every point from the first round is addressed, and the "don't merge foreign stamps at all" approach is the right call. It removes the clock question entirely rather than moving it somewhere else, and it is strictly safer than the failure mode it replaces (losing revokes). Happy to approve; the notes below are non-blocking.

### What I checked

- **Regression coverage is real.** I swapped `celery/` for `main`'s sources and re-ran the new tests against them: all of `test_add_orders_items_within_the_clock_resolution`, `test_add_at_time_zero`, `test_eq`, `test_hello_stamps_received_revoked_items_locally`, `test_start_stamps_received_revoked_items_locally`, `test_merge_stamps_saved_items_locally` and the updated `test_start` fail on `main` and pass on this branch. (`test_update_from_another_clock` passes on `main` too — it pins the existing `update(iterable)` vs `update(LimitedSet)` contract rather than the bug, which is fine, just not "all six".)
- **Mixed-version, both directions, through JSON.** An old worker's `_data` dict after a `json.dumps`/`loads` round-trip (`{id: [stamp, seq, id]}`) merges through `merge_revoked()` as ids only, stamped `<= monotonic()`. A new worker's `list` payload goes through the old `update()` iterable path and gets local stamps there. Nothing is lost or kept forever in either direction, so a rolling upgrade is safe and `--without-mingle` is no longer needed during the rollout.
- **§3 is resolved on the real path.** Merging 50 000 ids into a full `maxlen=50000` set takes ~33 ms locally, and `list(revoked)` on 50 000 ids ~4 ms. `s.update(s)` is fine again.
- **Payload.** For 50 000 uuid ids the `hello` payload goes from 5.28 MB (`_data` dict, 3-tuples) to 2.00 MB (list of ids) — a bit better than "halves", and a nice side effect for #6075.
- CI: all 377 checks green; single rebased commit. Branch is one commit behind `main` (a deps bump), which will not conflict.

### Non-blocking notes

1. **Integer stamps no longer survive a pickle/`update(dict)` round-trip.** Before, `add(item, now=0)` silently replaced `0` with the clock; now it stores `0` as an `int`, and `update(dict)` (which `__reduce__` goes through, i.e. `--statedb`) raises `ValueError: Expecting float timestamp, got type <class 'int'>`. Same for any integer `now`. Nobody in the tree passes an int, but since `add()` now honours what it is given, either coerce in `add()`:

   ```python
   now = time.monotonic() if now is None else float(now)
   ```

   or accept `numbers.Real` in `update()`. `test_add_at_time_zero` passes `now=0` (an int) and would cover the coercion if you add a `pickle.loads(pickle.dumps(s))` to it.

2. **`--statedb` now grants a fresh `REVOKE_EXPIRES` on every load, not just after a reboot.** With the old code a saved id that was 1 s from expiry stayed 1 s from expiry after a plain restart on the same host; now its age resets to 0. That is bounded by `maxlen` and clearly better than the reboot case, so I am fine with it, but the docs sentence only covers the mingle path. One more sentence in the `--statedb` paragraph of *Persistent revokes* would save a support question later. If someone wants exact expiry across restarts later, the state db is the one place where it is cheap to do properly: save `time.time()` alongside `{id: age}` and re-stamp as `monotonic() - (age + elapsed)` on load. Follow-up material, not this PR.

3. **Worth a comment: the order of the ids on the wire matters.** `list(revoked)` iterates oldest-first, and the receiver's sequence numbers preserve that order, so when the receiver is already at `maxlen` it is the sender's *oldest* ids that get evicted first — exactly what you want. That is a nice property that a later "optimisation" to `list(revoked._data)` could quietly weaken; a one-liner in `hello` / `send_hello` would protect it.

4. **`merge_revoked(None)` raises `TypeError`.** All three callers guard with `if revoked:` / `if zrevoked:` so this cannot trigger today, but the function is now in `__all__`. Either an `if not task_ids: return` or a note in the docstring that the argument must be an iterable.

5. **Release notes.** Two things a downstream consumer could trip on that the PR title alone will not convey: the `hello` reply's `revoked` value is now a list rather than a dict, and `LimitedSet._data`/`_heap` entries are 3-tuples. If maintainers assemble the changelog from PR titles at release time that is their call; I would just make sure both make it into the 5.7 notes.

6. Per @auvipy's comment, please push any further changes as follow-up commits rather than force-pushing; I re-reviewed the whole diff this time, but incremental diffs make the next round easier.

### Verdict

**Approve.** The fix is minimal, the analysis in the description is correct and confirmed by my own checks, and the mixed-version story is now better than `main` in both directions. Points 1 and 3 are the only ones I would actually push for before merge, and both are a few lines.

GangEunzzang (commented):
@auvipy Tested with real Redis workers (main/PR mingle in both directions and `--statedb` restarts), and the revoked ETA tasks are now discarded as expected — LGTM.

yauhpl-alt:
Thanks for the re-review and the measurements. Pushed 98c51ce as a follow-up commit (no force push — noted, @auvipy):

1. `add()` coerces `now` to a float; `test_add_at_time_zero` now includes the pickle round-trip, which raised the `ValueError` you describe on the previous commit.
2. A sentence on the `--statedb` paragraph of *Persistent revokes* in the workers guide.
3. Comments in `hello` and `send_hello` on the oldest-first order, and a line about it in the `merge_revoked()` docstring.
4. `merge_revoked()` returns on an empty argument, with tests for `None`/`[]`/`{}`.
5. Added a *Notes for the 5.7 changelog* section to the description with the `hello` list-vs-dict change and the `LimitedSet` entry arity, and retitled the PR after the symptom for the squash commit.

Unit suite unchanged otherwise (same pre-existing `test_prefork` failures on macOS as `main`), pre-commit clean.

auvipy (changes requested):
thanks for the new changes. i think it would be safer to have proper smoke tests and if possible integrations tests as well to thoroughly test the changes introduced in the worker... if the changes are not already covered in smoke and integration tests

yauhpl-alt:
@auvipy added in 512a99a:

- **Smoke**, `t/smoke/tests/test_revoked_sync.py`: a worker with `CELERY_WORKER_REVOKES_MAX=3` gets a `hello` from a "neighbour" whose stamps are a million seconds ahead of its clock; an id revoked afterwards must stay in `inspect().revoked()`, a revoked ETA task must be discarded rather than run, and the reply must be a list. And a cluster of the current source plus the last PyPI release, each restarted after the other revoked an id: both must have it after mingle, in both directions.
- **Integration**, `test_Inspect::test_hello_stamps_received_revoked_ids_locally`: the same `hello` to the session worker, whose set is in the test process — the received ids must carry local stamps.

Both fail on `main` on the bug itself (checked by swapping `celery/` for `main`'s under the running containers): the smoke test with the new id evicted, the integration one with the foreign stamp kept.

Heads-up on CI: the mixed-version test is pinned to the Redis broker because the last release cannot mingle on RabbitMQ 4.3 — `rabbitmq:latest` is 4.3.6 since 2026-09-19, and 5.6.3 still declares its pidbox reply queue transient and non-exclusive (fixed in `main` by #10290, unreleased). The existing smoke tests that run `celery_latest_worker` against RabbitMQ will break as soon as a runner pulls the current image.

auvipy:
thanks for the explanations. we will have an alpha release very soon. may be this week

auvipy (approved):

auvipy:
> Heads-up on CI: the mixed-version test is pinned to the Redis broker because the last release cannot mingle on RabbitMQ 4.3 — `rabbitmq:latest` is 4.3.6 since 2026-09-19, and 5.6.3 still declares its pidbox reply queue transient and non-exclusive (fixed in `main` by #10290, unreleased). The existing smoke tests that run `celery_latest_worker` against RabbitMQ will break as soon as a runner pulls the current image.

do you have any suggestion to change that? @yauhpl-alt

yauhpl-alt:
> > Heads-up on CI: the mixed-version test is pinned to the Redis broker because the last release cannot mingle on RabbitMQ 4.3 — `rabbitmq:latest` is 4.3.6 since 2026-09-19, and 5.6.3 still declares its pidbox reply queue transient and non-exclusive (fixed in `main` by #10290, unreleased). The existing smoke tests that run `celery_latest_worker` against RabbitMQ will break as soon as a runner pulls the current image.
> 
> do you have any suggestion to change that? @yauhpl-alt

Yes — set `control_queue_exclusive = True` and `event_queue_exclusive = True` explicitly in the smoke `default_worker_app` (`t/smoke/conftest.py`). pytest-celery builds the container's `app.py` from that app's `conf.changes`, and `celery_latest_worker` mounts the same volume, so the released worker declares exclusive queues too — what a 5.6 user on RabbitMQ 4.3 has to do anyway. The broker stays strict, and the defaults of `main` are still covered by the integration suite. Once 5.7.0 is on PyPI the two lines become no-ops. 

With that in place, the Redis pin on `test_mingle` in `t/smoke/tests/test_revoked_sync.py` can be dropped as well.

PR #10669, opened 2026-09-18
Queues.__setitem__() should apply the default routing key

phanky1:
## Description

`Queues.add()` applies the default routing key, but `Queues.__setitem__()` doesn't. I think these two operations should behave the same.

`_add()` also replaces an empty-name exchange with `self.default_exchange`, even when the default exchange is None. An empty exchange name means RabbitMQ's built-in unnamed exchange, not "no exchange" so the fallback should only be applied when self.default_exchange is not None.


celery/app/amqp.py
```diff
@@ -202,7 +202,7 @@ def deselect(self, exclude):
     def new_missing(self, name):
         queue_arguments = None
         if self.create_missing_queue_type and self.create_missing_queue_type != "classic":
-            if self.create_missing_queue_type not in ("classic", "quorum"):
```
auvipy, line 205: shouldnt we keep the classic for compatibility?

phanky1: This is just a readability refactor, it doesn't change the behavior. `classic` is still allowed because the outer if already skips this block when the queue type is classic.

auvipy:
@GangEunzzang @MehrazRumman please review

GangEunzzang (commented):
One thing: `queues['x'] = Queue('x')` now gets the default routing key but still not the default exchange, because `__setitem__` checks `not queue.exchange` and Kombu's default `Exchange('')` is truthy. Using the same `exchange is None or exchange.name == ''` check as `_add()` would make the two paths match. Otherwise looks good.

phanky1:
> One thing: `queues['x'] = Queue('x')` now gets the default routing key but still not the default exchange, because `__setitem__` checks `not queue.exchange` and Kombu's default `Exchange('')` is truthy. Using the same `exchange is None or exchange.name == ''` check as `_add()` would make the two paths match. Otherwise looks good.

I hesitated to make that change in the first place because IMO, it could be a breaking change for users who explicitly want to use an empty-name exchange. I think a change like that should be left for a later major release like 6.0.

auvipy:
I am thinking, should we put this change in v5.7 or should postpone for v5.8? may be we can take a risk to add on v5.7 alpha first...

phanky1:
> I am thinking, should we put this change in v5.7 or should postpone for v5.8? may be we can take a risk to add on v5.7 alpha first...

To clarify, I only consider the proposed empty-exchange behavior change potentially breaking. The current routing-key fix should be fine for 5.7.

auvipy:
from co pilot investigation:

The job fails because of one flaky Windows test, not dependency installation:

```text
FAILED t/unit/backends/test_asynchronous.py::test_Drainer::test_drain_catches_and_logs_oserror
AssertionError: assert False
where False = <MagicMock name='warning'>.called
```

The failing matrix entry is Python 3.14 on Windows. The `9009` messages for the Microsoft Store Python shims are non-fatal; tox successfully creates the environment and runs 4,170 tests.

## Cause

`test_Drainer::test_drain_catches_and_logs_oserror` relies on real threads and wall-clock sleeps:

- the worker thread fulfills the promise after `interval * 2`;
- `drain_events_until()` sleeps after catching `OSError`;
- the test then asserts that `logging.warning` was called.

On the Windows runner, scheduling/timing can allow the promise to become ready before the mocked failure path is observed consistently. The production handler itself is present in [`celery/backends/asynchronous.py`](https://github.com/celery/celery/blob/cc0c570b955ade5055b8b271ce4b43a86a1a2c35/celery/backends/asynchronous.py):

```python
except OSError:
    logging.warning(
        'Drainer: connection error during drain_events, '
        'will retry on next loop iteration.',
        exc_info=True,
    )
    time.sleep(interval)
```

## Recommended fix

Make the default-drainer test deterministic instead of using a real thread and real time. The lower-level test immediately above it already tests this behavior reliably using a controlled `wait` callback. Replace the integration-style test in [`t/unit/backends/test_asynchronous.py`](https://github.com/celery/celery/blob/cc0c570b955ade5055b8b271ce4b43a86a1a2c35/t/unit/backends/test_asynchronous.py) with something like:

```python
def test_drain_catches_and_logs_oserror(self, app):
    consumer = _make_consumer(app)
    drainer = consumer.drainer
    result = promise()
    calls = 0

    def wait(timeout=None):
        nonlocal calls
        calls += 1
        if calls == 1:
            raise OSError("simulated broker restart")
        result("done")

    with patch(
        "celery.backends.asynchronous.logging.warning"
    ) as mock_warn, patch(
        "celery.backends.asynchronous.time.sleep"
    ) as mock_sleep:
        list(
            drainer.drain_events_until(
                result,
                wait=wait,
                interval=0.01,
                timeout=1,
            )
        )

    assert result.ready
    mock_warn.assert_called_once()
    assert "connection error during drain_events" in mock_warn.call_args.args[0]
    mock_sleep.assert_called_once_with(0.01)
```

Alternatively, remove the duplicate `test_Drainer` integration test and retain the existing deterministic test:

```python
def test_drain_catches_oserror_and_logs(self, app):
    ...
```

That test explicitly controls the two `OSError` calls and verifies the retry behavior without relying on Windows thread scheduling.

The workflow definition confirms that this is a normal tox unit-test failure in the Python 3.14 Windows job; no workflow or dependency change is required.

auvipy (approved):

Issue #4670, opened 2018-04-19
The pubsub of redis is not thread safe, cause race condition error.


ideascf:
## Envrionment
* redis: 2.10.6
* celery: 4.0.2
* gevent: 1.0.2/1.2.1

## problem
I'm using celery for run massive task, with `CELERY_RESULT_BACKEND` was enabled.  The launch command is `celery -A server worker -P gevent -l info -E`. It work fine, but sometime got `AssertionError`. 

The stacktrace:

```
  File "/home/qfpay/agency_statistics/python/lib/python2.7/site-packages/celery/app/task.py", line 412, in delay
    return self.apply_async(args, kwargs)
  File "/home/qfpay/agency_statistics/python/lib/python2.7/site-packages/celery/app/task.py", line 535, in apply_async
    **options
  File "/home/qfpay/agency_statistics/python/lib/python2.7/site-packages/celery/app/base.py", line 736, in send_task
    self.backend.on_task_call(P, task_id)
  File "/home/qfpay/agency_statistics/python/lib/python2.7/site-packages/celery/backends/redis.py", line 180, in on_task_call
    self.result_consumer.consume_from(task_id)
  File "/home/qfpay/agency_statistics/python/lib/python2.7/site-packages/celery/backends/redis.py", line 76, in consume_from
    self._consume_from(task_id)
  File "/home/qfpay/agency_statistics/python/lib/python2.7/site-packages/celery/backends/redis.py", line 82, in _consume_from
    self._pubsub.subscribe(key)
  File "/home/qfpay/agency_statistics/python/lib/python2.7/site-packages/redis/client.py", line 2482, in subscribe
    ret_val = self.execute_command('SUBSCRIBE', *iterkeys(new_channels))
  File "/home/qfpay/agency_statistics/python/lib/python2.7/site-packages/redis/client.py", line 2404, in execute_command
    self._execute(connection, connection.send_command, *args)
  File "/home/qfpay/agency_statistics/python/lib/python2.7/site-packages/redis/client.py", line 2408, in _execute
    return command(*args)
  File "/home/qfpay/agency_statistics/python/lib/python2.7/site-packages/redis/connection.py", line 610, in send_command
    self.send_packed_command(self.pack_command(*args))
  File "/home/qfpay/agency_statistics/python/lib/python2.7/site-packages/redis/connection.py", line 585, in send_packed_command
    self.connect()
[...]

PR #10671, opened 2026-09-19
Redis result backend: serialize pubsub access between the drainer and result cleanup

goingforstudying-ctrl:
## Description

Got bitten by the #4670 race again on 5.5.x: `AsyncResult.__del__` runs on whatever thread the GC happens to pick, and it walks into `remove_pending_result` -> `cancel_for` -> `pubsub.unsubscribe()` while the result drainer is sitting in `get_message()` on the very same pubsub object. redis-py pubsub isn't safe for concurrent use, and with gevent this blows up as `ConcurrentObjectUseError: This socket is already used by another greenlet` (AssertionError on older versions). The #4666 / #6416 fixes added the unsubscribe but never serialized it against the reader, so the race just moved around.

What I did: gave `ResultConsumer` an RLock and took it around every touch of the shared pubsub — start/stop, subscribe, unsubscribe, the `get_message` poll in `drain_events`, reconnect, and the fork cleanup. RLock because the drainer itself calls `cancel_for` from `on_state_change` when a ready meta comes in, and the reconnect path re-enters from the error handler. The idle sleep when there's no pubsub yet stays outside the lock so `apply_async` isn't stalled by nothing.

One thing I wasn't 100% sure about: fork safety. The child inherits the lock possibly still held by a thread that didn't survive the fork, so `on_after_fork` swaps in a fresh one before closing the old pubsub.

There's a real trade-off here worth spelling out: while the drainer is mid-poll (up to its 1s timeout), a subscribe/unsubscribe from another thread has to wait. I think that's strictly better than crashing, and in the pure fire-and-forget producer case the drainer isn't running so there's no contention at all. If maintainers would rather have the drainer own the pubsub exclusively via a command queue, that's a bigger redesign — figured I'd start here.

Tests: the main one is a race test that hammers subscribe/unsubscribe from 4 threads while a drainer thread polls; the fake pubsub records any overlapping calls. It fails immediately on main (`assert not ['subscribe', 'subscribe', 'get_message', 'unsubscribe', ...]`) and is clean with the lock. Plus deterministic checks that each pubsub op holds the lock, that the idle sleep doesn't, and that `on_after_fork` replaces the lock. `t/unit/backends/test_redis.py` + `test_asynchronous.py` pass locally (167 passed, 9 skipped), flake8 clean.


auvipy:
can you also check this pr https://github.com/celery/celery/pull/10326 ? and is it related? or fixing a different problem?

auvipy (changes requested):
it would be great to have additional smoke or integration tests for this change

goingforstudying-ctrl:
Checked #10326. Same root cause (the shared pubsub from #4670), different interleaving. That one is same-thread: cyclic GC fires AsyncResult.__del__ mid-subscribe and the nested cancel_for re-enters the same connection. This one is cross-thread: __del__ runs on the GC thread and calls unsubscribe while the drainer is blocked in get_message on the same pubsub. The RLock here can't fix theirs (a re-entrant acquire succeeds and the writes still interleave), and their depth counter can't fix this one, so both are needed.

Added integration tests in c2e107a (t/integration/test_redis_backend.py): concurrent AsyncResult.get() round-trips through a worker on the shared pubsub, plus subscribe/unsubscribe churn on the ResultConsumer while the drainer polls. Asserts no errors and no leaked channels.

auvipy:
@GangEunzzang @MehrazRumman can you try and verify this please?

GangEunzzang (commented):
Tried it against a real Redis with a drainer thread and four threads subscribing/unsubscribing: under plain threads neither tree errors, but under gevent `main` throws `ConcurrentObjectUseError` and redis protocol `ValueError`s within seconds (11 to 392 per 4s run) and this branch stays at zero. The unit race test fails on `main` and passes here. LGTM.

auvipy (approved):

phanky1:
Hi @goingforstudying-ctrl @auvipy @GangEunzzang, I run an AI to review this, and it shows me a quite serious regression issue, that is:

> When using gevent with Redis as the result backend, the new lock in `drain_events()` might keep getting reacquired by the drainer, so other greenlets can’t get a turn. This could make `result.get(timeout=...)` hang even after the result is received, without respecting the timeout.


I haven't spent time to look into that, so I don't really know if this is real or not. Can you take a quick look?

auvipy:
@goingforstudying-ctrl can you please cross check the above comment?

GangEunzzang:
@phanky1 Taking a look at this now.

GangEunzzang:
@phanky1 It's real, and it's worse than a delay. With a real worker and Redis, a gevent-patched client calling `result.get()` on a task that already finished never returns on `main` (waited 25s+, six runs), while the same call takes 0.07s on the commit before this PR and 0.1s on `main` without gevent. The greenlet dump shows the caller stuck in `consume_from` waiting for `_pubsub_lock`, and the drainer holding it inside `pubsub.get_message(timeout=1)`; gevent's semaphore doesn't hand off on release, so the drainer re-takes the lock every loop before the waiter runs.

Yielding after the release only moves the problem: `AsyncResult.__del__` → `cancel_for` then blocks on the lock inside a finalizer and gevent raises `RuntimeError: greenlet is being finalized`. So the blocking `get_message()` can't sit under the same lock that `cancel_for` takes from `__del__`. My LGTM here was wrong: the stress test I ran counted errors, and the starved threads simply did nothing.

@auvipy this is on `main` now, so probably worth a look before the alpha. Happy to open a PR for it (keep the lock for the pubsub calls, but wait for messages outside it, and make `cancel_for` non-blocking from `__del__`) if you'd rather not wait for the author.

PR #10672, opened 2026-09-19
Store failures raised as BaseException subclasses with the JSON serializer

GangEunzzang:
`Backend.encode_result()` only runs `prepare_exception()` when the result is an `Exception`, so a task that fails with a bare `BaseException` or `asyncio.CancelledError` reaches the JSON encoder as the raw object and `mark_as_failure()` dies with `EncodeError: Object of type BaseException is not JSON serializable`. No result is written and the caller waits for its own timeout. Pickle is unaffected since it serializes the object as is.

This came up while looking at Billiard 4.3.0 in #10659: the worker loop now reports any `BaseException` a task raises instead of turning it into `WorkerLostError`. `SystemExit` and `KeyboardInterrupt` are handled in #10667; this covers the rest.

Checking `BaseException` instead is enough, `exception_to_python()` already accepts them on the way back. With a prefork worker against Redis, both cases now come back as the raised exception in well under a second where they used to time out.


t/unit/backends/test_base.py
```diff
@@ -1,3 +1,4 @@
+import asyncio
```
auvipy, line 1: why asyncio?

GangEunzzang: `asyncio.CancelledError` is a `BaseException` subclass, so it hit the same `EncodeError`; it's the real-world case from #10659.

auvipy: ok

auvipy (approved):

PR #10673, opened 2026-09-20
Refresh AMQP task routes for all configuration update forms

phanky1:
## Description
`AMQP._handle_conf_update()` only recognizes `task_routes` when `app.conf.update()` is called with keyword arguments. When a mapping or iterable of pairs is passed instead, the cached routes and router are not refreshed.

https://github.com/celery/celery/blob/0ec71f929742938cbcbe996b74cabf5b6cd54655/celery/app/amqp.py#L660-L664

For example:
`app.conf.update({'task_routes': routes})`
Here, args is:
```python
({'task_routes': routes},)
```
  It is a tuple containing the mapping, so `'task_routes' in args` is False.


auvipy:
@GangEunzzang

GangEunzzang (commented):
Tried it out: on `main`, `app.conf.update({'task_routes': ...})` leaves the old router in place and tasks still land on the default queue, and this fixes that for mappings, pairs and the legacy `CELERY_ROUTES` key alike. LGTM.

auvipy (approved):

PR #10676, opened 2026-09-21
Add a smoke test workflow that runs against py-amqp, Kombu and Billiard main

### LINKED ISSUE #10675: Run the smoke tests against Billiard/Kombu main on a schedule
Billiard 4.3.0 shipped two behavior changes that broke Celery's smoke tests (#10659, fixed in #10667 and #10672). The commit behind them, celery/billiard#452, was merged on August 16; Celery only saw it on September 17 when the release landed on `main`.

That gap exists because nothing in Celery's CI runs against the Git `main` of Billiard, Kombu or py-amqp. `requirements/dev.txt` has those, but it is only used to build the dev Docker image. The unit tests mock Billiard, so the smoke tests are the ones that catch this kind of change. Kombu already does the equivalent for its own upstreams: its default tox env installs py-amqp `main` and vine `master` from `requirements/dev.txt`.

Proposal: a scheduled workflow (weekly, plus `workflow_dispatch`) that runs `t/smoke/tests/test_worker.py` and `t/smoke/tests/test_tasks.py` with `requirements/dev.txt` installed both on the host and in the smoke worker image. The image side needs a build arg in `t/smoke/workers/docker/dev` and a `buildargs()` override in `SmokeWorkerContainer`; the workflow side is an `upstream_deps` input on `smoke-tests.yml`. Nothing changes for regular PR runs. Running it on a schedule rather than on every PR keeps an upstream breakage from turning every Celery PR red.

I can put the PR together if this sounds useful.

GangEunzzang:
Closes #10675.

Adds `.github/workflows/smoke-tests-dev.yml`, a separate workflow on the same push/PR triggers as the main one. It calls the reusable `smoke-tests.yml` with a new `dev_requirements` input, which selects the `smoke-dev` tox env: the `smoke` env plus `requirements/dev.txt`, so the host gets py-amqp, Kombu and Billiard from Git `main`. That env also sets `CELERY_DEV_REQUIREMENTS=1`, which `SmokeWorkerContainer.buildargs()` passes to the worker image, where a new build arg installs `dev.txt` on top of the editable Celery install. Without the flag the image build is unchanged, and `python-package.yml` is untouched.

The job covers `test_worker.py` and `test_tasks.py` (the two files that caught the Billiard 4.3.0 changes in #10659) on one Python version, so it is two extra jobs rather than another full matrix.

Checked locally on macOS with Docker: the `3.12-smoke-dev` tox env ends up with py-amqp, Kombu and Billiard from Git (`direct_url.json` points at the three repos), the worker image does too with the build arg and stays on PyPI without it, and both files pass in that mode (`test_tasks.py` 22 passed, `test_worker.py` 62 passed / 6 skipped, same as with the released versions).

Two things worth knowing. The image install needs `--force-reinstall`: pip clones the three repos but keeps the PyPI copy when the checkout reports the same version string. And `default.txt` pins `billiard==4.3.0`, so once Billiard `main` bumps its version the dev job will fail at install time until the pin moves, which is the kind of thing this job exists to surface.


tox.ini
```diff
@@ -5,6 +5,7 @@ envlist =
     {3.11,3.12,3.13,3.14,3.15,pypy3}-unit
     {3.11,3.12,3.13,3.14,3.15,pypy3}-integration-{rabbitmq_redis,rabbitmq,redis,dynamodb,azureblockblob,cache,cassandra,elasticsearch,docker}
     {3.11,3.12,3.13,3.14,3.15,ypy3}-smoke
+    {3.11,3.12,3.13,3.14,3.15,pypy3}-smoke-dev
```
auvipy, line 8: should we run all python version in dev? or the latest 2 should be fine?

auvipy (changes requested):

tox.ini
GangEunzzang: That tox line just defines the env for every version, the workflow only picked 3.13. Latest two sounds right, switched it to 3.14 and 3.15.

auvipy:
i can not see the jobs, can you please point me to them?

GangEunzzang:
They run as a separate workflow, "Smoke Tests (dev requirements)": https://github.com/celery/celery/actions/runs/35589666299 — four jobs, `test_tasks.py` and `test_worker.py` on 3.14 and 3.15. The `test_tasks.py` ones have already passed, the `test_worker.py` ones are still running. They also show up under the PR's checks as `Smoke-tests-dev (...)`.

auvipy:
things are passing here but the kombu 5.7.0a1 compatibility check pr got some failures here https://github.com/celery/celery/pull/10677

auvipy (approved):

PR #10677, opened 2026-09-21
Update kombu version to 5.7.0a1

auvipy:


auvipy:
The job is failing consistently in the Redis-backend smoke test for the **group-body chord** case:

- The header contains one successful task and one failing task.
- Redis correctly detects the header failure and creates a `ChordError`.
- Error handling then revokes the two tasks belonging to the group body.
- `result.get()` observes `TaskRevokedError: revoked`, while the test expects the original `ExpectedException`.

The failure is therefore in chord error propagation for a group callback, not in Redis startup or the GitHub Actions configuration.

Relevant code:

- Test: [`t/smoke/tests/test_canvas.py:155-192`](https://github.com/celery/celery/blob/230a311104598ab69dd42d829e80fffeeae58e92/t/smoke/tests/test_canvas.py#L155-L192)
- Redis chord handling: [`celery/backends/redis.py:692-716`](https://github.com/celery/celery/blob/230a311104598ab69dd42d829e80fffeeae58e92/celery/backends/redis.py#L692-L716)

## Recommended fix

Update the common chord-error handler, rather than the Redis backend, so a group callback is failed with the original chord exception instead of being revoked.

The relevant logic is called here:

```python
except ChordError as exc:
    logger.exception('Chord %r raised: %r', request.group, exc)
    return self.chord_error_from_stack(callback, exc)
```

`chord_error_from_stack()` should special-case a group callback:

```python
def chord_error_from_stack(self, callback, exc=None):
    callback = maybe_signature(callback, app=self.app)

    # A group callback has no single task result to fail. Propagate the
    # chord failure to each child result instead of revoking the children.
    if isinstance(callback, group):
        for child in callback.tasks:
            child = maybe_signature(child, app=self.app)

            if child.id is None:
                child.freeze()

            self.mark_as_failure(
                child.id,
                exc,
                request=Context(
                    id=child.id,
                    group=callback.options.get("group"),
                ),
            )
        return

    # Existing handling for single-task, chain, and chord callbacks.
    return self._chord_error_from_stack(callback, exc)
```

Use the project’s existing request/context construction and error serialization helpers rather than introducing a new `Context` shape if `base.py` already has an internal helper for this. The important behavior is:

1. Preserve the original exception as the failure result.
2. Assign IDs to group-body signatures before storing results.
3. Do not revoke group-body result IDs when the header has failed.
4. Leave the existing single-task, chain, and nested-chord handling unchanged.

If the current implementation deliberately revokes body tasks to prevent execution, the equivalent safer implementation is to revoke them only after storing a `FAILURE` result for each child. Otherwise `GroupResult.get()` will always surface `TaskRevokedError` instead of the header exception.

Add a regression assertion covering the failing body type:

```python
with pytest.raises(ExpectedException, match="Task expected to fail"):
    result.get(timeout=RESULT_TIMEOUT)
```

The existing test already covers this scenario; once group callback error propagation stores the original failure instead of producing revoked child results, the `3.13-smoke` job should pass without changing the workflow or increasing retries.

auvipy:
@phanky1 is this failure related to your finding in a redis related merged pr here https://github.com/celery/celery/pull/10671? or it is unrelated

auvipy:
@GangEunzzang do you have time working on the fix? or investigate the kombu compatibility issues?

GangEunzzang:
@auvipy Yes, looking into it now.

GangEunzzang:
@auvipy Looked into the `body is a group` failure. It isn't a Kombu incompatibility as such: the same test fails for me locally with Kombu 5.6.2 too, and the worker log shows why. On a header failure `_handle_group_chord_error` revokes the body task ids and then stores `FAILURE` for each of them, but since #9869 the worker's revoke handler also writes `REVOKED` to the backend for any id it isn't running, and it does that ~15ms later from the main process. Whichever write lands last wins, so the client sees `TaskRevokedError` instead of the header's exception. Kombu 5.7.0a1 just shifts the broadcast timing enough for that to lose on CI.

Skipping the `REVOKED` write in `_revoke` when the backend already holds a ready state fixes it here: the failing case passes 3/3 with 5.7.0a1 (it failed every time before), the other body types still pass, and `t/unit/worker` is unchanged. I have that ready as a PR against `main` if you agree; the version bump itself looks fine.

auvipy:
please come with the pr

auvipy:
intentionally merging this. lets fix the CI failures in another pr

GangEunzzang:
Opened as #10679.

PR #10679, opened 2026-09-21
Keep a finished result when a revoke arrives for the task

GangEunzzang:
Since #9869 the revoke command writes `REVOKED` to the result backend for every id the worker isn't currently running. When a chord header fails and the body is a group, `_handle_group_chord_error()` revokes the body ids and stores `FAILURE` for each of them; the revoke is a broadcast, so the worker's main process writes `REVOKED` a few milliseconds later and replaces that `FAILURE`. Whichever write lands last wins, and callers get `TaskRevokedError` instead of the header's exception. This is what makes `test_chord_error_propagation_with_different_body_types[body is a group]` fail on #10677: Kombu 5.7.0a1 only changes the broadcast timing enough for the wrong write to land last on CI.

Skip the `REVOKED` write when the backend already holds a ready state for the task, and store the chord failures before broadcasting the revoke so that check always sees them. Pending tasks, the case #9869 was about, are still marked. The check is one extra backend read per revoked id in the control command.

Checked with Kombu 5.7.0a1 pinned: the `body is a group` case passes 3 out of 3 runs where it failed every time before, all four body types pass on both brokers, and `t/unit/worker` is unchanged apart from the new test.

Fixes the smoke failure that came with the Kombu 5.7.0a1 bump (#10677).


auvipy:
there is also consumers test failures beside this one

auvipy:
my AI analysis:

The job fails because the smoke test expects prefetch-reduction logs, but the worker correctly skips prefetch reduction when the broker uses per-consumer QoS.

The decisive log is:

```text
Skipping prefetch count reduction after connection restart because
per-consumer QoS (apply_global=False) is in effect
```

This branch is implemented in [`Consumer.on_connection_error_after_connected`](https://github.com/celery/celery/blob/2a8f3ad23d1d7357a638184cdb6d681b03a8b6f8/celery/worker/consumer/consumer.py#L451-L490). When `qos_global is False`, Celery does not emit either of the messages expected by the test:

- `Temporarily reducing the prefetch count...`
- `Resuming normal operations following a restart...`

The failing tests unconditionally expect those messages in [`t/smoke/tests/test_consumer.py`](https://github.com/celery/celery/blob/2a8f3ad23d1d7357a638184cdb6d681b03a8b6f8/t/smoke/tests/test_consumer.py#L32-L75). The failures are therefore caused by a test/broker QoS mismatch, not by the connection retry itself.

## Recommended fix

Run these prefetch-reduction tests with a classic RabbitMQ queue, where `basic.qos` updates apply globally. Set the queue type explicitly in the test app fixture:

```python
@pytest.fixture
def default_worker_app(default_worker_app: Celery) -> Celery:
    app = default_worker_app
    app.conf.worker_prefetch_multiplier = WORKER_PREFETCH_MULTIPLIER
    app.conf.worker_concurrency = WORKER_CONCURRENCY
    app.conf.worker_enable_prefetch_count_reduction = True

    # These tests verify global basic.qos prefetch reduction.
    # Quorum/per-consumer QoS cannot exercise that behavior.
    app.conf.task_default_queue_type = "classic"

    app.conf.visibility_timeout = 3600
    if app.conf.broker_url.startswith("redis"):
        app.conf.broker_transport_options = {
            "visibility_timeout": app.conf.visibility_timeout,
            "polling_interval": 1,
        }
    if app.conf.result_backend.startswith("redis"):
        app.conf.result_backend_transport_options = {
            "visibility_timeout": app.conf.visibility_timeout,
            "polling_interval": 1,
        }
    return app
```

If the pytest-Celery worker queue is declared independently of `task_default_queue_type`, declare that queue explicitly as classic in the smoke-test broker setup:

```python
from kombu import Queue

app.conf.task_queues = (
    Queue(
        "celery",
        queue_arguments={"x-queue-type": "classic"},
    ),
)
```

Use the actual worker queue name if pytest-Celery generates a dedicated queue.

## Alternative test correction

If the intended broker configuration is quorum queues, the tests should validate the new supported behavior instead of expecting reduction/restoration logs:

```python
celery_setup.broker.restart()

celery_setup.worker.assert_log_exists(
    "Skipping prefetch count reduction after connection restart "
    "because per-consumer QoS (apply_global=False) is in effect"
)
```

The existing assertions should then be conditional on queue QoS mode. Do not remove the `qos_global is False` branch from `consumer.py`; removing it would reintroduce a stale reduced prefetch count after reconnect, exactly the condition the code is designed to avoid.

The workflow itself merely dispatches the smoke suite through [`python-package.yml`](https://github.com/celery/celery/blob/2a8f3ad23d1d7357a638184cdb6d681b03a8b6f8/.github/workflows/python-package.yml#L129-L138); the fix belongs in the smoke-test broker/queue configuration or in the test expectations.

auvipy:
should we fix the consumer test failure in a separate pr? or you prefer doin it here?

GangEunzzang:
@auvipy This one comes from Kombu: celery/kombu#2481 makes `qos_semantics_matches_spec()` return per-channel QoS for every RabbitMQ 4.x, so the worker skips the prefetch reduction after a reconnect. Global QoS still works on classic queues on RabbitMQ 4 (only quorum queues reject it), and these tests run on classic queues. The fix is to take that `>= 4.0` branch back out in Kombu. OK if I open a separate PR there for it?

auvipy:
sure please! no prob!

GangEunzzang:
Opened celery/kombu#2665 for the Kombu side.

GangEunzzang:
@auvipy Follow-up on the consumer failures: I had this backwards. RabbitMQ 4.3 denies `global_qos` by default and silently turns `basic.qos(global=true)` into per-consumer QoS (details in celery/kombu#2665), so Kombu's per-consumer default for 4.x is correct and I've closed that PR. What the red tests show is real: on RabbitMQ 4.3+ the prefetch-count reduction can't work through global QoS any more. Two ways to go: adapt the tests to expect the "Skipping prefetch count reduction" path on RabbitMQ 4, as your analysis suggested, or reimplement the reduction for per-consumer QoS (cancel and re-consume with the new prefetch). Which would you prefer? I'll open a separate PR for it either way.

auvipy:
I’d prefer adapting the consumer tests in a separate PR to expect the “Skipping prefetch count reduction” path on RabbitMQ 4.3+.

Reimplementing reduction for per-consumer QoS by canceling and re-consuming would be considerably more invasive and risky:

    it could affect delivery ordering and in-flight messages;
    cancel/re-consume during reconnect introduces additional race conditions;
    it would duplicate broker-specific QoS behavior that RabbitMQ 4.3 intentionally enforces;
    the current skip behavior avoids restoring a stale reduced count, so it is the safer default.

The separate PR should make the expectations conditional on the broker/QoS mode and retain coverage for global QoS where it is still supported. I’d also keep this PR focused on preserving finished chord results and avoid coupling it to the consumer-test changes.

auvipy:
although right now Im not fully sure which approach will be best in long term

auvipy (commented):
can you check if anything else can be done here? or can we do anything to harden the tests? going to merge this separately

GangEunzzang:
@auvipy Pushed one more test-only commit here (63d65d1): the guard is now checked for every READY and UNREADY state plus a failing state lookup, and there's a test that `_handle_group_chord_error()` stores all the failures before it broadcasts the revoke, since the guard depends on that order. Without the fix, the three ready-state cases and the ordering test fail. The integration case that first exposed this (`test_chord_error_propagation_with_different_body_types[body is a group]`) fails on main and passes here, so it stays as the end-to-end check. Nothing else I'd change in this PR. The consumer smoke tests are in #10682.

auvipy (approved):

PR #10680, opened 2026-09-21
fix(backends/couchbase): missing document raises instead of returning None

simpleqt:
## Description

The `KeyValueStoreBackend` contract across celery is: `get()` returns `None` for an absent key, and `delete()`/forgetting an absent key is a no-op — that's how redis, memcached, mongodb, filesystem, etc. behave, and it's what `get_task_meta()`/`AsyncResult` rely on to render an expired or unknown task id as a state rather than crashing.

`CouchbaseBackend` violates both halves:

- `get()` does `self.connection.get(key).content` with no guard. When the document has expired (auto-expiry is on: `supports_autoexpire = True`) or was never stored, the SDK raises `DocumentNotFoundException`, which propagates out of `get_task_meta()` — so `AsyncResult(id).state` on a purged id raises `DocumentNotFoundException` instead of reading as unknown/PENDING.
- `delete()` does `self.connection.remove(key)` with no guard, so `AsyncResult.forget()` on an already-expired or already-forgotten result raises `DocumentNotFoundException`.

Reproduce (against a reachable Couchbase with an empty/default collection):

```python
from celery import Celery, uuid
app = Celery('t', backend='couchbase://localhost/default')
app.AsyncResult(uuid()).state          # DocumentNotFoundException, not 'PENDING'
app.AsyncResult(uuid()).forget()       # DocumentNotFoundException
```

## What does this fix

Bug fix — `get()` now catches `DocumentNotFoundException` and returns `None`; `delete()` catches it and returns. Two mock-based regression tests added to `t/unit/backends/test_couchbase.py` (`test_get_missing_document_returns_None`, `test_delete_missing_document_is_noop`), following the file's existing mock-the-connection pattern.

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)

## Verification

- With the fix: `pytest t/unit/backends/test_couchbase.py` → 17 passed (couchbase SDK 4.6.3).
- Fix reverted, tests kept: the 2 new tests fail, the other 15 pass — they fail for exactly this behavior, not incidentally.
- Repo-pinned `pre-commit run --files celery/backends/couchbase.py t/unit/backends/test_couchbase.py` (flake8, isort, mypy, yesqa, codespell, pyupgrade): all pass.


simpleqt:
Heads-up on the red `codecov/patch`: the unit-test CI job doesn't install the `couchbase` extra, so `t/unit/backends/test_couchbase.py` is `importorskip`-ed in CI and none of the backend's lines (new or pre-existing) get coverage there — the file already had 0% before this change. The two new regression tests do exercise both new branches and pass locally with the SDK installed (couchbase 4.6.3): 17 passed; with the source fix stashed they fail as expected. Happy to restructure the tests to be coverage-visible if maintainers prefer.

MehrazRumman (changes requested):
## Review

Correct fix, narrowly scoped, and the verification is exactly what I'd want to see — in particular running the new tests with the fix reverted to show they fail for this reason and not incidentally. The contract claim checks out: `KeyValueStoreBackend._get_task_meta_for` treats a falsy `get()` as `PENDING`, so returning `None` is what makes `AsyncResult(id).state` read correctly, and catching only `DocumentNotFoundException` (rather than a broad SDK error) is the right width.

A few things before merge, one of them code.

---

### 1. The import fallback should be a placeholder class, in its own `try`

```python
try:
    from couchbase.auth import PasswordAuthenticator
    from couchbase.cluster import Cluster
    from couchbase.exceptions import DocumentNotFoundException
except ImportError:
    Cluster = PasswordAuthenticator = None
    DocumentNotFoundException = None
```

Two problems with this shape:

* **Coupling.** Adding the exception to the same `try` means that if that one name ever fails to import — an SDK release that moves or renames it — `Cluster` becomes `None` too, and the backend refuses to start with "You need to install the couchbase library", which is untrue and sends users looking in the wrong place. Within the supported `couchbase>=3` range the name exists, so this is about robustness rather than a current bug, but it's cheap to avoid.
* **`except None:` is a `TypeError`** if it's ever evaluated. It's unreachable today because `__init__` raises when `Cluster is None`, but that's an invariant in a different method keeping this one safe.

The MongoDB backend already has the pattern for this — a stand-in exception class when the driver isn't importable:

```python
try:
    from couchbase.exceptions import DocumentNotFoundException
except ImportError:  # pragma: no cover
    class DocumentNotFoundException(Exception):
        """Stand-in so ``except`` clauses stay valid without the SDK."""
```

Every `except DocumentNotFoundException:` is then always valid, and a missing name can't disable the whole backend.

### 2. `mget()` is the third leg of the same contract

The description says the backend violates "both halves" of the contract, but `mget()` is part of it too — it's what `get_many()` and native group joins go through, and it has to cope with absent keys as well:

```python
def mget(self, keys):
    return self.connection.get_multi(keys)
```

On the 4.x SDK, `get_multi` returns a `MultiGetResult` rather than a list of values, with per-key failures collected on the result object instead of raised. I'd be surprised if `_mget_to_results` handles that shape correctly for missing keys — or for present ones. Out of scope for this PR, but worth checking, and worth an issue if it's broken, since `GroupResult.get()` on this backend would go through it.

### 3. Tests

The two new tests are fine and follow the file's existing pattern. Three suggestions:

* **They never run in CI.** As you noted, the unit job doesn't install the `couchbase` extra, so the whole module is `importorskip`-ed and the backend sits at 0% CI coverage — before and after this PR. That's not something to fix here, but it's worth maintainers knowing: it's how a contract violation this basic survived, and nothing stops it regressing. Either installing the extra in the unit job or patching `Cluster` so the module runs without the SDK would do; the second is a bigger restructure than this PR should carry.
* **Test the contract, not just the method.** The user-visible claim is that `AsyncResult(id).state` reads `PENDING` and `forget()` doesn't raise. One test going through `backend.get_task_meta(...)` with the mocked `get` raising — asserting `status == PENDING` — pins the actual behaviour the PR is about, and would catch a future change in `_get_task_meta_for` that stops treating `None` as pending.
* **Pin the width of the catch.** A test that a different SDK error (a timeout, say) still propagates from `get()` stops a future "while I'm here" change from widening the `except` and silently turning outages into `PENDING`.

### 4. Worth mentioning in the description and changelog

The fix reaches further than `state` and `forget()`:

* `GroupResult.restore()` on an expired group goes through `get()` via `_restore_group`, so it now returns `None` instead of raising.
* If the key-value implementation of `task_result_exists()` goes through `get()`, this also fixes `AsyncResult.exists()` — new in 5.7.0, and whose entire purpose is reporting the never-stored case that currently raises on this backend. If so, that's the strongest argument for getting this into 5.7.x.

### Nits

* The description lists MongoDB among the backends following the `KeyValueStoreBackend` contract; the Mongo backend isn't a key-value backend, though it does report missing results as `PENDING`. The behavioural point stands; the wording would just be off in a changelog.
* `test_get_missing_document_returns_None` could also assert `get.assert_called_once_with('1f3fab')`, matching `test_get` right above it.

---

**Verdict:** approve after point 1. Points 2–4 are follow-ups or description changes, not blockers — though the contract-level test in point 3 is cheap and I'd add it here.

simpleqt:
Thanks for the thorough review — every point addressed on 66a7a13:

**1. Import fallback** — done as suggested: `DocumentNotFoundException` now lives in its own `try` with a stand-in exception class (mirroring `mongodb.py`), so a future SDK rename can't turn `Cluster` into `None`, and no `except None` can exist.

**2. `mget()`** — checked, and it's worse than "surprising": on 4.6.3, `MultiGetResult` has no `.items()` and `MultiResult` defines no `__iter__`, so `_mget_to_results()` raises `TypeError` — `get_many()`/group restore simply crash on SDK 4. Filed #10694 with the shape analysis and a suggested translation (per-key `DocumentNotFoundException` → `None`, other per-key errors re-raised), and flagged it as out of scope in the PR description.

**3. Tests** — added both:
- `test_get_task_meta_pending_for_missing_document` goes through `backend.get_task_meta()` and asserts `status == PENDING`, pinning the contract rather than the method;
- `test_get_unrelated_sdk_errors_propagate` feeds a `TimeoutException` and asserts it still propagates, pinning the catch width.
Red/green re-verified: with the fix reverted, the missing-document tests fail with `DocumentNotFoundException` propagating through `get()`/`get_task_meta()`; with the fix, 19/19 pass.

On CI coverage: agreed, and I've left it as you framed it — fixing it properly (extra in the unit job, or SDK-free restructure) is bigger than this PR, so it's for maintainers to weigh.

**4. Scope** — description now records both extra beneficiaries (`GroupResult.restore()` via `_restore_group()`, and `AsyncResult.exists()` via `task_result_exists()` = `bool(self.get(...))`), plus the MongoDB wording fix.

Nits: `assert_called_once_with('1f3fab')` added to `test_get_missing_document_returns_None`.

auvipy:
i think we should fix the test skip of couchbase in this pr as well...

simpleqt:
@auvipy both done on 988d372:

**Test skip** — `requirements/test-ci-default.txt` re-enables the couchbase install for the unit-test job. It was commented out because the SDK ships no wheels for part of the matrix, so it is re-enabled with markers that exclude exactly those cells: PyPy (no wheels), CPython 3.15 (no cp315 wheels; sdist needs libcouchbase), and the Windows job (which only runs 3.14, also no wheel). On ubuntu 3.11–3.14 the SDK now installs and `t/unit/backends/test_couchbase.py` runs instead of being `importorskip`-ed.

**mget()** — fixed as discussed in #10694 (`Fixes #10694`): `GetMultiOptions(return_exceptions=True)` + per-key translation into the plain dict `_mget_to_results` expects (missing document → `None`, other per-key errors re-raised). Five regression tests added including an end-to-end `get_many()` with one ready and one missing document. Red/green verified: with `mget()` reverted all five fail (`TypeError` through `_mget_to_results`); with the fix `pytest t/unit/backends/test_couchbase.py` → 24 passed (SDK 4.6.3). Repo-pinned flake8/isort green.

auvipy:
thats really great!

auvipy (approved):

Issue #9512, opened 2025-01-23
Worker loss on reconnection with RabbitMQ with QoS disabled


Osmoskour:
# Checklist

- [x] I have verified that the issue exists against the `main` branch of Celery.
- [ ] This has already been asked to the [discussions forum](https://github.com/celery/celery/discussions) first.
- [x] I have read the relevant section in the
  [contribution guide](https://docs.celeryq.dev/en/main/contributing.html#other-bugs)
  on reporting bugs.
- [x] I have checked the [issues list](https://github.com/celery/celery/issues?q=is%3Aissue+label%3A%22Issue+Type%3A+Bug+Report%22+-label%3A%22Category%3A+Documentation%22)
  for similar or identical bug reports.
- [x] I have checked the [pull requests list](https://github.com/celery/celery/pulls?q=is%3Apr+label%3A%22PR+Type%3A+Bugfix%22+-label%3A%22Category%3A+Documentation%22)
  for existing proposed fixes.
- [x] I have checked the [commit log](https://github.com/celery/celery/commits/main)
  to find out if the bug was already fixed in the main branch.
- [x] I have included all related issues and possible duplicate issues
  in this issue (If there are none, check this box anyway).
- [   I have tried to reproduce the issue with [pytest-celery](https://docs.celeryq.dev/projects/pytest-celery/en/latest/userguide/celery-bug-report.html) and added the reproduction script below.

## Mandatory Debugging Information

- [x] I have included the output of ``celery -A proj report`` in the issue.
    (if you are not able to do this, then at least specify the Celery
     version affected).
- [x] I have verified that the issue exists against the `main` branch of Celery.
- [x] I have included the contents of ``pip freeze`` in the issue.
- [x] I have included all the versions of all the external dependencies required
  to reproduce this bug.

## Optional Debugging Information

- [ ] I have tried reproducing the issue on more than one Python version
  and/or implementation.
- [ ] I have tried reproducing the issue on more than one message broker and/or
  result backend.
- [ ] I have tried reproducing the issue on more than one version of the message
  broker and/or result backend.
- [ ] I have tried reproducing the issue on more than one operating system.
- [ ] I have tried reproducing the issue on more than one workers pool.
- [ ] I have tried reproducing the issue with autoscaling, retries,
  ETA/Countdown & rate limits disabled.
- [ ] I have tried reproducing the issue after downgrading
  and/or upgrading Celery and its dependencies.

#### Related Issues

- None

#### Possible Duplicates

- None

[...]

PR #10682, opened 2026-09-21
Expect the prefetch reduction to be skipped under per-consumer QoS

GangEunzzang:
Since the Kombu 5.7.0a1 bump (#10677), `t/smoke/tests/test_consumer.py` fails on `rabbitmq:latest` in six cases: `test_reducing_prefetch_count[rabbitmq-1..5]` and `test_prefetch_count_restored[rabbitmq]`. Kombu 5.7 reports per-consumer QoS for RabbitMQ 4 (celery/kombu#2481), and in that mode the worker skips the prefetch-count reduction after a connection restart (#9512), so the "Temporarily reducing…" and "restored to the maximum" log lines the tests wait for never appear.

This makes the expectation conditional on the QoS mode, decided the same way `Tasks.qos_global()` decides it (`not connection.qos_semantics_matches_spec`): under per-consumer QoS the tests expect the "Skipping prefetch count reduction after connection restart" log; under global QoS the existing assertions run unchanged.

Verified locally against `rabbitmq:latest` (4.3.5): the whole module passes with Kombu 5.7.0a1 (10 passed, the 8 pre-existing Redis xfails unchanged), the six cases going through the skip branch, and `test_reducing_prefetch_count[rabbitmq-2]` and `test_prefetch_count_restored[rabbitmq]` still pass through the global branch on a tree pinned to Kombu 5.6.2.

Background in #10679; this keeps that PR on the revoke fix only, as requested.


auvipy:
just curious why the chord related tests are also green here? are the related?

GangEunzzang:
@auvipy Not related, this only touches test_consumer.py. The chord failure is a race between the chord error handler's FAILURE write and the revoke broadcast's REVOKED write, so it doesn't show up every run: it hit 1 of the 20 test_canvas jobs on #10677's run, 0 of 20 on main after the merge, and 0 of 20 here, while locally it failed 5 out of 5 runs. #10679 removes the race.

auvipy:
OK cool

auvipy (approved):

PR #10689, opened 2026-09-22
Preserve propagate in nested native result joins

GangEunzzang:
With native joins, `GroupResult.get(propagate=False)` still raises when a task inside a nested group fails. `join_native()` calls each child's `get()` without passing `propagate`, so it falls back to the default of `True`.

This PR forwards `propagate` to child results and adds regression tests for nested failures with propagation both enabled and disabled.

I also looked into the dropped `timeout` argument, but simply passing it to each child would give each one a fresh timeout rather than enforcing a timeout for the whole join.

One option would be to use a shared deadline and pass the remaining time to each child:

```python
deadline = None if timeout is None else time.monotonic() + timeout

...

remaining = (
    None if deadline is None
    else max(deadline - time.monotonic(), 0.0)
)

value = children_result.get(
    timeout=remaining,
    propagate=propagate,
)
```

This is just a sketch, though. Backend waits would need to share the same timeout budget, and we'd need tests to make sure ready results still work with `timeout=0`.

I've kept this PR focused on `propagate` for now, since handling nested timeouts properly would be a broader change. If you'd prefer to address both here, I'm happy to work on the timeout handling as well.


auvipy (commented):
any relevant issue or discussion you are targeting to fix with this?

GangEunzzang:
@auvipy I found this while testing nested groups and opened #10692 with a reproducer.

MehrazRumman (changes requested):
## Review

Correct fix, good tests, and credit for noticing the `timeout` problem rather than quietly forwarding it. The trace checks out: for a nested set, `iter_native` yields `(id, [child results])` instead of a meta dict, and the list branch in `join_native` called `children_result.get()` bare, so every nested child fell back to `propagate=True` regardless of what the caller asked for. Forwarding `propagate` fixes it, and because a nested `GroupResult.get()` routes back through `join_native`, the fix recurses to any depth — which the `depth=2` case confirms.

My main comment is about scope, and it's in the author's favour: the "broader change" you describe is mostly already written.

---

### 1. `propagate` isn't the only argument the nested branch drops — and `join()` already solved this

Compare the nested branch with `join()`, a few dozen lines up in the same class:

```python
# join()
remaining = max(timeout - (time.monotonic() - time_start), 0.0)
value = result.get(
    timeout=remaining, propagate=propagate,
    interval=interval, no_ack=no_ack, on_interval=on_interval,
    disable_sync_subtasks=disable_sync_subtasks,
)
```

```python
# join_native(), nested branch, after this PR
value.append(children_result.get(propagate=propagate))
```

The shared-deadline sketch in the PR description is essentially `join()`'s existing code, including its comment on what a spent budget means ("each result is polled once and raises if it is not ready"). So the design question — and the `timeout=0` question you raised — has already been settled for the non-native path. Mirroring it here is consistency, not new design.

Of the arguments still dropped, two are worth more than consistency:

* **`disable_sync_subtasks` turns into a hard error.** Inside a task, `group_result.get(disable_sync_subtasks=False)` passes the outer `assert_will_not_block()`, but the nested `children_result.get()` defaults to `True`, calls `assert_will_not_block()` again, and raises `RuntimeError: Never call result.get() within a task!` from one level down. Same class of bug as the one this PR fixes — an explicitly passed flag silently reset at the nested level — except it raises rather than just returning differently. (Callers using `allow_join_result()` are unaffected, since that sets the flag globally.)
* **`timeout` becomes unbounded.** Every nested child's `.get()` runs with `timeout=None`, so `group_result.get(timeout=10)` can block forever on a single stuck task inside a nested group. That's arguably worse than the `propagate` bug: that one raises unexpectedly, this one hangs.

`no_ack`, `interval` and `on_interval` are consistency fixes — `no_ack=False` in particular currently gets acked anyway for nested children.

Suggested shape:

```python
time_start = time.monotonic()
...
if isinstance(meta, list):
    value = []
    for children_result in meta:
        remaining = None
        if timeout is not None:
            remaining = max(timeout - (time.monotonic() - time_start), 0.0)
        value.append(children_result.get(
            timeout=remaining,
            propagate=propagate,
            interval=interval,
            no_ack=no_ack,
            on_interval=on_interval,
            disable_sync_subtasks=disable_sync_subtasks,
        ))
```

Two caveats so this doesn't over-correct:

* **Don't forward `callback` or `on_message`.** `callback` fires per top-level member with the nested list as its value — forwarding it would double-report every nested child. The current code is right to leave it off; a comment saying so would stop the next person "fixing" it.
* **This bounds the nested waits, not the whole call.** `iter_native` still gets the original `timeout` for its own `get_many`, so the worst case becomes roughly nested time plus that timeout rather than exactly `timeout`. Still a large improvement over unbounded. Making it exact would mean restructuring `iter_native`, and *that* is the genuinely broader change — reasonable to leave for a follow-up.

If maintainers would rather keep this PR to `propagate` only, I'd still push for `disable_sync_subtasks` to go in alongside it, since it's the same one-keyword fix for the same bug shape and the failure mode is an exception rather than a wrong value.

### 2. Tests

The parametrised matrix is well chosen: method × depth × callback covers both entry points, recursion, and the callback path's different return shape, and `test_get_nested_propagate_raises` pins the default so this can't be "fixed" by flipping it.

* **Please confirm `method='get'` actually exercises `join_native` in this class.** `get()` only dispatches there when `supports_native_join` is true; if this class's backend doesn't support it, `get` goes through `join()`, which already forwarded `propagate` — so half the matrix would pass on `main` and pin nothing. A one-line `assert ts.supports_native_join` makes the test self-documenting either way.
* The double `for _ in range(depth - 1)` — once to unwrap `values`, once to wrap `expected` — takes a moment to follow. Building the expected structure alongside the result structure at construction time would read more directly. Minor.
* If you take point 1, a test that nested `get()` respects `disable_sync_subtasks=False` inside a `denied_join_result()` block would pin the one that currently raises.

### 3. Housekeeping

* The issue link isn't a closing reference — the sidebar shows no linked issues. Adding `Fixes #10692` to the description will close it on merge.
* Changelog entry: nested `GroupResult.get(propagate=False)` returning exceptions instead of raising is a user-visible behaviour change, even if nobody should be relying on the old one.
* One merge commit from `main`; fine assuming a squash.

---

**Verdict:** mergeable as-is — it's a correct fix with real regression tests. I'd much prefer it went in with the fuller argument forwarding from point 1, since `join()` already defines the right behaviour and the `disable_sync_subtasks` case is the same bug with a worse symptom. Happy either way on whether exact whole-call timeout semantics happen here or in a follow-up.

GangEunzzang:
@MehrazRumman Thanks for the review! I’ve added `disable_sync_subtasks` forwarding and regression tests, along with an explicit native-join assertion. I’d like to leave timeout handling and the other backend options for a follow-up, where we can test their behavior more thoroughly.

auvipy (approved):

auvipy (approved):

